PR #136, opened 2013-12-12
Ensure translatable strings, and string domains are not dynamic

### LINKED ISSUE #50: Check for proper usage of translation functions
I had read that variables were discouraged from being used in the i18n functions. I assumed this was just to discourage them from being changed. Well, it's actually because some automated i18n tools will [tokenize](http://i18n.trac.wordpress.org/browser/tools/trunk/add-textdomain.php#L60) the PHP files to extract the textdomain from PHP without actually evaluating the file. See @markjaquith's blog post [Translating WordPress Plugins and Themes: Don’t Get Clever](http://markjaquith.wordpress.com/2011/10/06/translating-wordpress-plugins-and-themes-dont-get-clever/).

So a sniff should be written which makes sure that translation functions have a text domain which is a literal string. The sniff should also ensure that variables are not being used for the text to be translated, or inside of the text to be translated.

shadyvb:
Fixes #50.


westonruter:
> May I ask why you would disallow dynamic string domains ?

@jrfnl: See https://markjaquith.wordpress.com/2011/10/06/translating-wordpress-plugins-and-themes-dont-get-clever/

jrfnl:
@westonruter See my original comment in which I already respond to that article. Mark presumes in his article that people don't know how to generate .pot files outside of using the tooling in the WP repo admin.
I think an error is out of place, a warning may be more appropriate.

GaryJones:
At this moment in time, even the WP i18n tools don't actually limit the strings collected by the text domain i.e. the final argument (string, variable, constant, something else) is not actually read. This is why it also works for core i18n strings which obviously don't have that arg. The precaution here is trying to make future developments which may consider the text domain, easier.

jrfnl:
Thanks for that addition @GaryJones.

Sounds to me like all the more reason to differentiate between the messages about the translatable string and messages about the domain, to make messages about the domain a warning and to make sure they have a separate error code, so people can turn them off without turning messages about the translatable strings off.

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -0,0 +1,93 @@
...
+        // Get arguments
+        for ( $i = $nextToken + 1; $i < $tokens[$nextToken]['parenthesis_closer'] - 1; $i++ ) {
+            if ( in_array( $tokens[$i]['code'], array( T_WHITESPACE, T_COMMA,T_CONSTANT_ENCAPSED_STRING ) ) ) {
+                continue;
+            }
+
+            if ( $tokens[$i]['code'] === T_DOUBLE_QUOTED_STRING ) {
+                $string = $tokens[$i]['content'];
+                if ( preg_match( '#\$#', $string ) > 0 ) {
+                    $phpcsFile->addError( 'Translatable strings should not contain variables, found ' . $tokens[$i]['content'], $i );
```
westonruter: Per @jrfnl's note on this PR:

> Sounds to me like all the more reason to differentiate between the messages about the translatable string and messages about the domain, to make messages about the domain a warning and to make sure they have a separate error code, so people can turn them off without turning messages about the translatable strings off.

This can be change to a warning, and the third argument can he supplied to the method to provide the warning with an identifier which can be used to target the warning in a custom ruleset.

GaryJones: As a side question, shouldn't _all_ our errors and warnings have identifiers so folks can target them as they desire?

westonruter: > As a side question, shouldn't all our errors and warnings have identifiers so folks can target them as they desire?

@GaryJones yes, absolutely.

westonruter:
I checked with @ocean90, and he says if it is hosted on WordPress.org, the string literal text domain must match the theme/plugin slug. So…

- [ ] another check we can is to see if the string literal corresponds to a directory name in the file's directory hierarchy as a basic way to check for conformance to this.

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Shady Sharaf <shady@x-team.com>
+ */
+class WordPress_Sniffs_WP_I18nSniff implements PHP_CodeSniffer_Sniff {
+
+	public $i18n_functions = array(
+		'translate',
+		'translate_with_gettext_context',
```
ocean90: translate() and translate_with_gettext_context() are low-API functions, they shouldn't be used by plugins IMO.

Tests/WP/I18nUnitTest.php
```diff
@@ -0,0 +1,49 @@
...
+     * @return array(int => int)
+     */
+    public function getErrorList()
+    {
+        return array(
+            3 => 1,
+            7 => 1,
+            9 => 1,
+            );
+
```
westonruter: The comment was incorrectly saying it was bad, however. It is actually good.

ocean90:
Since #50 is about "proper usage of translation functions", should this include a warning when `_()` (single underscore) gets used? 

See also https://github.com/WordPress/WordPress/blob/bd4c250078deb4ade9b6b1d3ec884c63d6595e2e/wp-includes/compat.php#L9-L14.

westonruter:
- [ ] Warn when using `_()`, a single underscore.
- [ ] Implement proper checks for `_n()` and friends.
- [ ] Add discrete error codes.

jrfnl:
Just an idea: Looking at the added tests - what about an optional extra check to verify that a `text-domain` has been provided ? Would be off by default (core), but could be turned on either via the extra rules or via a custom `phpcs.xml` ruleset.
With a recommendation for plugins and themes to turn it on ?

westonruter:
@jrfnl I like that. Actually, I think we can accomplish this via a `MissingTextDomain` error code which for Core can just be suppressed in the default ruleset.

WordPress-Extra/ruleset.xml
```diff
@@ -33,6 +33,7 @@
 	<rule ref="WordPress.WP.PreparedSQL"/>
 	<rule ref="WordPress.Variables.GlobalVariables"/>
 	<rule ref="WordPress.PHP.StrictComparisons" />
+	<rule ref="WordPress.WP.I18n" />
```
westonruter: Should this go into `WordPress-Extra` or `WordPress-Core`? I'm just concerned about use with _actual_ core, and how a lot of warnings will start appearing since no text domain is supplied.

westonruter:
Ready for review.

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -0,0 +1,256 @@
...
+	/**
+	 * Check if supplied tokens represent a translation text string literal.
+	 *
+	 * @param PHP_CodeSniffer_File $phpcs_file The file being scanned.
+	 * @param string $arg_name
+	 * @param int $stack_ptr
+	 * @param array $tokens
+	 * @return bool
+	 */
+	protected function check_literal_string_text_tokens( $phpcs_file, $arg_name = '$text', $stack_ptr, $tokens = array() ) {
```
westonruter: There's an unfortunate amount of duplication between `check_literal_string_text_tokens`, `check_literal_string_context_tokens`, and `check_string_domain_tokens`. However, an immediate refactoring wasn't apparent which would allow custom messages as well as error codes for each argument type.

westonruter:
@JDGrimes would you be able to reviewmerge?

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -0,0 +1,256 @@
...
+ *
+ * Makes sure internationalization functions are used properly
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @author    Shady Sharaf <shady@x-team.com>
+ */
+class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
+
+	public $i18n_functions = array(
```
JDGrimes: Wouldn't it be better to flip this array and have the function names as keys (this is standard in most other sniffs)? Then you can use `isset()` instead of `in_array()`. And perhaps more importantly, you could let the nature of each function be specified by the values, rather than [hardcoding this below](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/136/files#diff-df68b36623d243f82931be87d069e1efR106). So you could have:

```php
array(
    '__' => 'simple',
    '_x' => 'context',
    '_n' => 'plural',
    '_nx' => 'plural_context',
)
```

ect.

Because right now if someone ever wanted to sub-class this and add their own custom translation functions (bad idea, I know), they'd have to duplicate all of the code rather than just add them to this list, because how each function is handled has to be hard-coded below.

Anyway, I just prefer `isset()` to `in_array()`. :smile:

westonruter: Great points! See 9ebd183.

WordPress/Sniffs/WP/I18nSniff.php
JDGrimes: > There's an unfortunate amount of duplication between check_literal_string_text_tokens, check_literal_string_context_tokens, and check_string_domain_tokens. However, an immediate refactoring wasn't apparent which would allow custom messages as well as error codes for each argument type.

Aren't the messages and error codes identical other than the arg names though? ("translatable text", "context", "text domain") I think it could be refactored to a single method that would just accept those strings and build the error messages and codes from them?

westonruter: Also, some are errors and some are warnings.

It could be refactored to try to abstract these out more by passing in params to common methods if you think it is worthwhile.

westonruter: @JDGrimes here you go: a47029e7af7902174f1b77ce3e27339bc5522bb4

grappler:
The text domain can be a variable even for the WordPress makepot or on WordPress.org. The only condition for wordpress.org is that the text domain must match the theme/plugin slug.

I have tested it. Anyone who has says differently has not tested it. What is best practice is another disucssion.

ocean90:
@grappler Nobody is saying that it doesn't work… See also https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/136#issuecomment-101508414. 

Not sure why you think that it's just a "best practice" but it's definitely a WordPress coding standard which is enforced by several handbook pages, for example https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/.

grappler:
> It can’t read variables or constants. It only reads strings. So your text domain strings needs to stay hardcoded as actual quoted strings. 'my-plugin-name'.
Quoting from the post https://markjaquith.wordpress.com/2011/10/06/translating-wordpress-plugins-and-themes-dont-get-clever/

Yes, I can accept it is WordPress coding standard. :smile:

korobochkin:
Is it ok if I can't use variable (constant) as argument for translation functions? I have a lot of string needed to be translated and to make my work and writing code faster I'm using `Plugin::NAME` constant as text domain. But code sniffer generates an error:

>The $domain arg must be a single string literal, not "Plugin::NAME".

PR #210, opened 2014-07-12
Account for agreed upon $_SERVER vars

### LINKED ISSUE #191: False positive - VIP.ValidatedSanitizedInput
This looks like a false positive to me:

`Detected usage of a non-validated input variable: $_SERVER`

``` php
$domain = str_replace( 'www.', '', sanitize_text_field( $_SERVER['SERVER_NAME'] ) );
```

shadyvb:
Closes #191


WordPress/Sniffs/VIP/ValidatedSanitizedInputSniff.php
```diff
@@ -45,6 +45,26 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr )
...
+		 *
+		 * @see https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/191
+		 */
+		$server_vars = array(
+			'AUTH_TYPE', 'CONTENT_LENGTH', 'CONTENT_TYPE', 'GATEWAY_INTERFACE',
+			'PATH_INFO', 'PATH_TRANSLATED', 'QUERY_STRING', 'REMOTE_ADDR',
+			'REMOTE_HOST', 'REMOTE_IDENT', 'REMOTE_USER', 'REQUEST_METHOD',
+			'SCRIPT_NAME', 'SERVER_NAME', 'SERVER_PORT', 'SERVER_PROTOCOL',
+			'SERVER_SOFTWARE',
+		);
```
westonruter: I wonder if these are always available in a WP-CLI context, however.

JDGrimes: > I wonder if these are always available in a WP-CLI context, however.

I'm not sure, but some server vars can always be expected within WordPress thanks to [`wp_fix_server_vars()`](https://developer.wordpress.org/reference/functions/wp_fix_server_vars/). Specifically, it makes sure `SERVER_SOFTWARE` and `REQUEST_URI` are always available, and assumes that `SCRIPT_NAME` and `PHP_SELF` exist.

I think that there are other places where WordPress assumes the existence of different `$_SERVER` vars, however, I've found some of these to give notices when running PHPUnit tests.

I think the point of this sniff is that superglobals are untrusted data that could be modified by anything. So, I kind of favor wontfix'ing this. Even if we could establish that all servers always make x server var available, one can still argue that it is best not to expect it, since something else might be un/intentionally unsetting it. Sure, that probably isn't a good idea, but they are superglobals, and so, like all globals, they are fragile (only more-so).

JDGrimes: Related: 
- [[33775]](https://core.trac.wordpress.org/changeset/33775) - Avoid PHP notices in `redirect_canonical()` and `_wp_menu_item_classes_by_context()` if `$_SERVER['HTTP_HOST']` is not set.
- [[32839]](https://core.trac.wordpress.org/changeset/32839) - Make sure `$_SERVER['SERVER_NAME']` is set whenever `wp_mail()` is called in PHPUnit tests.
- [[32785]](https://core.trac.wordpress.org/changeset/32785) - Set `$_SERVER['REQUEST_METHOD']` for unit tests.

In other words, WordPress assumes that `REQUEST_METHOD` and `SEVER_NAME` will always be set, and monkey-patches them when running the PHPUnit tests.

It doesn't assume that `HTTP_HOST` is set, and it is not in this list either.

JDGrimes: Related: [#33877](https://core.trac.wordpress.org/ticket/33877) - Initialise `$_SERVER[ 'REMOTE_ADDR' ]` in the unit test bootstrap

I'd be OK with merging this at some point, but with the following modifications:
- It should be possible to turn off ignoring these server vars.
- Maybe we should consider limiting the list to only those that WordPress assumes will exist.
- In that case, we'll probably want to make it possible for the list to be extended via XML config.

JDGrimes: Related: [#34394](https://core.trac.wordpress.org/ticket/34394) – Initialise `$_SERVER['SERVER_NAME']` in the unit test bootstrap

GaryJones:
Closing due to conflict and lack of movement. Can be re-opened if wanted,

PR #294, opened 2014-12-26
Support whitelisting for input sanitization

### LINKED ISSUE #293: Whitelisting for the input sanitization sniff
In the XSS output escaping sniff there is a check for a comment at the end of the line (like `// XSS OK`) which can be used to whitelist  a location. It would be nice to have this for the sanitized input sniff.

JDGrimes:
Whitelisting covers sanitization only. Validation can't be whitelisted away.

Fixes #293

### DIFF
--- a/WordPress/Sniffs/VIP/ValidatedSanitizedInputSniff.php
+++ b/WordPress/Sniffs/VIP/ValidatedSanitizedInputSniff.php
@@ -67,6 +67,10 @@ class WordPress_Sniffs_VIP_ValidatedSanitizedInputSniff implements PHP_CodeSniff
 			if ( in_array( $tokens[ key( $nested ) - 1 ]['code'], array( T_ISSET, T_EMPTY, T_UNSET ) ) )
 				return;
 		} else {
+			if ( $this->is_whitelisted_by_comment( $tokens, $stackPtr ) ) {
+				return;
+			}
+
 			// Search for casting
 			$prev = $phpcsFile->findPrevious( array( T_WHITESPACE ), $stackPtr -1, null, true, null, true );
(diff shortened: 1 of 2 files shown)


westonruter:
@JDGrimes would you have some more examples of how this would be used in practice?

westonruter:
I suppose I have tended to pass any input vars through `sanitize_text_field()` all the time, in part to ensure that this sniff is not flagged, for instance:

```php
if ( isset( $_GET['value'] ) ) {
    $value = wp_unslash( sanitize_text_field( $_GET['value'] ) );
}
```

Is this what you're wanting to add a means of whitelisting?

JDGrimes:
In this case the input parameter is an array of values. I'm doing this:
```php
$value = array_map( 'sanitize_text_field', $_GET['value'] );
```

Maybe a better solution would be to roll my own `sanitize_text_deep()` function and add it to the list of sanitizing functions via the XML config.

However, there is also another case similar to this but where HTML may be allowed. I can't use `sanitize_text_field()` because it will destroy any HTML. The input is validated/sanitized, but it isn't known what types of values the array might contain. The function to validate the input is chosen dynamically. It is a case similar to saving the array of settings for a Widget; the widget class has a method that sanitizes and validates the settings for that type of widget before it is saved. So everything is sanitized, but there is no way for static analysis to figure that out.

westonruter:
Yeah, case for array input vars is a good one. We could potentially look for `array_map()` and ensure that a valid sanitizing function is used as a callback there.

For HTML, shouldn't `wp_kses` be used? It will satisfy the WPCS check as well.

lkraav:
How far away is this from landing based on #339?

JDGrimes:
@lkraav I'll try to get this into 0.4.0, so, by the end of the month.

Issue #297, opened 2015-01-11
Sniff errors/warnings not all named


jrfnl:
Each error/warning can be named:
```php
$phpcsFile->addError($error, $stackPtr, 'SpacingAfter');

//vs the unnamed variant:

$phpcsFile->addError($error, $stackPtr);
```

If they are, they can be selectively disabled:
```xml
<exclude name="WordPress.VIP.RestrictedFunctions.user_meta" />
```

A number of the WP specific sniffs at this moment don't have all errors/warnings named which means that if an errors/warnings is buggy or a project for whatever reason chooses not to use that errors/warnings, they have to disable the complete sniff as the specific errors/warnings can't be disabled.

I'd suggest that naming all errors/warnings generated would be good practice. Opinions ?

JDGrimes:
Duplicate: #140

westonruter:
:+1:

jrfnl:
PR #298 addressed this.

westonruter:
PR merged. Nice work.

jrfnl:
Thanks ;-)

Issue #140, opened 2013-12-23
Add custom error name to addError function


jonathanbardo:
It would be great if addError function in WPCS ([example here](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/b46eedae7fb41aebee2632520e84f0ced47debd1/Sniffs/Arrays/ArrayDeclarationSniff.php#L425)) function would implement the 3rd argument of phpcs [Ref Squiz Lab](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/File.php#L734).

This could add more granularity in custom ruleset based on top of this standard like so : http://www.squizlabs.com/view?a=6160.

Right now we can only exclude a sniff but not an error message inside a sniff.

westonruter:
PR merged into `develop`.

PR #298, opened 2015-01-12
Add granular error/warning names - issue #297, #140

jrfnl:
I've added granular error/warning names in most places.
* I've ignored some sniff files which only throw one specific error as those can be disabled by just disabling the sniff itself.
* For error/warning names, I've used the Squiz and PEAR coding standards for inspiration and stuck to the names used in there for the same error if one existed.
* In 1 instance (VIP/CronIntervalSniff), the existing error name did not comply with the CamelCaps convention for the names. I've changed the name. This will break backward compatibility for any rulesets referencing the 'old' name.


GaryJones:
+1

westonruter:
Nice job!

PR #299, opened 2015-01-12
Change array key spacing message from warning to error

jrfnl:
As it's a hard rule mentioned in the core guide and most other spacing rules are enforced with errors, I suggest changing the enforcement of this rule from warning to error.


westonruter:
Maybe then `should` should become `must` in the error messages?

Please refresh the branch as the merge of #298 now creates a conflict with this.

jrfnl:
Updated the wording & rebased. Should be fine now.

PR #303, opened 2015-01-18
Account for ternary operator in XSS sniff

### LINKED ISSUE #281: XSS escaping and the ternary operator
I got this error:

> 211 | ERROR | Expected next thing to be a escaping function, not '('

The offending line:
```
value="<?php echo ( $args['placeholders'] && 'hidden' !== $field['type'] ) ? '<% if ( typeof ' . sanitize_key( $name ) . ' !== "undefined" ) { print( ' . sanitize_key( $name ) . ' ); } %>' : esc_attr( $value ); ?>"
```
While on the one hand this is messy, I chose to do it anyway because it was the best way to do the escaping immediately before output.

When dealing with the ternary operator, I think we need to ignore everything before the `?`.

JDGrimes:
When a ternary operator is present in an `echo` statement, it is safe
to skip over everything before the `?`, since that part isn’t output.

Includes unit tests.

Fixes #281

### DIFF
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -310,15 +310,17 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff implements PHP_CodeSniffer_Sniff
 			}
 		}
 
+		$end_of_statement = $phpcsFile->findNext( T_SEMICOLON, $stackPtr );
+
+		// Check for the ternary operator.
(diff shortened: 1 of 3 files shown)


jrfnl:
Nice :+1:

westonruter:
:+1: 

Way to go.

It probably won't handle nested ternaries, but you have bigger problems if you try that :wink:

JDGrimes:
> It probably won't handle nested ternaries, but you have bigger problems if you try that 
> 
Yeah, I made a conscious decision that I wasn't going to go there. :-)=

jrfnl:
Rightfully so: https://github.com/dseguy/clearPHP/blob/master/rules/no-nested-ternary.md

PR #304, opened 2015-01-18
Don't give an error for mixed key/keyless arrays

### LINKED ISSUE #233: "Key specified for array entry; first entry has no key"
Adding some more unit tests for #134 and tried something which simplifies down to:

~~~php
$keyless_then_keyed_value = array(
	'value',
	'key' => 'value',
);
~~~

That will trigger the error at https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blame/develop/WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php#L242

Why is this an error?

At some unknown time during refactoring, this was changed from checking for [keyed then keyless](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/commit/04fd547c691ca2baae3fa8e195a46b0c9dd671c5#diff-cc87c471fa9290ab5f67791d61e775c7R213) and [keyless then keyed](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/commit/04fd547c691ca2baae3fa8e195a46b0c9dd671c5#diff-cc87c471fa9290ab5f67791d61e775c7R250). 

As such, since the change, reversing the items in the array above does not trigger any error.

Very related is [this](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/commit/d4263757260d793a7425bc1d5c16a01054177cd7) which changes only one of the pre-refactored errors into warnings, suggesting that both should have been warnings at most.

Since it's perfectly fine to have keyed and keyless entries in PHP, and the tax query uses them as an example, should this check be dropped altogether?

JDGrimes:
This isn’t a part of WordPress’s coding standards. Actually, meta
queries require this for setting the query’s `relation`.

Fixes #233

### DIFF
--- a/WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
+++ b/WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
@@ -188,7 +188,6 @@ class WordPress_Sniffs_Arrays_ArrayDeclarationSniff implements PHP_CodeSniffer_S
         $nextToken  = $stackPtr;
         $lastComma  = $stackPtr;
         $keyUsed    = false;
-        $singleUsed = false;
         $lastToken  = '';
         $indices    = array();
         $maxLength  = 0;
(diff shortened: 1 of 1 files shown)


westonruter:
I also amended the unit test to be sure this was accounted for: 71ad45a50a11c57b0883459716a527fe8ea41a00

PR #307, opened 2015-02-05
Allow camel caps method names in child classes

### LINKED ISSUE #278: Don't flag camelCase naming on subclasses
If we're subclassing and existing class, we have no choice but use the naming convention of the parent class. For instance, subclasses of `WP_UnitTestCase` should not be flagged by PHP_CodeSniffer when we override the `setUp` method.

JDGrimes:
Fixes #278


westonruter:
Thanks so much!! :heart_eyes:

PR #308, opened 2015-02-09
Remove BOM that breaks JSON output parsing

leewillis77:
WordPress/Sniffs/WhiteSpace/OperatorSpacingSniff.php contains byte-order marks before the opening PHP tags. This causes output reports to also include this output. Unfortunately this breaks parsing of the JSON reports (See https://github.com/Block8/PHPCI/issues/768). 

This patch just removes the extra characters before the opening PHP tags.


westonruter:
Thanks!

GaryJones:
Nice catch Lee :+1:

leewillis77:
No worries guys - confused me for a bit, thanks for the work on this - it's a great resource!

PR #309, opened 2015-02-09
Allow integers in XSS sniff

JDGrimes:
Currently the XSS sniff will flag this:

```php
echo 5;
```

Of course, that’s harmless.


westonruter:
On ya!

Issue #170, opened 2014-05-01
Flag unescaped variables passed to printf, fwrite, etc


westonruter:
There are functions other than the `echo` and `print` statements which output data. The `printf` function, for example, we've seen used at times to get around PHPCS warnings.

JDGrimes:
We now flag `printf()`. Do you still think `fwrite()` be added to the list?

westonruter:
Perhaps, but in practice I don't think I've ever seen `fwrite()` used in any theme or plugin code, so it seems like a very low priority.

GaryJones:
Also, use of the Filesystem API should be encouraged.

grappler:
`fwrite()` is now included in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/633

Issue #212, opened 2014-07-12
Track sanitization of sprintf params in an echo statement.


shadyvb:
From #189 reported by @jrfnl, these usages should be handled by `EscapeOutputSniff`

```php
echo '<p>' . sprintf( esc_html__( 'Some text -> %sLink text%s', 'textdomain' ), '<a href="' . esc_url( add_query_arg( array( 'page' => 'my_page' ), admin_url( 'admin.php' ) ) ) . '">', '</a>' ). '</p>';

echo '<br/><strong>' . sprintf( esc_html__( 'Found %d results', 'textdomain' ), $result_count ) . '</strong><br/><br/>';
```

PR #312, opened 2015-02-15
XSS sniff: require escaping for die(), exit(), printf()

JDGrimes:
Other functions we should possibly add: https://stackoverflow.com/questions/4673747/what-php-functions-create-output

See #170

### DIFF
--- a/WordPress/Tests/XSS/EscapeOutputUnitTest.inc
+++ b/WordPress/Tests/XSS/EscapeOutputUnitTest.inc
@@ -67,3 +67,5 @@ exit( esc_html( $foo ) ); // OK
 
 die( $foo ); // Bad
 die( esc_html( $foo ) ); // OK
+
+printf( 'Hello %s', $foo ); // Bad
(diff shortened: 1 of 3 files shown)


WordPress/Tests/XSS/EscapeOutputUnitTest.inc
```diff
@@ -60,4 +60,12 @@ echo $foo ? $foo : 'no foo'; // Bad
...
\ No newline at end of file
+echo 4; // OK
+
+exit( $foo ); // Bad
+exit( esc_html( $foo ) ); // OK
+
+die( $foo ); // Bad
+die( esc_html( $foo ) ); // OK
+
+printf( 'Hello %s', $foo ); // Bad
```
westonruter: @JDGrimes I think `printf` here is a tricky one, and it is discussed in #212. We basically need recognize that what you have here is indeed bad, but whitelist the following as OK:

```php
printf( 'Hello %s', esc_html( $foo ) ); // OK
```

That will require some sophistication, so I think this that `printf` and `vprintf` should be removed from the scope of this PR, unless you want to tackle #212 as well :smile:

JDGrimes: I've just pushed some improved tests in 39881e8 that demonstrate that this works for `printf()`. It also addresses `vprintf()`. `sprintf()` is still another story, which I/someone will tackle at a later date. :-)

westonruter:
Impressive!

corvannoorloos:
I believe this gives an error on:
```php
if ( ! defined( 'ABSPATH' ) ) {
	exit;
}
```
As I probably don't comprehend the complete change, would you suggest also escaping this?

JDGrimes:
@corvannoorloos That is a bug. I'll create a PR to fix it shortly.

corvannoorloos:
@JDGrimes thanks!

PR #313, opened 2015-02-16
XSS: don't flag code after an empty exit

JDGrimes:
This bug would happen when there was code like this:

```php
if ( ! defined( ‘ABSPATH’ ) ) {
exit; // We skipped over the semicolon, assuming the next thing would be an opening parenthesis.
}

// This code would get flagged as needing escaping.
other_code();
```

See #312


westonruter:
Great work!

PR #316, opened 2015-02-19
XSS sniff: Add wp_die() to the list of printing functions

JDGrimes:
`wp_die()` takes three parameters, and [all of their values may be
printed to the
user](https://developer.wordpress.org/reference/functions/_default_wp_di
e_handler/), without any escaping.


westonruter:
Good on you. But could you update the unit test to show the new usages that are caught?

JDGrimes:
I've updated the unit tests (and fixed several issues they turned up).

westonruter:
:+1:

PR #318, opened 2015-02-21
Support for echo with comma-separated args

JDGrimes:
`echo` accepts an unlimited number of arguments, separated by commas.
This was previously supported, however, there was a bug that any
arguments after the first escaping function would be ignored.

Recent changes (specifically e329860329df49f54a727aa2125b217b06a50e1f),
have made it so that now any commas themselves are being reported as
needing sanitization.

The fix is a simple change on line 373 so that we always wake up for
commas. All of the arguments given to `echo` are always checked, and
the commas themselves are ignored.

However, this causes an issue with the casting of function calls with
multiple arguments.

```php
echo (int) get_post_meta( $post_id, ‘some_integer’, true );
```

The first comma in the casted function would cause us to wake up
prematurely, and report the later arguments as needing escaping. I’ve
fixed this issue by watching for an opening parenthesis when the value
is being casted, and jumping to the end of the casted function. There
is already a test (on line 29) that demonstrates that this is working.

### DIFF
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -352,6 +352,7 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff implements PHP_CodeSniffer_Sniff
 			// Skip to the end of a function call if it has been casted to a safe value.
 			if ( T_OPEN_PARENTHESIS === $tokens[ $i ]['code'] && $in_cast ) {
 				$i = $tokens[ $i ]['parenthesis_closer'];
+				$in_cast = false;
 				continue;
 			}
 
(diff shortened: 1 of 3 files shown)


westonruter:
Please refresh the patch. The PR #320 seems to conflict.

JDGrimes:
> Please refresh the patch. The PR #320 seems to conflict.

Updated.

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -347,6 +349,12 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr )
 				continue;
 			}
 
+			// Skip to the end of a function call if it has been casted to a safe value.
+			if ( T_OPEN_PARENTHESIS === $tokens[ $i ]['code'] && $in_cast ) {
+				$i = $tokens[ $i ]['parenthesis_closer'];
```
westonruter: @JDGrimes should this then reset `$in_cast` to be `false`?

JDGrimes: Yeah. But strangely it seems to work anyway.

JDGrimes: dcd3c49 fixes this.

PR #319, opened 2015-02-21
Yoda Sniff: Use tokenizer instead of regex

### LINKED ISSUE #290: Yoda comparisons not flagged against constants
The lack of Yoda comparisons not flagged against constants, only against strings

```php
    if ( $taxonomy === MyClass::TAXONOMY_SLUG ) { // not flagged
        $link = true;
    } elseif ( $taxonomy === 'post_tag' ) { // flagged
        $link = false;
    }
```

If you correct the flagged errors, you end up with:

```php
    if ( $taxonomy === MyClass::TAXONOMY_SLUG ) { 
        $link = true;
    } elseif (  'post_tag' === $taxonomy ) { // flagged
        $link = false;
    }
```

which is inconsistent, and looks like a bug.

JDGrimes:
This fixes issues with comparison to constants not being flagged, and
probably will be less buggy overall.

There is one side-effect:

```php
// Before this was good:
if ( $true == strtolower( $check ) ) {}

// Now it’s bad, but this is OK:
if ( strtolower( $check ) == $true  ) {}
```

I think this is the correct behavior.

Another thing to consider is when comparing a variable to a variable. I
don’t remember what the previous behavior was, but this is current
behavior:

```php
if ( $foo == $bar ) {} // Bad
if ( ( $foo ) == $bar ) {} // OK
```

Finally, the matching code is no longer included in the error message.
It’s possible to add that back in, but it would take some work.

Feedback welcome. :-)

Fixes #290

### DIFF
--- a/WordPress/Sniffs/PHP/YodaConditionsSniff.php
+++ b/WordPress/Sniffs/PHP/YodaConditionsSniff.php
@@ -29,8 +29,10 @@ class WordPress_Sniffs_PHP_YodaConditionsSniff implements PHP_CodeSniffer_Sniff
 	public function register()
 	{
 		return array(
-			T_IF,
-			T_ELSEIF,
+			T_IS_EQUAL,
+			T_IS_NOT_EQUAL,
+			T_IS_IDENTICAL,
+			T_IS_NOT_IDENTICAL,
 		);
 
 	}//end register()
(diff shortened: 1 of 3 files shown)


westonruter:
There's a unit test failure:

```
There was 1 failure:
1) WordPress_Tests_PHP_YodaConditionsUnitTest::getErrorList
[LINE 74] Expected 0 error(s) in YodaConditionsUnitTest.inc but found 1 error(s). The error(s) found were:
 -> Use Yoda Condition checks, you must (WordPress.PHP.YodaConditions)
/tmp/phpcs/tests/Standards/AbstractSniffUnitTest.php:169
/tmp/phpcs/tests/TestSuite.php:48
FAILURES!
Tests: 30, Assertions: 0, Failures: 1.
```

GaryJones:
> `if ( strtolower( $check ) == $true  ) {}`

According to the description (_"When doing logical comparisons, **always put the variable on the right side**, constants or literals on the left."_), so this is technically correct. 

> `if ( ( $foo ) == $bar ) {} // OK`

I'm really not keen on that. It feels like a hack just to get around the check.
Can you count the variables (or set `$has_variable` to true), instead of breaking at the first instance, and then set `$needs_yoda` and break if there's a second variable?

westonruter:
> `if ( ( $foo ) == $bar ) {} // OK`

I'm not too concerned with this. Most wouldn't think to add parens there anyway. But, in this situation, where the L-value and the R-value are *both* variables, the sniff needs to not flag it at all because obviously the Yoda condition doesn't make any sense since there is nothing reversed about saying `$foo == $bar` vs `$bar == $foo`. They are identical, and both should be fine.

GaryJones:
Sorry, ambiguity on my part.

Can you count the variables (or set `$has_variable` to true) and set `$needs_yoda` to `true`, instead of breaking at the first instance, and then set `$needs_yoda` _to `false`_ and break if there's a second variable?

Basically, I'm agreeing Weston, that `$var1 == $var2` should not be flagged, without resorting to using paretheses. I'm not concerned, at this stage, about someone's intended use of redundant parentheses which then passes, but about any unfortunate encouragement of them, as a way to pass this sniff. I'd rather get this in and then address edge cases later.

JDGrimes:
> $var1 == $var2 should not be flagged, without resorting to using paretheses.

OK, I'm fine with that. My rationale was just that it is in the spirit of yoda conditions. From [the standards](https://make.wordpress.org/core/handbook/coding-standards/php/#yoda-conditions):

>In the above example, if you omit an equals sign (admit it, it happens even to the most seasoned of us), you’ll get a parse error, because you can’t assign to a constant like true. If the statement were the other way around ( $the_force = true ), the assignment would be perfectly valid, returning 1, causing the if statement to evaluate to true, and you could be chasing that bug for a while.

I was just thinking that the idea is for any conditions involving variables to be invalid if the equality operator is accidentally converted to an assignment operator. Adding parentheses around the right-hand variable does that.

----

Something else I just thought of. Previously this sniff was hooked to the `T_IF` and `T_ELSEIF` tokens. This will hook it to the equality operators directly, which will check their use elsewhere as well, like in ternaries, etc.

GaryJones:
While I get your point (which, to be honest, I hadn't thought of in the way you'd described), I think if someone can remember to add parentheses around one variable when it's just this type of condition, then they can also just look at the equals signs.

----
Perhaps a combined  approach is to add a Warning check to WordPress Extra when an assignment is made inside a condition, since some prefer to avoid condition assignments as a matter of good coding anyway?

JDGrimes:
> I think if someone can remember to add parentheses around one variable when it's just this type of condition, then they can also just look at the equals signs.

Right, but if someone can remember to use yoda conditions, they can just check the equals sign, too. :smile: 

> Perhaps a combined approach is to add a Warning check to WordPress Extra when an assignment is made inside a condition, since some prefer to avoid condition assignments as a matter of good coding anyway?

I'm actually now thinking that the whole yoda conditions thing is silly, and that it makes more sense for the standard to just disallow assignments in conditions. Or, require that condition assignments by accompanied by a comment flag. Not that one doesn't get used to yoda conditions, but I think with today's codesniffing tools it is really no longer useful, just confusing for newbies and something else adding to the learning curve.

JDGrimes:
596b75e makes `if ( $foo == $bar ) {} // OK`.

JDGrimes:
I think this is ready to be merged. It now preserves the originally behavior for var to var comparison. The only significant change that isn't a fixed bug, is that it no longer displays the piece of offending code in the error message. But that can be added back if needed.

---

Another bug this fixes:

```php
// Before, this was OK:
if ( $array['key'] !== '' ) {
}
```

PR #320, opened 2015-02-21
XSS sniff: allow for negative numbers

JDGrimes:
I have quite a few instances of `wp_die( -1 );` in some of my code.
This is used a lot in WordPress too.


westonruter:
You are on an incredible roll.

PR #324, opened 2015-03-02
Fix EscapeOutput regression

westonruter:
A regression was introduced in the `EscapeOutputSniff`, probably as part of #312, #314, #315, #316, or #317.

The second two of the three OK lines added to this unit test are failing. It seems not adding a semicolon to the first instance causes the second OK line to croak, but the third OK line seems to fail regardless of prior semicolons.


westonruter:
@JDGrimes perfect! Works so much better now!

PR #327, opened 2015-03-08
Ignore docblock comments within arrays

### LINKED ISSUE #230: Array trailing comma on comment lines
Related: #59 

~~~php
add_theme_support( 'tha_hooks', array(

	/**
	 * As a Theme developer, use the 'all' parameter, to declare support for all
	 * hook types.
	 * Please make sure you then actually reference all the hooks in this file,
	 * Plugin developers depend on it!
	 */
	'all',

	/**
	 * Themes can also choose to only support certain hook types.
	 * Please make sure you then actually reference all the hooks in this type
	 * family.
	 *
	 * When the 'all' parameter was set, specific hook types do not need to be
	 * added explicitly.
	 */
	'html',
	'body',
	'head',
	'header',
	'content',
	'entry',
	'comments',
	'sidebars',
	'sidebar',
	'footer',

	/**
	 * If/when WordPress Core implements similar methodology, Themes and Plugins
	 * will be able to check whether the version of THA supplied by the theme
	 * supports Core hooks.
	 */
//	'core'
) );
~~~

This is from the Theme Hooks Alliance, specifically a [PR](https://github.com/zamoose/themehookalliance/pull/15) that aims to fix up the CS.

One false-positive seems to be occurring on that penultimate line ( `// 'core'` ) - it's asking for a trailing comma.

> Comma required after last value in array declaration (WordPress.Arrays.ArrayDeclaration)

Since this is a comment, there shouldn't be a comma, and the previous non-comment line does have a trailing comma. Moving the block comment to last doesn't change the false positive. Moving `'footer',` down naturally fixes the problem, so it strongly suggests it is related to having a comment / block comment as the last item.

Does the sniff need to exclude the trailing comma check for comment / block comment tokens?

JDGrimes:
When checking that the last line of an array ends in a comma, both whitespace ( `T_WHITESPACE`) and regular comments (`T_COMMENT`) were ignored, but not docblock comments (those beginning with `/**`, `T_DOC_COMMENT`). This fixes that.

Fixes #230

### DIFF
--- a/WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
+++ b/WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
@@ -314,7 +314,7 @@ class WordPress_Sniffs_Arrays_ArrayDeclarationSniff implements PHP_CodeSniffer_S
             $count     = count($indices);
             $lastIndex = $indices[($count - 1)]['value'];
 
-            $trailingContent = $phpcsFile->findPrevious(array(T_WHITESPACE, T_COMMENT), ($arrayEnd - 1), $lastIndex, true);
+            $trailingContent = $phpcsFile->findPrevious(PHP_CodeSniffer_Tokens::$emptyTokens, ($arrayEnd - 1), $lastIndex, true);
             if ($tokens[$trailingContent]['code'] !== T_COMMA) {
                 $error = 'Comma required after last value in array declaration';
                 $phpcsFile->addError($error, $trailingContent, 'NoCommaAfterLast');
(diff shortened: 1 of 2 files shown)


westonruter:
@JDGrimes can you wait to merge at least until Travis finishes its checks? :smile:

JDGrimes:
> @JDGrimes can you wait to merge at least until Travis finishes its checks? :smile:

Oops. Yes. Next time I'll just open the PR and add `#reviewmerge` at the bottom—then someone else can review and merge at will.

westonruter:
Beauty. Perfect.

GaryJones:
> Next time I'll just open the PR and add #reviewmerge at the bottom—then someone else can review and merge at will.

I've just added the option of a [reviewmerge](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/labels/reviewmerge) label, so you can add that instead.

PR #331, opened 2015-03-12
Removes content_width as non-overwritable global. Fixes #276

### LINKED ISSUE #276: Stop flagging the override of $content_width
Themes are supposed to be setting this `$content_width` global variable, so we should not be flagging it as an error.

See #26 and #213.

allan23:
I removed content_width from the globals array which makes it so it won't get flagged. #276


westonruter:
Good call! This is supposed to be overwritten.

PR #334, opened 2015-03-15
abstract WordPress_Sniff class

### LINKED ISSUE #246: abstract WordPress_Sniff
As per my 5th point of https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/243#issuecomment-56791336, if we're going to have more sniffs bypassable via end-of-line whitelisting comments, we should create a way for all our sniffs to access that feature, without duplicating the method to do it.

We can do that in an abstract class that sits implements `PHP_CodeSniffer_Sniff` then update our sniffs which use the features to extend our new abstract class.

I propose something like:

~~~php
abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
	/**
	 * Get the last pointer in a line
	 *
	 * @since 1.0.0
	 *
	 * @param array   $tokens   Tokens.
	 * @param integer $stackPtr The position of the current token in the	stack passed in $tokens.
	 *
	 * @return integer Next pointer position.
	 */
	protected function get_last_ptr( $tokens, $stackPtr ) {
		$currentLine = $tokens[$stackPtr]['line'];
		$nextPtr = $stackPtr;
		while ( isset( $tokens[$nextPtr + 1]['line'] ) && $tokens[$nextPtr + 1]['line'] == $currentLine ) {
			$nextPtr++;
			// Do nothing, we just want the last token of the line.
		}

		return $nextPtr;
	}
	
	/**
	 * Find whitelisting comment.
	 *
	 * Comment must be at the end of the line, and use // format.
	 * It can be prefixed or suffixed with anything e.g.
	 * ... // foobar okay
	 * ... // WPCS: foobar okay.
	 * are both fine, the latter meeting starting capital and trailing full stop
	 * that single comments may require.
	 *
	 * @since ...
	 *
	 * @param string  $comment  Comment to find.
	 * @param array   $tokens   Tokens.
	 * @param integer $stackPtr The position of the current token in the stack passed in $tokens.
	 *
	 * @return boolean True if whitelisting comment was found, false otherwise.
	 */
	protected function has_whitelist_comment( $comment, array $tokens, $stackPtr ) {
		$lastPtr = $this->get_last_ptr( $tokens, $stackPtr );

		return $tokens[$lastPtr]['code'] === T_COMMENT &&
			preg_match( '#' . preg_quote( $comment ) .'#i', $tokens[$lastPtr]['content'] ) > 0
		);
	}
}

// Change sniffs that need the functionality to
// extend WordPress_Sniff instead of PHP_CodeSniffer_Sniff
class WordPress_Sniffs_PHP_FoobarSniff extends WordPress_Sniff {
	function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr )
	{
		$tokens = $phpcsFile->getTokens();

		if ( ! $this->has_whitelist_comment( 'foobar okay', $tokens, $stackPtr ) {
			$error = 'Found: ' . $tokens[$stackPtr]['content'] . '. Use foobar.';
			$phpcsFile->addWarning($error, $stackPtr);
		}
	}
}
~~~

Untested, but you should get the idea.

As you can see, the `WordPress_Sniff` class also includes any other common patterns that occur over our sniffs - not just finding if a comment is the last thing on a line, but a method for getting the last pointer, etc. Of itself, these methods should be fairly easy to be unit testable too.

JDGrimes:
Adds an abstract class, `WordPress_Sniff` to provide a bootstrap for our sniffs. Includes a method for checking for whitelisting comments. I've converted the sniffs that allow for whitelisting comments to use this method, however, the logic is a bit different than previously used, so it'd be helpful if some folks would test this against projects where they use whitelisting comments heavily.

Feedback would also be appreciated regarding https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/246#issuecomment-80720786 and  a9ad818.

Closes #246

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -1,0 +1,86 @@
+<?php
+
+/**
+ * Represents a PHP_CodeSniffer sniff for sniffing WordPress coding standards.
+ *
+ * PHP version 5
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ */
+
+/**
+ * Represents a PHP_CodeSniffer sniff for sniffing WordPress coding standards.
+ *
+ * Provides a bootstrap for the sniffs, to reduce code duplication.
+ *
+ * @category  PHP
+ * @package   PHP_CodeSniffer
+ * @version   0.4.0
+ * @link      http://pear.php.net/package/PHP_CodeSniffer
+ */
+abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
+
+	/**
+	 * Get the last pointer in a line.
+	 *
+	 * @since 0.4.0
+	 *
+	 * @param array   $tokens   Tokens.
+	 * @param integer $stackPtr The position of the current token in the stack passed
+	 *                          in $tokens.
+	 *
+	 * @return integer Position of the last pointer on that line.
+	 */
+	protected function get_last_ptr_on_line( array $tokens, $stackPtr ) {
+
+		$currentLine = $tokens[ $stackPtr ]['line'];
+		$nextPtr = $stackPtr + 1;
+
+		while ( isset( $tokens[ $nextPtr ] ) && $tokens[ $nextPtr ]['line'] === $currentLine ) {
+			$nextPtr++;
+			// Do nothing, we just want the last token of the line.
+		}
+
+		// We've made it to the next line, back up one to the last in the previous line.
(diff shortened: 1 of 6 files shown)


JDGrimes:
I've gone ahead and merged this. If any whiltelisting comments stop working, open a new issue. (Or revert and slap my hand :smile:)

PR #341, opened 2015-03-27
XSS sniff: require escaping for error functions

### LINKED ISSUE #340: XSS sniff: Add _doing_it_wrong() to list of output functions
WordPress provides some error-reporting functions, like `_doing_it_wrong()`, which won't generate output by default. However, they do generate output when `WP_DEBUG` and `WP_DEBUG_DISPLAY` are on. In that case, passing un-escaped user input to these functions may result in XSS.

Of course, it is bad practice to enable error display on production servers. I'd like to hear any thoughts on whether escaping should be required here anyway.

FWIW, in the few instances in WordPress core where untrusted input is passed to `_doing_it_wrong()`, it is put through `esc_html()`.

Note that `_doing_it_wrong()` is just the scapegoat here, representative of a broader range of functions:

```
_deprecated_argument()
_deprecated_function()
_deprecated_file()
trigger_error()
user_error()
error_log()
debug_print_backtrace()
...
```

JDGrimes:
#reviewmerge

Fixes #340


GaryJones:
Couple of indentation tweaks needed in DiscouragedFunctionsSniff.php, but looks fine otherwise.

JDGrimes:
> Couple of indentation tweaks needed in DiscouragedFunctionsSniff.php, but looks fine otherwise.

Fixed.

PR #342, opened 2015-03-28
XSS sniff: better handling for sprintf()

### LINKED ISSUE #212: Track sanitization of sprintf params in an echo statement.

JDGrimes:
`sprintf()` is often used for formatting translation strings, and it is common practice to escape the individual parameters passed to it as needed instead of escaping the entire result. This is especially true when the string being formatted contains HTML, which makes escaping the full result more difficult.

This provides better handling for `sprintf()` by ignoring the function itself and only requiring escaping for its parameters.

Note that `sprintf()` also provides limited sanitization for the values passed to it when number formatting is performed (i.e., for all but the `%s` placeholder). I had originally wanted to take this into account, but it proved too complex. So all parameters passed to `sprintf()` need to be escaped, even when they will be formatted using, e.g., `%d`, and are really safe.

#reviewmerge

Fixes #212

### DIFF
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -264,6 +264,7 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff
 	public static $formattingFunctions = array(
 		'sprintf',
 		'vsprintf',
+		'wp_sprintf',
 	);
 
 	/**
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -439,11 +439,11 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff
 			else {
 				$functionName = $tokens[$i]['content'];
 
-				if ( in_array( $functionName, self::$formattingFunctions ) ) {
-					continue;
-				}
+				$is_formatting_function = in_array( $functionName, self::$formattingFunctions );
 
 				if (
+					! $is_formatting_function
+					&&
 					in_array( $functionName, self::$autoEscapedFunctions ) === false
 					&&
 					in_array( $functionName, self::$sanitizingFunctions ) === false
@@ -454,7 +454,15 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff
 
 				// Skip pointer to after the function
 				if ( $_pos = $phpcsFile->findNext( array( T_OPEN_PARENTHESIS ), $i, null, null, null, true ) ) {
-					$i = $tokens[$_pos]['parenthesis_closer'];
+
+					// If this is a formatting function we just skip over the opening
+					// parenthesis. Otherwise we skip all the way to the closing.
+					if ( $is_formatting_function ) {
+						$i = $_pos + 1;
+						$watch = true;
+					} else {
+						$i = $tokens[ $_pos ]['parenthesis_closer'];
+					}
 				}
 				continue;
 			}
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -248,6 +248,29 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff
 		'wp_die',
 	);
 
+	/**
+	 * Functions that format strings.
+	 *
+	 * These functions are often used for formatting translation strings, and it is
+	 * common practice to escape the individual parameters passed to them as needed
+	 * instead of escaping the entire result. This is especially true when the string
+	 * being formatted contains HTML, which makes escaping the full result more
+	 * difficult.
+	 *
+	 * @since 0.4.0
+	 *
+	 * @var array
+	 */
+	public static $formattingFunctions = array(
+		'sprintf',
+		'vsprintf',
+	);
+
+	/**
+	 * Whether the custom functions were added to the default lists yet.
+	 *
+	 * @var bool
+	 */
 	public static $addedCustomFunctions = false;
 
 	/**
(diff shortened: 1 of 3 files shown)


WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -248,6 +248,29 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff
...
+	 * being formatted contains HTML, which makes escaping the full result more
+	 * difficult.
+	 *
+	 * @since 0.4.0
+	 *
+	 * @var array
+	 */
+	public static $formattingFunctions = array(
+		'sprintf',
+		'vsprintf',
```
westonruter: @JDGrimes there's also the seldom-used `wp_sprintf`

westonruter:
Amazing how little code is added here to do such a complicated thing :smile:

JDGrimes:
@westonruter I've added `wp_sprintf()` to the list of formatting functions.

---

> Amazing how little code is added here to do such a complicated thing :smile:

I really didn't do the complicated part. :smile: 

Just to clarify, the first parameter (the format string), still requires sanitization (as do the other parameters). Essentially what it does is make it so that with this code:

```php
// In real life this is more complex or a little different and doesn't look as stupid. ;-)
echo sprintf( esc_html__( 'Hello %s' ), esc_html( $foo ) );
```

You wont get this error:

> Expected a sanitizing function (see Codex for 'Data Validation'), but instead saw 'sprintf'

It will pass instead.

If you have this:

```php
echo sprintf( esc_html__( 'Hello %s' ), $foo ); // Bad: $foo isn't escaped.
```

You will get a more helpful error:

> Expected a sanitizing function (see Codex for 'Data Validation'), but instead saw '$foo'

Instead of `sprintf()` being reported as the problem.

GaryJones:
How about the following (all taken from the Genesis Framework):

~~~php
echo sprintf( __( 'Welcome to Genesis %s', 'genesis' ), PARENT_THEME_BRANCH );

$patterns[] = sprintf( '~(<(?:%s)[^>]*)\s+%s=[\\\'"][^\\\'"]+[\\\'"]([^>]*[^>]*>)~', $elements_pattern, $attribute );

$output = sprintf( translate_nooped_plural( $units[$i][1], $count, 'genesis' ), $count );

	$output .= sprintf(
			'<label class="box%2$s" for="%5$s"><span class="screen-reader-text">%1$s </span><img src="%3$s" alt="%1$s" /><input type="radio" name="%4$s" id="%5$s" value="%5$s" %6$s class="screen-reader-text" /></label>',
			esc_attr( $data['label'] ),
			esc_attr( $class ),
			esc_url( $data['img'] ),
			esc_attr( $args['name'] ),
			esc_attr( $id ),
			checked( $id, $args['selected'], false )
		);
~~~

JDGrimes:
@GaryJones, the last three are not proceeded by an output function (e.g., `echo`) so aren't checked by the sniff. The first one is checked by the sniff, and should fail on two counts: `esc_html__()` should be used instead of `__()`, and `PARENT_THEME_BRANCH` should be wrapped in `esc_html()`. (Alternatively, the whole thing could be wrapped in `esc_html()`.)

I just realized that there is a bug which causes the first parameter of the formatting function to be ignored. Only `PARENT_THEME_BRANCH` was being reported. I'm fixing that now.

JDGrimes:
> I just realized that there is a bug which causes the first parameter of the formatting function to be ignored. Only PARENT_THEME_BRANCH was being reported. I'm fixing that now.

Done.

GaryJones:
Thanks for the clarification. Here's two Genesis instances with `echo sprintf(...)`:

~~~php
echo sprintf( '<div %s>', genesis_attr( 'entry-comments' ) ) . $no_comments_text . '</div>';

$heading = ( genesis_a11y( 'headings' ) ? 'h3' : 'h2' );
echo sprintf( '<%1$s class="widgettitle">%2$s</%1$s>', $heading, esc_html( $name ) );
~~~

How would they fare? (I'm not saying Genesis is right btw - just happens to be a repo I have quick access to for real life examples in fairly wide usage that isn't WP core.)

The first one can't be escaped as HTML, and the `genesis_attr()` already includes `esc_attr()` as it builds the string of attribute="values".

The second is similar, but uses placeholders. I guess the `$heading` arg could have `esc_html()`, but since it's not from user input, and can only be a choice of `h3` or `h2`, escaping it seems extremely redundant.

Also, how about `var_dump( sprintf( ... ) )`?

Just trying to think of some edge cases for you to test against :smiling_imp:

JDGrimes:
Those will both fail. Of course, they would have failed already but instead of `sprintf()` being reported as the culprit, `genesis_attr()` and `$heading` will:

```
 -> Expected a sanitizing function (see Codex for 'Data Validation'), but instead saw 'genesis_attr' (WordPress.XSS.EscapeOutput.OutputNotSanitized)
 -> Expected next thing to be a escaping function, not '$no_comments_text' 
-----
 -> Expected next thing to be a escaping function, not '$heading' (WordPress.XSS.EscapeOutput.OutputNotEscaped)
```

(Really, both of these seem like strange uses of `sprintf()`. For starters, `printf()` could be used instead of `echo sprintf()`. Also, when HTML is involved, it usually makes more sense to me to just break out of PHP.)



> Also, how about var_dump( sprintf( ... ) )?

`var_dump()` is currently only flagged by the discouraged functions sniff, and is not in the list of output functions recognized by the XSS sniff. That should be changed though, since it's possible that one sniff would be used with the other disabled.



> Just trying to think of some edge cases for you to test against :smiling_imp:

That is helpful. I do run my changes against some of my own projects, but there weren't any cases of `echo sprintf()` in them that would be flagged, so no help here.

JDGrimes:
> (Really, both of these seem like strange uses of sprintf(). For starters, printf() could be used instead of echo sprintf()...)

This will probably be true of most cases where `sprintf()` is used. The issue that is addressed by this PR is that in some cases there is proper escaping, but the developer has used `echo sprintf()` instead of `printf()`. This PR will save them from getting what will seem like a bunch of false alarm errors when they first run the codesniffer over a project where `echo sprintf()` is used heavily.

GaryJones:
I'm :+1: for getting this merged, and then handle any other issues with it later on.

Good work as always @JDGrimes :star2:

JDGrimes:
> Good work as always @JDGrimes :star2:

Thanks!

PR #344, opened 2015-04-04
Improve performance of the Files.FileName sniff

JDGrimes:
Instead of searching through the tokens of a file for each opening PHP
tag that is encountered, we keep a cache of the files that have been
checked, with lightning-fast lookup when we’re checking if a file has
already been processed.

Also removes a useless call to `ucfirst()` on a string that already has
an uppercase first letter.

### DIFF
--- a/WordPress/Sniffs/Files/FileNameSniff.php
+++ b/WordPress/Sniffs/Files/FileNameSniff.php
@@ -55,7 +55,7 @@ class WordPress_Sniffs_Files_FileNameSniff implements PHP_CodeSniffer_Sniff
                 $phpcsFile->addError($error, $stackPtr, 'UnderscoresNotAllowed');
         }
 
-		$phpcsFile->removeTokenListener( $this, $this->register() );
+        $phpcsFile->removeTokenListener( $this, $this->register() );
 
     }//end process()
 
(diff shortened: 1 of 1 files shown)


JDGrimes:
> Instead of searching through the tokens of a file for each opening PHP
> tag that is encountered, we keep a cache of the files that have been
> checked, with lightning-fast lookup when we’re checking if a file has
> already been processed.

You know, I just realized that there is a better way to do this. I think the sniff can call `$phpcsFile->removeTokenListener( $this, $this->register() );`, after it is finished checking the file name. Then it wont get called anymore for that file at all. That would probably improve performance even further.

I'll update the PR.

JDGrimes:
> I'll update the PR.

Done. #reviewmerge

GaryJones:
Indentation on line 58 looks off.

This sniff seems generic enough that it could be added to the Generic standard in PHPCS itself (comparing it against [this](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Generic/Sniffs/Files/LowercasedFilenameSniff.php).)? cc @gsherwood.

JDGrimes:
> Indentation on line 58 looks off.

Fixed. Caused by mixed tabs/spaces for indenting.

JDGrimes:
> This sniff seems generic enough that it could be added to the Generic standard in PHPCS itself (comparing it against this.)?

Looking at that sniff again, I just realized that there is [another (better?) way to skip the rest of the file](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Generic/Sniffs/Files/LowercasedFilenameSniff.php#L67-L68).

gsherwood:
> This sniff seems generic enough that it could be added to the Generic standard in PHPCS itself

Sure. If a PR is submitted, I'll take a look.

> Looking at that sniff again, I just realized that there is another (better?) way to skip the rest of the file.

Yes, that is the most efficient way to skip to the end of the file, or pause the sniff until any point within it.

PR #349, opened 2015-04-09
Yoda sniff: fix issues with return statements

JDGrimes:
With the updates to the Yoda sniff (#319), a bug was introduced that
would cause this code to be flagged:

```php
return 0 === strpos( $foo, 'a' );
```

This is fixed by adding `return` to the list of tokens that, when
encountered without being preceded by a variable, indicate a condition
either already implements, or does not require, Yodaism.

#reviewmerge


westonruter:
Nice work!

PR #243, opened 2014-09-23
Strict Comparisons Sniff and Unit Tests

mattyrob:
Linked to:
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/242

Issue #256, opened 2014-10-12
Settle on coding convention for WPCS snfifs


westonruter:
See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/commit/a446a38416b563393a248562a3c204b5c60ecfe5#commitcomment-8129539.

> But in reality this is the least of the problems in the coding style of the WPCS sniffs. It's currently a mismash of the conventions outlined in the PHPCS coding standards and the WPCS coding standards. There needs to be a comprehensive sweep done to fix them up (after all PRs are merged).

JDGrimes:
> WPCS will be problematic due to the camelCase methods of PHPCS. So I think we'll have to use the PHPCS standards.

I'd really much prefer if we use the WordPress standards, at least as far as whitespace goes. An added benefit of using WPCS itself, is that with an IDE like phpStorm you can see what is failing/succeeding when editing the `.inc` test files before you even run the tests.

> It will also make it consistent when looking at Generic / Squiz and WP sniffs.

Honestly, I don't think that'd be such a good thing :smiling_imp:. Seriously though, I think parity with WordPress's coding standards would be much easier on contributors, since that is what they'll be used to and probably expect.

GaryJones:
Just like PHPCS has a PHPCS Standard that itself follows, we can do the same with a WPCS standard, where we can mostly use the WordPress standard we're familiar with, but allow for camel case method names and other things needed as we extend PHPCS.

westonruter:
+1

JDGrimes:
We already allow for camelCase in subclasses anyway (#307).

So, are we saying, "`WordPress-Core` is our standard *going forward*", or do we want to clean up all of the sniffs now?

westonruter:
Either way, but for the cleanup of all existing sniffs we just wouldn't want to make any headaches for any open pull requests.

GaryJones:
To be able to check it properly, I think all the sniffs have to be cleaned up, but minimising the effect on open PRs. If we're waiting on improved PRs, give PR authors a nudge to get it improved in the next couple of weeks. If the PRs is ready, get it merged into 0.4.0, then let's make 0.5.0 the version where we meet our own standards.

FWIW,  we only have six open PRs - three for @JDGrimes , one for @westonruter  and two for @shadyvb so this is pretty much an internal nudge anyway.

JDGrimes:
[...]

PR #351, opened 2015-04-10
Strict Comparisons

mattyrob:
Sniffs and Unit Tests for Strict Comparisons so === not == and !== not
!=

Replaces ticker #243 that got scrubbed inadvertently

### DIFF
--- a/WordPress/Sniffs/PHP/StrictComparisonsSniff.php
+++ b/WordPress/Sniffs/PHP/StrictComparisonsSniff.php
@@ -40,7 +40,7 @@ class WordPress_Sniffs_PHP_StrictComparisonsSniff implements PHP_CodeSniffer_Sni
 	{
 		$tokens = $phpcsFile->getTokens();
 
-		if ( ! $this->has_whitelist_comment( 'loose comparison okay', $tokens, $stackPtr ) ) {
+		if ( ! $this->has_whitelist_comment( 'loose comparison, OK', $tokens, $stackPtr ) ) {
 			$error = 'Found: ' . $tokens[$stackPtr]['content'] . '. Use strict comparisons (=== or !==).';
 			$phpcsFile->addWarning($error, $stackPtr);
 		}
--- a/WordPress/Sniffs/PHP/StrictComparisonsSniff.php
+++ b/WordPress/Sniffs/PHP/StrictComparisonsSniff.php
@@ -40,14 +40,10 @@ class WordPress_Sniffs_PHP_StrictComparisonsSniff implements PHP_CodeSniffer_Sni
 	{
 		$tokens = $phpcsFile->getTokens();
 
-		if ( $tokens[$stackPtr]['code'] !== T_IS_EQUAL && $tokens[$stackPtr]['code'] !== T_IS_NOT_EQUAL) {
-        	return;
-    	} else {
-			if ( ! $this->has_whitelist_comment( 'loose comparison okay', $tokens, $stackPtr ) ) {
-        		$error = 'Found: ' . $tokens[$stackPtr]['content'] . '. Use strict comparisons (=== or !==).';
-				$phpcsFile->addWarning($error, $stackPtr);
-    		}
-    	}
+		if ( ! $this->has_whitelist_comment( 'loose comparison okay', $tokens, $stackPtr ) ) {
+			$error = 'Found: ' . $tokens[$stackPtr]['content'] . '. Use strict comparisons (=== or !==).';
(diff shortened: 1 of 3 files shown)


GaryJones:
Line 44, 45, 47 (and maybe others) have incorrect indentation. Part of this will be because we've not clarified which coding standards these sniffs should be following, but let's at least try to the choice of indentation character consistent within each file :-)

File's should not close with `?>` and should have a clear line at the end.

@JDGrimes can better advise, but I think we now have a `WordPress_Sniff` class you can inherit from, which already includes the method for checking for a whitelist comment. This class was added since your first PR.

Tests seem to use `Good`, but I think existing tests use `OK`?

#reviewmerge

JDGrimes:
> File's should not close with ?> 

30 of them do though (lack of standards again).



> I think we now have a WordPress_Sniff class you can inherit from, which already includes the method for checking for a whitelist comment.

Yes. If you extend `WordPress_Sniff` you can use its [`has_whitelist_comment()` method](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniff.php#L107). You just need to call `$this->init( $phpcsFile )` at the top of `process()`, and then call `has_whitelist_comment()` like so:

```php
$this->has_whitelist_comment( 'loose comparison', $stkPtr );
```

Note that it takes only the two arguments, and that you should omit the `okay` suffix from the comment.



> Tests seem to use Good, but I think existing tests use OK?

We break that rule 36 times. We need some standards for the standards, and some standard testing standards. :smile: 

---
@mattyrob If you don't want to tackle the changes, just say the word and I'll merge and make any modifications needed.

---
Note: There is a [`Squiz.Operators.ComparisonOperatorUsage` sniff](https://github.com/squizlabs/PHP_CodeSniffer/blob/5b63fd3fe960e2020770c845e04254e5ca2b722a/CodeSniffer/Standards/Squiz/Sniffs/Operators/ComparisonOperatorUsageSniff.php), but it is much more complex, because it disallows implicit boolean comparisons (`if ( $var )/if ( ! $var )`).

WordPress/Sniffs/PHP/StrictComparisonsSniff.php
```diff
@@ -0,0 +1,87 @@
...
+	 * @param int			$stackPtr  The position of the current token in the
+	 *						stack passed in $tokens.
+	 *
+	 * @return void
+	 */
+	public function process(PHP_CodeSniffer_File $phpcsFile, $stackPtr)
+	{
+		$tokens = $phpcsFile->getTokens();
+
+		if ( $tokens[$stackPtr]['code'] !== T_IS_EQUAL && $tokens[$stackPtr]['code'] !== T_IS_NOT_EQUAL) {
```
JDGrimes: I don't think that this condition is necessary. PHPCS will already be limiting the sniff to these tokens based on `register()`, won't it?

GaryJones:
> 30 of them do though (lack of standards again).

That's a WP and a PSR standard, and the one file in PHPCS I just checked also drops it, so there's no reason not to follow it too.

> We break that rule 36 times. We need some standards for the standards, and some standard testing standards.

Do we have an existing ticket for our own code standards? Thought we did, but it may just be comments on another ticket. Just like PHPCS has a PHPCS Standard that itself follows, we can do the same with a WPCS standard, where we can mostly use the WordPress standard we're familiar with, but allow for camel case method names and other things needed as we extend PHPCS.

JDGrimes:
> Do we have an existing ticket for our own code standards?

#256

mattyrob:
I've made some changes based on the suggestions above. Feel free to make further ones as you need.

GaryJones:
The tests are currently failing, as you've changed "loose comparison okay" whitelisting comment to "loose comparison, OK", without using the `WordPress_Sniff` defined class to define the whitelisting comment method.

mattyrob:
Fix for the synced

JDGrimes:
Great, I'll take it from there.

mattyrob:
@JDGrimes - thanks for that ;)

westonruter:
> The tests are currently failing, as you've changed "loose comparison okay" whitelisting comment to "loose comparison, OK", without using the WordPress_Sniff defined class to define the whitelisting comment method.

I don't think we need a whitelisting comment. The ambiguity should be resolved with code, for instance if `$value` is a string `'1'` then equality should be checked via:

```
if ( 1 === intval( $value ) ) {
    // ...
}
```

This eliminates all ambiguity.

PR #352, opened 2015-04-10
XSS sniff: Support for ternaries within escaping functions

JDGrimes:
In #281 I introduces support for the ternary operator in cases like
this:

```php
echo ( $var ) ? esc_html( $var ) : ‘something’;
```
But sometimes it makes more since to put the ternary condition inside
of the escaping function:

```php
echo esc_html( $var ? $var : $other_var );
```

Up to now, both `$var` and `$other_var` would be flagged as needing
escaping. This PR fixes that.

#reviewmerge


westonruter:
Good on _you_:exclamation:

Issue #33, opened 2013-04-08
Make the Escaped Output sniff smarter


westonruter:
The sniff needs more logic to determine safe output

```
echo 'Some Raw String';  //Causes no escaping function warning
_e( $some_nasty_var );  //No warning
echo apply_filters( 'the_content', $post->post_content );  //Causing no escaping function warning
```

Reported by @c3mdigital

PR #356, opened 2015-04-15
XSS sniff: Mark _e() and _ex() as unsafe printing functions

JDGrimes:
In #33, a list of localization functions like these were flagged as
needing escaping. In #315 I amended this list, removing non-printing
localization functions like `__()`, among other things. These changes
caused a regression in the treatment of the `_e()` and `_ex()`
functions, which remained on the list. They were previously flagged as
needing escaping, even when their parameters were “safe” values
(encapsed strings). That behavior was intentional, and was added
because these functions incorporate other values into their final
output, which need to be escaped before being printed. After #315,
however, these functions themselves were not flagged, only their
parameters when deemed unsafe (e.g., `_e( $var )`). This restores the
previous behavior, by classifying these as ‘unsafe printing functions’,
and flagging them as such when they are used.

For performance, I’ve opted to skip checking the parameters of these
functions when they are flagged, since the recommended replacements
will remedy any issues there as well. If the error for the function
itself is disabled though, any non-safe values incorporated into the
parameters will still be flagged.

### DIFF
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -249,6 +249,16 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff
 	);
 
 	/**
+	 * Printing functions that incorporate unsafe values.
+	 *
+	 * @var array
+	 */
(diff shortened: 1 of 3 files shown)


JDGrimes:
#reviewmerge

PR #371, opened 2015-05-01
Add in_array() to list of auto-escaped functions

JDGrimes:
Say you have code like this:

```php
if ( isset( $_REQUEST['foo'] ) && in_array( $_REQUEST['foo'], $array ) ) {
	$foo = sanitize_key( $_REQUEST['foo'] );
}
```
This is safe, but the use of the raw `$_REQUEST['foo']` value in `in_array()` would be flagged. Adding `in_array()` to the list of auto-escaped functions avoids this.


JDGrimes:
I've added `is_array()` as well:  83aa3fe

PR #373, opened 2015-05-01
Add check for lowercase "as" in foreach

GaryJones:
Copied from https://github.com/Yoast/yoastcs/blob/master/Yoast/ruleset.xml#L55-L61

The Squiz sniff actually checks whitespace for foreach loops as well as "as" being lowercase. We already [check whitespace](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/WhiteSpace/ControlStructureSpacingSniff.php#L66) for `foreach`, so this is very much duplicating a bunch of checks.

I'm happy for this PR to be rejected if we want to have a separate Sniff for checking case, though it may make more sense to switch to the Squiz sniff for all `foreach` checks, if it has enough properties for us to set what our standard should be.

Edit: Added to Core, but may be more suited as an Extra.

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -38,6 +38,14 @@
 	<rule ref="Squiz.Strings.DoubleQuoteUsage.ContainsVar">
 		<severity>0</severity>
 	</rule>
+	
+	<!-- Added for lowercase "as" keyword check -->
+	<rule ref="Squiz.ControlStructures.ForEachLoopDeclaration">
+		<properties>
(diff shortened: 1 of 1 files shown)


JDGrimes:
I think this is appropriate for Core, since `true`, `false`, etc., are expected to be lowercase.

JDGrimes:
As far a duplication goes, we probably need to go through our sniffs an check that we can't use upstream sniffs for some of them. And maybe some of them should be ported upstream as well.

PR #376, opened 2015-05-02
 Santized sniff: don't flag comparisons 

### LINKED ISSUE #186: Input var validation sniff flags vars even when value isn't being assigned
Say you have this code:

```php
if ( isset( $_POST['some_var'] ) && '-1' === $_POST['some_var'] ) {
     // do stuff
}
```

The second reference to the variable will be flagged as not sanitized, even though its value is not being assigned to another variable, or passed to a function.

JDGrimes:
Comparison of a super global to another value (e.g. `if ( 'a' === $_POST['foo'] )` or `switch ( $_POST['foo'] )`) doesn’t require sanitization. Previously it was flagged anyway, but this fixes that,
and introduces a new method to check for comparisons.

Fixes #186

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -623,10 +623,10 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 		// Find the previous non-empty token. We check before the var first because
 		// yoda conditions are usually expected.
 		$previous_token = $this->phpcsFile->findPrevious(
-			PHP_CodeSniffer_Tokens::$emptyTokens
-			, $stackPtr - 1
-			, null
-			, true
+			PHP_CodeSniffer_Tokens::$emptyTokens,
+			$stackPtr - 1,
+			null,
+			true
 		);
 
 		if ( in_array( $this->tokens[ $previous_token ]['code'], PHP_CodeSniffer_Tokens::$comparisonTokens ) ) {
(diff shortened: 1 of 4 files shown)


WordPress/Sniff.php
```diff
@@ -446,6 +446,218 @@ protected function is_sanitized( $stackPtr ) {
...
+	 * @return string|false The array index key whose value is being accessed.
+	 */
+	protected function get_array_access_key( $stackPtr ) {
+
+		// Find the next non-empty token.
+		$open_bracket = $this->phpcsFile->findNext(
+			PHP_CodeSniffer_Tokens::$emptyTokens
+			, $stackPtr + 1
+			, null
+			, true
```
westonruter: Minor point, but putting commas in front isn't a pattern found in Core, at least as far as I have seen.

JDGrimes: Yeah, it's a personal preference that's become a habit. I've found that it really does reduce confusion between function calls and arrays (since arrays will always have the commas in back). I can (try to) refrain from it if you'd like. :smile:

GaryJones: I have to agree with @westonruter on this one. 
Save the front-placed commas for SQL please :-)

JDGrimes: I've moved the commas to  the back in 1521b3f. I guess this can be merged.

westonruter: I think you neglected the commas here in the `get_array_access_key` method :-)

JDGrimes: Argh! I can't believe I missed those. Fixed now.

PR #378, opened 2015-05-03
Add support for error fixing to ArrayDeclaration

JDGrimes:
This basically synchronizes it with `Squiz.Arrays.ArrayDeclaration`.
The upstream version is similar, except that we exclude a few errors.
Unfortunately we have to actually comment out the code rather than just
using the upstream sniff and `<exclude>` in our ruleset, due to a bug
(squizlabs/PHP_CodeSniffer#582). (I’ve also included a fix for another
bug, squizlabs/PHP_CodeSniffer#584.) Because of this, we cannot yet
eliminate duplicated logic from this child sniff.

I think this pretty much maintains the previous behavior, except that I
have added a warning for extra whitespace at the edges of multiline
arrays.

Another note, the whitespace messages are warnings:

```php
array(1, 2); // missing spaces
array(      1, 2        ) // extra spaces
```
Maybe these should be converted to errors? I'm not sure why they are just warnings (it goes all the way back to Urban Giraffe).

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -66,7 +66,9 @@
 		<severity>0</severity>
 	</rule>
 
-	<rule ref="WordPress.Arrays.ArrayDeclaration"/>
+	<rule ref="WordPress.Arrays.ArrayDeclaration">
+		<exclude name="WordPress.Arrays.ArrayDeclaration.SingleLineNotAllowed" />
+	</rule>
 	<rule ref="WordPress.Arrays.ArrayKeySpacingRestrictions"/>
 	<rule ref="WordPress.Classes.ValidClassName"/>
 	<rule ref="WordPress.Files.FileName"/>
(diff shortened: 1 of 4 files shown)


JDGrimes:
Note: I don't think this will be compatible withe PHPCS < 2.0. Are we trying to maintain compatibility with 1.5? There is this [note in the reademe](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards#standalone):

>Do ensure, if for example you're using VVV, that you have the latest version of CodeSniffer (earlier versions, e.g. ~1.5.5, may warn about incorrect line indentation on every single line even if your code is actually correct.)

[`composer.json`](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/composer.json#L18) says that we require `~2.0`. Maybe we should make this note more explicit about that.

westonruter:
@JDGrimes I don't think we should be trying to maintain compatibility with PHPCS 1.5, no. That readme note is just an old one. There's no need to incur the burden of supporting older versions of PHPCS. Maybe the base sniff class could do a check for the PHPCS version and error-out if it detects that an older version is being used.

GaryJones:
> I don't think we should be trying to maintain compatibility with PHPCS 1.5, no.

+1.

I know Yoast has currently tied itself to PHPCS 2.2.0 (latest tag is 2.3.2) because one of the tools they use hasn't been updated yet, but there's no need to support PHPCS lower than that.

JDGrimes:
I've converted the warnings to errors in e2c5160. This should be ready to merge.

Issue #241, opened 2014-09-20
Test documentation standards


ScreamingDev:
Please add tests for documentation standards:
https://make.wordpress.org/core/handbook/inline-documentation-standards/php-documentation-standards/

westonruter:
See @Rarst's tweet in regards to incorporating the `Squiz.Commenting` sniffs: https://twitter.com/rarst/status/567237335495221248

GaryJones:
I'm going to change my mind again. I think a WordPress-Docs ruleset is best, but include that in the overall WordPress ruleset, so that it's still on by default.

GaryJones:
I've started https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/compare/temp/241-commenting?expand=1

Lots of exclusions from the default Squiz.Commenting, and not looked to see if Generic might be better.

Plenty of cases where we're going to need to write custom sniffs, if there aren't parameters to tweak what we need.

I've created a [quick test file](https://gist.github.com/GaryJones/2ae4e95d38c3de08562e) and asked the Docs channel on Slack to see if this is right so we can check the perfect case doesn't generate false positives.

GaryJones:
I realise that it would be better to whitelist which checks we want, instead of excluding, so that we're not at the mercy of PHPCS introducing a check that isn't relevant. I'll change this.

JDGrimes:
> I'd like to whitelist which messages are reported, but this isn't possible

See https://github.com/squizlabs/PHP_CodeSniffer/issues/586.

JDGrimes:
@GaryJones I think this can be closed now that `WordPress-Docs` ruleset was added in #381?

GaryJones:
No @JDGrimes - that was the first pass / easy wins. There are still plenty of aspects of comments / inline docs that are not yet tested, so I'd like to keep this open.

Alternatively, it can be closed, and a new one opened if that makes sense for the release paper trail.

PDTechService:
Just wanted to comment, as it resulted in us changing how we used the WordPress coding standards.

According to here:  https://make.wordpress.org/core/handbook/inline-documentation-standards/php-documentation-standards/

Wordpress makes an attempt to adopt PSR-5 standards, but they are not intending it to be PSR-5 complaint.  Further, the language of the drafted standards here use SHOULD and RECOMMENDED throughout, but this changes the intent to almost a MUST.

[...]

PR #381, opened 2015-05-08
Add Squiz and Generic Commenting with exclusions

GaryJones:
There are still plenty of checks not made, but this first pass takes advantage of the existing code sniffs that match the WordPress handbook for inline documentation and commenting.

The exclusions are for specific checks that don't match that expected in WordPress-related code.

See #241

### DIFF
--- a/WordPress-Docs/ruleset.xml
+++ b/WordPress-Docs/ruleset.xml
@@ -95,8 +95,8 @@
 		<exclude name="Generic.Commenting.DocComment.ContentBeforeClose"/>
 
 		<!-- WP allows @todo's in comments -->
-	    <exclude name="Generic.Commenting.Todo.CommentFound"/>
+		<exclude name="Generic.Commenting.Todo.CommentFound"/>
 		<!-- WP allows @todo's in comments -->
-	    <exclude name="Generic.Commenting.Todo.TaskFound"/>
+		<exclude name="Generic.Commenting.Todo.TaskFound"/>
 	</rule>
 </ruleset>
--- a/WordPress-Docs/ruleset.xml
+++ b/WordPress-Docs/ruleset.xml
@@ -1,0 +1,102 @@
+<?xml version="1.0"?>
+<ruleset name="WordPress Docs">
+	<description>WordPress Coding Standards for Inline Documentation and Comments</description>
+
+	<rule ref="Squiz.Commenting">
+		<!-- Excluded to allow /* translators: ... */ comments -->
+		<exclude name="Squiz.Commenting.BlockComment.SingleLine"/>
+		<!-- Sniff seems to require indenting with spaces -->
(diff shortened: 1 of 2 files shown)


WordPress-Docs/ruleset.xml
```diff
@@ -0,0 +1,102 @@
...
+		<exclude name="Generic.Commenting.DocComment.ContentAfterOpen"/>
+		<!-- Exclude to allow duplicate hooks to be documented -->
+		<exclude name="Generic.Commenting.DocComment.SpacingBeforeShort"/>
+		<!-- Exclude to allow duplicate hooks to be documented -->
+		<exclude name="Generic.Commenting.DocComment.ContentBeforeClose"/>
+
+		<!-- WP allows @todo's in comments -->
+	    <exclude name="Generic.Commenting.Todo.CommentFound"/>
+		<!-- WP allows @todo's in comments -->
+	    <exclude name="Generic.Commenting.Todo.TaskFound"/>
```
JDGrimes: Mixed spaces/tabs indenting this line.

GaryJones: Fixed.

JDGrimes:
I've run this over a project of mine, and I have several suggestions.

---
Could we exclude `Commenting.FileComment.SpacingAfterOpen`? It prevents a space coming before the docblock for the file, like this:

```php
<?php

/**
 * File Docblock.
 */
```

I don't think there is anything on that in [the standard](https://make.wordpress.org/core/handbook/inline-documentation-standards/php-documentation-standards/#6-file-headers).

---

I think that `Squiz.Commenting.BlockComment.HasEmptyLineBefore` should be excluded, because it prevents a newline at the start of a block comment. However, the [example in the standard](https://make.wordpress.org/core/handbook/inline-documentation-standards/php-documentation-standards/#5-2-multi-line-comments) starts with a newline:

```php
/* 
 * This is a comment that is long enough to warrant being stretched over
 * the span of multiple lines. You'll notice this follows basically
 * the same format as the PHPDoc wrapping and comment block style.
 */
```

GaryJones:
> I don't think there is anything on that in the standard.

Yet another hole in the standard - it doesn't say _to_ include a space there either, so in this case, we look at the files. Of the non-3rd-party files I checked, all of them have no space between the `<?php` and file header. In this case, do as they do, not as they say.



> I think that Squiz.Commenting.BlockComment.HasEmptyLineBefore should be excluded

This [check](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Sniffs/Commenting/BlockCommentSniff.php#L289) is about a blank line before (or after) the whole comment block, not on the first line (`/*`) which is handled via a different check. It may be that some of the checks do need to be excluded, but just wanted to highlight your example seems different to what I believe that check is looking for.

JDGrimes:
@GaryJones you are correct.

PR #382, opened 2015-05-08
Add Core and Extra sniffs

GaryJones:
Adds a few Generic sniffs for Core, and a whole bunch more for Extra. See the diff for which sniffs.

### DIFF
--- a/WordPress-Extra/ruleset.xml
+++ b/WordPress-Extra/ruleset.xml
@@ -6,17 +6,23 @@
 	<rule ref="Generic.PHP.ForbiddenFunctions"/>
 	<rule ref="Generic.Functions.CallTimePassByReference"/>
 	<rule ref="Generic.Formatting.DisallowMultipleStatements"/>
-	<rule ref="Generic.Formatting.MultipleStatementAlignment"/>
 	<rule ref="Generic.CodeAnalysis.EmptyStatement" />
 	<rule ref="Generic.CodeAnalysis.ForLoopShouldBeWhileLoop"/>
 	<rule ref="Generic.CodeAnalysis.ForLoopWithTestFunctionCall"/>
 	<rule ref="Generic.CodeAnalysis.JumbledIncrementer"/>
 	<rule ref="Generic.CodeAnalysis.UnconditionalIfStatement"/>
 	<rule ref="Generic.CodeAnalysis.UnnecessaryFinalModifier"/>
-	<rule ref="Generic.CodeAnalysis.UnusedFunctionParameter"/>
 	<rule ref="Generic.CodeAnalysis.UselessOverridingMethod"/>
 	<rule ref="Generic.Classes.DuplicateClassName"/>
 	<rule ref="Generic.Strings.UnnecessaryStringConcat"/>
+	
+	<!-- This sniff is not refined enough for general use -->
+	<!-- https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/382#discussion_r29970107 -->
+	<!--<rule ref="Generic.Formatting.MultipleStatementAlignment"/>-->
+	
+	<!-- Hook callbacks may not use all params -->
+	<!-- https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/382#discussion_r29981655 -->
+	<!--<rule ref="Generic.CodeAnalysis.UnusedFunctionParameter"/>-->
 
 	<rule ref="WordPress-Core"/>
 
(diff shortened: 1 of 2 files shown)


WordPress-Extra/ruleset.xml
```diff
@@ -2,6 +2,22 @@
 <ruleset name="WordPress Extra">
 	<description>Best practices beyond core WordPress Coding Standards</description>
 
+	<rule ref="Generic.PHP.DeprecatedFunctions"/>
+	<rule ref="Generic.PHP.ForbiddenFunctions"/>
+	<rule ref="Generic.Functions.CallTimePassByReference"/>
+	<rule ref="Generic.Formatting.DisallowMultipleStatements"/>
+	<rule ref="Generic.Formatting.MultipleStatementAlignment"/>
```
westonruter: I really don't think we should include `MultipleStatementAlignment`. It's really annoying, for instance:

class-https-resource-proxy_php_-_wordpress-develop_-____projects_vvv_www_wordpress-develop_

JDGrimes: I agree, `MultipleStatementAlignment` is a good practice within reason, but the sniff isn't refined enough.

WordPress-Extra/ruleset.xml
```diff
@@ -2,6 +2,22 @@
...
+	<rule ref="Generic.CodeAnalysis.EmptyStatement" />
+	<rule ref="Generic.CodeAnalysis.ForLoopShouldBeWhileLoop"/>
+	<rule ref="Generic.CodeAnalysis.ForLoopWithTestFunctionCall"/>
+	<rule ref="Generic.CodeAnalysis.JumbledIncrementer"/>
+	<rule ref="Generic.CodeAnalysis.UnconditionalIfStatement"/>
+	<rule ref="Generic.CodeAnalysis.UnnecessaryFinalModifier"/>
+	<rule ref="Generic.CodeAnalysis.UnusedFunctionParameter"/>
```
JDGrimes: Because of the way WordPress's filters/actions API works, `UnusedFunctionParameter` will return a lot of false positives. I often need to hook into a filter and need the 1st and 5th parameters, for example, but you still have to specify the 2nd-4th parameters (or use `func_get_arg()`, which is slow).

WordPress-Extra/ruleset.xml
```diff
@@ -2,6 +2,22 @@
...
+	<rule ref="Generic.CodeAnalysis.UselessOverridingMethod"/>
+	<rule ref="Generic.Classes.DuplicateClassName"/>
+	<rule ref="Generic.Strings.UnnecessaryStringConcat"/>
```
JDGrimes: I will probably disable `UnnecessaryStringConcat` in my custom ruleset if it gets added, because sometimes I use concatenation to wrap a long string (which that sniff will report as an error).

WordPress-Extra/ruleset.xml
GaryJones: In this case, I'd optionally put `$redirect_vars_pairs` above line 129, and align the `=` out.

In other cases, just adding a clear line above 134 would do - that the separates the predefined data in `$redirect_vars`, and the variable that's only populated from the `foreach`.

It is [part of the standard](https://make.wordpress.org/core/handbook/coding-standards/php/#indentation).

However, I'm open to removing it if you both still insist.

WordPress-Extra/ruleset.xml
GaryJones: OK, I'll remove it. Do we want a custom sniff that does a similar check, but doesn't report unused (e.g. 2nd-4th) variables if a later one (e.g. 5th) is used?

WordPress-Extra/ruleset.xml
westonruter: > It is part of the standard.

Kinda. It says “spaces can be used mid-line for alignment”. It doesn't say that that alignment is required.

westonruter: We should probably have in the `README` a list of other sniffs that users may like to opt-in to, such as this, but which are perhaps even more opinionated than Extra already is.

WordPress-Extra/ruleset.xml
westonruter: My standard practice has been to just `unset( $unused_var )`, and this makes PhpStorm happy, and I assume it would make this sniff happy as well. Having a sniff that skips warning if the last argument is used but the intermediate ones are not used, is a great idea. This seems like it should be a core property on `Generic.CodeAnalysis.UnusedFunctionParameter`.

GaryJones:
Updated.

WordPress-Extra/ruleset.xml
jrfnl: I've opened an upstream PR which addresses this: https://github.com/squizlabs/PHP_CodeSniffer/pull/1318

Issue #189, opened 2014-06-27
False positive - XSS.EscapeOutput


jrfnl:
The following snippet triggers an error while IMHO it shouldn't:

``` php
echo str_repeat( "\t", 2 );
```

PR #383, opened 2015-05-08
Separate sanitizing and escaping functions

### LINKED ISSUE #343: Correct terminology: "sanitization" vs "escaping"
@GaryJones has [pointed out before that the XSS sniff uses the wrong term](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/167#issue-32151032) for preparation of untrusted data for output. Instead of expecting an "escaping" function, it expects a "sanitizing" function:

> Expected a sanitizing function (see Codex for 'Data Validation'), but instead saw '$foo'

*Input* is sanitized, *output* is escaped.

This should be corrected in the error message. Unfortunately, it also affects the `sanitizingFunctions` and `customSanitizingFunctions` properties. Ideally we'd rename this to `escapingFunctions` and `customEscapingFunctions` respectively. 

We may also want to review the `escapingFunctions` list, and possibly split the sanitization functions from it. I propose that these lists then be moved to `WordPress_Sniff`, where they can be maintained in a single location and are accessible to all the sniffs.

Changing the name of the `customSanitizingFunctions` property isn't backward compatible though, because that property is used for configuration (e.g., via a custom XML config file). However, we can introduce the `customEscapingFunctions` property and just deprecate `customSanitizingFunctions`, while still providing support for it until it is eventually removed.

JDGrimes:
Historically there has a been a list of “sanitizing” functions in the
XSS sniff. This list could be added to via the
`customSanitizingFunctions` property. This list of functions was also
used by the `ValidatedSanitizedInput` sniff.

However, what the XSS sniff is really checking for is that the output
is *escaped*. This is slightly different from *sanitization*, and is
(usually) performed by different functions.

In this commit, I’ve split the `sanitizingFunctions` list in the XSS
sniff, and moved it to the main sniff. This will allow the list to be
accessed from all sniffs more easily. I’ve also flipped the list so
that the functions are the keys instead of values, because this allows
for quicker lookup with `isset()` instead of having to use
`in_array()`. The values currently have no meaning, though it’s
possible we’ll use them for something in the future. (I’ve also done
this for the other function lists from the XSS sniff, and the list of
nonce verification functions.)

I’ve split the list of “sanitizing” functions into the
`sanitizingFunctions` and `escapingFunctions` properties. A few
functions are found on both lists, but for the most part there is
pretty clear separation. We will tweak this in the future as needed.

I’ve completely removed the `balanceTags()` function from both lists,
because it doesn’t actually strip scripts or anything like that. There
may be a case for adding it back to the sanitizing functions list, and
you’re welcome to make it.

I’ve also moved `ent2ncr()` to the list of formatting functions, since
it doesn't actually escape things, just format them. (It was added in
#189.)

The `customSanitizingFunctions` property of the XSS sniff has been
deprecated, and a `customEscapingFunctions` property has been added in
its place. When the deprecated property is used, a warning will be
given.

I’ve also added a `customSanitizingFunctions` property to the
`ValidatedSanitizedInput` sniff. This is where custom sanitizing
functions should be added now. Adding them to either of the properties
of the XSS sniff won’t have the desired effect.

Finally, there is one issue yet to be resolved, so the tests are
intentionally failing. There is currently no provision made for code
like this, which takes user input and directly escapes it for display:

```php
echo esc_html( $_POST[‘foo’] );
```

With a little work, we could make this permissible. But perhaps there
is a case to be made that sanitization should still be applied even in
a case like this. Please let your opinion be known.

Fixes #343

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -29,9 +29,292 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	 * @var array
 	 */
 	public static $nonceVerificationFunctions = array(
-		'wp_verify_nonce',
-		'check_admin_referer',
-		'check_ajax_referer',
+		'wp_verify_nonce' => true,
+		'check_admin_referer' => true,
+		'check_ajax_referer' => true,
+	);
+
+	/**
+	 * Functions that escape values for display.
+	 *
+	 * @since 0.5.0
+	 *
+	 * @var array
+	 */
+	public static $escapingFunctions = array(
+		'absint' => true,
+		'esc_attr__' => true,
+		'esc_attr_e' => true,
+		'esc_attr_x' => true,
+		'esc_attr' => true,
+		'esc_html__' => true,
+		'esc_html_e' => true,
+		'esc_html_x' => true,
+		'esc_html' => true,
+		'esc_js' => true,
+		'esc_sql' => true,
+		'esc_textarea' => true,
+		'esc_url_raw' => true,
+		'esc_url' => true,
+		'filter_input' => true,
+		'filter_var' => true,
+		'intval' => true,
+		'json_encode' => true,
+		'like_escape' => true,
+		'number_format' => true,
+		'rawurlencode' => true,
+		'sanitize_html_class' => true,
+		'sanitize_user_field' => true,
+		'tag_escape' => true,
+		'urlencode_deep' => true,
+		'urlencode' => true,
+		'wp_json_encode' => true,
+		'wp_kses_allowed_html' => true,
+		'wp_kses_data' => true,
+		'wp_kses_post' => true,
+		'wp_kses' => true,
+	);
+
+	/**
+	 * Functions whose output is automatically escaped for display.
+	 *
+	 * @since 0.5.0
+	 *
+	 * @var array
+	 */
+	public static $autoEscapedFunctions = array(
+		'allowed_tags' => true,
+		'bloginfo' => true,
+		'body_class' => true,
+		'calendar_week_mod' => true,
+		'cancel_comment_reply_link' => true,
+		'category_description' => true,
+		'checked' => true,
+		'comment_author_email_link' => true,
+		'comment_author_email' => true,
+		'comment_author_IP' => true,
+		'comment_author_link' => true,
+		'comment_author_rss' => true,
+		'comment_author_url_link' => true,
+		'comment_author_url' => true,
+		'comment_author' => true,
+		'comment_class' => true,
+		'comment_date' => true,
+		'comment_excerpt' => true,
+		'comment_form_title' => true,
+		'comment_form' => true,
+		'comment_id_fields' => true,
+		'comment_ID' => true,
+		'comment_reply_link' => true,
+		'comment_text_rss' => true,
+		'comment_text' => true,
+		'comment_time' => true,
+		'comment_type' => true,
+		'comments_link' => true,
+		'comments_number' => true,
+		'comments_popup_link' => true,
+		'comments_popup_script' => true,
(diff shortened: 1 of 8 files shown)


JDGrimes:
And I guess I forgot to mention that I also fixed the error message in the XSS sniff to use the word "escaping" rather than "sanitization". While I was at it I did a slight refactoring so that there is now only a single error message, rather than two.

westonruter:
There's a [unit test error](https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/61826560#L190):

```
PHPUnit 4.5.0 by Sebastian Bergmann and contributors.
........................F.......
Tests generated 57 unique error codes; 11 were fixable (19.3%)
Time: 1.28 seconds, Memory: 51.50Mb
There was 1 failure:
1) WordPress_Tests_VIP_ValidatedSanitizedInputUnitTest::testSniff
[LINE 7] Expected 1 error(s) in ValidatedSanitizedInputUnitTest.inc but found 2 error(s). The error(s) found were:
 -> Detected usage of a non-validated input variable: $_POST (WordPress.VIP.ValidatedSanitizedInput.InputNotValidated)
 -> Detected usage of a non-sanitized input variable: $_POST (WordPress.VIP.ValidatedSanitizedInput.InputNotSanitized)
[LINE 51] Expected 0 error(s) in ValidatedSanitizedInputUnitTest.inc but found 1 error(s). The error(s) found were:
 -> Detected usage of a non-sanitized input variable: $_GET (WordPress.VIP.ValidatedSanitizedInput.InputNotSanitized)
[LINE 72] Expected 0 error(s) in ValidatedSanitizedInputUnitTest.inc but found 1 error(s). The error(s) found were:
 -> Detected usage of a non-sanitized input variable: $_POST (WordPress.VIP.ValidatedSanitizedInput.InputNotSanitized)
/tmp/phpcs/tests/Standards/AbstractSniffUnitTest.php:172
/tmp/phpcs/tests/TestSuite.php:48
```

JDGrimes:
> There's a unit test error:

Yes, that's intentional:



> Finally, there is one issue yet to be resolved, so the tests are
> intentionally failing. There is currently no provision made for code
> like this, which takes user input and directly escapes it for display:
> 
> ```php
> echo esc_html( $_POST[‘foo’] );
> ```
>
> With a little work, we could make this permissible. But perhaps there
> is a case to be made that sanitization should still be applied even in
> a case like this. Please let your opinion be known.

westonruter:
@JDGrimes Right, so the non-validated issue would be addressed adding an `isset` check:

```php
if ( isset( $_POST['foo'] ) ) {
    echo esc_html( $_POST['foo'] ); // Note: this should technically include wp_unslash()
}
```

Bit then there would be `non-sanitized input variable` errors because `sanitize_text_field()` or another sanitizing function isn't being used. Why would we want the sniff to allow direct outputting of an input var to output? It seems sanitization should be used here. If sanitization isn't desired, then a whitelist comment could be used.

JDGrimes:
> Why would we want the sniff to allow direct outputting of an input var to output? It seems sanitization should be used here. If sanitization isn't desired, then a whitelist comment could be used.

It would only be allowed in cases where the value was being escaped before output. In other words:

```php
echo $_POST['var']; // NOT allowed.
echo esc_html( $_POST['var'] ); // OK.
```

This has been the behavior of the sniffs up to now. The question is whether this PR should change the second line to be not allowed either.

westonruter:
```php
echo esc_html( $_POST['var'] );
```

I don't think this should be allowed by the sniff, no.

JDGrimes:
> echo esc_html( $_POST['var'] );
> 
> I don't think this should be allowed by the sniff, no.

Great, I've fixed the tests to reflect that. This should be ready for merge.

PR #391, opened 2015-05-20
Allow camelCase methods in classes that implement others

westonruter:


WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
```diff
@@ -133,7 +133,50 @@ protected function processTokenWithinScope(PHP_CodeSniffer_File $phpcsFile, $sta
...
+	 * @param PHP_CodeSniffer_File $phpcsFile The stack position of the class.
+	 *
+	 * @see PEAR_Sniffs_NamingConventions_ValidFunctionNameSniff::findExtendedClassName()
+	 *
+	 * @todo This needs to be upstreamed and made part of PHP_CodeSniffer_File.
+	 *
+	 * @return string
+	 */
+	public function findImplementedClassName( $stackPtr, $phpcsFile ) {
+        $tokens = $phpcsFile->getTokens();
```
JDGrimes: It looks like this line is indented with spaces instead of tabs.

WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
```diff
@@ -133,7 +133,50 @@ protected function processTokenWithinScope(PHP_CodeSniffer_File $phpcsFile, $sta
...
+		if ( isset( $tokens[ $stackPtr ] ) === false ) {
+			return false;
+		}
+		if ( $tokens[ $stackPtr ]['code'] !== T_CLASS ) {
+			return false;
+		}
+		if ( isset( $tokens[ $stackPtr ]['scope_closer'] ) === false ) {
+			return false;
+		}
+		$classCloserIndex = $tokens[ $stackPtr ]['scope_closer'];
```
JDGrimes: I know this is just following the original function, but I'm not sure why `scope_closer` is used here instead of `scope_opener`. If `scope_opener` refers to the position of the opening brace, it should be used instead. Otherwise, when there is no interface being implemented, the code searches through the whole class looking for the `T_IMPLEMENTS`.

WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
```diff
@@ -133,7 +133,50 @@ protected function processTokenWithinScope(PHP_CodeSniffer_File $phpcsFile, $sta
...
+	 * @param int $stackPtr The stack position of the class.
+	 * @param PHP_CodeSniffer_File $phpcsFile The stack position of the class.
+	 *
+	 * @see PEAR_Sniffs_NamingConventions_ValidFunctionNameSniff::findExtendedClassName()
+	 *
+	 * @todo This needs to be upstreamed and made part of PHP_CodeSniffer_File.
+	 *
+	 * @return string
+	 */
+	public function findImplementedClassName( $stackPtr, $phpcsFile ) {
```
JDGrimes: I'd prefer the function name `findImplementedInterfaceName()`. Also, this could be moved to the base `WordPress_Sniff` class so it will be available to our other sniffs in the future.

WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
westonruter: @JDGrimes it's because it is being passed as the third argument (`$end`) to `PHP_CodeSniffer_File::findNext( $types, $start, $end, $exclude, $value, $local )`, meaning it should stop looking once it gets to that point. Right?

WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
westonruter: I renamed it in 0c1a73e.

> Also, this could be moved to the base `WordPress_Sniff` class so it will be available to our other sniffs in the future.

Actually, the inclusion here is just a stop gap until we can get the method added directly to `PHP_CodeSniffer_File` upstream.

JDGrimes:
I've committed  1c95ca7, which changes to using `scope_opener`. This change should also be made upstream to the `PHP_CodeSniffer_File::findExtendedClassName()` method.

I've also realized that we're making the code do a bit of extra work to actually get the class/interface names. In this case we just need to check if the `extends` or `implements` token exists. Maybe that part of the method could be factored out into a separate method for this purpose?

westonruter:
> I've committed 1c95ca7, which changes to using scope_opener. This change should also be made upstream to the PHP_CodeSniffer_File::findExtendedClassName() method.

Are you sure this is right? Per my comment above:

> it's because it is being passed as the third argument (`$end`) to `PHP_CodeSniffer_File::findNext( $types, $start, $end, $exclude, $value, $local )`, meaning it should stop looking once it gets to that point. Right?

JDGrimes:
> Are you sure this is right?

Yes. I think the reason it was originally done using `scope_closer` is because if you just substitute `scope_opener` it doesn't work. This is because of this line:

```php
$end  = $phpcsFile->findNext( $find, ( $extendsIndex + 1 ), $classOpenerIndex, true );
```

When the `scope_opener` is used, this will mysteriously return false. I'm assuming that rather than figure out why this was happening, the original author just decided to use `scope_closer`. However, I was able to figure out the reason this was ending up as `false`: no non-matching tokens would be found because the `$end` parameter is non-inclusive. Changing it to this works as expected:

```php
$end  = $phpcsFile->findNext( $find, ( $extendsIndex + 1 ), $classOpenerIndex + 1, true );
```

Note that 1 is added to `$classOpenerIndex`.

This line isn't actually the one where the performance gain is though, it is the other use of `$phpcsFIle->findNext()` above it.

Issue #172, opened 2014-05-20
Enforce wp_slash() on functions that expect slashed data


westonruter:
Since WordPress forces global input vars to get magic quoted, any access to them should pass through `wp_unslash()`. Likewise, any data data sent into a function that expects pre-slashed input should require an explicit `wp_slash()`.

For example, `wp_unslash()` and `wp_slash()` could be enforced in situations like this:

``` php
$title = sanitize_text_field( wp_unslash( $_POST['title'] ) );
// ...
wp_insert_post( wp_slash( array(
    'post_title' => $title,
) ) );
```

As a `WordPress-Extra` rule, this will help enforce a discipline of unslashing, sanitizing, and slashing when slashing is required (e.g. in `wp_update_post()`, `update_post_meta()`, etc). It's easy to forget and for slashing to sneak in or to get stripped out, `\\o/` `o/`, yay.

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/395 implements the sniff for `wp_unslash()`

GaryJones:
> Since WordPress forces global input vars to get magic quoted, any access to them should pass through `wp_unslash()`

I remember reading some of the discussions, but is there any documentation on the Codex or otherwise about this?

JDGrimes:
Original ticket: https://core.trac.wordpress.org/ticket/21767
I couldn't find anything about it on the codex or developer handbook.

JDGrimes:
> I'm tempted to sniff for wp_unslash() when accessing an input var, and then checking for wp_slash() when passing a variable into update_post_meta() and other functions that (ridiculously) expect pre-slashed data.

Yeah, we'll add a separate sniff that checks that the data is slashed for all functions that expect slashed data.

johnbillion:
Related: https://core.trac.wordpress.org/ticket/41593

JDGrimes:
Note that I still have some old code for a sniff here, that I just haven't finished yet. The sniff itself can be found in this gist: https://gist.github.com/JDGrimes/7a29ec88d533459345565ae3caabe7d2

It needs to be updated before it can actually be added to WPCS, but it does contain some research in regard to functions that require slashing and ones that don't. 

The list is incomplete, because basically what I've been doing is finding functions that require slashing, adding them to the list, and then running the sniff over core again. This reveals other functions that use those functions, and that therefore also require data passed to them to be slashed.

[...]

PR #395, opened 2015-05-30
Require unlashing of input superglobals

JDGrimes:
This updates the validated/sanitized input sniff to also check for
slashing. This could have been made into another sniff instead,
however, it would have required lots of duplicated logic and this sniff
would need to be updated to accommodate the use of `wp_unslash()`
anyway.

Currently only `wp_unslash()` is recognized as an unlashing function,
but this can be changed in the future if needed.

The sniff currently requires that `wp_unslash()` be used *before* the
data is passed through a sanitizing function. Sanitizing first and then
wrapping that in `wp_unslash()` is not accepted.

The error for missing the use of `wp_unslash()` is independent of the
missing sanitizing function error, so an error will be given for
missing use of an unslashing function whether or not a sanitizing
function is used, and vice versa.

Unslashing is not required when sanitization is provided via casting,
or when certain sanitization functions are used which implicitly or
explicitly perform an unslash or for which unslashing isn’t necessary.
`absint()` implicitly unslashes, and `sanitize_key()` will remove
slashes explicitly. And unslashing isn’t necessary when testing a value
with `is_array()`. This list can be expanded in the future, and is
configurable via the `customUnslashingSanitizingFunctions` property.

See #172

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -272,6 +272,19 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	);
 
 	/**
+	 * Sanitizing functions that implicitly unslash the data passed to them.
+	 *
+	 * @since 0.5.0
+	 *
+	 * @var array
+	 */
+	public static $unslashingSanitizingFunctions = array(
+		'absint' => true,
+		'is_array' => true,
+		'sanitize_key' => true,
+	);
+
+	/**
 	 * Functions that format strings.
 	 *
 	 * These functions are often used for formatting values just before output, and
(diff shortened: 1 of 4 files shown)


GaryJones:
I don't see any `$_POST` or `$_GET` values being assigned with explicitly slashes values / values that would become slashed, and checked in tests? Not necessarily to unit test the functionality of `wp_unslash()` but to integration test with different data for our existing tests.

JDGrimes:
@GaryJones I'm not sure I follow. There are some places where `wp_unslash()` is not used, to test that the error is given, if that's what you mean.

GaryJones:
Nevermind - think I'm confusing myself.

PR #429, opened 2015-08-25
Add do_shortcode to list of autoescaped functions

### LINKED ISSUE #428: do_shortcode() flagged as needing escaping
`do_shortcode()` is currently flagged as needing escaping.

I discussed this with VIP support (#44195). David, after conferring with another team member, said that it's unnecessary, as we do the escaping where the HTML is emitted (in the shortcode's code).

paulschreiber:
Fixes #428.


JDGrimes:
@paulschreiber — I think you forked from `master` instead of `develop`. That's OK, but I just thought I'd point it out, 'cause in some cases it could make merging difficult. This small change wasn't a problem, but you might want to keep it in mind for next time. :smile:

PR #432, opened 2015-08-29
Only check T_USE tokens associated with closures

### LINKED ISSUE #431: No space after opening parenthesis is probihited
I'm not sure why it's happening but it's showing this error on the `use` statements:

```
<?php

namespace Foo;

use Foo\Admin\Admin;
use Foo\Admin\User;
use Foo\Admin\Work;
```

JDGrimes:
The `T_USE` token is also used to import traits to classes, and to
import classes into namespaces.

Fixes #431

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -1041,6 +1041,40 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 
 		return false;
 	}
+
+	/**
+	 * Check what type of 'use' statement a token is part of.
+	 *
+	 * The T_USE token has multiple different uses:
+	 *
+	 * 1. In a closure: function () use ( $var ) {}
+	 * 2. In a class, to import a trait: use Trait_Name
+	 * 3. In a namespace, to import a class: use Some\Class;
+	 *
+	 * This function will check the token and return 'closure', 'trait', or 'class',
+	 * based on which of these uses the use is being used for.
+	 *
+	 * @param int $stackPtr The position of the token to check.
+	 *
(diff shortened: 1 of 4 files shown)


WordPress/Sniff.php
```diff
@@ -1041,6 +1041,40 @@ protected function is_comparison( $stackPtr ) {
...
+
+		// USE keywords inside closures.
+		$next = $this->phpcsFile->findNext( T_WHITESPACE, $stackPtr + 1, null, true );
+
+		if ( T_OPEN_PARENTHESIS === $this->tokens[ $next ]['code'] ) {
+			return 'closure';
+		}
+
+		// USE keywords for traits.
+		if ( $this->phpcsFile->hasCondition( $stackPtr, array( T_CLASS, T_TRAIT ) ) ) {
```
GaryJones: Is this just checking that the file has a class that uses a trait?

How about:

```
<?php
namespace = Foo;

use Bar; // Testing this one.

class Baz {
    use MyTrait;
}
```

JDGrimes: > Is this just checking that the file has a class that uses a trait?

No. Each token has a list of blocks (class/trait/function/if/while/foreach/etc.) that it is in. This checks if the `use` token is in a class or trait.

The code is based on [this function in PHPCS](https://github.com/squizlabs/PHP_CodeSniffer/blob/9dd0ada629d115e6ac9fbea7cac944f2775539ec/CodeSniffer/Standards/PSR2/Sniffs/Namespaces/UseDeclarationSniff.php#L136-L162).

PR #436, opened 2015-08-30
Changelog

GaryJones:
Some change log improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
-# Change Log
+# Change Log for WordPress Coding Standards
+
 All notable changes to this project will be documented in this file.
 
 This projects adheres to [Semantic Versioning](http://semver.org/) and [Keep a CHANGELOG](http://keepachangelog.com/).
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -123,5 +124,23 @@ property should now be used instead for the `EscapeOutput` sniff.
...
+
+## [2013-10-06]
+
+See the comparison for full list.
+
+## 2013-06-11
+
+Initial tagged release.
+
+[Unreleased]: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/compare/0.7.0...HEAD
```
JDGrimes: Not all unreleased changes are in `develop` though. The ones I specifically mention in that list are unmerged pull requests.

GaryJones: In which case, they shouldn't have made it to the Change log yet - Unreleased section should be empty.

JDGrimes: I don't think that [Keep a CHANGELOG](http://keepachangelog.com/#how-can-i-minimize-the-effort-required-) is really clear on this either way. 

I'm going to go ahead and merge this just like it is so we won't get conflicts when I add the change log for 0.7.1. If you want to discuss the Unreleased section further, go ahead and open up another issue/PR for it.

PR #437, opened 2015-08-31
Update maintainers to include JDGrimes and GaryJones

westonruter:


GaryJones:
I'm assuming that was going to be #reviewmerge ?

westonruter:
Yes :smile:

PR #440, opened 2015-08-31
Default to 0 spaces before closure parentheses 

### LINKED ISSUE #439: Spacing in closures
Many of my Travis-tested repos just started failing due to the merge of #417. I see that WPCS now wants:

```php
add_filter( 'excerpt_length', function() {
	return 40;
} );
```

to be:

```php
add_filter( 'excerpt_length', function () {
	return 40;
} );
```

I see no mention of this in the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/). Is there a secondary guide that WPCS is using to make decisions like this? If it's a matter of opinion, I'd love to see the discussion so I can better understand why this format was chosen. I personally think `function() {` makes more, as it follows the non-closure function standard more closely, where the parens are adjacent to the function name like `function foo() {`. If it were just a matter of opinion, I would hide it in my ruleset, though it doesn't look like this can be ignored without also affecting other control structures (please correct me if I'm wrong) covered by `WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceAfterStructureOpen` and `WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceBeforeOpenParenthesis`.

Thanks!

JDGrimes:
I forked this from `master`, but let's go ahead and just merge it into `develop`. Since there haven't been any new features merged into `develop` yet, we can go ahead and merge `develop` back into `master` before tagging 0.7.1.

Fixes #439


GaryJones:
Merged - knock up the Changelog and then go ahead and release 0.7.1.

PR #444, opened 2015-09-05
XSS sniff: support implode() and array_map()

### LINKED ISSUE #442: implode() use causes a false positive
The following code triggers an `Expected a sanitizing function...` warning, despite the output of `implode()` being safe:

```php
$items = array( 'foo', 'bar' );
echo implode( '<br>', esc_html( $items ) );
```

JDGrimes:
Because `implode()` is mostly useless without `array_map()`.

Fixes #442

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -304,6 +304,7 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	 */
 	public static $formattingFunctions = array(
 		'ent2ncr' => true,
+		'implode' => true,
 		'sprintf' => true,
 		'vsprintf' => true,
 		'wp_sprintf' => true,
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -305,6 +305,7 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	public static $formattingFunctions = array(
 		'ent2ncr' => true,
 		'implode' => true,
+		'join' => true,
 		'sprintf' => true,
 		'vsprintf' => true,
 		'wp_sprintf' => true,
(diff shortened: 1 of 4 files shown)


WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -274,19 +274,41 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr )
...
+				if ( $function_opener ) {
+
+					if ( 'array_map' === $functionName ) {
+
+						// Get the first parameter (name of function being used on the array).
+						$mapped_function = $this->phpcsFile->findNext(
+							PHP_CodeSniffer_Tokens::$emptyTokens
+							, $function_opener + 1
+							, $tokens[ $function_opener ]['parenthesis_closer']
+							, true
```
westonruter: @JDGrimes I think we previously talked about this comma-first style not being the style for WPCS?

GaryJones: That's correct Weston.

JDGrimes: Oops! I actually copied this from another part of the code (which I must have written :smile:), but I'll correct it.

JDGrimes: Fixed in 6b786e9.

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -274,19 +274,41 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr )
 			// Now check that next token is a function call.
 			if ( T_STRING === $this->tokens[ $i ]['code'] ) {
 
+				$ptr = $i;
+
 				$functionName = $this->tokens[ $i ]['content'];
+
+				$function_opener = $this->phpcsFile->findNext( array( T_OPEN_PARENTHESIS ), $i + 1, null, null, null, true );
```
westonruter: Gotta love positional arguments :smile:

WordPress/Sniff.php
```diff
@@ -304,6 +304,7 @@
 	 */
 	public static $formattingFunctions = array(
 		'ent2ncr' => true,
+		'implode' => true,
```
westonruter: I think `join` should be added here too, as it is an alias of `implode`.

westonruter:
:clap: excellent work as always, @JDGrimes

PR #468, opened 2015-10-03
XSS sniff: fix ternary detection bug

### LINKED ISSUE #466: Bug: Expects escape function ternary operation inside echo statement
```php
<?php

function select_options( array $options, $selected ) {
	foreach ( $options as $name => $value ) {
		echo '<option name="' . esc_attr( $name ) . '"' .
		     ( $name === $selected ? ' selected' : '' ) .
		     '>' . esc_html( $value )
		     . '</option>';
	}
}
```

The error:

```
 6 | ERROR | Expected next thing to be an escaping function (see
   |       | Codex for 'Data Validation'), not '$name'
```

JDGrimes:
The logic was supposed to be looking for a second set of parentheses
within an outer set of parentheses. But because it passed `$i` to
`findNext()`, the current opening parenthesis was being found, since
`$i` points to the opening parenthesis at that point, and `findNext()`
starts with the _current_ token.

The solution is to start looking for the next opening parenthesis at
`$i + 1` instead.

Fixes #466

### DIFF
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -208,7 +208,7 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff
 
 					if ( $ternary ) {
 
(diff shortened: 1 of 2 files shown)


vohof:
Nice!

PR #477, opened 2015-10-21
Add Squiz.PHP.CommentedOutCode to WordPress-VIP

westonruter:
https://vip.wordpress.com/documentation/code-review-what-we-look-for/#commented-out-code-debug-code-or-output

> ## Commented out code, Debug code or output
> […] The use of commented out code should be avoided. Having code that is not ready for production on production is bad practice and could easily lead to mistakes while reviewing (since the commented out code might not of been reviewed and the removing on a comment might slip in accidently).

### DIFF
--- a/WordPress-VIP/ruleset.xml
+++ b/WordPress-VIP/ruleset.xml
@@ -11,4 +11,7 @@
 	<rule ref="WordPress.PHP.StrictComparisons" />
 	<rule ref="WordPress.WP.PreparedSQL" />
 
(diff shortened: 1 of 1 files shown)


westonruter:
We may need to adjust the default `maxPercentage` be something higher. I seem to be getting a lot of false positives.

mboynes:
Leaving this note for posterity: one of my projects (which was fine previously) started triggering an infinite loop in phpcs due to this addition. Filed an issue in squizlabs/PHP_CodeSniffer#862.

PR #480, opened 2015-10-28
Improve WordPress.WP.PreparedSQL.NotPrepared

westonruter:
Account for cases where '$' is used in double-quoted strings and is not referring to a variable.

### DIFF
--- a/WordPress/Sniffs/VIP/ValidatedSanitizedInputSniff.php
+++ b/WordPress/Sniffs/VIP/ValidatedSanitizedInputSniff.php
@@ -90,11 +90,14 @@ class WordPress_Sniffs_VIP_ValidatedSanitizedInputSniff extends WordPress_Sniff 
 
 		// Handling string interpolation
 		if ( T_DOUBLE_QUOTED_STRING === $tokens[ $stackPtr ]['code'] ) {
-			foreach ( $superglobals as $superglobal ) {
-				if ( false !== strpos( $tokens[ $stackPtr ]['content'], $superglobal ) ) {
-					$phpcsFile->addError( 'Detected usage of a non-sanitized, non-validated input variable: %s', $stackPtr, null, array( $tokens[ $stackPtr ]['content'] ) );
-					return;
-				}
+			$interpolated_variables = array_map(
+				function ( $symbol ) {
+					return '$' . $symbol;
+				},
+				$this->get_interpolated_variables( $tokens[ $stackPtr ]['content'] )
+			);
+			foreach ( array_intersect( $interpolated_variables, $superglobals ) as $bad_variable ) {
+				$phpcsFile->addError( 'Detected usage of a non-sanitized, non-validated input variable %s: %s', $stackPtr, null, array( $bad_variable, $tokens[ $stackPtr ]['content'] ) );
 			}
 
 			return;
--- a/WordPress/Sniffs/WP/PreparedSQLSniff.php
+++ b/WordPress/Sniffs/WP/PreparedSQLSniff.php
@@ -74,23 +74,23 @@ class WordPress_Sniffs_WP_PreparedSQLSniff extends WordPress_Sniff {
 
 			if ( T_DOUBLE_QUOTED_STRING === $tokens[ $i ]['code'] ) {
 
-				$string = str_replace( '$wpdb', '', $tokens[ $i ]['content'] );
-
-				// Check if the $ is followed by a valid name, and that it is not preceded by an escape sequence.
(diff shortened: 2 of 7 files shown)


WordPress/Sniffs/WP/PreparedSQLSniff.php
```diff
@@ -76,7 +76,8 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 
 				$string = str_replace( '$wpdb', '', $tokens[ $i ]['content'] );
 
-				if ( false !== strpos( $string, '$' ) ) {
+				// Check if the $ is followed by a valid name, and that it is not preceded by an escape sequence.
+				if ( preg_match( '/(\\\\*)\$\w+/', $string, $matches ) && strlen( $matches[1] ) % 2 === 0 ) {
```
JDGrimes: Can you add some tests that demonstrate that this correctly handles the case where the variable is proceeded by an even number of backslashes and therefore is not escaped?

Also, what happens if there are multiple matches in a single string? I think if the first one is escaped the second one will be ignored, will it not? Tests please.

westonruter: @JDGrimes great catch! Points addressed in 32450cc.

JDGrimes:
I just happened to think, maybe we should make that into a helper function that would return an array of the interpolated variables. Then we could also use it here: 

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/e9bb0a2baf5564f701d3b365ae61c02677d5b6a2/WordPress/Sniffs/VIP/ValidatedSanitizedInputSniff.php#L94 

(That appears to be the only other place we'd need it right now.)

westonruter:
> I just happened to think, maybe we should make that into a helper function that would return an array of the interpolated variables

@JDGrimes good idea. Implemented in 494864f6.

JDGrimes:
Looks good @westonruter! :smile:

olefredrik:
After this PR was merged, I started getting Travis CI build failures. 

```Parse error: syntax error, unexpected T_FUNCTION, expecting ')' in /tmp/wordpress/wordpress-coding-standards/WordPress/Sniffs/VIP/ValidatedSanitizedInputSniff.php on line 94```

https://travis-ci.org/olefredrik/FoundationPress/jobs/88868838#L181

I'm using the same travis.yml configuration as the Underscores theme. So I guess there will be build failures there as well, next time there is a pull request on that repo. 

Would appreciate if any of you could look into it.

File: WordPress/Sniffs/VIP/ValidatedSanitizedInputSniff.php [hunk 2]
```diff
@@ -90,11 +90,14 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 
 		// Handling string interpolation
 		if ( T_DOUBLE_QUOTED_STRING === $tokens[ $stackPtr ]['code'] ) {
-			foreach ( $superglobals as $superglobal ) {
-				if ( false !== strpos( $tokens[ $stackPtr ]['content'], $superglobal ) ) {
-					$phpcsFile->addError( 'Detected usage of a non-sanitized, non-validated input variable: %s', $stackPtr, null, array( $tokens[ $stackPtr ]['content'] ) );
-					return;
-				}
+			$interpolated_variables = array_map(
+				function ( $symbol ) {
```
westonruter: @olefredrik Here's the problem: I used a closure. I assumed PHP_CodeSniffer required PHP 5.3 or above, but it seems it is compatible back to 5.1. So we'll need to adjust our Travis config to test in 5.1 as well, since we test in 5.3 or above only.

JDGrimes:
@olefredrik It's because you are running on PHP 5.2 but the PR introduces some uses of closures, which require PHP 5.3. We don't actually officially support PHP 5.2, only 5.3+, which is why we didn't catch this change. I'm actually surprised that you have been able to use it on 5.2, but I guess we haven't been using many 5.3 features.

olefredrik:
@westonruter , @JDGrimes : Aha. Actually I'm not sure why PHP 5.2 is included in my travis.yml file, if it's not officially supported. I copied the Travis CI configuration from the Underscores theme. What you're saying makes sense, though, because all tests passed on PHP 5.3 and above.

westonruter:
> Aha. Actually I'm not sure why PHP 5.2 is included in my travis.yml file, if it's not officially supported. I copied the Travis CI configuration from the Underscores theme. What you're saying makes sense, though, because all tests passed on PHP 5.3 and above.

@olefredrik it makes sense actually that you would have PHP 5.2 in your `.travis.yml` because WordPress supports PHP 5.2, so the WordPress-Coding-Standards should be able to run in that environment as well.

Issue #176, opened 2014-06-25
VariableRestrictionsSniff can not be run on PHP5.2


jrfnl:
Found when running a travis build: https://travis-ci.org/Yoast/wordpress-seo/builds/28375064

Considering the minimum requirement for WP is still (unfortunately) PHP 5.2.6, I'd suggest that the codesniffs should also be able to work under PHP 5.2.

Parse error: syntax error, unexpected T_FUNCTION in /home/travis/.phpenv/versions/5.2.17/pear/PHP/CodeSniffer/Standards/WordPress/Sniffs/Variables/VariableRestrictionsSniff.php on line 140

PR #483, opened 2015-11-02
Fix compatibility with PHP 5.2

westonruter:
See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/480#issuecomment-153154449


JDGrimes:
Related: #176

westonruter:
@JDGrimes because we're just temporarily not using a closure. I think it's silly to have methods like `prepend_array_items_with_dollar_sign()` and `filter_out_wpdb_from_array_items()` when clearly lambdas make more sense for `array_filter()` and `array_map()` in these cases, IMHO.

westonruter:
@JDGrimes ok, at least WPCS won't cause syntax errors anymore. I did prevent PHPUnit tests, however, since the fixer logic was failing when running on code with namespaces in 5.2.

.travis.yml
```diff
@@ -14,6 +15,7 @@ env:
   - PHPCS_BRANCH=2.2.0
 
 matrix:
+  fast_finish: true
```
westonruter: This didn't seem to help any.

JDGrimes: It doesn't cause the build to stop at the first failure, it just causes the notification to be sent as soon as there is a failure instead of waiting for the build to finish. (In case that was what you meant by it not having any effect.)

JDGrimes:
I've re-enabled the PHPUnit tests on 5.2, though two of the tests have to be skipped. All tests pass.

.travis.yml
westonruter: Ah, yes, I misunderstood it then. Thank you.

westonruter:
@JDGrimes thanks!

PR #488, opened 2015-11-14
Proposed updates as per updated VIP guidelines

sboisvert:
I've added the functions that we check in the VIP scanner here. This is based on the documentation http://vip.wordpress.com/documentation/uncached-functions/
and
http://vip.wordpress.com/documentation/code-review-what-we-look-for/


sboisvert:
I think I accidentally removed get_category_by_slug, it should go under get_term_by

sboisvert:
Ugh, just found I put things in the wrong place. Lots of this stuff is in other files. closing request

PR #499, opened 2015-12-16
Require yoda conditions for object properties

### LINKED ISSUE #497: Yoda comparison not flagged / new testcase
The below code did not get flagged, not sure why (haven't looked in depth).

```php
if ( $GLOBALS['wpdb']->num_rows === 0 ) {}
```

JDGrimes:
(I added the unrelated function assignment check because there wasn’t
one in the tests yet and I wanted to make sure that we didn’t change
behavior on that.)

Fixes #497

### DIFF
--- a/WordPress/Tests/PHP/YodaConditionsUnitTest.inc
+++ b/WordPress/Tests/PHP/YodaConditionsUnitTest.inc
@@ -84,3 +84,7 @@ return 0 == $foo; // OK
 return $foo == 0; // Bad
 
 if ( (int) $a['interval'] === (int) $b['interval'] ) {} // OK
+
+if ( $GLOBALS['wpdb']->num_rows === 0 ) {} // Bad
+
+if ( $true == strtolower( $check ) ) {} // Bad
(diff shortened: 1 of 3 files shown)


WordPress/Tests/PHP/YodaConditionsUnitTest.inc
```diff
@@ -84,3 +84,7 @@ return 0 == $foo; // OK
 return $foo == 0; // Bad
 
 if ( (int) $a['interval'] === (int) $b['interval'] ) {} // OK
+
+if ( $GLOBALS['wpdb']->num_rows === 0 ) {} // Bad
+
+if ( $true == strtolower( $check ) ) {} // Bad
```
jrfnl: If you want to test function assignment, shouldn't this be one `=` ?

Oh and shouldn't that kind of assignment within a condition be blacklisted as bad practice ? (in another sniff)

JDGrimes: I guess I misspoke, I actually meant comparison. :blush: There was a test for the function on the left but not on the right.

PR #506, opened 2016-01-05
Remove PrivateNoUnderscore and PublicHasUnderscore from ValidVariableName

### LINKED ISSUE #502: Naming convention sniffs over-zealous ?
The naming convention sniffs introduced in #500 seem to me to be a bit over-zealous.

The handbook doesn't mention anything about private properties needing to start with an underscore.

And to be quite fair, IMHO, that shouldn't become part of the standard either as that is an outdated PHP4 practice from when you couldn't indicate the visibility of a property.

Oh and there is a message error on line 113:
```php
$error = 'Variable "%s" is not in valid camel caps format';
```

should probably read:
```php
$error = 'Variable "%s" is not in valid snake_case format';
```

westonruter:
Also fixes message error.

Fixes #502


jrfnl:
Looks good to me :+1:

PR #515, opened 2016-01-22
Fix false positives and false negatives in PreparedSQL sniff

### LINKED ISSUE #508: False positive: Use placeholders and $wpdb->prepare()
I have a few lines of code that make use of ```$wpdb->prepare()```, but no matter how I structure it I receive an error.  My code:

    $sql = sprintf( 'SELECT `post_id`, `meta_value` FROM `%s` WHERE `meta_key` = "sort_order" AND `post_id` IN (%s)',
    			$wpdb->postmeta,
    			implode( ',', array_fill( 0, count( $post_ids ), '%d' ) )
    		);
    $sql = $wpdb->prepare( $sql, $post_ids );
    $all_post_meta = $wpdb->get_results( $sql );

The error:  ```Use placeholders and $wpdb->prepare(); found $sql```.

If I condense it all into one line, so the placeholders and ```prepare()``` call are all in the same place, it still produces the same error.

    $all_post_meta = $wpdb->get_results( $wpdb->prepare( sprintf( 'SELECT `post_id`, `meta_value` FROM `%s` WHERE `meta_key` = "sort_order" AND `post_id` IN (%s)',
    		$wpdb->postmeta,
    		implode( ',', array_fill( 0, count( $post_ids ), '%d' ) )
    	),
    	$post_ids )
    );

JDGrimes:
The prepared SQL sniff was incorrectly flagging code that was properly escaping SQL without using the `$wpdb->prepare()` function. I've updated it not to flag this code, but perhaps it should be possible to toggle this feature (in case we want to force `$wpdb->prepare()` to be used)?

The sniff was also incorrectly only looking for variables being incorporated into SQL queries, not constants, function calls, etc. This is now fixed, and anything being incorporated into an SQL query will be flagged.

Fixes #508

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -308,6 +308,7 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	 * @var array
 	 */
 	public static $formattingFunctions = array(
+		'array_fill' => true,
 		'ent2ncr' => true,
 		'implode' => true,
 		'join' => true,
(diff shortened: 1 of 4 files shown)


westonruter:
Nice work :heart:

PR #528, opened 2016-02-10
Add SublimeLinter-phpcs installation instructions

corysimmons:
Atom with linting ended up being insanely slow. SublimeLinter ain't as pretty but it gets the job done. =)


westonruter:
I would merge this, but I don't use Sublime so I can't verify it works :)

LoreleiAurora:
:thumbsup: Works for me xXx

Issue #535, opened 2016-03-08
ConcatenationSpacing?


samwilson:
My understanding is that the string concatenation operator `.` should be surrounded by whitespace in WordPress code. Something like the opposite of `Squiz.Strings.ConcatenationSpacing`.

The [docs read](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage): "Always put spaces after commas, and on both sides of logical, comparison, string and assignment operators."

westonruter:
You are right.

samwilson:
:) I'll send a PR?

westonruter:
Yes you will! :)

samwilson:
Hurrah! :) Thanks for being speedy. (I'm in UTC+8 and normally feel a bit alone...)

samwilson:
PR submitted.

PR #536, opened 2016-03-09
Add sniff and tests for concatenation operator spacing

samwilson:
This change adds the ConcatenationSpacingSniff, which checks that the concatenation operator is correctly surrounded by spaces. Tests are also included.

The relevant WordPress coding standard [reads](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage): "Always put spaces after commas, and on both sides of logical, comparison, string and assignment operators."

Both the sniff and test classes are simply extensions of the `Squiz.Strings.ConcatenationSpacing` code (which kindly already supported the functionality we require here).

Refs #535.

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -55,6 +55,12 @@
 
 	<!-- https://make.wordpress.org/core/handbook/coding-standards/php/#space-usage -->
 	<rule ref="Generic.Formatting.SpaceAfterCast"/>
+	<rule ref="Squiz.Strings.ConcatenationSpacing">
+		<properties>
+			<property name="spacing" value="1"/>
+			<property name="ignoreNewlines" value="true"/>
+		</properties>
+	</rule>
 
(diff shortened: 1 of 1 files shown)


westonruter:
Humm. Looks like the tests fail only when the `PHPCS_BRANCH` is `2.2.0`. Wonder why that is.

samwilson:
Yes, I was just looking into that. I'm a bit confused so far though. Any ideas?

samwilson:
Actually, scratch this whole thing. The right way to do this is perhaps to add the following to `WordPress-Core/ruleset.xml` —

	<rule ref="Squiz.Strings.ConcatenationSpacing">
		<properties>
			<property name="spacing" value="1"/>
			<property name="ignoreNewlines" value="true"/>
		</properties>
	</rule>

samwilson:
Ok, I've updated the PR. I don't think there's any need to have tests for this, because the spacing behaviour is already tested in PHP_CodeSniffer.

westonruter:
Perfect! :heart_eyes:

samwilson:
Thanks :)

PR #545, opened 2016-03-24
Merge multi-line string literal tokens into single token

### LINKED ISSUE #50: Check for proper usage of translation functions

westonruter:
Fixes #50.

### DIFF
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -86,6 +86,18 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 				$argument_tokens = array();
 				continue;
 			}
+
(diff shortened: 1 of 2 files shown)


GaryJones:
Good for me :+1:

PR #546, opened 2016-03-26
This adds support for standards installation through Composer.

schlessera:
This depends on the following pull request for the `squizlabs/php_codesniffer` package, and should only be merged once that pull request has been merged as well:
https://github.com/squizlabs/PHP_CodeSniffer/pull/936

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,6 @@
 	"name"       : "wp-coding-standards/wpcs",
 	"description": "PHP_CodeSniffer rules (sniffs) to enforce WordPress coding conventions",
 	"type"       : "phpcs-standard",
-	"version": "9.0",
 	"keywords"   : ["phpcs", "standards", "WordPress"],
 	"license"    : "MIT",
 	"authors"    : [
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -1,6 +1,8 @@
 {
 	"name"       : "wp-coding-standards/wpcs",
 	"description": "PHP_CodeSniffer rules (sniffs) to enforce WordPress coding conventions",
+	"type"       : "phpcs-standard",
+	"version": "9.0",
```
GaryJones: Where does this version 9.0 come from?

schlessera: TBH, I have no idea. Was certainly not my intention to introduce that one here (and the formatting is off as well).
I'll adapt the pull request...

Rarst:
I think by now #855 (using third party installer plugin) is more realistic way to go since upstream PR doesn't seem to be happening.

jrfnl:
@schlessera I believe this PR was superseded by https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/855 ? Do you agree or is there still something we're missing ?

schlessera:
@jrfnl I agree and will close this one.

PR #552, opened 2016-04-01
Ignore Squiz.Commenting.FunctionComment.ScalarTypeHintMissing

### LINKED ISSUE #550: ScalarTypeHintMissing sniff targets PHP 7
Unless I got that wrong `string` scalar is only available starting with PHP 7, which is a little too new for WP. :D http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration
```
  40 | ERROR | Type hint "string" missing for $class
     |       | (Squiz.Commenting.FunctionComment.ScalarTypeHintMissing)
```

westonruter:
Fixes #550.

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -102,7 +102,4 @@
 	<rule ref="WordPress.WhiteSpace.CastStructureSpacing"/>
 	<rule ref="WordPress.PHP.YodaConditions"/>
 
-	<rule ref="Squiz.Commenting.FunctionComment.ScalarTypeHintMissing">
-		<severity>0</severity>
-	</rule>
 </ruleset>
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -102,4 +102,7 @@
 	<rule ref="WordPress.WhiteSpace.CastStructureSpacing"/>
 	<rule ref="WordPress.PHP.YodaConditions"/>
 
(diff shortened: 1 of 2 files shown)


Rarst:
Is there a practical difference between setting severity 0 (which I didn't use before) and using exclude tag (which I tend to do) ?

Does severity approach imply that sniff still runs (read it consumes time) and just result is ignored?

westonruter:
I think `exclude` relates to an entire sniff whereas setting the severity relates to suppressing an individual error code _within_ a sniff. In other words, the `Squiz.Commenting.FunctionComment` sniff isn't to be excluded and should still be run, but the `ScalarTypeHintMissing` error code should be suppressed.

Rarst:
This is how I tend to write it downstream:

```xml
<rule ref="WordPress-Docs">
	<exclude name="Squiz.Commenting.FunctionComment.ScalarTypeHintMissing" />
</rule>
```

Just trying to figure out if there is a practical difference between approaches.

westonruter:
Interesting. I didn't know that worked. Maybe they are two ways of telling PHPCS the same thing?

JDGrimes:
> Maybe they are two ways of telling PHPCS the same thing?

I think so, I don't think that there is really any practical difference between them. I tend to use the same format as @Rarst though, because its more expressive.

westonruter:
OK, I've fixed this in c1af3a9. I put it in the wrong standard anyway. Once I opened `WordPress-Docs` it became clear which style should be used :smile:

PR #554, opened 2016-04-08
Fix _n_noop() and _nx_noop()

westonruter:
These don't take a `$number` arg.

See #50 and #136.

### DIFF
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -75,7 +75,7 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 		$argument_tokens = array();
 
 		// Look at arguments.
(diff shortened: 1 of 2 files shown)


westonruter:
@GaryJones any concerns with this?

GaryJones:
@westonruter Assuming the tests pass, then it would be fine, but I do have two suggestions.

Take a test string like:

```
_n_noop( 'I have a cat.', 'I have cats.', "illegal $string" );
```

I'd like to see test string variants for each relevant i18n function that include:
* a `%s` placeholder in both the single and plural strings - it shouldn't make any difference to the logic, but strings with placeholders for the number (i.e. `'I have %s cat.', 'I have %s cats.'`) is more like real-world usage. `%d` may be a better choice, but I've got a feeling I read something about there being a drawback to trying to force an int compared to allowing any string.
* a variable (or at least a dollar sign) in a single string text domain: 
  ```
  _n_noop( 'I have a cat.', 'I have cats.', 'illegal-$string' ); // OK
  ```

  I don't know what the rules are regarding dollar signs in text domains, and that's probably a set of separate checks, but if so, it shouldn't trigger the same error as a straight variable or a variable inside double-quotes.

PR #558, opened 2016-04-11
[VariableName Sniff] Improve error message differentiation in the sniff.

### LINKED ISSUE #555: [snakecase sniff] Unclear error msg when using static property
Just ran into an unclear error message.

```php
$classes = array_udiff( $classes, Prefix_PHP_Classes::$PHP_classes, 'strcasecmp' );
```

Throws a `Variable "PHP_classes" is not in valid snake_case format` while I would expect it to say `Property "PHP_classes" is not in valid snake_case format`

Also wondering if there is a way for me to whitelist this as an exception ?

Related #549

[Edited: quoted wrong line from source code originally, now corrected]

jrfnl:
Ensures that static class properties get the `Object property... ` version of the error message rather than the `Variable ... ` version.

Includes additional unit tests to verify that the differentiation works properly.

Also:
- Slightly simplifies the `in_class` check
- Fixes one `// Bad` on the wrong line issue in the unit test file
- Fixes one inconsistent error message (missing "valid" in the sentence).

Closes #555


jrfnl:
*ping* anyone willing to review this PR ?

PR #560, opened 2016-04-26
Validate $domain arg in WordPress.WP.I18n

westonruter:
* Introduce `WordPress_Sniffs_WP_I18nSniff::$text_domain` property.
* Skip reporting issues for missing `$domain` if `WordPress_Sniffs_WP_I18nSniff::$text_domain` is not defined.
* Increase severity of domain issues from warning to error.
* Add `WordPress.WP.I18n` to WordPress-Core.

A project's ruleset can opt-in to doing `$domain` validation via adding this to there PHPCS `ruleset.xml`, here for the Customize Posts plugin:

```xml
...
		<properties>
			<property name="text_domain" value="customize-posts" />
		</properties>
	</rule>
...
```

See #50.


westonruter:
@GaryJones what do you think of this?

GaryJones:
Nothing of concern standing out to me right now. :+1:

PR #563, opened 2016-05-03
Update with latest VIP checks

sboisvert:
added check for 
get_intermediate_image_sizes()
serialize() / unserialize()
error_log(), var_dump(), trigger_error(), set_error_handler()
wp_redirect(),
wp_is_mobile()
urlencode()

As well as added get_previous_post_link() and get_next_post_link() to the adjacent_post check


westonruter:
@sboisvert thanks!

PR #566, opened 2016-06-01
Update PHPCS version dependency

GaryJones:
Should we require the latest minor release of PHPCS?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,9 +11,10 @@
 		}
 	],
 	"require"    : {
-		"squizlabs/php_codesniffer": "~2.2"
+		"squizlabs/php_codesniffer": "^2.6"
 	},
 	"minimum-stability" : "RC",
+	"prefer-stable": true,
 	"support"    : {
 		"issues": "https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues"
 	},


jrfnl:
This would need a change in the Travis version matrix as well as that has PHPCS `2.2.0` hard-coded in.

westonruter:
@jrfnl would you then also update `PHPCS_BRANCH` to `2.4.0`?

jrfnl:
@westonruter You mean in PR #591 ? I can't make any changes to this branch.

GaryJones:
I'm not going to have time, so feel free to replicate my PR, update it as needed, and close this PR.

jrfnl:
I've changed the minimum version to `2.4.0` in PR #591 in all the right places AFAICS. That PR needs PHPCS 2.4.0 to function properly.

@GaryJones was there a particular reason to pick `2.6.0` as the new minimum ?

GaryJones:
At the time of opening the PR, it was the latest minor branch. There's little reason for developers to be using an older version IMO.

Plus, [WPCS expects `include|require(_once)` lines to be documented](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#3-requires-and-includes), but that's only [supported](https://github.com/squizlabs/PHP_CodeSniffer/pull/581) in 2.6.0. In effect, this is no different to the bump in #591.

(Aside: I genuinely forgot that I had contributed to PHPCS, and was surprised to my name in the release notes for 2.6.0!)

jrfnl:
Ok - please decide upon #591 first and let's handle the upping of the minimum version to `2.6.0` after that with the argumentation included in the commit message.
After all there is a separate and distinct reason for that change (and it would conflict with the current open PR).

JDGrimes:
> At the time of opening the PR, it was the latest minor branch. There's little reason for developers to be using an older version IMO.

Agreed, but at one time we were sticking with an older version because some folks (Yoast, I think) who had extended the sniffs weren't compatible with the latest version yet. Of course, that may no longer be true, and we don't *have* to take that into account when determining what we'll support.

PR #567, opened 2016-06-08
Add more i18n sniffs

deBhal:
This PR is to add another test to the I18nSniff.php to ensure that multiple placeholders are orderable.

e.g. from the PHP: sprintf page:
`__( 'The %s contains %d monkeys' );` is a problem because the variables will need to be the other order in some languages, and the string should be:
 `__( 'The %1$s contains %2$d monkeys' );`

The PR includes tests and a fixer.

One thing worth discussing is that the fixer could potentially get it wrong if:
1. It's a pluralised string
2. The order of the placeholders is different between the two strings
3. The placeholders are unordered.

The test file gives an example of the sort of string I'm talking about:
`_n( 'There is %d monkey in the %s', 'In the %s there are %d monkeys', $number, 'my-slug' );`

My assessment here is that this is a bug in the code, and this sniff won't fix it, but it also won't make it any worse than it already was.  I was tempted to not apply the fix in plural cases, but think the good outweighs the bad, so for now the fix does run on plurals.

### DIFF
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -194,6 +194,7 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 		$arg_name = $context['arg_name'];
 		$method = empty( $context['warning'] ) ? 'addError' : 'addWarning';
 		$content = $tokens[0]['content'];
+		$fixable_method = empty( $context['warning'] ) ? 'addFixableError' : 'addFixableWarning';
 
 		if ( 0 === count( $tokens ) ) {
 			$code = 'MissingArg' . ucfirst( $arg_name );
@@ -228,12 +229,20 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 					$suggestions[ $i ] = substr_replace( $unordered_matches[ $i ], ( $i + 1 ) . '$', 1, 0 );
 				}
 
-				$phpcs_file->$method(
+				$fix = $phpcs_file->$fixable_method(
 					'Multiple placeholders should be ordered. Expected \'%s\', but got %s.',
 					$stack_ptr,
 					'UnorderedPlaceholders',
 					array( join( ', ', $suggestions ), join( ',', $unordered_matches ) )
 				);
+
+				if ( true === $fix ) {
+					$fixed_str = str_replace( $unordered_matches, $suggestions, $content );
+
+					$phpcs_file->fixer->beginChangeset();
+					$phpcs_file->fixer->replaceToken( $stack_ptr, $fixed_str );
+					$phpcs_file->fixer->endChangeset();
+				}
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


JDGrimes:
This looks great, thanks @deBhal! In regard to the plurals and placeholders in the wrong order, I don't think that we really need to worry about that. As you say, it is a bug in the plugin itself. Though I suppose we could check if the order is the same in both of the strings, and report in error if it isn't. In some cases it would be undetectable though, if the user was just using `%s` both times instead of `%d`.

deBhal:
Thanks for the quick feedback!

I've added another new sniff for the singular problem described in https://codex.wordpress.org/I18n_for_WordPress_Developers#Plurals
`_n( 'This is a subtle problem', 'These are subtle problems', $number )`.
"Singular" means exactly one in English, but that's not the case in several languages.

I've also added a warning for mismatched placeholders, which will catch at least some of the cases we talked about above.

I'm adding these to this PR rather than opening separate ones because the way the tests work would force annoying re-edits if they came from separate PRs.

Quick question:  I've used `$this->compare_single_and_plural_arguments()`, but the existing code uses `call_user_func()` instead of calling a member function directly.  Is there a reason for this?  PHP versions or something?

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -70,8 +70,14 @@ public function register() {
 	 */
 	public function process( PHP_CodeSniffer_File $phpcs_file, $stack_ptr ) {
 		$tokens = $phpcs_file->getTokens();
-		$token  = $tokens[ $stack_ptr ];
+		static $only_once = true;
+		if ( $only_once ) {
+			// echo( '$tokens = ' . var_export( $tokens, true ) . PHP_EOL ); //DEBUG
```
JDGrimes: The debug code here (and below on line 80) can be removed. 😃 

(And just FYI, if you have never used an IDE with support for xdebug built in, you are truly missing out! 😃)

JDGrimes:
> "Singular" means exactly one in English, but that's not the case in several languages.

Thanks for this, I didn't realize that (or maybe had just forgotten).

> Quick question: I've used $this->compare_single_and_plural_arguments(), but the existing code uses call_user_func() instead of calling a member function directly. Is there a reason for this? PHP versions or something?

I'm pretty sure that this is just a leftover from when the code called one of several different methods in that spot, but it has been refactored to just call that one now. So there is really no reason for `call_user_func()` to be used there anymore, it is just a relic.

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -179,6 +183,14 @@ public function process( PHP_CodeSniffer_File $phpcs_file, $stack_ptr ) {
 			}
 			call_user_func( array( $this, 'check_argument_tokens' ), $phpcs_file, $argument_assertion_context );
 		}
+
+		// for _n*() calls, compare the singular and plural strings
```
GaryJones: Capital to start comments, and a full stop to end please.

deBhal:
Thanks, @GaryJones.  I've fixed that comment and several others, and squashed the feedback changes into logical commits.

deBhal:
I've added one more test that checks for translations that are just placeholders ( `'%s` just wastes everyone's time).

I'm not expecting to add any more tests now, but I'll still respond to feedback.

PR #571, opened 2016-06-11
Added support for catching functions that will be deprecated once php7 updates

alexiskulash:
Functions include ereg, eregi, ereg_replace, and split

### DIFF
--- a/WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php
+++ b/WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php
@@ -294,6 +294,38 @@ class WordPress_Sniffs_VIP_RestrictedFunctionsSniff extends WordPress_Sniffs_Fun
 				),
 			),
 
+			'ereg' => array(
+				'type' => 'error',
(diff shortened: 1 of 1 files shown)


alexiskulash:
cc @sboisvert

westonruter:
Are these deprecated functions restricted by VIP?

sboisvert:
Yup we just finished removing them from all the repos. I'll add it to the documentation early next week.

jrfnl:
This PR raises two questions for me:
1. Should cross-PHP version compatibility issues be a concern which should be addressed by WPCS ?
2. And if so, wouldn't it make more sense to incorporate the far more comprehensive  [PHP Compatibility](https://github.com/wimg/PHPCompatibility) sniffs rather than reinvent the wheel ?

sboisvert:
>1. Should cross-PHP version compatibility issues be a concern which should be addressed by WPCS ?
I can't speak for the project as a whole but VIP is moving to requiring PHP7 compatibility, these functions are the ones we've seen used multiple times so it was a sort of quick win.

>2. And if so, wouldn't it make more sense to incorporate the far more comprehensive PHP Compatibility sniffs rather than reinvent the wheel ?

If the project's answer to 1 is yes then I would agree with 2.

PR #587, opened 2016-07-11
Add dev-lib and use for dogfooding rulesets (WordPress-Core) in Travis CI builds

### LINKED ISSUE #256: Settle on coding convention for WPCS snfifs

westonruter:
Fixes #256.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "dev-lib"]
+	path = dev-lib
+	url = https://github.com/xwp/wp-dev-lib.git
+	branch = master
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,12 +31,17 @@ matrix:
     - php: hhvm
     - env: PHPCS_BRANCH=3.0
 
-before_script:
+install:
+    - export DEV_LIB_PATH=dev-lib
+    - export DEV_LIB_SKIP=composer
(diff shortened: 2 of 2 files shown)


JDGrimes:
This appears to be failing on HHVM: https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/143952804

`develop` doesn't currently fail on HHVM, so that would be a regression.

westonruter:
Humm. It seems to be failing during the install step but it doesn't indicate what the failure is. I'm not sure actually if dev-lib has run on the HHVM environment yet.

GaryJones:
Can you please summarise what this change is doing?

westonruter:
It's just using wp-dev-lib's builtin PHP syntax check and PHPCS running. The key feature of wp-dev-lib here is the ability to restrict error reporting to the changes in a PR. However, with #590 it is not necessary since the whole codebase is now getting cleaned.

ntwb:
> `develop` doesn't currently fail on HHVM, so that would be a regression.

Things might be different after #595 which updates the HHVM job to use the latest HHVM

JDGrimes:
@westonruter Feel free to open another PR if you want to still use the dev-lib for syntax checking, etc.

PR #590, opened 2016-07-12
Fix codestyle.

### LINKED ISSUE #256: Settle on coding convention for WPCS snfifs

jrfnl:
Complies with WP-Core.
As there are only two warnings left when run against WP-Extra, I've activated that as well. Might as well lead by example ;-)

Still would need some work to comply with WP-Docs. When run against Docs PHPCS currently yields some 100 or so issues mostly to do with file vs class docblocks.

Also:
* Removed some large commented out code blocks.
* Minor spelling and punctuation corrections.
* Few minor fixes to do with variable comparisons, mainly changing loose to strict comparisons.
* Applied some best practices from the Docs and Extra standards.
* Variable and array value alignment when I saw it.

Closes #256

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -1126,5 +1126,6 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 			}
 		}
 		return $variables;
-	}
+	} // end get_interpolated_variables()
+
 }
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -1,5 +1,4 @@
 <?php
-
 /**
  * Represents a PHP_CodeSniffer sniff for sniffing WordPress coding standards.
  *
@@ -29,9 +28,9 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	 * @var array
 	 */
 	public static $nonceVerificationFunctions = array(
-		'wp_verify_nonce' => true,
+		'wp_verify_nonce'     => true,
 		'check_admin_referer' => true,
-		'check_ajax_referer' => true,
+		'check_ajax_referer'  => true,
 	);
 
 	/**
@@ -42,37 +41,37 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	 * @var array
 	 */
 	public static $escapingFunctions = array(
-		'absint' => true,
-		'esc_attr__' => true,
-		'esc_attr_e' => true,
-		'esc_attr_x' => true,
-		'esc_attr' => true,
-		'esc_html__' => true,
-		'esc_html_e' => true,
-		'esc_html_x' => true,
-		'esc_html' => true,
-		'esc_js' => true,
-		'esc_sql' => true,
-		'esc_textarea' => true,
-		'esc_url_raw' => true,
-		'esc_url' => true,
-		'filter_input' => true,
-		'filter_var' => true,
-		'intval' => true,
-		'json_encode' => true,
-		'like_escape' => true,
-		'number_format' => true,
-		'rawurlencode' => true,
-		'sanitize_html_class' => true,
-		'sanitize_user_field' => true,
-		'tag_escape' => true,
-		'urlencode_deep' => true,
-		'urlencode' => true,
-		'wp_json_encode' => true,
+		'absint'               => true,
+		'esc_attr__'           => true,
+		'esc_attr_e'           => true,
+		'esc_attr_x'           => true,
+		'esc_attr'             => true,
+		'esc_html__'           => true,
+		'esc_html_e'           => true,
+		'esc_html_x'           => true,
+		'esc_html'             => true,
+		'esc_js'               => true,
+		'esc_sql'              => true,
+		'esc_textarea'         => true,
+		'esc_url_raw'          => true,
+		'esc_url'              => true,
+		'filter_input'         => true,
+		'filter_var'           => true,
+		'intval'               => true,
+		'json_encode'          => true,
+		'like_escape'          => true,
+		'number_format'        => true,
+		'rawurlencode'         => true,
+		'sanitize_html_class'  => true,
+		'sanitize_user_field'  => true,
+		'tag_escape'           => true,
+		'urlencode_deep'       => true,
+		'urlencode'            => true,
+		'wp_json_encode'       => true,
 		'wp_kses_allowed_html' => true,
-		'wp_kses_data' => true,
-		'wp_kses_post' => true,
-		'wp_kses' => true,
+		'wp_kses_data'         => true,
+		'wp_kses_post'         => true,
+		'wp_kses'              => true,
 	);
 
 	/**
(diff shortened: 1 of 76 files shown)


WordPress/Sniff.php
```diff
@@ -42,37 +41,37 @@
...
+		'urlencode_deep'       => true,
+		'urlencode'            => true,
+		'wp_json_encode'       => true,
 		'wp_kses_allowed_html' => true,
-		'wp_kses_data' => true,
-		'wp_kses_post' => true,
-		'wp_kses' => true,
+		'wp_kses_data'         => true,
+		'wp_kses_post'         => true,
+		'wp_kses'              => true,
```
JDGrimes: Honestly, I'm not big on these array alignment. It can make things look better, but IMO it isn't necessary when all of the values are currently the same, and it will inevitably get out of sync next time a longer function is added to the list. When that happens we're either going to realign the whole array, or we'll end up with a partly aligned array, which after a while will put us back in square one. :-) I'm not going to complain if others disagree, but I just thought I'd mention it in case anybody wants to give a second opinion.

For the record, the variable alignments look good, I'm just talking about these array alignments.

Or do we still sniff for array alignments? I don't think that we do. Or else maybe I've disabled that (but I don't think so)...

WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
```diff
@@ -10,136 +10,134 @@
...
+		'wakeup',
+		'toString',
+		'set_state',
+		'clone',
+		'invoke',
+		'debugInfo',
+	);
+
+
+	/**
```
JDGrimes: We might as well remove these extra newlines between methods. I don't know that a single blank line is a written part of the standard, but it seems to be the rule that WordPress follows. Most of our sniffs follow it as well, so we might as well be consistent. 😄

WordPress/Sniffs/Arrays/ArrayAssignmentRestrictionsSniff.php
```diff
@@ -84,56 +82,57 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr )
...
-		// $foo['bar'] = $taz;
-		if ( in_array( $token['code'], array( T_CLOSE_SQUARE_BRACKET, T_DOUBLE_ARROW ) ) ) {
-			if ( $token['code'] == T_CLOSE_SQUARE_BRACKET ) {
-				$operator = $phpcsFile->findNext( array( T_EQUAL ), $stackPtr + 1 );
-			} elseif ( $token['code'] == T_DOUBLE_ARROW ) {
+		/*
+		   Covers:
+		   $foo = array( 'bar' => 'taz' );
+		   $foo['bar'] = $taz;
+		 */
```
JDGrimes: It looks like you've converted multi-line comments in some places but not others. Just wondering if there is any rhyme or reason to it. Can't remember if that is part of `extra` or not.

WordPress/Sniff.php
jrfnl: > Or do we still sniff for array alignments?

No, we don't.

WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
jrfnl: It was very inconsistent between files. Most of them had two lines between functions (bleed over from PHPCS files probably) which is why I settled on that.

But you're right that WP mostly uses one line between, though it's not checked for.

So one line it is ?

WordPress/Sniffs/Arrays/ArrayAssignmentRestrictionsSniff.php
jrfnl: > Just wondering if there is any rhyme or reason to it.

I ran the codebase against WP-Docs as well and fixed anything which was a 'quick' fix to make.

In this case - if I remember correctly -, it flagged this as a multi-line comment not using multi-line comment syntax.

WordPress/Sniff.php
GaryJones: I'd prefer alignment for the multiple times code like this will be read by developers, over the relative few instances that someone might need to adjust the spacing because of an addition that has more characters. FWIW, in PHPStorm you can highlight, hit a keystroke (mine is CMD+Shift+L), and they will all align anyway - no need for manual alignment.

WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
GaryJones: Definitely. One never needs more than a single blank line.

There might be a sniff (not WP) that checks for this (or maybe it's a PHPStorm thing) - if so, maybe include this sniff as part of `Extra`?

WordPress/Sniffs/Arrays/ArrayAssignmentRestrictionsSniff.php
GaryJones: I'm good with this change.

WordPress/Sniff.php
grappler: I think it would complicate things if we say we don't do variable alignments if the value is the same otherwise we we do it.

JDGrimes: OK, we can keep them aligned. My issue isn't that it will be trouble to realign them (I use PHPStorm as well), but that doing so makes history harder to follow. But I guess it does make sense to prioritize the most common use, which is devs looking at the lists just as they are.

WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
jrfnl: > One never needs more than a single blank line.

As a dyslectic, I prefer more space anytime which is why I'm quite fond of the WPCS as it's generally speaking quite spacious compared to some.
I always found it weird and kind of inconsistent that the WPCS "more space" philosophy did not go so far as to have the `else` on a new line by itself and two lines between functions.

If given the choice, I will personally always use two lines between functions. For me, it really increases the scan-ability - one line break = within a function, two line break = hang on! new function coming up.

The `Squiz.WhiteSpace.FunctionSpacing` sniff looks for number of lines before and after each function and can be tweaked to look for 1 line rather than 2, so would be a good candidate for potentially adding to `extra` (and I'll just happily turn it off again for any project I work on).

This rule does conflict with the `WordPress.WhiteSpace.ControlStructureSpacing.BlankLineAfterEnd` rule for the last function in a class if the function does not have a `} // end myfunction()` comment.
Still - the fact that it throws a message `Blank line found after control structure` when finding a blank line after the last method in a class might actually be a bug....

jrfnl: Added commit which now consistently applies 'one empty line before and after each function/method'.

I've - for now - resorted to adding `// end` comments after the closing brace of the last method in a class to bypass the above mentioned rule clash with `WordPress.WhiteSpace.ControlStructureSpacing.BlankLineAfterEnd`.

jrfnl:
Added two more commits:

Minor touch-up.

* Fix some missed "class opening brackets on different line" issues. There currently is no sniff checking for this, so visual fix.
* Fixed one class extending the wrong parent.

Documentation fixes.

* Added generic file docblock to sniff files which didn't have one.
* Added/replaced generic file docblock to all unit test files.
* Use `@link` rather than `@see` for links to external documentation about a sniff in the file/class docblock.
* Consistent placement of the `@link` tag if available and pointing to WPCS relevant info.
* Removed `PHP version 5` notes. No longer relevant.

JDGrimes:
As far as I am concerned, this is ready for merge. Any other comments or objections?

GaryJones:
:shipit:

PR #591, opened 2016-07-12
Add sniff for (no) closing PHP tag at end of file to `core` ruleset.

### LINKED ISSUE #588: Check for (no) PHP close tag at end of file ?
While I can't seem to find the rule in the handbook, not having a PHP closing tag if it's at the end of a file is an unwritten rule AFAIK.

So... should we add it ? `PSR2.Files.ClosingTag` should do the trick or alternatively `Zend.Files.ClosingTag`.

As it's not in the handbook, probably in `extra` ?

jrfnl:
Closes #588

Added to `extra` as the handbook mentions it as a best practices, not a **_must do_**.

Ref: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#remove-trailing-spaces

Code style PR has already run against this rule, so should cause no issue when merging in either order.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ php:
 
 env:
   - PHPCS_BRANCH=master
-  - PHPCS_BRANCH=2.2.0
+  - PHPCS_BRANCH=2.4.0
 
 matrix:
   fast_finish: true
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -14,6 +14,7 @@
 
 	<!-- http://make.wordpress.org/core/handbook/coding-standards/php/#remove-trailing-spaces -->
 	<rule ref="Squiz.WhiteSpace.SuperfluousWhitespace"/>
+	<rule ref="PSR2.Files.ClosingTag"/>
 
 	<!-- http://make.wordpress.org/core/handbook/coding-standards/php/#no-shorthand-php-tags -->
 	<rule ref="Generic.PHP.DisallowShortOpenTag"/>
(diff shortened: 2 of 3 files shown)


jrfnl:
Travis run fails as the upstream sniff was introduced in PHPCS > 2.2, making this PR dependent on #566 being merged.

jrfnl:
Looks like the sniff was originally introduced in PHPCS 2.3.0 and made far less buggy in 2.4.0, so any PHPCS minimum version above that would work.

https://github.com/squizlabs/PHP_CodeSniffer/commits/master/CodeSniffer/Standards/PSR2/Sniffs/Files/ClosingTagSniff.php

jrfnl:
Changed the PHPCS minimum requirement to 2.4.0 as per https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/566#issuecomment-232183416

westonruter:
@jrfnl actually, I think this should be in `WordPress-Core` not `WordPress-Extra`. See: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#remove-trailing-spaces

> Remove trailing whitespace at the end of each line of code. Omitting the closing PHP tag at the end of a file is preferred. If you use the tag, make sure you remove trailing whitespace.

jrfnl:
@westonruter I'd be happy to adjust the PR. My reasoning for `extra` was that - in the line you quoted - it says "preferred".

westonruter:
IMHO it should be required to omit.

jrfnl:
Ok. Updated the PR.

PR #592, opened 2016-07-12
Add sniff for bracketed arithmetics to `extra` ruleset.

### LINKED ISSUE #589: Add arithmetic bracketing sniff ?
As operator precedence is a mystery to a lot of programmers, demanding brackets around arithmetics is good practice and prevents bugs.

So, should we add it ? `Squiz.Formatting.OperatorBracket.MissingBrackets` does the trick.

Optionally this sniff could be forked to throw a warning instead of an error.

As it's not in the handbook, should probably go in `extra`.

Opinions ?

jrfnl:
Closes #589

Code style PR has already run against this rule, so should cause no issue when merging in either order.


jrfnl:
Let's close this for now as the sniff would need adjustment for WPCS - see discussion in #589.

PR #596, opened 2016-07-14
Add (new) sniff for class opening brace on same line to `core` ruleset.

jrfnl:
Checks that the opening brace of a class or interface is on the same line as the class declaration.
Also checks that the brace is the last thing on that line and has precisely one space before it.

Review appreciated.

Also: needs testing of the fixer code (doesn't work on Windows so haven't been able to test this myself.


GaryJones:
I think the test for exactly one space before the `{` would already be covered with other sniffs?

If your new sniff is named as checking for opening brace on same line, then that's only what it should cover, nothing else. Keeping the sniffs small and focused allows user flexibility when choosing which they want to use in their project.

As such, isn't there a PHPCS sniff that ensures a certain arrangement of braces? If not, these seems generic enough that the fix should be made there, rather than here, as it's not really WP specific.

jrfnl:
> I think the test for exactly one space before the `{` would already be covered with other sniffs?

It is not.

> If your new sniff is named as checking for opening brace on same line, then that's only what it should cover, nothing else. Keeping the sniffs small and focused allows user flexibility when choosing which they want to use in their project.

The messages have all been given a code, so people can disable them modularly.

What about if I change the sniff name to something like `ClassOpeningStatement` ?

> As such, isn't there a PHPCS sniff that ensures a certain arrangement of braces?

There are a few, but all of them enforce the opposite of what the WP standard is (braces on new line).

> If not, these seems generic enough that the fix should be made there, rather than here, as it's not really WP specific.

Makes sense to me, I'll open an issue there to see if there is interest.

All the same, it might be a while before something like this gets merged into PHPCS and then another while before the WPCS PHPCS dependency minimum version goes up.... so should this be maybe be merged anyway ? It can always be removed later if and when it would make it into PHPCS.

jrfnl:
Upstream issue opened: https://github.com/squizlabs/PHP_CodeSniffer/issues/1068

jrfnl:

jrfnl:
Renamed the sniff as per https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/596#issuecomment-232587368

jrfnl:
Got a positive response regarding this sniff upstream: https://github.com/squizlabs/PHP_CodeSniffer/issues/1068

PR has been opened: https://github.com/squizlabs/PHP_CodeSniffer/pull/1070

jrfnl:
Anything that would still need to change for this to be mergeable ?

How do we move forward on this ?

Wait for upstream ? May take quite a while as it looks like they're currently focused on PHPCS 3 and not merging things for 2.

Merge it and remove when it would be merged upstream and the WPCS minimum PHPCS version has caught up ?

JDGrimes:
> Merge it and remove when it would be merged upstream and the WPCS minimum PHPCS version has caught up ?

+1. I'd like to see a couple of tests added though, like:

```php
class A_Class_With_Really_Long_Name
    extends Another_Class_With_A_Really_Long_Name {

}

class A_Class_With_Really_Long_Name_2
    extends Another_Class_With_A_Really_Long_Name
    implements Some_Interface_With_A_Long_Name,
          Another_Interface_With_A_Long_Name {

}
```

jrfnl:
> I'd like to see a couple of tests added though, like:

@JDGrimes I'll add these tests. I'm in doubt about what the sniff should say about these though as the sniff is really only about the brace. Should it give an error because the brace is not on the same line as the `class` keyword or should these be passed as ok as the brace is on the same line as the last part of the class opening statement ?
Please advise.

JDGrimes:
> should these be passed as ok as the brace is on the same line as the last part of the class opening statement ?

Yes. Sorry I wasn't clear. 

I think this is the correct behavior because AFAIK WordPress doesn't really address the issue of wrapping here, so it should be permitted. Especially because I do it. 😈

GaryJones:
Yes, wrapping should be allowed. Though not explicit, line-length suggestions would come into play here.

jrfnl:
> I think this is the correct behavior because AFAIK WordPress doesn't really address the issue of wrapping here, so it should be permitted.

Makes sense to me, though this will involve a rewrite of the part of the sniff which checks this.

jrfnl:
Ok, sorted. Unit tests added and minor code fix to allow for multi-line class declarations.

jrfnl:
FYI: Upstream PR (including the last change) has just been merged, but - similar to the other PR - is not in a released version yet and when it is, the WPCS PHPCS minimum version requirement would need to be upped.
So I suggest merging this for now and removing it again at a later point in time.

PR #597, opened 2016-07-14
Fix bug in control structures whitespace sniff triggered on last method of class

jrfnl:
As reported here:

> This rule does conflict with the `WordPress.WhiteSpace.ControlStructureSpacing.BlankLineAfterEnd` rule for the last method in a class if the function does not have a `} // end myfunction()` comment.
Still - the fact that it throws a message `Blank line found after control structure` when finding a blank line after the last method in a class might actually be a bug....

Added separate commits for unit test and fix to let travis demonstrate the bug and the fix.


jrfnl:
Freaky weird that the bug is not triggered by the unit test when running this on travis. I can consistently reproduce this on my machine on several different PHP versions.

jrfnl:
Rebased for merge conflict.

PR #599, opened 2016-07-14
Add new sniff to check whether a hook name is valid.

jrfnl:
Covers the `action` part of this rule in the handbook:
> Use lowercase letters in variable, **action**, and function names (never camelCase). Separate words via underscores.

Review appreciated.

Also: I've made non-lowercase hook names an `error` and non-underscore punctuation a `warning`. Should that also be an `error` ?

### DIFF
--- a/WordPress/Sniffs/NamingConventions/ValidHookNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidHookNameSniff.php
@@ -10,6 +10,11 @@
 /**
  * Use lowercase letters in action and filter names. Separate words via underscores.
  *
+ * This sniff is only testing the hook invoke functions as when using 'add_action'/'add_filter'
+ * you can't influence the hook name.
+ *
+ * Hook names invoked with `do_action_deprecated()` and `apply_filters_deprecated()` are ignored.
+ *
  * @link     https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#naming-conventions
  *
  * @category PHP
@@ -53,21 +58,6 @@ class WordPress_Sniffs_NamingConventions_ValidHookNameSniff implements PHP_CodeS
 	protected $punctuation_regex = '`[^\w%s]`';
 
 	/**
-	 * Functions we're interested in.
-	 *
-	 * Only testing the hook call functions as when using 'add_action'/'add_filter' you can't influence
-	 * the hook name.
-	 *
-	 * @var array
-	 */
-	protected $hook_functions = array(
-		'do_action',
-		'do_action_ref_array',
-		'apply_filters',
-		'apply_filters_ref_array',
-	);
-
-	/**
 	 * Register this sniff.
 	 *
 	 * Prepares the punctuation regex and returns an array of tokens this test wants to listen for.
@@ -96,7 +86,12 @@ class WordPress_Sniffs_NamingConventions_ValidHookNameSniff implements PHP_CodeS
 		$regex  = $this->prepare_regex();
 
 		// Check if one of the hook functions was found.
-		if ( false === in_array( $tokens[ $stackPtr ]['content'], $this->hook_functions, true ) ) {
+		if ( ! isset( WordPress_Sniff::$hookInvokeFunctions[ $tokens[ $stackPtr ]['content'] ] ) ) {
+			return;
+		}
+
+		// Ignore deprecated hook names.
+		if ( strpos( $tokens[ $stackPtr ]['content'], '_deprecated' ) > 0 ) {
 			return;
 		}
 
--- a/WordPress/Sniffs/NamingConventions/ValidHookNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidHookNameSniff.php
@@ -1,0 +1,247 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     https://make.wordpress.org/core/handbook/best-practices/coding-standards/
+ */
+
+/**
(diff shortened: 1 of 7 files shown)


GaryJones:
Slight aside from here, but maybe worth adding as a valid unit test to check the new sniff doesn't complain, is hook names with `/` in them, like `acf/save_post`.

jrfnl:
> adding as a valid unit test to check the new sniff doesn't complain, is hook names with / in them

@GaryJones That would throw a warning based on the below phrase in the rule as stated in the handbook:

> Separate words via underscores.

I choose `warning` for that part as core often doesn't comply with this rule either - they often use dashes instead of underscores - and changing hook names would break backward compatibility.

westonruter:
See also https://github.com/humanmade/WordPress-Importer/pull/10#issuecomment-157680263

jrfnl:
@westonruter Same goes there - it would throw a warning, not an error.

jrfnl:

GaryJones:
It's not so much a separate word, as a prefix.

JDGrimes:
I wonder if core is actually strict about this rule. Personally I like to use a hyphen or something other than an underscore in dynamic hook names to prevent potential conflicts:

```php
// These could conflict unexpectedly:
do_action( "prefix_{$var}" );
do_action( 'prefix_something' );

// This prevents that:
do_action( "prefix-{$var}" );
```

But if core is going to be strict about this, then I guess we do need to add this check to `core`, and if some of us don't want to use it we can silence it.

jrfnl:
> I wonder if core is actually strict about this rule. Personally I like to use a hyphen or something other than an underscore in dynamic hook names to prevent potential conflicts:

I had a look at [this hook database](http://adambrown.info/p/wp_hooks/hook) when I created the sniff and based on visual inspection, it seemed that core *is* strict about the lower case - only 13 exceptions out of 2000+ hook names AFAICS.
Exceptions I found were:
* `attachment_innerHTML`
* `xmlrpc_call_success_blogger_deletePost` up to `xmlrpc_call_success_wp_newComment` (12 hooks around line 2088)

Core however is inconsistent about applying the second part of the rule about using `_` for punctuation. Dashes are most frequently used as an alternative often even combined with underscore within the same hook name. Often also when they refer to a function name within the hook - the 'words' within the function name is then separated with dashed while the other 'words' are separated with underscores.

All the more reason why I made a difference in error levels between the two parts. Error for non-lowercase, Warning for non-underscore punctuation.

jrfnl:
Just thinking - I could add an option for people to provide one-character punctuation exceptions via the `phpcs.xml` config.

What do you think ?

jrfnl:
New commit added - this creates the ability to provide additional allowed word delimiters via `phpcs.xml` or via the inline `@codingStandardsChangeSetting` directive.

Example usage:
```xml
<rule ref="WordPress.NamingConventions.ValidHookName">
   <properties>
     <property name="additionalWordDelimiters" value="-"/>
   </properties>
</rule>
```

Commit includes unit tests.

I've still left the error level for the punctuation at `warning` or should this now be raised to `error` ?

WordPress/Tests/NamingConventions/ValidHookNameUnitTest.php
```diff
@@ -0,0 +1,128 @@
...
+					74 => 1,
+					75 => 1,
+					76 => 1,
+					77 => 1,
+					78 => 1,
+					79 => 1,
+					80 => 1,
+					81 => 1,
+				);
+				break;
```
GaryJones: This `break`, and the `break` below are redundant (dead code).

jrfnl: Good point. Artifact left from upstream code style. Removed.

WordPress/Sniffs/NamingConventions/ValidHookNameSniff.php
```diff
@@ -0,0 +1,289 @@
...
+	 * the hook name.
+	 *
+	 * @var array
+	 */
+	protected $hook_functions = array(
+		'do_action',
+		'do_action_ref_array',
+		'apply_filters',
+		'apply_filters_ref_array',
+	);
```
JDGrimes: We generally make the function names keys so that `isset()` can be used instead of `in_array()`. Also, most of these lists should go in the main sniff class so that they can be reused across sniffs.

Also, `do_action_deprecated()` and `apply_filters_deprecated()` could be added. https://core.trac.wordpress.org/changeset/37861

jrfnl: > We generally make the function names keys so that isset() can be used instead of in_array().

Good point, changing now.

> Also, most of these lists should go in the main sniff class so that they can be reused across sniffs.

I was thinking about that, but then couldn't think of a re-use case. Will move them anyway.

> Also, do_action_deprecated() and apply_filters_deprecated() could be added. https://core.trac.wordpress.org/changeset/37861

Nice one! Hadn't seen that commit yet. Darn! This means I'm going to add that to the `Debug Bar` as well (https://github.com/dd32/debug-bar/pull/17).

For this particular sniff, I'm not sure they should be added. Should deprecated hook names also comply with the naming conventions ? They might have been deprecated because of hook name improvement.
What do you think ?

JDGrimes: > Should deprecated hook names also comply with the naming conventions ? They might have been deprecated because of hook name improvement.
> What do you think ?

Good point, I hadn't thought of that. I guess it really doesn't make sense to flag something that is deprecated for any naming convention violations. So yeah, skip them for this sniff.

jrfnl:
Rebased to solve merge conflict.

jrfnl:
Anyone got anymore feedback on this ?

Issue #90, opened 2013-10-11
Flag use of extract()


westonruter:
- [x] Flag any instance of the function `extract()`

> <h4 id="using-extract">Using extract() <a href="http://vip.wordpress.com/documentation/code-review-what-we-look-for/#using-extract" class="anchor">#</a></h4>
> 
> <p><code>extract()</code> should never be used because it is too opaque and difficult to understand how it will behave under a variety of inputs. It makes it too easy to introduce new variables into a function’s scope without knowing about it, potentially causing unintended and difficult to debug conflicts.</p>

Issue #582, opened 2016-07-11
Cleanup of discouraged functions 


grappler:
We should have better separation of the different function types

A possible list
- [ ] PHPRestricted (ereg, eval and the like)
- [ ] PHPAlternativePreferred/PHPDiscouraged (create_function, curl, split and the like)
- [ ] PHPDeprecated (ereg and the like)
- [ ] PHPDevelopment (print_r and the like)
- [ ] WPDiscouraged (query_posts and the like)
- [ ] WPDeprecated (get_link and the like)
- [ ] WPCacheRestricted (get_pages, get_post_terms and the like)
- [ ] WPAlternatives (json_encode, parse_url and the like)
- [ ] VIPAlternatives (all those which have wpcom_...() alternative functions)
- [ ] BatcacheRestricted (cookies and the like)
- [ ] Theme (a few theme specific sniffs)

Currently we have the following sniffs
- [WordPress.PHP.DiscouragedFunctions](https://github.com/WPTRT/WordPress-Coding-Standards/blob/feature/theme-review-sniffs/WordPress/Sniffs/PHP/DiscouragedFunctionsSniff.php) related: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/581
- [WordPress.VIP.FileSystemWritesDisallow](https://github.com/WPTRT/WordPress-Coding-Standards/blob/feature/theme-review-sniffs/WordPress/Sniffs/VIP/FileSystemWritesDisallowSniff.php)
- [WordPress.VIP.RestrictedFunctions](https://github.com/WPTRT/WordPress-Coding-Standards/blob/feature/theme-review-sniffs/WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php)
- [WordPress.VIP.SessionFunctionsUsage](https://github.com/WPTRT/WordPress-Coding-Standards/blob/feature/theme-review-sniffs/WordPress/Sniffs/VIP/SessionFunctionsUsageSniff.php)

Some of the functions being checked are duplicated or would better in another sniff.that is more general or a new sniff.

jrfnl:
Marked #583 as duplicate.

> Some of the functions being checked are duplicated.

I'm not too worried about lines triggering several different errors from different sniff for different reasons. Sniffs can be disabled on a per sniff per project basis, so if there are several reasons a function should not be used, I'm quite happy to have it in several different sniffs.

grappler:
[...]

PR #604, opened 2016-07-15
Separate out the sniffing for `extract()`.

jrfnl:
Currently only VIP checked for the usage of `extract()`, even though it is an explicit rule for core.

Ref: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#dont-extract

Solves #90 for real ;-)

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -111,5 +111,6 @@
 	<rule ref="WordPress.WhiteSpace.CastStructureSpacing"/>
 	<rule ref="WordPress.PHP.YodaConditions"/>
 	<rule ref="WordPress.WP.I18n"/>
+	<rule ref="WordPress.Functions.DontExtract"/>
 
 </ruleset>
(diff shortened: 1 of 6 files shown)


JDGrimes:
Why not give the sniff a more generic name, so that other functions can be added in the future? Maybe just `Restricted` instead of `DontExtract`?

JDGrimes:
I guess this relates to #582.

jrfnl:
> Why not give the sniff a more generic name, so that other functions can be added in the future

@JDGrimes I followed the handbook - it has a separate section all to itself, so it seemed appropriate to name the rule after the section.

I'd rather have these kind of sniffs more modular and pluggable as - yes - noted in #582, rather than have all of them jumbled together.
That will also make for less backward compatibility breaks for these kind of sniffs.

JDGrimes:
@jrfnl I've decided that you are right and we should keep these sniffs more modular. However, I would add that this will impact performance as we reprocess the tokens for every single child sniff. This ought not so to be. We should refactor the parent sniff so that it only processes each token once. It could keep an index of functions that it found, and then if a child sniff flagged a particular function, it would have stack pointer for all of the times in the file that that function was used already saved from when it processed the functions previously. (Or possibly there is a better way to do this?) Anyway, that is for its own issue.

jrfnl:
@JDGrimes Made a very small start with some efficiency improvement for the AbstractFunctionRestrictionsSniff in #615

Issue #575, opened 2016-06-27
Suggestion: PHP Compatibility checks


jrfnl:
Would it be an idea to suggest for people to add the [PHP compatibility](https://github.com/wimg/phpcompatibility) sniffs ?

The PHPCompatibility sniffs check for cross-PHP version compatibility - and while not 100% feature complete - does a good job detecting usage of functions/classes/parameters etc which are not cross-version compatible.

PR #608, opened 2016-07-16
Better cover the regex related rules.

jrfnl:
From the handbook:
> Perl compatible regular expressions (PCRE, preg_ functions) should be used in preference to their POSIX counterparts. Never use the /e switch, use preg_replace_callback instead.
>
> It’s most convenient to use single-quoted strings for regular expressions since, contrary to double-quoted strings, they have only two metasequences: \' and \\.

Ref: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#regular-expressions

This can be split into three distinct rules:
1. Don't use `POSIX` functions
2. Don't use the `/e` modifier
3. Prefer single quoted stings.

Rule 1 was up to now not covered in core. Parts of it were covered in `extra` and in `vip`, but the implementations were inconsistent and both were incomplete.
Rule 2 was not covered at all.
Rule 3 is already covered by the single vs double quotes sniff.

This PR fixes the issues with covering rule 1 and 2 with the following two commits:

Split off `POSIX` Functions restrictions to its own sniff.

* Removes similar checks from existing sniffs (`PHP.DiscouragedFunctions` and `VIP.RestrictedFunctions`).
* Makes sure that *all* functions in the `POSIX` extension are covered by this check - they previously were not.
* Add unit tests for the new sniff covering *all* `POSIX` functions.

Copy in the `PregReplaceEModifierSniff` from the `PHPCompatibility` Sniff library.

* Add a new `PHPCompatibility` folder to the `Sniff` folder.
* Copy in the `PregReplaceEModifierSniff`.
* Copy in the related parent class `PHPCompatibility_Sniff`.
* Renamed both classes to follow the PSR1 autoloading pattern.
* Added a readme to the `PHPCompatibility` sniff folder about its usage.
* Excluded the `PHPCompatibility` folder from the PHPCS code style checks - this way we can just drop-in replace the file (ok, except for the class renaming) when there is a new version available.

### DIFF
--- a/WordPress/Sniffs/PHP/DiscouragedFunctionsSniff.php
+++ b/WordPress/Sniffs/PHP/DiscouragedFunctionsSniff.php
@@ -31,13 +31,6 @@ class WordPress_Sniffs_PHP_DiscouragedFunctionsSniff extends Generic_Sniffs_PHP_
 	 * @var array(string => string|null)
 	 */
 	public $forbiddenFunctions = array(
-		// Deprecated.
-		'ereg_replace'             => 'preg_replace',
-		'ereg'                     => null,
-		'eregi_replace'            => 'preg_replace',
-		'split'                    => null,
-		'spliti'                   => null,
-
 		// Development.
 		'print_r'                  => null,
 		'debug_print_backtrace'    => null,
--- a/WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php
+++ b/WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php
@@ -285,38 +285,6 @@ class WordPress_Sniffs_VIP_RestrictedFunctionsSniff extends WordPress_AbstractFu
 				),
 			),
 
-			'ereg' => array(
-				'type'      => 'error',
-				'message'   => '%s is prohibited, please use preg_match() instead. See http://php.net/manual/en/function.ereg.php',
-				'functions' => array(
-					'ereg',
-				),
-			),
-
-			'eregi' => array(
-				'type'      => 'error',
-				'message'   => '%s is prohibited, please use preg_match() with i modifier instead. See http://php.net/manual/en/function.eregi.php',
-				'functions' => array(
-					'eregi',
-				),
-			),
-
-			'ereg_replace' => array(
-				'type'      => 'error',
-				'message'   => '%s is prohibited, please use preg_replace() instead. See http://php.net/manual/en/function.ereg-replace.php',
-				'functions' => array(
-					'ereg_replace',
-				),
-			),
-
-			'split' => array(
-				'type'      => 'error',
-				'message'   => '%s is prohibited, please use explode() or preg_split() instead. See http://php.net/manual/en/function.split.php',
-				'functions' => array(
-					'split',
-				),
-			),
-
 			'runtime_configuration' => array(
 				'type'      => 'error',
 				'message'   => '%s is prohibited, changing configuration at runtime is not allowed on VIP Production.',
--- a/WordPress/Tests/PHP/POSIXFunctionsUnitTest.inc
+++ b/WordPress/Tests/PHP/POSIXFunctionsUnitTest.inc
@@ -1,0 +1,26 @@
+<?php
+
+// These are ok.
+$regex = preg_quote( 'cool', '#' ); // Good
+$title = preg_match( 'cool', get_the_title() ); // Good
+$title = preg_match_all( 'cool', get_the_title(), $matches ); // Good
+$title = preg_replace( 'cool', 'not cool', get_the_title() ); // Good
+$title = preg_replace_callback( 'cool', 'callback_function', get_the_title() ); // Good
+$title = preg_split( 'cool', get_the_title() ); // Good
+
+
+// These should all trigger an error.
+if ( ereg( '[A-Za-z]+', $title, $regs ) ) // Bad, ereg deprecated. Use preg_match instead.
+	die( $regs );
+
+if ( eregi( '[a-z]+', $title, $regs ) ) {} // Bad, eregi deprecated. Use preg_match instead.
+
+$title = ereg_replace( 'cool', 'not cool', get_the_title() ); // Bad, ereg_replace has been deprecated. Use preg_replace instead.
+
+$title = eregi_replace( 'cool', 'not cool', get_the_title() ); // Bad, eregi_replace also deprecated. Use preg_replace instead.
+
+list( $year, $month, $day ) = split( ':', $date ); // Bad, split has been deprecated. Use preg_split or explode instead.
+
+$title_parts = spliti( ' ', get_the_title(), 4 ); // Bad, spliti also deprecated. Use preg_split instead.
+
+sql_regcase( 'Foo - bar.'); // Bad. Deprecated.
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -118,6 +118,6 @@
 	<rule ref="WordPress.DB.RestrictedFunctions"/>
 	<rule ref="WordPress.DB.RestrictedClasses"/>
 
-		<rule ref="WordPress.PHP.POSIXFunctions" />
+	<rule ref="WordPress.PHP.POSIXFunctions" />
 
 </ruleset>
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -116,7 +116,8 @@
 	<rule ref="WordPress.Functions.DontExtract"/>
 	<rule ref="WordPress.NamingConventions.ValidHookName"/>
 	<rule ref="WordPress.DB.RestrictedFunctions"/>
-
 	<rule ref="WordPress.DB.RestrictedClasses"/>
+
+		<rule ref="WordPress.PHP.POSIXFunctions" />
 
 </ruleset>
--- a/WordPress/Sniffs/PHP/POSIXFunctionsSniff.php
+++ b/WordPress/Sniffs/PHP/POSIXFunctionsSniff.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     https://make.wordpress.org/core/handbook/best-practices/coding-standards/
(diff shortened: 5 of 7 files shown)


WordPress/Sniffs/PHP/DiscouragedFunctionsSniff.php
```diff
@@ -31,13 +31,6 @@ class WordPress_Sniffs_PHP_DiscouragedFunctionsSniff extends Generic_Sniffs_PHP_
 	 * @var array(string => string|null)
 	 */
 	public $forbiddenFunctions = array(
-		// Deprecated.
```
GaryJones: Maybe leave a comment that this is handled with the new sniff?

WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php
```diff
@@ -293,38 +293,6 @@ public function getGroups() {
 				),
 			),
 
-			'ereg' => array(
```
GaryJones: Maybe leave a comment that this is handled with the new sniff?

WordPress/Tests/PHP/POSIXFunctionsUnitTest.inc
```diff
@@ -0,0 +1,26 @@
...
+
+$title = ereg_replace( 'cool', 'not cool', get_the_title() ); // Bad, ereg_replace has been deprecated. Use preg_replace instead.
+
+$title = eregi_replace( 'cool', 'not cool', get_the_title() ); // Bad, eregi_replace also deprecated. Use preg_replace instead.
+
+list( $year, $month, $day ) = split( ':', $date ); // Bad, split has been deprecated. Use preg_split or explode instead.
+
+$title_parts = spliti( ' ', get_the_title(), 4 ); // Bad, spliti also deprecated. Use preg_split instead.
+
+sql_regcase( 'Foo - bar.'); // Bad. Deprecated.
```
GaryJones: I don't see any reference to this above?

jrfnl: *grin* yeah, that's because the previous sniffs weren't complete.
The handbook talks about the whole POSIX extension and that's another - albeit rarely used - function from that group: http://php.net/manual/en/ref.regex.php
As the POSIX extension was deprecated in 5.3 and removed in 7.0, being complete seemed like the right thing to do.

Or do you mean in the sniff itself ? It's there - in the first group of functions. I grouped them based on the advised replacement function.

WordPress/Sniffs/PHP/DiscouragedFunctionsSniff.php
jrfnl: In the code will become very untidy in combination with the upcoming #582 work. I'd rather suggest at some point also re-ordering the `vip` ruleset according to their documentation and making a note in the ruleset about rules like this which are already covered in `core`. Ok ?

JDGrimes:
Is there no way that we can just have the [PHPCompatibility sniff library](https://github.com/wimg/PHPCompatibility) as a dependency? Maybe it is best to just copy the sniff over, but I shy away from anything that might require routine maintenance. 😄

jrfnl:
> Is there no way that we can just have the PHPCompatibility sniff library as a dependency?

See #575 as well where I originally suggested that.

The only way I can think of would be to make it a dependency via composer and add it to the ruleset. Not sure how well that would work for the re-hashing to get PHPCS to recognize the installed standards, but I'd leave that up to @Rarst to advise on.

This is not fool-proof no matter what. Not everyone uses Composer, so this could mean people will start complaining about WPCS not working as it breaks on the missing ruleset.

> Maybe it is best to just copy the sniff over, but I shy away from anything that might require routine maintenance. 

As for now, for the handbook coverage, we'd only need the one sniff, that seemed like the simplest way to handle it.
As noted above, I've excluded it from the WPCS code standards check to make copy/paste maintenance as easy as possible.

If/when WPCS would want to start incorporating more sniffs from PHPCompatibility, then this should be revisited IMHO..

JDGrimes:
> If/when WPCS would want to start incorporating more sniffs from PHPCompatibility, then this should be revisited IMHO..

OK, sounds good to me.

jrfnl:
Rebased for merge conflicts.

Rarst:
I am concerned about licensing here, that library is LGPL. By copy/pasting pieces of it here (into MIT project) we are making a potential mess of that.

jrfnl:
Would it help if we had formal permission of the project owner ?

Rarst:
Not really, it's just a... GPL mess in such case. Using it as a library would be _clear_ separation of stuff, which LGPL allows without issue.

jrfnl:
P.S.: current travis failure has nothing to do with the PR, but hick-up in travis itself.

jrfnl:
FYI: Working on an alternative solution for the `e` modifier sniff.

jrfnl:
What with the upcoming release and the work @grappler is doing in #633, I'm removing the commit which added the `e` modifier sniff from PHPCompatibility for now, so that the splitting off of the POSIX functions to their own file can be merged.

I'm working on an abstract class to detect regular expressions and concrete classes for the `e` modifier sniff and best practices sniffs, but that will need more time and more testing, so won't be ready for the 0.10.0 release.

JDGrimes:
@jrfnl Just to confirm—this is ready for merge?

jrfnl:
@JDGrimes Absolutely, it now only contains the splitting off of the POSIX check from the VIP sniff to an individual sniff and adding that rule to the `core` ruleset where it belongs.

The other part of the regex rules will be covered in a separate PR.

Rarst:
To be clear phpcombatibility got thrown out of this, so I can stop worry about licensing stuff? ;)

jrfnl:
@Rarst You can (for now) ;-)

jrfnl:
All and @Rarst in particular:

I've been doing quite some work on the PHPCompatibility sniff library in the mean time and have created a number of utility functions which would also be useful for WPCS and could get rid of a sh*tload of semi-duplicate code and allow us to check certain things with higher precision.

As I created these functions - though sometimes based on existing code - and contributed them to PHPCompatibility, will it be ok license-wise if I port them over to WPCS ?

At a later time - once these functions have been used a bit more and gone through the wrangler, I'd even like to PR them to PHPCS itself.

Rarst:
> As I created these functions - though sometimes based on existing code - and contributed them to PHPCompatibility, will it be ok license-wise if I port them over to WPCS ?

If by "based on existing code" you mean copy/paste/modify of its code then it's a GPL derivative work and ugh ugh ugh mess for us.

I am not familiar with specific case of LGPL into GPL move to say confidently about it. As above it would be clearly fine if LGPL lib was kept as dependency, but I don't really know about mixing their code.

jrfnl:
By "based on existing code", I mean inspired by. The actual code was rewritten.

All the same, I find this whole licensing thing very confusing. If **_I_** write code and choose to contribute the same code to three different projects, it now sounds like the licensing of the first project to merge the code would rule over it all. That just feels wrong and discourages from contributing at all.

I also wonder if code snippets like that can be licensed or even copyrighted at all, after all, writing it from scratch someone else will just end up with 90% the same code with maybe some minor personalizations. 
Licensing IMHO is more about the library as a whole and the concept of it, rather than bits and pieces of the code itself (though I know that's not how it works, but it is how I think it should work). 

Being able to re-use bits and pieces will encourage best of breed coding much more than everyone re-inventing the wheel.

On a side-note: I remember seeing somewhere a note in the WPCS code base about a piece of code which was taken from WP Core. As WP Core is GPL, according to the licensing, that code shouldn't have been allowed in in that case. Correct ?

Rarst:
> By "based on existing code", I mean inspired by. The actual code was rewritten.

_If_ you wrote the code from scratch then it's fine. It's _your_ work to be licensed as you see fit.

> If I write code and choose to contribute the same code to three different projects, it now sounds like the licensing of the first project to merge the code would rule over it all.

Nope. Since it's your code you would be free to contribute it to three different project and _license_ it differently for each one. _After_ the code is contributed to the project it's subject to its license, but _you_ as copyright holder of _your_ code are free to license it differently for another project.

> I also wonder if code snippets like that can be licensed or even copyrighted at all, after all, writing it from scratch someone else will just end up with 90% the same code with maybe some minor personalizations. 
> Licensing IMHO is more about the library as a whole and the concept of it, rather than bits and pieces of the code itself (though I know that's not how it works, but it is how I think it should work).

Yes, it's vague and often people close eyes at technicalities. My _completely personal_ rule of a thumb — copy/paste of actual code is a line behind which it becomes a clear unarguable derivative code. Many people (WordPress, _cough_) would disagree though and draw a line differently.

> On a side-note: I remember seeing somewhere a note in the WPCS code base about a piece of code which was taken from WP Core. As WP Core is GPL, according to the licensing, that code shouldn't have been allowed in in that case. Correct ?

Correct. If it's still the case we should get rid of it.

jrfnl:
@Rarst Thanks for your detailed reply! Will keep this in mind :+1:

khacoder:
So we find ways to not do something rather then ways to do something?

What about dual licensing code sniffer to make it easier to expand it?

Rarst:
> So we find ways to not do something rather then ways to do something?

Welcome to GPL.

> What about dual licensing code sniffer to make it easier to expand it?

Dual licensing makes things harder, not easier. Now you have _two_ licenses to be compatible with _at the same time_.

khacoder:
Actually no :) Dual licensing means either one applies, from everything that I have read.

The user would pick the one to use. In the case of GPLv2 or later versus MIT, WordPress would then pick GPLv2 or later.

Rarst:
> Actually no :) Dual licensing means either one applies, from everything that I have read.

Yes, user can pick which license to apply, good for them. And user can do that _because_ project is obliged to have project's code comply with _both_.

Think about it. I don't like GPL? I put GPL into dual license project and say it's now MIT... Doesn't work.

khacoder:
So what code in Code Sniffer is MIT specific?

Rarst:
> So what code in Code Sniffer is MIT specific?

All of it, as long as that's the license project declares.

I am not too keen to continue general licencing chatter in this ticket. If you have specific licensing concerns or proposals for the project please open a dedicated issue.

khacoder:
Nah I won't pursue it but thanks for asking :)

PR #609, opened 2016-07-16
Fix bug in function name comparison.

jrfnl:
`AbstractFunctionRestrictionsSniff` did not allow for case-insensitive comparison of function names.

> **Note**: Function names are case-insensitive, though it is usually good form to call functions as they appear in their declaration.

Ref: http://php.net/manual/en/functions.user-defined.php

Also:
* Better be safe than sorry, so `preg_quote()`-ing the function names passed to this class.
* Minor memory management improvement: no need to remember the matched function name.

The first commit adds a unit test proving the bug. The second commit fixes the bug.

### DIFF
--- a/WordPress/AbstractFunctionRestrictionsSniff.php
+++ b/WordPress/AbstractFunctionRestrictionsSniff.php
@@ -105,10 +105,10 @@ abstract class WordPress_AbstractFunctionRestrictionsSniff implements PHP_CodeSn
 				continue;
 			}
 
-			$functions = implode( '|', $group['functions'] );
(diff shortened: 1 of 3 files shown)


jrfnl:
Interesting. I just found another bug in this class which I had apparently accidentally fixed as well.

The wildcard replacement was not just replacing the `*`, but also the character before it. While for now, wildcard function groups was only used for `curl_*`, I expect this will be used more soon.

Example of the effect of the bug:
`curlyhair();` should not match, but did.

Added unit test to that effect.

PR #612, opened 2016-07-17
Disallow direct database calls.

jrfnl:
From the handbook:
> Avoid touching the database directly. If there is a defined function that can get the data you need, use it. Database abstraction (using functions instead of queries) helps keep your code forward-compatible and, in cases where results are cached in memory, it can be many times faster.

Ref: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#database-queries

This section was, AFAICS, so far not covered.

This PR covers checking for any calls made to PHP functions which can approach mysql type databases.
If any such function is found, an `error` will be thrown.

In a future PR, a similar check should be created for calls to native PHP DB Classes, such as PDO.

To that end, I've created a new sniff sub-folder called `DB` to hold both this as well as the class check sniff at a later date.

------
FYI: travis will fail on this PR because of the second bug in #609 as described here: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/609#issuecomment-233163109

Once #609 is merged and this would be rebased, the unit tests will pass.

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -113,4 +113,6 @@
 	<rule ref="WordPress.WP.I18n"/>
 	<rule ref="WordPress.Functions.DontExtract"/>
 
+	<rule ref="WordPress.DB.RestrictedFunctions"/>
+
 </ruleset>
--- a/WordPress/Sniffs/DB/RestrictedFunctionsSniff.php
+++ b/WordPress/Sniffs/DB/RestrictedFunctionsSniff.php
@@ -1,0 +1,59 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @link     https://make.wordpress.org/core/handbook/best-practices/coding-standards/
+ */
+
+/**
+ * Verifies that no database related PHP functions are used.
+ *
+ * "Avoid touching the database directly. If there is a defined function that can get
+ *  the data you need, use it. Database abstraction (using functions instead of queries)
+ *  helps keep your code forward-compatible and, in cases where results are cached in memory,
+ *  it can be many times faster."
+ *
+ * @link     https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#database-queries
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @author   Juliette Reinders Folmer <wpplugins_nospam@adviesenzo.nl>
+ */
+class WordPress_Sniffs_DB_RestrictedFunctionsSniff extends WordPress_AbstractFunctionRestrictionsSniff {
+
+	/**
+	 * Groups of functions to restrict.
+	 *
+	 * Example: groups => array(
+	 * 	'lambda' => array(
+	 * 		'type'      => 'error' | 'warning',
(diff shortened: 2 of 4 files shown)


GaryJones:
I'm not sure this should be an error, as sometimes there might be no choice but to craft a custom call, though I guess where there is no choice but to do direct DB calls, this sniff could be disabled.
Nothing covering PDO or PostgreSQL?

jrfnl:
> I'm not sure this should be an error, as sometimes there might be no choice but to craft a custom call, though I guess where there is no choice but to do direct DB calls, this sniff could be disabled.

Or selectively disabled for one particular file.

> Nothing covering PDO or PostgreSQL?

See my remark above:
> > In a future PR, a similar check should be created for calls to native PHP DB Classes, such as PDO.

I'm working on creating an abstract sniff to check for classes and class related calls which can then be used as the basis for this.

JDGrimes:
> I'm not sure this should be an error

IMO using anything beside `$wpdb` to touch the database should definietely be an error.

Actually though, I think it is somewhat ambiguous exactly what that passage in the handbook means. At first I assumed that it was talking about using `$wpdb` directly, rather than, e.g., `get_posts()`. I'm really not sure, as there are no examples there. That'd be another check for another sniff though, if so.

jrfnl:
> Actually though, I think it is somewhat ambiguous exactly what that passage in the handbook means.

Maybe I should have also included the second paragraph from the handbook:
> If you must touch the database, get in touch with some developers by posting a message to the wp-hackers mailing list. They may want to consider creating a function for the next WordPress version to cover the functionality you wanted.

To me, that implied - touching the DB using WP functions is ok, touching the DB using PHP functions is not.

jrfnl:
Rebased so travis run will pass.

JDGrimes:
>> If you must touch the database, get in touch with some developers by posting a message to the wp-hackers mailing list. They may want to consider creating a function for the next WordPress version to cover the functionality you wanted.
> 
> To me, that implied - touching the DB using WP functions is ok, touching the DB using PHP functions is not.

I thought that paragraph ambiguous as well. It could just as easily mean, e.g., "if there is no function to get a post, ask the developers to introduce `get_post()` in the next release".

jrfnl:
> I thought that paragraph ambiguous as well.

It **_is_** ambiguous and I agree it should be clarified. The question is how far this rule goes.

However, I also believe that we can all agree that *if* a database call is made, it should *always* be done via WP functions/classes/methods and never via direct PHP DB calls. That's kind of the bare minimum which would be covered by the rule.
And that's all that's currently covered in this PR and PR #615.

If it would, at a later point in time, become clear that part of the WP db functions should also be discouraged, either an additional sniff could be added or these sniffs could be added to.

GaryJones:
I've got a client site, which touches non-WP tables inside the WP database, and other tables in a new DB. The former uses PDO, though could be converted to use `$wpdb`, but the latter couldn't.

jrfnl:
> I've got a client site, which touches non-WP tables inside the WP database, and other tables in a new DB. The former uses PDO, though could be converted to use $wpdb, but the latter couldn't. 

IMHO that would be a typical exception situation where you would either turn off this particular sniff for the files making those calls or lower the level from error to warning via the config.

PR #613, opened 2016-07-17
Update WPCS testing procedure

GaryJones:
I don't have `phpcs` installed globally, but PHPCS now has it as a requirement, so `vendor/bin/phpcs` can be used.


JDGrimes:
@GaryJones I think you meant:

> I don't have phpunit installed globally, but PHPCS now has it as a requirement, so vendor/bin/phpunit can be used.

GaryJones:
`phpunit`, `phpcs`, it's all the same :-D

PR #615, opened 2016-07-18
Add sniff for Restricted Database Classes.

jrfnl:
* Added `AbstractClassRestrictionsSniff` class which can sniff for the use of (namespaced) classes. The abstract class does currently not support `use` statement aliased namespaces.
* Refactored the `AbstractFunctionRestrictionsSniff` somewhat to make parts more reusable and extended the `AbstractClassRestrictionsSniff` from this class.
	* This refactoring includes a small efficiency improvement where the function regular expressions are no longer rebuild for every token, but only build once.
* Added `DB/RestrictedClassesSniff` to actually sniff for the Restricted DB Classes.
* Added extensive unit tests which test both the `DB/RestrictedClassesSniff` as well as the complete functionality of the Abstract class.

If at some point in the future the `AbstractFunctionRestrictionsSniff` would need to support namespaces too, this will not be too hard to do. Some functions would need to be moved from the `AbstractClassRestrictionsSniff`  to the `AbstractFunctionRestrictionsSniff` and the functionname determination would need to start using them, but the underlying logic has already been build.

This is the sister-PR to #612 which covers the function sniffing for direct DB calls.

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -114,7 +114,8 @@
 	<rule ref="WordPress.WP.I18n"/>
 	<rule ref="WordPress.Functions.DontExtract"/>
 	<rule ref="WordPress.NamingConventions.ValidHookName"/>
-
 	<rule ref="WordPress.DB.RestrictedFunctions"/>
+
+	<rule ref="WordPress.DB.RestrictedClasses"/>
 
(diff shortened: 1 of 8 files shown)


jrfnl:
Rebased for merge conflict (also done for the others which were conflicting)

PR #621, opened 2016-07-19
Add some more PHP magic constants which can be skipped …

jrfnl:
… to the EscapeOutputSniff.

### DIFF
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -75,6 +75,21 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 	public static $addedCustomFunctions = false;
 
 	/**
+	 * List of names of the tokens representing PHP magic constants.
+	 *
+	 * @var array
+	 */
+	private $magic_constant_tokens = array(
+		'T_CLASS_C'  => true, // __CLASS__
+		'T_FILE'     => true, // __FILE__
+		'T_FUNC_C'   => true, // __FUNCTION__
+		'T_LINE'     => true, // __LINE__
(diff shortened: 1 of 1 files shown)


WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -226,7 +226,7 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 			}
 
 			// Handle magic constants for debug functions.
-			if ( in_array( $tokens[ $i ]['code'], array( T_METHOD_C, T_FUNC_C, T_FILE, T_CLASS_C ), true ) ) {
+			if ( in_array( $tokens[ $i ]['code'], array( T_CLASS_C, T_FILE, T_FUNC_C, T_LINE, T_METHOD_C, T_NS_C, T_TRAIT_C ), true ) ) {
```
JDGrimes: Maybe we should also move these to a property. Make them the keys so that we can use `isset()` here instead of `in_array()`. Not sure if there would be any use for this list outside of this sniff or not; if you don't think so then just have it be a property of this sniff instead of the base sniff class.

jrfnl: When I made the change I checked that:
a) there wasn't already a (static) property in the PHPCS Tokens class we could use for this (there isn't)
b) whether any of these tokens are used anywhere else in WPCS (they are not).

I thought about making it a property, but am a bit wary what with the fact that the actual values these constants represent are integers which change between PHP versions. It shouldn't be an issue, but still.

Thinking now - I might as well make it a property with the string name as keys and check against the `$tokens[ $i ]['type']` value instead of `code`. That should work and be stable.

jrfnl: Done.

PR #622, opened 2016-07-19
Add two new function to the `$printingFunctions` array.

jrfnl:
WP 4.3 introduced `_deprecated_constructor()`.
WP 4.6 will introduce `_deprecated_hook()`.


JDGrimes:
It looks like you added tests for `_deprecate_hook()` but not `_deprecated_constructor()`. Was that intentional? I'm not sure that it is necessary for us to test every single function, but I thought that I would point it out anyway.

jrfnl:
> Was that intentional?

Someone is paying attention :sunglasses: 

Yes, that was intentional, I didn't really see the need as 99% of the time the function will be called like this:
```php
_deprecated_constructor( __CLASS__, '4.5.0' );
```
Ref: https://developer.wordpress.org/reference/functions/_deprecated_constructor/

Still easy enough to add one if you like.

JDGrimes:
> Still easy enough to add one if you like.

Nope, that's fine. :shipit:

PR #623, opened 2016-07-19
Update `ValidFunctionNames` Sniff to allow for non-valid names in extended classes.

### LINKED ISSUE #507: False positive: only PHP magic methods should be prefixed with a double underscore
The following code:

    class Child extends SoapClient {
    	public function __doRequest( $request, $location, $action, $version, $one_way = 0 ) {
    		// some custom stuff
    	}
    }

Produces the following ERROR:

    Method name "EventsSoapClient::__doRequest" is
    invalid; only PHP magic methods should be prefixed
    with a double underscore

```SoapClient``` is part of an extension provided as part of core PHP that makes heavy use of ```__``` named functions with valid use cases for overloading them.

jrfnl:
Fixes #507 

This allows for classes which extend a parent class or implement an interface to use method names which do not comply with the "don't start methods with a double underscore" and snake case rules so they can overload parent methods without an issue being raised.
This was previously already implemented for snake case, but not for double underscore.

I've also synced the class with the parent class it uses and applied any relevant changes to the overloaded methods and removed some code which can defer to the parent.

Also:
* removed some code which was no longer needed (had to do with the private _ method names)
* updated the documentation here and there
* added more extensive unit tests
* fixed a wrongly tagged message and improved the wording of it

Please see the individual commits for more details.

There was one change in the parent class - which I have implemented -, which I was not sure about as it changes the result of one unit test.

Basically, before, function names (outside class scope) were allowed to start with a double underscore.
The change in the parent *does* allow for methods where the name only consists of underscores, like __(), but does not allow function names which also contain other characters to start with a double underscore unless they are a PHP magic function name, i.e. `__autoload()`.

### DIFF
--- a/WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
@@ -19,6 +19,27 @@
  */
 class WordPress_Sniffs_NamingConventions_ValidFunctionNameSniff extends PEAR_Sniffs_NamingConventions_ValidFunctionNameSniff {
 
+	/**
+	 * Additional double underscore prefixed methods specific to certain PHP native extensions.
+	 *
+	 * Currently only handles the SoapClient Extension.
+	 *
+	 * @link http://php.net/manual/en/class.soapclient.php
+	 *
+	 * @var array <string method name> => <string class name>
+	 */
+	private $methodsDoubleUnderscore = array(
+		'doRequest'              => 'SoapClient',
+		'getFunctions'           => 'SoapClient',
+		'getLastRequest'         => 'SoapClient',
+		'getLastRequestHeaders'  => 'SoapClient',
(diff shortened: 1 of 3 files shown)


JDGrimes:
> There was one change in the parent class - which I have implemented -, which I was not sure about as it changes the result of one unit test.
> 
> Basically, before, function names (outside class scope) were allowed to start with a double underscore.
> The change in the parent does allow for methods where the name only consists of underscores, like __(), but does not allow function names which also contain other characters to start with a double underscore unless they are a PHP magic function name, i.e. __autoload().

I think this change is correct. [PHP's docs](http://www.php.net/manual/en/language.oop5.magic.php) don't specify that only class methods with double-underscores are reserved, but rather *all functions*:

> **Caution** PHP reserves all function names starting with `__` as magical. It is recommended that you do not use function names with `__` in PHP unless you want some documented magic functionality.

PR #624, opened 2016-07-19
Add strict type checking for array_search() and array_keys().

### LINKED ISSUE #503: Require strict=true for array_search()
Require that `array_search()`'s third parameter, strict, be true as is done for `in_array()`.

jrfnl:
Will only check for the `$strict` parameter for `array_keys()` if the second parameter has been set.

Fixes #503.

The file/sniff name is now a bit dated as the sniff checks for more than just `in_array()`, but changing the file/sniff/classname would break backward compatibility.

What do you think ?

### DIFF
--- a/WordPress/Sniffs/PHP/StrictInArraySniff.php
+++ b/WordPress/Sniffs/PHP/StrictInArraySniff.php
@@ -8,7 +8,7 @@
  */
 
 /**
- * Flag calling in_array() without true as the third parameter.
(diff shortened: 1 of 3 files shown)


JDGrimes:
> The file/sniff name is now a bit dated as the sniff checks for more than just in_array(), but changing the file/sniff/classname would break backward compatibility.
> 
> What do you think ?

Yeah, we'll just leave it like it is.

PR #633, opened 2016-07-25
Modularizing the discouraged functions

grappler:
This is an intial start of fixing https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/582 & https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/576

Do you think I am on the right path?

TODO
- [x] Complete the list of deprecated functions.
- [x] Finish modularising `WordPress.VIP.RestrictedFunctions`
- [x] Find a place for the forbidden Theme Review functions. https://github.com/WPTRT/WordPress-Coding-Standards/issues/9 https://github.com/WPTRT/WordPress-Coding-Standards/pull/10
- [x] Update the unit tests to include the new functions.
- [x] Update the rulesets with the changes


WordPress/Sniffs/PHP/DeprecatedFunctionsSniff.php
```diff
@@ -0,0 +1,81 @@
...
+ * WordPress_Sniffs_PHP_DeprecatedFunctionsSniff.
+ *
+ * Restricts usage of deprecated PHP functions.
+ *
+ * @category PHP
+ * @package  PHP_CodeSniffer
+ * @author   Shady Sharaf <shady@x-team.com>
+ * @author   Ulrich Pogson <ulrich@pogson.ch>
+ */
+class WordPress_Sniffs_PHP_DeprecatedFunctionsSniff extends WordPress_AbstractFunctionRestrictionsSniff {
```
jrfnl: FYI: This whole file can be removed as this is covered in #608.

grappler: Thanks, I will rebase once it is merged.

grappler:
As I am going through DiscouragedFunctionsSniff I see a check for `register_globals()` but I cannot find any reference to that function in WordPress other than `WP::register_globals`. Is that what the sniff is trying to catch?

grappler:
A quick update. I have separated the different functions. Currently Deprecated function are not done yet. Waiting for a response from https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/576#issuecomment-250063696

Travis is failing becuase of another issue. My unit tests are passing.

This list of functions is still in `WordPress_Sniffs_VIP_RestrictedFunctionsSniff`. I think it should be moved out so that others have access to it but I am not sure where.
```php
			'runtime_configuration' => array(
				'type'      => 'error',
				'message'   => '%s is prohibited, changing configuration at runtime is not allowed on VIP Production.',
				'functions' => array(
					'dl',
					'error_reporting',
					'ini_alter',
					'ini_restore',
					'ini_set',
					'magic_quotes_runtime',
					'set_magic_quotes_runtime',
					'apache_setenv',
					'putenv',
					'set_include_path',
					'restore_include_path',
				),
			),
```

We also have a list of functions that are blocked by the theme review team that we could maybe add to the same sniff. https://github.com/WPTRT/WordPress-Coding-Standards/pull/10/files#diff-75f603ed5bb071632d90df9854203926R29

Would love some feedback! 😄

JDGrimes:
RE `register_globals()`: maybe that was a reference to [this](https://secure.php.net/manual/en/security.globals.php)?

grappler:
So `register_globals` is a deprecated PHP ini config. I think it is best left of the check. Do we need to document this in the class changelog?

jrfnl:
> So register_globals is a deprecated PHP ini config. I think it is best left of the check. Do we need to document this in the class changelog?

As the `WordPress_AbstractFunctionRestrictionsSniff` class which is used as the base only checks for function calls, this should probably be removed - or better yet: be moved to a separate sniff which checks for ini directives.

grappler:
The check is already included in PHPCompatibility [DeprecatedIniDirectivesSniff.php](https://github.com/wimg/PHPCompatibility/blob/master/Sniffs/PHP/DeprecatedIniDirectivesSniff.php)

What do you think of including the TRT restricted PHP functions in the PHP folder?

jrfnl:
> The check is already included in PHPCompatibility DeprecatedIniDirectivesSniff.php

The PHPCompatibility sniffs are a separate ruleset which does not get shipped with WPCS, so unless people add that to their custom ruleset, they won't get the benefit of it.

> What do you think of including the TRT restricted PHP functions in the PHP folder?

You mean including the array from the TR sniff in the `WordPress_Sniffs_PHP_RestrictedFunctionsSniff` ? Possibly, but only if it doesn't impact any of the other WP rulesets.
I'd be more inclined to add the function groups each in a separate sniff file which could then go in the PHP folder (so it is easier to include them in the rulesets).

For that matter - the splitting out of the files as you're doing in this PR will also affect some rulesets, so you may want to add an action item to _review & adjust the existing rulesets_ for the changes you are making.

grappler:
> The PHPCompatibility sniffs are a separate ruleset which does not get shipped with WPCS, so unless people add that to their custom ruleset, they won't get the benefit of it.

I am not that keen on duplicating checks just so that we have backwards compatibility. I expect in most WP projects there is no php.ini file. This check was looking for a function called `register_globals()` and not an ini directive.

> I'd be more inclined to add the function groups each in a separate sniff file which could then go in the PHP folder (so it is easier to include them in the rulesets).

This is what I meant but I am not sure do we want to create a separate sniff for each item or if we can bundle a few together.

> so you may want to add an action item to review & adjust the existing rulesets for the changes you are making.

Thought of that, I added an action item. 😄

jrfnl:
> > The PHPCompatibility sniffs are a separate ruleset which does not get shipped with WPCS, so unless people add that to their custom ruleset, they won't get the benefit of it.
> I am not that keen on duplicating checks just so that we have backwards compatibility. I expect in most WP projects there is no php.ini file. This check was looking for a function called register_globals() and not an ini directive.

So the original sniff was wrong, but that doesn't mean it shouldn't be sniffed for. `register_globals` is evil and should not be touched.
Whether a project has a `php.ini` file is less relevant as a) those won't be parsed by PHPCS anyway and b) `register_globals` could - when it was still in PHP - be changed via `ini_set()` AFAICR which is what *should* be sniffed for.
Maybe leave it out of this PR, but open a new issue for this ?

grappler:
The rulesets should be should checking the same things are were doing before the functions were modularized.

I have place all of the restricted functions in `WordPress.PHP.RestrictedFunctions`. It is better to have all of the restricted functions together. If a ruleset does not need a certain check then it can still be excluded like so:

```XML
<rule ref="WordPress.WP.AlternativeFunctions">
	<properties>
		<property name="exclude" value="curl,file_get_contents"/>
	</properties>
</rule>
```

The only remaining thing are the deprecated functions which are being discussed in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/576#issuecomment-250063696

grappler:
I have added a start for a check for deprecated based on https://github.com/WPTRT/WordPress-Coding-Standards/pull/77

@JDGrimes Would be interested on your inital feedback on the deprecated functions check. It is just a start.

WordPress/Sniffs/PHP/DiscouragedFunctionsSniff.php
```diff
@@ -7,65 +7,69 @@
...
+				),
+			),
 
-		// WordPress discouraged.
-		'query_posts'              => 'WP_Query',
-		'wp_reset_query'           => 'wp_reset_postdata',
-	);
+			'urlencode' => array(
+				'type'      => 'warning',
+				'message'   => '%s() should only be used when dealing with legacy applications rawurlencode should now de used instead. See http://php.net/manual/en/function.rawurlencode.php and http://www.faqs.org/rfcs/rfc3986.html',
```
JDGrimes: Small typo here, "de" instead of "be".

JDGrimes:
The deprecated functions check looks like it is going in the right direction. The main thing that I think could be improved is the minimum WP version. Just so that we're all on the same page, the purpose of having the minimum WP version would seem to be this:

Say that you have a plugin that supports WordPress 4.2-4.6. But a function was deprecated in 4.3. In WordPress 4.3+, you want to avoid that function. But in 4.2, it isn't deprecated, and you may actually *have* to use it, because the alternative may not have been available yet.

The ability to define the lowest version that is supported lets us continue to use such a function in  such a project, without getting errors for it (currently the sniff would still give a warning though).

In general I think this is a good feature, but I do think that it could be even better. For example, sometimes when a function is deprecated, the alternative may have already existed. So using the above example, if the deprecated function had an alternative, and that alternative was available in 4.2, then there is no reason that we shouldn't use it. Admittedly though, I'm not sure how common that situation is. If we find that it is at all common, we could check the minimum version against the version that the alternative function was introduced, instead of against the version that the deprecated function was deprecated.

However, a more general thing that we could do is check that whenever a deprecated function does need to be used on older versions of WordPress, that it would be surrounded by a WP version check, and only used on those older versions where it isn't deprecated yet. Of course, that feature could be added to the sniff later, I guess, it doesn't have to be in the initial version, especially since a warning will still be given.

WordPress/Sniffs/WP/DeprecatedFunctionsSniff.php
```diff
@@ -0,0 +1,586 @@
...
+	 * 		'message'   => 'Use anonymous functions instead please!',
+	 * 		'functions' => array( 'eval', 'create_function' ),
+	 * 	)
+	 * )
+	 *
+	 * @return array
+	 */
+	public function getGroups() {
+		foreach ( $this->depreacted_functions as $depreacted_function => $data ) {
+			if ( intval( $data['version'] < $this->minimum_supported_version ) ) {
```
westonruter: Shouldn't `version_compare()` be used?

grappler:
> The ability to define the lowest version that is supported lets us continue to use such a function in such a project, without getting errors for it (currently the sniff would still give a warning though).

Yes that is right. We need the warning system for the theme review so that theme authors know they have to switch the deprected function out in the future.

> sometimes when a function is deprecated, the alternative may have already existed.

This may be the case but I have not come across this. 

> However, a more general thing that we could do is check that whenever a deprecated function does need to be used on older versions of WordPress, that it would be surrounded by a WP version check, and only used on those older versions where it isn't deprecated yet. 

There are few different ways to do the check depending on the function so this would be a bit more involved. It would make sense to add support for somthing like this till a maximum of version 3.0. Some check if the new function exists and otherwise use the depreacted functions. Here is an example which checks for a function introduced in the version but is not the alternative function that should be used. https://make.wordpress.org/core/2015/10/20/document-title-in-4-4/

Thank you for the vote of confidence. I will get this finished up soon.

grappler:
I have completed the WordPress Deprecated functions sniff. Somehow the CURL functions check is failing when using PHPCS 2.7. I don't think it is due to my PR as I just moved the code around.

grappler:
I have worked out why travis test is failing with PHPCS 2.7. I have excluded two function groupd in the VIP ruleset as they don't apply there. When running the Unit Test in PHPCS 2.7 it expects an notice but there is non becuase the function groups have been exluded.

I wonder if there is a way to check when ruleset is being used and then change the expected value for the Unit Test.

In `WordPress_Sniffs_WP_AlternativeFunctionsSniff` the notice is set to Warning but for the Theme Review sniff we will need them to be Errors. Is there any reason not to set them to errors?  I kept them as Warnings as that is what they were in the VIP sniff.

jrfnl:
The unit tests test the sniff, not the ruleset, so in that case, the unit tests need to be corrected. They should test the complete sniff. See also #696.

jrfnl:
> In WordPress_Sniffs_WP_AlternativeFunctionsSniff the notice is set to Warning but for the Theme Review sniff we will need them to be Errors. 

That can be changed from the (Theme) ruleset anyway, so the sniff itself does not need to be changed for that.

grappler:
> The unit tests test the sniff, not the ruleset, so in that case, the unit tests need to be corrected. They should test the complete sniff.

The unit test is testing the complete sniff as that is why travis is passing when using PHPCS master but it is failing when using PHPCS 2.7. This is the reverse ofhttps://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/696

jrfnl:
@grappler Hmm.. I'd need to look in more detail. Haven't got time at the moment, will try next week. if it's still needed by then.

jrfnl:
@grappler I've had a look at this and found a way to solve it. I've send in a PR to this branch in your fork to sort it out. https://github.com/grappler/WordPress-Coding-Standards/pull/1

grappler:
Nearly there, now the travis test is failing because `eval` is being used in `VIP/CronIntervalSniff.php`. I don't think `WordPress_AbstractFunctionRestrictionsSniff` supports whitelisting comments.

jrfnl:
> Nearly there, now the travis test is failing because eval is being used in VIP/CronIntervalSniff.php

Not sure how the others feel about using this, but you could get the build to pass by adding `// @codingStandardsIgnoreLine` to the line.

So the line would becomes something like this (not sure how well the comment works in combination with the original comment, but a quick test should tell you quickly enough):
```php
	$interval = eval( "return ( $value );" ); // @codingStandardsIgnoreLine - No harm here.
```

grappler:
The build is now passing. Can the label `Status: Review Ready` be added?

grappler:
Any chance of a review? 😄

westonruter:
@jrfnl if you're good with this then I am good with it.

jrfnl:
@westonruter I'll try and do a thorough review later this week

WordPress/Sniffs/WP/DeprecatedFunctionsSniff.php
```diff
@@ -0,0 +1,1097 @@
...
+	 * Example: '4.3'
+	 *
+	 * @var string WordPress versions.
+	 */
+	public $minimum_supported_version = 4.3;
+
+	/**
+	 * List of deprecated functions with alternative when available.
+	 *
+	 * To be updated after every major release.
```
jrfnl: I would suggest to add a note which documents when the list was last updated and on which version / commit the last update was based, so the next person updating it knows where to start from.

WordPress/Sniffs/PHP/DevelopmentFunctionsSniff.php
```diff
@@ -0,0 +1,62 @@
...
+					'print_r',
+					'trigger_error',
+					'set_error_handler',
+					'debug_backtrace',
+					'debug_print_backtrace',
+					'wp_debug_backtrace_summary',
+				),
+			),
+
+			'prevent_path_disclosure' => array(
```
grappler: I am fine with changing it to `warning` but I believe they should belong in the category of development functions. We could always remove it from the `extra` ruelset. I don't think every plugin needs to use these functions and they definitely should not be used in themes.

By having the development functions together it can be easily enabled or disabled.

WordPress-VIP/ruleset.xml
```diff
@@ -14,6 +14,27 @@
 	<!-- https://vip.wordpress.com/documentation/code-review-what-we-look-for/#commented-out-code-debug-code-or-output -->
 	<rule ref="Squiz.PHP.CommentedOutCode" />
 
+	<!-- https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#eval-and-create_function -->
+	<!--https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#settings-alteration -->
+	<rule ref="WordPress.PHP.RestrictedFunctions" />
```
grappler: None of us can update the VIP documentation. There is a check in the VIP scanner that does the same thing that could backup including the two new sections. https://github.com/Automattic/vip-scanner/blob/master/vip-scanner/checks/ForbiddenPHPFunctionsCheck.php

grappler:
Thank you @jrfnl for the impressive review. The initial struggle was to work out how to separate the different checks. The reason I ended up having this huge PR is that if I was moving checks around I was missing functionality. I suppose I should have done the PR is stages. It is easier to see the structure in hindsight. 

I have started to make changes according to the inline comments.

I am fine changing the notice from`error` to `warning` where ever as PHPCS is configurable that it can be changed in the ruleset. If the notice is changed then we will need to move the functions from `RestrictedFunctions` to `DiscouragedFunctions`.

@sboisvert @david-binda could you check that the changes being made to the VIP ruleset is OK? They should be OK but they are not fully documented. It would be good if they are OK that they are documented too. Thanks! :) See the section "VIP ruleset" in the comment by @jrfnl

JDGrimes:
I disagree with @jrfnl regarding most of the functions that he's suggested should be removed from `extra`/downgraded to a warning (except for `create_function`, of course). While there may be perfectly valid use-cases for these functions, I don't think that is so much the point. I think that those are really special cases, that can be handled via a custom config for a project. Generally speaking, plugins that are using them are very probably `_doing_it_wrong()`. And in some cases hosts disable some of these functions anyway, and so as along as that is a somewhat common practice, those functions shouldn't be used without first checking if they are available.

Upon reflection, I guess maybe a warning makes some sense as opposed to an error for some of the cases, but I've always just turned all off the warnings off when I run PHPCS (I don't remember exactly why, maybe I should just stop doing this), so when I think of a warning, I think "most people will never see this or won't pay any attention to it." Which maybe is a completely wrong assumption. In which case, yeah, a warning is fine. But I still wouldn't support removing them from extra entirely.

sboisvert:
Thanks for the amazing PR and all the work that went into it!

The one thing VIP wouldn't necessarily want to sniff for is under section obfuscation: base64_decode, base64_encode. There are enough valid use cases for base64 encoding that I don't think this should be included. As for the others under obfuscation I don't have a strong opinion but if we do leave the sniffs in I think they should be warnings. @david-binda might have other suggestions.

As a sidenote I support the suggested modifications from @jrfnl.

grappler:
@jrfnl I have made all of the changes as requested. I removed the duplicate `curl` and `file_get_contents` checks and just changed the message instead.

The Unit Test is failing in PHPCS version 2.7 due to the notice being changed from warning to error. To prove this I have **temporarily** increased the version in the `.travis.yml` file. To fix this I have created a separate PR https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/733 which should be merged before this PR. Once PR #773 is merged I can rebase this PR and the tests should pass.

WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php
```diff
@@ -122,34 +113,16 @@ public function getGroups() {
...
 				'type'      => 'error',
-				'message'   => '%s is prohibited, please use wpcom_vip_attachment_url_to_postid() instead.',
+				'message'   => '%s() is prohibited, please use wpcom_vip_attachment_url_to_postid() instead.',
 				'functions' => array(
 					'attachment_url_to_postid',
 				),
 			),
 
-			// @link https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#remote-calls
-			'wp_remote_get' => array(
```
jrfnl: This section looks to be removed without it being brought back anywhere... Looks like a mistake ?

WordPress-Extra/ruleset.xml
```diff
@@ -29,7 +30,14 @@
 
 	<rule ref="WordPress.XSS.EscapeOutput"/>
 	<rule ref="WordPress.CSRF.NonceVerification" />
-	<rule ref="WordPress.PHP.DiscouragedFunctions"/>
+	<rule ref="WordPress.PHP.DevelopmentFunctions"/>
+	<rule ref="WordPress.PHP.DiscouragedPHPFunctions"/>
+	<!-- WP core still supports PHP 5.2+  -->
+	<exclude name="WordPress.WP.DiscouragedFunctions.create_function" />
```
jrfnl: XML syntax - I believe this should be like:
```xml
<rule ref="WordPress.PHP.DiscouragedPHPFunctions">
	<!-- WP core still supports PHP 5.2+-->
	<exclude name="WordPress.WP.DiscouragedFunctions.create_function" />
</rule>
```

WordPress-VIP/ruleset.xml
```diff
@@ -14,6 +14,41 @@
 	<!-- https://vip.wordpress.com/documentation/code-review-what-we-look-for/#commented-out-code-debug-code-or-output -->
 	<rule ref="Squiz.PHP.CommentedOutCode" />
 
+	<!-- https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#eval-and-create_function -->
+	<!-- https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#settings-alteration -->
```
jrfnl: This link no longer applies to this sniff. It should be moved down to the rule inclusion of `WordPress.PHP.DiscouragedPHPFunctions` / `WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration`

WordPress-VIP/ruleset.xml
```diff
@@ -14,6 +14,41 @@
...
+	<rule ref="WordPress.PHP.RestrictedFunctions" />
+
+	<!-- https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#eval-and-create_function -->
+	<!-- https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#serializing-data -->
+	<!-- https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#encoding-values-used-when-creating-a-url-or-passed-to-add_query_arg -->
+	<rule ref="WordPress.PHP.DiscouragedPHPFunctions" />
```
jrfnl: A link to documentation is still needed to cover the `system_calls` section. I'm happy for it to be the link to the VIP scanner which you previously posted - https://github.com/Automattic/vip-scanner/blob/master/vip-scanner/checks/ForbiddenPHPFunctionsCheck.php

WordPress-VIP/ruleset.xml
```diff
@@ -14,6 +14,41 @@
...
+	<exclude name="WordPress.PHP.DiscouragedPHPFunctions.obfuscation" />
+	<rule ref="WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration">
+		<type>error</type>
+	</rule>
+
+	<!-- https://vip.wordpress.com/documentation/code-review-what-we-look-for/#commented-out-code-debug-code-or-output -->
```
jrfnl: This link should probably also be added to cover the `prevent_path_disclosure` section: https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#settings-alteration

WordPress/Sniffs/PHP/DiscouragedPHPFunctionsSniff.php
```diff
@@ -0,0 +1,108 @@
...
+				'functions' => array(
+					'error_reporting',
+					'ini_alter',
+					'ini_restore',
+					'ini_set',
+					'apache_setenv',
+					'putenv',
+					'set_include_path',
+					'restore_include_path',
+					// This alias was DEPRECATED in PHP 5.3.0, and REMOVED as of PHP 7.0.0.
```
grappler: They are included because they touch the runtime. For PHP compatibility we can use the PHPcompatibility sniffs.

grappler:
I have fixed the last few issues. The travis test is now failing as I removed the commit to update PHPCS to `2.7.1`. Once #733 is merged I will rebase and then this will be ready to be merged.

jrfnl:
@grappler Thanks! Ping me once the blank file has been added (and maybe the last open comment has been addressed too https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/633#discussion_r93869725 ) and I'll approve the changes.

WordPress/Sniffs/PHP/DiscouragedFunctionsSniff.php
```diff
@@ -0,0 +1,52 @@
...
+ *                    The checks for the WP discouraged functions have been replaced by the
+ *                    stand-alone sniff WordPress_Sniffs_WP_DiscouragedFunctionsSniff.
+ *                    by the stand-alone sniff WordPress_Sniffs_WP_AlternativeFunctionsSniff.
+ *                    The checks for the PHP discouraged functions have been replaced by the
+ *                    stand-alone sniff WordPress_Sniffs_WP_DiscouragedPHPFunctionsSniff.
+ *                    The check for the `register_globals` has been removed as there is no such
+ *                    function. To check for `register_globals` ini directive use
+ *                    PHPCompatibility_Sniffs_PHP_DeprecatedIniDirectivesSniff from wimg/PHPCompatibility.
+ */
+class WordPress_Sniffs_PHP_DiscouragedFunctionsSniff extends WordPress_AbstractFunctionRestrictionsSniff {
```
jrfnl: If you remove the `extends WordPress_AbstractFunctionRestrictionsSniff`, you can remove the everything else in the class too, making it even clearer that it's deprecated.

grappler:
Thank you @jrfnl !

jrfnl:
Anyone else still wants to have a look at this before merging ? Please let me know, if not, I'll merge this tomorrow.

WordPress/Sniffs/PHP/DevelopmentFunctionsSniff.php
mundschenk-at: Both `trigger_error` and `set_error_handler` would appear to me to be legitimate functions for production code (especially when integrating third-party PHP libraries).

PR #637, opened 2016-07-26
Implement the extra rules as suggested in #607.

### LINKED ISSUE #607: Some extra rule inclusion suggestions
##### For `core` _(or `extra`)_:

To cover: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#clever-code
- `Squiz.Formatting.DisallowInlineIf`: Stops inline IF statements from being used.
- `Squiz.PHP.DisallowMultipleAssignments`: Ensures that there is only one value assignment on a line, and that it is the first thing on the line.
- [Move from `extra` to `core`] `Generic.Formatting.DisallowMultipleStatements`: Ensures each statement is on a line by itself.
##### For `extra` to further enhance the code analysis related sniffs:
- `Squiz.PHP.NonExecutableCode`: Warns about code that can never been executed. This happens when a function returns before the code, or a break ends execution of a statement etc.
- `Squiz.Operators.IncrementDecrementUsage`: Tests that the ++ operators are used when possible and not used when it makes the code confusing.
- `Squiz.Operators.ValidLogicalOperators`: Checks to ensure that the logical operators 'and' and 'or' are not used. Use the && and || operators instead.
- `Squiz.PHP.DisallowComparisonAssignment`: Ensures that the value of a comparison is not assigned to a variable. **_(probably not refined enough)_**
- One more: `Squiz.Functions.FunctionDuplicateArgument`: Checks that duplicate arguments are not used in function declarations.

Opinions ?

jrfnl:
To be merged **_after_** 0.10.0 has been released so we can test & fine-tune before it's released.

Fixes #607.

### DIFF
--- a/WordPress-Extra/ruleset.xml
+++ b/WordPress-Extra/ruleset.xml
@@ -8,7 +8,6 @@
 	<rule ref="WordPress.WP.DeprecatedFunctions"/>
 	<rule ref="Generic.PHP.ForbiddenFunctions"/>
 	<rule ref="Generic.Functions.CallTimePassByReference"/>
-	<rule ref="Generic.Formatting.DisallowMultipleStatements"/>
 	<rule ref="Generic.CodeAnalysis.EmptyStatement" />
 	<rule ref="Generic.CodeAnalysis.ForLoopShouldBeWhileLoop"/>
 	<rule ref="Generic.CodeAnalysis.ForLoopWithTestFunctionCall"/>
@@ -18,6 +17,13 @@
 	<rule ref="Generic.CodeAnalysis.UselessOverridingMethod"/>
 	<rule ref="Generic.Classes.DuplicateClassName"/>
 	<rule ref="Generic.Strings.UnnecessaryStringConcat"/>
+
+	<!-- More generic PHP best practices.
+		 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/607 -->
+	<rule ref="Squiz.PHP.NonExecutableCode"/>
+	<rule ref="Squiz.Operators.IncrementDecrementUsage"/>
+	<rule ref="Squiz.Operators.ValidLogicalOperators"/>
+	<rule ref="Squiz.Functions.FunctionDuplicateArgument"/>
 
 	<!-- This sniff is not refined enough for general use -->
 	<!-- https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/382#discussion_r29970107 -->
--- a/WordPress/AbstractFunctionRestrictionsSniff.php
+++ b/WordPress/AbstractFunctionRestrictionsSniff.php
@@ -140,30 +140,27 @@ abstract class WordPress_AbstractFunctionRestrictionsSniff implements PHP_CodeSn
 	 * @return void
 	 */
 	public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
-		$tokens = $phpcsFile->getTokens();
-		$token  = $tokens[ $stackPtr ];
+		$tokens        = $phpcsFile->getTokens();
+		$token         = $tokens[ $stackPtr ];
+		$token_content = strtolower( $token['content'] );
 
 		// Exclude function definitions, class methods, and namespaced calls.
-		if (
-			T_STRING === $token['code']
-			&&
-			( $prev = $phpcsFile->findPrevious( T_WHITESPACE, ( $stackPtr - 1 ), null, true ) )
-			&&
-			(
-				// Skip sniffing if calling a method, or on function definitions.
-				in_array( $tokens[ $prev ]['code'], array( T_FUNCTION, T_DOUBLE_COLON, T_OBJECT_OPERATOR ), true )
-				||
-				(
-					// Skip namespaced functions, ie: \foo\bar() not \bar().
-					T_NS_SEPARATOR === $tokens[ $prev ]['code']
-					&&
-					( $pprev = $phpcsFile->findPrevious( T_WHITESPACE, ( $prev - 1 ), null, true ) )
-					&&
-					T_STRING === $tokens[ $pprev ]['code']
-				)
-			)
-			) {
-			return;
+		if ( T_STRING === $token['code'] && isset( $tokens[ ( $stackPtr - 1 ) ] ) ) {
+			$prev = $phpcsFile->findPrevious( PHP_CodeSniffer_Tokens::$emptyTokens, ( $stackPtr - 1 ), null, true );
+
+			if ( false !== $prev ) {
+				// Skip sniffing if calling a same-named method, or on function definitions.
+				if ( in_array( $tokens[ $prev ]['code'], array( T_FUNCTION, T_DOUBLE_COLON, T_OBJECT_OPERATOR ), true ) ) {
+					return;
+				}
+
+				// Skip namespaced functions, ie: \foo\bar() not \bar().
+				$pprev = $phpcsFile->findPrevious( PHP_CodeSniffer_Tokens::$emptyTokens, ( $prev - 1 ), null, true );
+				if ( false !== $pprev && T_NS_SEPARATOR === $tokens[ $prev ]['code'] && T_STRING === $tokens[ $pprev ]['code'] ) {
+					return;
+				}
+			}
+			unset( $prev, $pprev );
 		}
 
 		$exclude = explode( ',', $this->exclude );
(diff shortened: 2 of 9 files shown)


jrfnl:
Due to a change in the GH interface, a rebase (or update _= back-merge resulting in the history being littered with unnecessary extra merges_) is now needed for every PR after every merge - **_even if the PR would merge cleanly_** -.
So ... please ping me when someone is ready to merge this and I'll rebase asap (instead of continuously doing so after every merge which gets old very fast).

Hopefully GH will reverse/improve this change soon as this is not very workable. (I've already contacted them about it, please do so too if you think this should change)

WordPress/AbstractArrayAssignmentRestrictionsSniff.php
```diff
@@ -144,7 +144,11 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 				continue;
 			}
 
-			$callback = ( isset( $group['callback'] ) && is_callable( $group['callback'] ) ) ? $group['callback'] : array( $this, 'callback' );
+			if ( isset( $group['callback'] ) && is_callable( $group['callback'] ) ) {
+				$callback = $group['callback'];
+			} else {
```
GaryJones: This `else` could be set as the default before the `if`.

```
callback = ...
if ( ... ) {
    callback = ...
}
```

WordPress/Sniff.php
```diff
@@ -663,7 +664,11 @@ protected function has_nonce_check( $stackPtr ) {
 		// If this is inside an isset(), check after it as well, all the way to the
 		// end of the scope.
 		if ( $in_isset ) {
-			$end = ( 0 === $start ) ? count( $tokens ) : $tokens[ $start ]['scope_closer'];
+			if ( 0 === $start ) {
+				$end = count( $tokens );
+			} else {
```
GaryJones: This `else` could be set as default before the `if`.

WordPress-Extra/ruleset.xml
```diff
@@ -16,6 +15,11 @@
 	<rule ref="Generic.CodeAnalysis.UselessOverridingMethod"/>
 	<rule ref="Generic.Classes.DuplicateClassName"/>
 	<rule ref="Generic.Strings.UnnecessaryStringConcat"/>
+	<rule ref="Squiz.PHP.NonExecutableCode"/>
+	<rule ref="Squiz.Operators.IncrementDecrementUsage"/>
+	<rule ref="Squiz.Operators.ValidLogicalOperators"/>
+	<rule ref="Squiz.PHP.DisallowComparisonAssignment"/>
```
GaryJones: This basically seems to forbid the use of ternary conditions?

jrfnl: It's been a while since I looked at this one. Will check & get back to you after the weekend (conference).

jrfnl: Sorry for the silence. Finally had a chance to look at this in properly.

By the looks of it, it does (forbid assignments using ternary conditions), though it is more specifically intended to catch things like `$var = ($foo === $bar)`.
See the unit tests.
https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Tests/PHP/DisallowComparisonAssignmentUnitTest.inc
https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Tests/PHP/DisallowComparisonAssignmentUnitTest.php

Would it be an idea to lower the error level for this sniff from `error` to `warning` ?

WordPress/AbstractArrayAssignmentRestrictionsSniff.php
jrfnl: Rebased & fixed.

GaryJones:
I think the ternary issue is the most contentious issue here - even the basic `$a = $b ? $c : $d;` throws an error. I'm generally fine with the rest.

jrfnl:
@GaryJones So what would you suggest - remove that sniff altogether (for now, possibly to be revisited later), or lower the errors from that sniff to warnings ?

GaryJones:
I'd say remove for now, but definitely have discussion for future review.

jrfnl:
@GaryJones Ok, I've had another look and I think it's not just `Squiz.PHP.DisallowComparisonAssignment` we'd need to exclude in that case, but also `Squiz.PHP.DisallowInlineIf` which also looks at those kind of statements.
See: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Tests/PHP/DisallowInlineIfUnitTest.inc

Either way, we're still not covering the example as used in the handbook which manages to bypass *both* these sniffs: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#clever-code

Shall I just remove both for now and open a new PR for these two sniffs ?
In that case, I will re-do this PR so as to remove the code adjustments which were made based on these two rules (and move those to the other PR).

GaryJones:
> Shall I just remove both for now and open a new PR for these two sniffs ?

I think that's the way to go for now.

jrfnl:
@GaryJones Ok, thanks for working with me on this. I've rebased & re-done the commits. This PR now only contains the rules which are without opinion on ternaries.
The more controversial part of the PR has been moved to PR #768

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -140,30 +140,27 @@ protected function setup_groups( $key ) {
...
+					return;
+				}
+
+				// Skip namespaced functions, ie: \foo\bar() not \bar().
+				$pprev = $phpcsFile->findPrevious( PHP_CodeSniffer_Tokens::$emptyTokens, ( $prev - 1 ), null, true );
+				if ( false !== $pprev && T_NS_SEPARATOR === $tokens[ $prev ]['code'] && T_STRING === $tokens[ $pprev ]['code'] ) {
+					return;
+				}
+			}
+			unset( $prev, $pprev );
```
grappler: If `$prev` is false then `$pprev` is never defined.

jrfnl: Correct, but that doesn't matter for an `unset()`. No notices will be thrown and if the variable *is* set, it will be correctly unset.

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -140,30 +140,27 @@ protected function setup_groups( $key ) {
 	 * @return void
 	 */
 	public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
-		$tokens = $phpcsFile->getTokens();
-		$token  = $tokens[ $stackPtr ];
+		$tokens        = $phpcsFile->getTokens();
+		$token         = $tokens[ $stackPtr ];
+		$token_content = strtolower( $token['content'] );
```
GaryJones: Is this `$token_content` not unrelated to the rest of the changes going on here? i.e. does it deserve it's own issue (to describe whatever bug this `strtolower()` is fixing) and PR?

WordPress/Sniffs/VIP/DirectDatabaseQuerySniff.php
```diff
@@ -167,7 +167,8 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 		}
 
 		$whitelisted_cache = false;
-		$cached            = $wp_cache_get = false;
+		$cached            = false;
+		$wp_cache_get      = false;
```
GaryJones: WordPress/Sniff.php, you have split out the multiple assignments as A = X, and then B = A.

Here, you have a pattern of A = X and B = X.

Is there any reason for this inconsistent approach? (There may well be, and that's fine if so)

WordPress/AbstractFunctionRestrictionsSniff.php
jrfnl: The quite convoluted `if()` statement below it was triggering the `Squiz.PHP.DisallowMultipleAssignments` sniff (twice), this probably slipped in as I was looking to simplify the code.

Function names are case-insensitive in PHP, so should always be lowercased before comparisons are done.
I'm happy to move that bit to another PR (I will probably include it in one of the review PRs which I've mentioned in some recent issues in that case).

WordPress/Sniffs/VIP/DirectDatabaseQuerySniff.php
jrfnl: Oh, there's a quite simple explanation for that:
In `WordPress/Sniff.php` we're assigning the output of a _function call_ to a variable and if a second variable needs the same value, there is no need to repeat the function call (efficiency).
In the above code, we're assigning the PHP native constant `false`, which carries no overhead in contrast to the function call.

GaryJones: I think there could also be an aspect of semantics here too - the last pointer IS the end of the line, where as the cached _thing_ is not the same as the WP cache get _thing_. If so, then the current code satisfies that, so all good.

jrfnl:
As all the more contentious extra rules have now been (re)moved from this PR, can this be merged ?

PR #646, opened 2016-07-28
Add rule to prohibit the use of the backtick shell execution operator…

jrfnl:
… to `extra`.

Execution of shell commands should be discouraged. A number of the other commands along the same lines are discouraged via function sniffs, however, the backtick operator was so far ignored.

For this sniff to be added, the minimum PHPCS version needs to be upped as it has only just been added to master and is expected to be in the 2.7 release.

See: https://github.com/squizlabs/PHP_CodeSniffer/pull/1073

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ php:
 
 env:
   - PHPCS_BRANCH=master
-  - PHPCS_BRANCH=2.6.0
+  - PHPCS_BRANCH=2.7.0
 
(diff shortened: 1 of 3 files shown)


jrfnl:
FYI: PHPCS 2.7 has been released.

GaryJones:
All good. I see no reason why we can't move forward with PHPCS 2.7.0 as a requirement.

PR #648, opened 2016-08-01
Sync the WhiteSpace_CastStructureSpacing sniff with its base class.

jrfnl:
The `WordPress_Sniffs_WhiteSpace_CastStructureSpacingSniff` is based upon the `Squiz_Sniffs_WhiteSpace_CastSpacingSniff`.

This PR contains the following changes:
* The class now extends the class it is based upon.
   This also means that the class now only contains the WP specific checks and defers to the parent for anything else.
* The check which is covered by the parent was already fixable there. This benefit is now carried over to this sniff.
* Made the two WP specific checks fixable as well and added a `.fixed` file to unit test this.
* Removed the class property which was at the bottom of the class which looks like it was left over from an earlier attempt to use the parent class in combination with making the check covered in the parent a `warning` instead of an `error`. The property did not have the desired effect.
  This is now solved by setting this in the ruleset.

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -150,6 +150,9 @@
 		<!-- Covers rule: When type casting, do it like so: $foo = (boolean) $bar; -->
 		<rule ref="Generic.Formatting.SpaceAfterCast"/>
 		<rule ref="WordPress.WhiteSpace.CastStructureSpacing"/>
+		<rule ref="WordPress.WhiteSpace.CastStructureSpacing.ContainsWhiteSpace">
+			<type>warning</type>
+		</rule>
 
 		<!-- Covers rule: ... array items, only include a space around the index if it is a variable. -->
 		<rule ref="WordPress.Arrays.ArrayKeySpacingRestrictions"/>
--- a/WordPress/Sniffs/WhiteSpace/CastStructureSpacingSniff.php
+++ b/WordPress/Sniffs/WhiteSpace/CastStructureSpacingSniff.php
@@ -7,6 +7,10 @@
  * @license https://opensource.org/licenses/MIT MIT
  */
 
+if ( ! class_exists( 'Squiz_Sniffs_WhiteSpace_CastSpacingSniff', true ) ) {
+	throw new PHP_CodeSniffer_Exception( 'Class Squiz_Sniffs_WhiteSpace_CastSpacingSniff not found' );
+}
+
 /**
  * Ensure cast statements don't contain whitespace, but *are* surrounded by whitespace, based upon Squiz code.
  *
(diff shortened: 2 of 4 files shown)


jrfnl:
Rebased for merge conflicts.

jrfnl:
Please ping me when someone is ready to merge this and I'll rebase asap - see https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/637#issuecomment-242906170

WordPress-Core/ruleset.xml
```diff
@@ -150,6 +150,9 @@
 		<!-- Covers rule: When type casting, do it like so: $foo = (boolean) $bar; -->
 		<rule ref="Generic.Formatting.SpaceAfterCast"/>
 		<rule ref="WordPress.WhiteSpace.CastStructureSpacing"/>
+		<rule ref="WordPress.WhiteSpace.CastStructureSpacing.ContainsWhiteSpace">
+			<type>warning</type>
```
JDGrimes: Why does this need to be a warning instead of an error? Is it not called for by WordPress's standards? I suggest that we let it be an error, since as I understand it it already is right now, no?

WordPress/Sniffs/WhiteSpace/CastStructureSpacingSniff.php
```diff
@@ -15,21 +19,13 @@
  * @package WPCS\WordPressCodingStandards
  *
  * @since   0.3.0
+ * @since   0.xx.0 This sniff now extends the Squiz_Sniffs_WhiteSpace_CastSpacingSniff class it was based upon.
+ * @since   0.xx.0 This sniff now has the ability to fix the issues it flags.
```
JDGrimes: Are you planning to fill out the version numbers just before merge, or at the time of release? (Just so that we're all on the same page.)

jrfnl: I left it like that when I submitted the PR as 0.10.0 was nearing release but not released yet, so it was unclear whether this would go in then or later. As 0.10.0 has been released now, I'll update these to read 0.11.0.

WordPress-Core/ruleset.xml
jrfnl: > Why does this need to be a warning instead of an error?

This PR was about syncing with the upstream sniff and was not intended to *change* existing functionality.
The WPCS version of the sniff had this as a `warning` previously, so I figured there must have been a reason at the time the sniff was copied into WPCS for this check to have been changed from `error` (upstream) to `warning` (WPCS), which is why I left it as such.

> Is it not called for by WordPress's standards?

It is: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage

> I suggest that we let it be an error, since as I understand it it already is right now, no?

As mentioned above, currently it's a `warning`, but I'd be happy to change it to an `error`.

WordPress/Sniffs/WhiteSpace/CastStructureSpacingSniff.php
```diff
@@ -41,37 +37,25 @@ public function register() {
...
-		$expected = str_replace( ' ', '', $content );
-		$expected = str_replace( "\t", '', $expected );
-
-		if ( $content !== $expected ) {
-			$error = 'Cast statements must not contain whitespace; expected "%s" but found "%s"';
-			$data  = array(
-				$expected,
-				$content,
-			);
-			$phpcsFile->addWarning( $error, $stackPtr, 'ContainsWhiteSpace', $data );
```
jrfnl: This is the old definition of the `ContainsWhiteSpace` check which differs from the upstream one by using `addWarning()` instead of `addError()`

WordPress-Core/ruleset.xml
jrfnl: Oh and another thing: If it would be decided that this should be an `error`, I would suggest to change the other two checks in this sniff to `error`s as well.

PR #649, opened 2016-08-01
Sync the Classes_ValidClassName sniff with its base class.

jrfnl:
The `WordPress_Sniffs_Classes_ValidClassNameSniff` is (currently) based upon the `Squiz_Sniffs_Classes_ValidClassNameSniff`.

This PR contains the following changes:
* Carried over the changes/fixes made in the Squiz sniff to the WP sniff.
* Documented when the class was last synced and why it doesn't extend the parent.

----

We may actually want to consider just switching over to the `PEAR.NamingConventions.ValidClassName` sniff as it looks like we can just include that rule in the ruleset without the need for an own sniff.

If run against the Pear sniff, the WP class name unit tests still pass (including some extra ones I added to test this).
The only thing which is "missing" from the Pear sniff compared to the Squiz sniff is that the Squiz sniff contains a check for missing class braces, but
a) that's covered in the syntax check and
b) that is not something which ought to be covered by a naming convention sniff anyway.

Opinions ?

### DIFF
--- a/WordPress/Sniffs/Classes/ValidClassNameSniff.php
+++ b/WordPress/Sniffs/Classes/ValidClassNameSniff.php
@@ -16,8 +16,11 @@
  *
  * @since   0.1.0
  *
- * Last synced with base class ?[unknown date]? at commit ?[unknown commit]?.
- * @link    https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Sniffs/Classes/ValidClassNameSniff.php
+ * Last synced with base class August 2016 at commit 0a474f9ad5756b00e50b9d5b0a8bdb0c8d6fcd99.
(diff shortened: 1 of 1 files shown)


jrfnl:
Please ping me when someone is ready to merge this and I'll rebase asap - see https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/637#issuecomment-242906170

JDGrimes:
>If run against the Pear sniff, the WP class name unit tests still pass (including some extra ones I added to test this).
> The only thing which is "missing" from the Pear sniff compared to the Squiz sniff is that the Squiz sniff contains a check for missing class braces, but
> a) that's covered in the syntax check and
> b) that is not something which ought to be covered by a naming convention sniff anyway.
> 
> Opinions ?

Yeah, let's just switch to the Pear sniff.

PR #658, opened 2016-08-03
Make the `StrictInArray` sniff fixable.

jrfnl:


westonruter:
Can this be auto-fixable though? What if there is code that relies on a form submission that sends `?foo=1` and the PHP you have does `in_array( $_GET['foo'], array( 1, 2, 3 ) )`. If this gets fixed, then this code will then break. Same goes for fixing `==` to `===`. Auto-fixing these may break the logic.

jrfnl:
> If this gets fixed, then this code will then break, Same goes for fixing == to ===. Auto-fixing these may break the logic.

True, but that indicates an underlying issue with the comparison logic which should be addressed anyway. Loose type checks are the cause of *lots* of bugs.

If you don't believe me, come join me for the big "Why Equal doesn't Equal" Quiz some time if you get the chance ;-)
http://phpcheatsheets.com/ FTW :sunglasses:

westonruter:
@jrfnl I absolutely agree. _But_ I think the underlying premise of `phpcbf` is that it can fix problems _automatically_ without any impact on functionality. If we try to automatically fix loose type checks, there is a good chance the automatic fix would break the logic. So I think it cannot be a fixable error. I would not feel comfortable using `phpcbf` if I knew it could break my code.

jrfnl:
Makes sense to me. Let's close this for now.

PR #668, opened 2016-08-22
flush_rewrite_rules should be flagged

sboisvert:
When we split these sniffs out this will only apply to wpcom VIP and not Go. We will want to sniff for it on Go but just make it a warning and make sure people aren't flushing on every page load.

### DIFF
--- a/WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php
+++ b/WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php
@@ -283,6 +283,15 @@ class WordPress_Sniffs_VIP_RestrictedFunctionsSniff extends WordPress_AbstractFu
 				),
 			),
 
+			// @link https://vip.wordpress.com/documentation/change-your-pretty-permalinks-or-add-custom-rewrite-rules/
+			'flush_rewrite_rules' => array(
+				'type'     => 'warning',
+				'message'   => '%s Using flush_rewrite_rules() is not required on WordPress.com VIP since it is done on every deploy and every theme switch. Using this will cause excessive writes to the database and slow down performance',
+				'functions' => array(
+					'flush_rewrite_rules',
+				),
+			),
+
 			// @link https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#mobile-detection
 			'wp_is_mobile' => array(
 				'type'      => 'error',


jrfnl:
Just checking: what do you mean by "Go" in this context ?

> make sure people aren't flushing on every page load.

As far as I can see, that would be impossible to check for via a static code parser like PHPCS.
Could be mentioned as part of the warning, but not checked for.

westonruter:
@jrfnl: https://vip.wordpress.com/documentation/vip-go/

And right, we could make `flush_rewrite_rules()` a hard error in `WordPress-VIP` (shared cloud hosting) but make it a warning on a yet-to-be-developed `WordPress-VIP-Go` standard.

sboisvert:
@jrfnl Yes my apologies it wasn't clear, "and make sure people aren't flushing on every page load." was more to be that on VIP Go it shouldn't be done, we can only flag it as a warning and let people know they shouldn't do it on every page load and rely on a human to make the call.

jrfnl:
@westonruter @sboisvert Thanks for the clarification.

WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php
```diff
@@ -283,6 +283,15 @@ public function getGroups() {
 				),
 			),
 
+			// @link https://vip.wordpress.com/documentation/change-your-pretty-permalinks-or-add-custom-rewrite-rules/
+			'flush_rewrite_rules' => array(
+				'type'     => 'warning',
+				'message'   => '%s Using flush_rewrite_rules() is not required on WordPress.com VIP since it is done on every deploy and every theme switch. Using this will cause excessive writes to the database and slow down performance',
```
grappler: The error outputted will be
> flush_rewrite_rules Using flush_rewrite_rules() is not required on WordPress.com VIP since it is done on every deploy and every theme switch. Using this will cause excessive writes to the database and slow down performance

What about
> Using %s() is not needed on WordPress.com VIP since it is done on every deploy and every theme switch. Using this will cause excessive writes to the database and slow down performance.

Make the function dynamic, instead of required use needed or necessary, add a fullstop to the end of the second sentence.

grappler:
As to VIP Go and VIP, you can easily create a custom ruleset for VIP GO where you can change the notice to a Warning and change the message. You can see how it can be done in the PR https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/633/files#diff-23b2b1d22ca2f4c52692ecf35d074f7b

jrfnl:
@sboisvert @tomjn Just checking - as the VIP sniff are going to be deprecated in WPCS, can this PR be closed ?

tomjn:
I'm in favour of closing this, but I think flush rewrite rules calls should flag as a warning for general WP usage, especially if it's possible to detect it's running on every page load. Perhaps a future sniff under a performance category?

PR #670, opened 2016-08-26
Fix bug in VIP restricted functions + unit tests

jrfnl:
:exclamation: **This PR should be merged before 0.10.0 is released as it contains a bug fix for a bug which was introduced in one of the 0.10.0 commits.**

The `urlencode` check was checking for `rawurlencode()` instead of `urlencode()` while at the same time advising to use `rawurlencode()`.
The grammar of the error message was incorrect as well.
Both are now fixed.

In the [same commit](e7707b2338e7823a2d14cfbdca8bdc88488d6081) as this was introduced, a number of other restricted functions were added, but none of them were accompanied by unit tests.
This is now fixed as well.


jrfnl:
@sboisvert Stéphane, could you have a quick look at this one as it concerns a commit you originally made ?

sboisvert:
@jrfnl Looks good, thanks for adding the unit tests and fixing urlencode / rawurlencode!

PR #675, opened 2016-08-30
Update PhpStorm documentation link

tnorthcutt:
The old link goes to a page that makes no mention of WordPress

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Will result in following output:
 
 ### PhpStorm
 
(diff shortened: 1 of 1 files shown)


westonruter:
@tnorthcutt actually, while the new link refers to WordPress it does _not_ refer to PHP_CodeSniffer. The “WordPress Code Style” in PhpStorm is not the same as “WordPress Coding Standards” via PHP_CodeSniffer. Really, both should be configured for PhpStorm. The code style configuration helps auto-format code as you write it, and then PHP_CodeSniffer checks code that others wrote and catches a lot more specifics that the built-in PhpStorm code style rules don't catch.

tnorthcutt:
more-you-know

PR #685, opened 2016-09-10
I18n: Always allow text domain of 'default' to facilitate feature plugins

westonruter:
When developing feature plugins, there is often a need to re-use strings in core. In such cases, the sniff should allow an explicit `default` text domain to allow for core strings to be re-used without being flagged.

For example: https://github.com/xwp/wp-customize-object-selector/blob/c7957ea0e7549a7d3fdc8ef0f811e3e05b4f46e4/php/class-plugin.php#L90

### DIFF
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -25,7 +25,7 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 	 * @todo Eventually this should be able to be auto-supplied via looking at $phpcs_file->getFilename()
 	 * @link https://youtrack.jetbrains.com/issue/WI-17740
 	 *
-	 * @var string
+	 * @var array
 	 */
 	public $text_domain;
 
@@ -102,6 +102,9 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 
 		if ( ! empty( self::$text_domain_override ) ) {
 			$this->text_domain = self::$text_domain_override;
+		}
+		if ( is_string( $this->text_domain ) ) {
+			$this->text_domain = explode( ',', $this->text_domain );
 		}
 
 		if ( '_' === $token['content'] ) {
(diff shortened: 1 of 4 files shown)


JDGrimes:
Would it be better to just let each project specify an array of textdomains instead of only one? Then projects that need to use the `default` textdomain could configure that in their own ruleset. Same for any time a project needs to use multiple text domains for some other reason.

westonruter:
@JDGrimes good idea. How's b8e0b5f?

grappler:
This will be usefull for with the theme review as we allow a second text domain to be used by a framework.

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -103,6 +103,9 @@ public function process( PHP_CodeSniffer_File $phpcs_file, $stack_ptr ) {
 		if ( ! empty( self::$text_domain_override ) ) {
 			$this->text_domain = self::$text_domain_override;
 		}
+		if ( is_string( $this->text_domain ) ) {
+			$this->text_domain = explode( ',', $this->text_domain );
```
jrfnl: You might want to add a `array_map( 'trim', $this->text_domain);` after this to allow for spaces in the array value like:
```xml
	    <properties>
	        <property name="text_domain" value="default, _s, tgmpa" type="array" />
	    </properties>
```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -253,8 +256,8 @@ protected function check_argument_tokens( PHP_CodeSniffer_File $phpcs_file, $con
 		}
 
 		if ( T_CONSTANT_ENCAPSED_STRING === $tokens[0]['code'] ) {
-			if ( 'domain' === $arg_name && ! empty( $this->text_domain ) && trim( $content, '\'""' ) !== $this->text_domain ) {
-				$phpcs_file->$method( 'Mismatch text domain. Expected \'%s\' but got %s.', $stack_ptr, 'TextDomainMismatch', array( $this->text_domain, $content ) );
+			if ( 'domain' === $arg_name && ! empty( $this->text_domain ) && ! in_array( trim( $content, '\'""' ), $this->text_domain, true ) ) {
+				$phpcs_file->$method( 'Mismatch text domain. Expected \'%s\' but got %s.', $stack_ptr, 'TextDomainMismatch', array( join( "' or '", $this->text_domain ), $content ) );
```
jrfnl: Just wondering: why use the alias `join()` when the actual function used is `implode()` ?

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -268,8 +271,8 @@ protected function check_argument_tokens( PHP_CodeSniffer_File $phpcs_file, $con
 			if ( ! empty( $interpolated_variables ) ) {
 				return false;
 			}
-			if ( 'domain' === $arg_name && ! empty( $this->text_domain ) && trim( $content, '\'""' ) !== $this->text_domain ) {
-				$phpcs_file->$method( 'Mismatch text domain. Expected \'%s\' but got %s.', $stack_ptr, 'TextDomainMismatch', array( $this->text_domain, $content ) );
+			if ( 'domain' === $arg_name && ! empty( $this->text_domain ) && ! in_array( trim( $content, '\'""' ), $this->text_domain, true ) ) {
+				$phpcs_file->$method( 'Mismatch text domain. Expected \'%s\' but got %s.', $stack_ptr, 'TextDomainMismatch', array( join( "' or '", $this->text_domain ), $content ) );
```
jrfnl: dito

WordPress/Sniffs/WP/I18nSniff.php
westonruter: Just a habit of mine I guess (and influence from JavaScript). Happy to use `implode` instaed.

WordPress/Sniffs/WP/I18nSniff.php
westonruter: Great point.

PR #689, opened 2016-09-13
Defer to upstream PEAR `ValidClassName` sniff.

jrfnl:
As suggested in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/649#issue-168558657

Closes #649


westonruter:
It looks like the PEAR sniff doesn't check for `CamelCaps`.

jrfnl:
> It looks like the PEAR sniff doesn't check for CamelCaps.

It doesn't use the CamelCaps function, but it *does* check for them - see: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/PEAR/Sniffs/NamingConventions/ValidClassNameSniff.php#L59

westonruter:
OK. It looks like it uses a more generic `Invalid` error code rather than the more explicit `NotCamelCase` that is in WPCS.

jrfnl:
True, but that shouldn't really matter or does it ? The error message itself is a bit more generic as well, but we could alter that via the config.

Message in WP sniff:
`%s name "%s" is not in camel caps format` where 1: token type (class, interface) and 2: the class name.

Message in PEAR sniff:
`%s name is not valid; consider %s instead` where 1: the class name and 2: CamelCaps version of the class name.

Issue #549, opened 2016-04-01
NotSnakeCaseMemberVar complains about core cat_ID


Rarst:
`cat_ID` is still returned with that name in core's category objects (as opposed to `term_id`).

I've tried adding it to `$whitelisted_mixed_case_member_var_names`, but for some reason it didn't take so might be issue in implementation.

```
135 | ERROR | Variable "cat_ID" is not in valid snake_case format
     |       | (WordPress.NamingConventions.ValidVariableName.NotSnakeCaseMemberVar)
```

westonruter:
@Rarst I think this is by design. The sniff is designed to whitelist object properties that have mixed cases, but variables themselves are supposed to have the proper case. I suppose the only place where you'd be forced to use mixed case for a variable name is if you use `compact()` to construct an array to pass into a core function.

Take a look at #551. This includes the missing `cat_ID` for the whitelisted names, but it also allows all these whitelisted names to be used as variable names in addition to object properties. Is this something we want to do? Maybe we should introduce a variant error code for `NotSnakeCase` like `NotSnakeCaseCore` which could be used if a variable is snake case but it is among the whitelisted terms.

Rarst:
Hm, to be clear it _was_ a property name like `$category->cat_ID`, so the sniff's message isn't entirely precise. Maybe that's the reason for the quirk, it confuses variable/property distinction somehow?

westonruter:
@Rarst you're right. The error message was not great. I've fixed this in d8d7942da1516c78e76eddee43ad44c0b48210a5. I've also amended the PR to make it clear that using `$cat_ID` is bad, but `$category->cat_ID` is ok. Please review #551.

PR #691, opened 2016-09-14
Apply snakeCase naming whitelist to vars

JDGrimes:
Not just properties as former logic did (inadvertently, I think).

### DIFF
--- a/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
@@ -147,7 +147,6 @@ class WordPress_Sniffs_NamingConventions_ValidVariableNameSniff extends PHP_Code
 					$error      = 'Object property "%s" is not in valid snake_case format';
 					$error_name = 'NotSnakeCaseMemberVar';
 				} else {
-					var_dump( $this->whitelisted_mixed_case_member_var_names );
 					$error      = 'Variable "%s" is not in valid snake_case format';
 					$error_name = 'NotSnakeCase';
 				}
--- a/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
@@ -142,12 +142,15 @@ class WordPress_Sniffs_NamingConventions_ValidVariableNameSniff extends PHP_Code
 		}
 
 		if ( self::isSnakeCase( $var_name ) === false ) {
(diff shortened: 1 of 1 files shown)


JDGrimes:
OK, it looks like the tests were checking for the old behavior, so the build is failing. I wonder whether we should allow the properties and variables to be whitelisted separately? We could even use a single list, but prefix the variables with `$`. Thoughts?

jrfnl:
Related #551 #549 

My two pennies:

With object properties you may not always have a choice which case to use as you can extend native PHP classes or classes from other sources which will force you to use the code style of the parent class, so whitelisting makes sense.

For variables, you (should) always have a choice and these therefore should comply with the WP coding standards without exception (ok, with the exception of native PHP variables which do not comply, but those are whitelisted anyway)

JDGrimes:
Yeah, after thinking over this more, I guess maybe it is better not to allow whitelisting of variables. My reason for expecting it was that the name of the whitelist is `customVariablesWhitelist`. Shouldn't that be changed to `customPropertiesWhitelist` instead? (although that would break back-compat, unless we just deprecate `customVariablesWhitelist` for now). The reason that I wanted to use it was that I have some Codeception tests for a plugin, and these heavily use `$I` to hold the user object. However, I can just exclude this sniff from running over that directory. The other variable that I wanted to whitelist was `$EZSQL_ERROR`. It is an ancient remnant of the merge of ezsql into core as `wpdb`, and I use it in my PHPUnit tests to catch database errors. We can add that to the built-in variable whitelist, or else I'll just exclude those files for this sniff in my ruleset as well.

jrfnl:
> Shouldn't that be changed to customPropertiesWhitelist instead? (although that would break back-compat, unless we just deprecate customVariablesWhitelist for now).

*grin* I'd had that same _"hang on !?"_ moment a while back and decided against sending in a PR to avoid breaking backward compatibility, but a PR for it does make sense as the property name is currently confusing.


> The other variable that I wanted to whitelist was $EZSQL_ERROR. It is an ancient remnant of the merge of ezsql into core as wpdb

If `$EZSQL_ERROR` is part of WP - even though rarely used in userland - whitelisting it makes sense to me.

jrfnl:
Just thinking: we could even be stricter/more precise with the property name check and only allow deviations when the class extends another class or interface (or if the property is whitelisted).

jrfnl:
@JDGrimes Just doing some clean up. Can I delete this branch ?

JDGrimes:
> @JDGrimes Just doing some clean up. Can I delete this branch ?

@jrfnl done.

PR #700, opened 2016-10-04
Whitelist $EZSQL_ERROR in variable naming conventions sniff

JDGrimes:
Though obscure, this global variable is used by WordPress core in `wpdb`, as a remnant of when it was forked from EzSQL. See https://core.trac.wordpress.org/browser/branches/4.6/src/wp-includes/wp-db.php#L1350.

See #691 for previous discussion.

### DIFF
--- a/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
@@ -46,6 +46,15 @@ class WordPress_Sniffs_NamingConventions_ValidVariableNameSniff extends PHP_Code
 	);
 
 	/**
+	 * Mixed-case variables used by WordPress.
+	 *
+	 * @var array
+	 */
+	protected $wordpress_mixed_case_vars = array(
+		'EZSQL_ERROR' => true,
+	);
+
+	/**
 	 * List of member variables that can have mixed case.
 	 *
 	 * @var array
@@ -96,6 +105,11 @@ class WordPress_Sniffs_NamingConventions_ValidVariableNameSniff extends PHP_Code
 
 		// If it's a php reserved var, then its ok.
 		if ( in_array( $var_name, $this->php_reserved_vars, true ) ) {
+			return;
+		}
+
+		// Likewise if it is a mixed-case var used by WordPress core.
+		if ( isset( $this->wordpress_mixed_case_vars[ $var_name ] ) ) {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


westonruter:
Tests are failing on PHPCS `master` and on PHP 5.2. 

In `master` it looks like the `phpcbf` output format has changed, and for PHP 5.2 it looks like its choking on a namespace separator.

JDGrimes:
@westonruter See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/696 as to why this is failing on `master`. `develop` is currently failing on `master` as well.

WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
```diff
@@ -57,6 +57,15 @@ class WordPress_Sniffs_NamingConventions_ValidVariableNameSniff extends PHP_Code
 	);
 
 	/**
+	 * Mixed-case variables used by WordPress.
+	 *
+	 * @var array
+	 */
+	public $wordpress_mixed_case_vars = array(
+		'EZSQL_ERROR',
```
jrfnl: 1. `public` properties within sniffs can be overruled by setting the property in the ruleset. I think the visibility here needs to change to either `protected` or `private`.
2. The pattern of `array( 'string_key' =>true )` for array properties would allow for checking using `isset()` rather than the slower `in_array()`.

JDGrimes: I was just following the pattern already set with the sniff's other similar properties. Obviously, using `protected` and `isset()` is better, though. Changing the other properties would break back-compat, so I haven't modified them.

jrfnl: As the `$php_reserved_vars` and `$whitelisted_mixed_case_member_var_names` properties look like they were never intended to be overloadable, I would have no problem breaking backward compatibility in this case.
If someone would currently be overloading them, they are effectively overwriting our defaults which would be a pretty bad idea.

JDGrimes: OK, but let's make a separate PR for that?

jrfnl: Fine by me ;-)

jrfnl:
@JDGrimes Could you rebase (& squash) so that we can get the travis checks to pass ? I can't merge until they do :disappointed:

PR #701, opened 2016-10-04
Rename "variables" whitelist to "properties"

JDGrimes:
Only properties can be whitelisted using this, so the name "variables" is confusing. See #691.

The term "member var" is used elsewhere within this sniff, so possibly that should be used here for consistency, rather than "property".

I've kept the old whitelist around for back-compat, but I've marked it as deprecated. The old internal flag `$addedCustomVariables` I removed completely rather than deprecating, though I suppose in theory it is possible that someone has a custom sniff extending from this one that might be referencing it.

### DIFF
--- a/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
@@ -60,22 +60,33 @@ class WordPress_Sniffs_NamingConventions_ValidVariableNameSniff extends PHP_Code
 	);
 
 	/**
-	 * Custom list of variables which can have mixed case.
+	 * Custom list of properties which can have mixed case.
+	 *
+	 * @since 0.11.0
+	 *
+	 * @var string[]
+	 */
+	public $customPropertiesWhitelist = array();
+
+	/**
+	 * Whether the custom whitelisted properties were added to the default list yet.
 	 *
 	 * @since 0.10.0
+	 * @since 0.11.0 Name changed from $addedCustomVariables.
+	 *
+	 * @var bool
+	 */
+	protected $addedCustomProperties = false;
+
+	/**
+	 * Custom list of properties which can have mixed case.
+	 *
+	 * @since 0.10.0
+	 * @deprecated 0.11.0 Use $customPropertiesWhitelist instead.
 	 *
 	 * @var string[]
 	 */
 	public $customVariablesWhitelist = array();
-
-	/**
-	 * Whether the custom whitelisted variables were added to the default list yet.
-	 *
-	 * @since 0.10.0
-	 *
-	 * @var bool
-	 */
-	protected $addedCustomVariables = false;
 
 	/**
 	 * Processes this test, when one of its tokens is encountered.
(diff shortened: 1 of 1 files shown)


westonruter:
Differences in `phpcbf` output is again causing a build failure here. Should we remove PHPCBF `master` from being tested?

jrfnl:
> Should we remove PHPCBF master from being tested?

As there are fixable sniffs within the WPCS native sniff set, excluding PHPCBF from testing would IMHO be a bad idea.

jrfnl:
How will the user be informed of the property deprecated ? Should the sniff throw a warning when the property is encountered and not empty ?
I imagine something similar to:  https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/XSS/EscapeOutputSniff.php#L129

WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
```diff
@@ -236,12 +246,30 @@ public static function isSnakeCase( $var_name ) {
...
+	 * @param PHP_CodeSniffer_File $phpcs_file The file being scanned.
+	 *
 	 * @return void
 	 */
-	protected function mergeWhiteList() {
-		if ( false === $this->addedCustomVariables && ! empty( $this->customVariablesWhitelist ) ) {
-			$this->whitelisted_mixed_case_member_var_names = array_merge( $this->whitelisted_mixed_case_member_var_names, $this->customVariablesWhitelist );
-			$this->addedCustomVariables = true;
+	protected function mergeWhiteList( $phpcs_file ) {
+		if ( ! empty( $this->customVariablesWhitelist ) ) {
```
jrfnl: This first block is also only needed once, so should either also be enclosed in a `if( false === $this->addedCustomProperties )` or probably better - start the method with the below and remove the check for this from line 267:

```php
if ( true === $this->addedCustomProperties ) {
	return;
}
```

WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
```diff
@@ -60,22 +60,32 @@ class WordPress_Sniffs_NamingConventions_ValidVariableNameSniff extends PHP_Code
...
 	 */
-	public $customVariablesWhitelist = array();
+	public $customPropertiesWhitelist = array();
 
 	/**
-	 * Whether the custom whitelisted variables were added to the default list yet.
+	 * Whether the custom whitelisted properties were added to the default list yet.
 	 *
-	 * @since 0.10.0
+	 * @since 0.11.0
```
jrfnl: I'd still leave this at `0.10.0`, but may be add a note that the property name was adjusted.

WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
JDGrimes: Ahh, good catch, I intended that but forgot to double check it in the end result.

jrfnl:
@JDGrimes Could you rebase (& squash) so that we can get the travis checks to pass ? I can't merge until they do :disappointed:

PR #703, opened 2016-10-05
Fix failing unit tests / extended upstream classes

### LINKED ISSUE #696: Don't extend upstream sniffs.
This is related to a buggy upstream change and has been reported upstream: https://github.com/squizlabs/PHP_CodeSniffer/issues/1177

jrfnl:
Fixes #696 

#### TL;DR: This PR fixes unnecessary dependencies on upstream sniffs and handles necessary dependencies more gracefully. Once this PR is merged, the unit tests will no longer fail.

---
There are four upstream sniffs which were being extended by one or more WP sniffs.
This could cause problems if something changes upstream and/or if the WPCS tests are not purely testing for _WPCS_ additions, but also testing upstream checks - as was the case for the `ArrayDeclaration` sniff which was causing the unit tests to fail.

I've reviewed all relevant sniffs.

### `Generic.PHP.ForbiddenFunctions`
There are four sniffs which extend this:
* WordPress.PHP.DiscouragedFunctions
* WordPress.VIP.FileSystemWritesDisallow
* WordPress.VIP.SessionFunctionsUsage
* WordPress.VIP.TimezoneChange

None of these duplicate the logic of the original class, they only provide the class with an array of functions to check for and in some cases, an adjusted error message.
The test files related to these sniffs only cover the functions provided by the child class.

#### Conclusion: usage as-is is safe.

Related #582, #633 

### `PEAR.NamingConventions.ValidFunctionName`

The `WordPress.NamingConventions.ValidFunctionName` sniff extends this.
In this case, the exact opposite applies: the functional logic is not used at all, only the class properties of the original class is used to prevent duplication of the maintenance of these (native PHP) exceptions to the rules.
The test files in this case do test that the properties from the parent are handled appropriately, so there is some bleed through.

As the properties we're talking about are a list of the PHP magic methods, I don't expect any to be removed any time soon, but we will automatically get the benefit of any being added upstream if PHP would add more magic methods.

### `Squiz.WhiteSpace.CastSpacing`

The `WordPress.WhiteSpace.CastStructureSpacing` sniff was based upon this sniff and extended it.
This was a recent change. Previously the logic of the parent was duplicated. This duplication was removed early in the 0.11.0 cycle in favour of extending the sniff.
The WP version adds two additional checks to the parent.
Upon further review, the parent does not need to be extended, but can be added to the ruleset as is. The two additional checks become a stand-alone sniff. The only code duplication involved is the `register()` method.
The tests did contain bleed through, testing both the parent as well as the child. This has been cleaned up.

### `Squiz.Arrays.ArrayDeclaration`

This one is the most awkward and complicated one and is extended by the `WordPress.Arrays.ArrayDeclaration` sniff.
The parent class does some initial processing which already contains three checks and then - depending on the kind of array - passes off to either the `processSingleLineArray()` or the `processMultiLineArray()` method.
The logic of the `processSingleLineArray()` method of the parent is used as-is, one check contained there-in is excluded via the ruleset and four additional checks are added in the child.
The logic of the `processMultiLineArray()` method is however duplicated in the child with some code removed. This is/was needed as the parent `return`s early for some checks which we want to exclude. However the excluding from the ruleset does not prevent the `return` from being enacted, which would have the unintended side-effect of later checks being ignored.

This was already reported upstream in https://github.com/squizlabs/PHP_CodeSniffer/issues/582 by @JDGrimes over a year ago and while the intention (upstream) is to split this sniff into more re-usable parts which will live in the `Generic` standard, this is a dormant ticket.

The problem here is that this is the only upstream sniff dealing with array token walking, that the logic is sound and that we want/need 70% of the checks being done by the sniff.

The test file contains a *_lot_** of bleed through though - 99% of all tests relate to the upstream checks.

My take after investigating was as follows:
* Create a new `WordPress.Arrays.ArrayDeclarationSpacing` sniff for the additional checks in `processSingleLineArray()`. This sniff **will** extend the parent, but only for the process logic to get to the `processSingleLineArray()` method. The checks in the initial processing will be excluded from the ruleset to prevent duplicate messages being thrown.
The tests for the new sniff will only test the WP specific additional checks without bleed through.
* For the generic checks in the initial parent `process()` and select checks in the `processMultiLineArray()`, the `WordPress.Arrays.ArrayDeclaration` remains.
The class will still extend the parent and the `processMultiLineArray()` will still be duplicated with a select number of checks commented out.
As the class now doesn't *add* anything to the parent - just removes checks - no separate test file is needed as all checks are already being tested upstream.
As this sniff will from time to time need to be synced with upstream to keep up with fixes made there, I've largely reverted the WPCS code style changes  for this file and excluded this file from being sniffed for code style. This will facilitate _much_ easier syncing with upstream.

#### Other notes

Where necessary I've added additional documentation about this for anyone looking at these sniffs in the future.

---
Ok, so this has become quite a long tale, but it seemed prudent to document this carefully and in detail.

For anyone who's read this far: kudo's & let's merge this baby ;-)


WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
```diff
@@ -424,48 +454,132 @@ public function processMultiLineArray( PHP_CodeSniffer_File $phpcsFile, $stackPt
...
+						$phpcsFile->fixer->replaceToken( ( $index['value'] - 1 ), str_repeat(' ', $expected ) );
+						$phpcsFile->fixer->endChangeset();
+					} else if ($found === 0) {
+						$phpcsFile->fixer->addContent( ( $index['value'] - 1 ), str_repeat(' ', $expected ) );
+					} else {
+						$phpcsFile->fixer->replaceToken( ( $index['value'] - 1 ), str_repeat(' ', $expected ) );
+					}
+				}
+			} // end if
+			*/
```
westonruter: Should this commented-out code just be removed?

WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
```diff
@@ -424,48 +454,132 @@ public function processMultiLineArray( PHP_CodeSniffer_File $phpcsFile, $stackPt
...
+
 			// Check each line ends in a comma.
 			$valueLine = $tokens[ $index['value'] ]['line'];
 			$nextComma = false;
 			for ( $i = $index['value']; $i < $arrayEnd; $i++ ) {
 				// Skip bracketed statements, like function calls.
-				if ( T_OPEN_PARENTHESIS === $tokens[ $i ]['code'] ) {
+				if ($tokens[$i]['code'] === T_OPEN_PARENTHESIS) {
```
westonruter: Why the de-yodification?

WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
```diff
@@ -221,7 +187,7 @@ public function processMultiLineArray( PHP_CodeSniffer_File $phpcsFile, $stackPt
 						);
 
 						$fix = $phpcsFile->addFixableError( $error, $nextToken, 'SpaceBeforeComma', $data );
-						if ( true === $fix ) {
+						if ($fix === true) {
```
westonruter: Why the removal of the whitespace and the yoda conditions?

WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
```diff
@@ -328,28 +333,51 @@ public function processMultiLineArray( PHP_CodeSniffer_File $phpcsFile, $stackPt
...
+						$fix = $phpcsFile->addFixableError( $error, $value['value'], 'ValueNotAligned', $data );
+						if ($fix === true) {
+							if ($found === 0) {
+								$phpcsFile->fixer->addContent( ( $value['value'] - 1 ), str_repeat(' ', $expected ) );
+							} else {
+								$phpcsFile->fixer->replaceToken( ( $value['value'] - 1 ), str_repeat(' ', $expected ) );
+							}
+						}
+					}
+					*/
```
westonruter: Shouldn't this commented-out code just be removed?

WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
```diff
@@ -275,25 +248,57 @@ public function processMultiLineArray( PHP_CodeSniffer_File $phpcsFile, $stackPt
...
+
+					$phpcsFile->fixer->replaceToken( $i, '' );
+				}
+
+				$phpcsFile->fixer->endChangeset();
+			}
+
+			return;
+		} // end if
+		*/
```
westonruter: Shouldn't this commented-out code be removed?

WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
```diff
@@ -235,14 +201,21 @@ public function processMultiLineArray( PHP_CodeSniffer_File $phpcsFile, $stackPt
...
+			}//end if
 
-			if ( T_DOUBLE_ARROW === $tokens[ $nextToken ]['code'] ) {
+			if ($tokens[$nextToken]['code'] === T_DOUBLE_ARROW) {
+				/*
+				if ($singleUsed === true) {
+					$error = 'Key specified for array entry; first entry has no key';
+					$phpcsFile->addError( $error, $nextToken, 'KeySpecified' );
+					return;
+				}
```
westonruter: Can this commented-out code be removed?

WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
```diff
@@ -205,8 +163,16 @@ public function processMultiLineArray( PHP_CodeSniffer_File $phpcsFile, $stackPt
...
 				}
 
-				if ( false === $keyUsed ) {
-					if ( T_WHITESPACE === $tokens[ ( $nextToken - 1 ) ]['code'] ) {
+				/*
+				if ($keyUsed === true && $tokens[$lastToken]['code'] === T_COMMA) {
+					$error = 'No key specified for array entry; first entry specifies key';
+					$phpcsFile->addError( $error, $nextToken, 'NoKeySpecified' );
+					return;
+				}
```
westonruter: Can this commented-out code be removed?

WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
```diff
@@ -124,17 +60,37 @@ public function processMultiLineArray( PHP_CodeSniffer_File $phpcsFile, $stackPt
...
+
+			$fix = $phpcsFile->addFixableError( $error, $arrayEnd, 'CloseBraceNotAligned', $data );
+			if ($fix === true) {
+				if ($found === 0) {
+					$phpcsFile->fixer->addContent( ( $arrayEnd - 1 ), str_repeat(' ', $expected ) );
+				} else {
+					$phpcsFile->fixer->replaceToken( ( $arrayEnd - 1 ), str_repeat(' ', $expected ) );
+				}
+			}
+			*/
```
westonruter: Can this commented-out code be removed?

WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
jrfnl: See above

WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
jrfnl: See above

WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
jrfnl: dito

WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
jrfnl: dito

WordPress/Sniffs/Arrays/ArrayDeclarationSniff.php
jrfnl: See above

Issue #706, opened 2016-10-22
Include `strip_tags` in list of escaping functions.


schlessera:
The PHP function `strip_tags` is currently not recognized as a valid escaping function.

The function "returns a string with all NULL bytes, HTML and PHP tags stripped" (http://php.net/manual/en/function.strip-tags.php).

It should be added to the list of escaping functions to be a valid way of getting rid of unneeded "Expected next thing to be an escaping function" errors.

schlessera:
As @westonruter has correctly pointed out, `strip_tags` does not work correctly as an escaping function.

PR #707, opened 2016-10-22
Add `strip_tags` to the list of valid escaping functions. See #706

schlessera:
When escaping output with `strip_tags`, WPCS throws an error, as it does not recognize `strip_tags` as an escaping function. This PR adds the function to avoid this false positive.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -61,6 +61,7 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 		'rawurlencode'         => true,
 		'sanitize_html_class'  => true,
 		'sanitize_user_field'  => true,
+		'strip_tags'           => true,
 		'tag_escape'           => true,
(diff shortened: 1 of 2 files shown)


westonruter:
Should this be considered an escaping function or a sanitizing function? I should think this function would be used on input not output.

schlessera:
I would say it can also be used for escaping, as a cheaper alternative to more involved methods.

Either way, if it is used for escaping, it does the job, as nothing "executable" is left.

westonruter:
> Either way, if it is used for escaping, it does the job, as nothing "executable" is left.

Not quite. Consider:

```
<input type=search value="<?php echo strip_tags( wp_unslash( $_GET['s'] ) ) ?>">
```

And consider an `s` GET param that looks like: `Malicious " onclick="alert('Evil')`. So `strip_tags` allows for malicious attributes to be injected into a tag.

schlessera:
Yes, you are right, I stand corrected.

Issue #708, opened 2016-10-22
Include `htmlspecialchars` in list of escaping functions.


schlessera:
The PHP function `htmlspecialchars` is currently not recognized as a valid escaping function.

The function "converts special characters into HTML entities" (http://php.net/manual/en/function.htmlspecialchars.php).

It is actually what `esc_textarea()` is built upon.

It should be added to the list of escaping functions to be a valid way of getting rid of unneeded "Expected next thing to be an escaping function" errors.

johnbillion:
What's your use case for using `htmlspecialchars()` instead of one of the `esc_*()` functions?

thenbrent:
@johnbillion we are using `htmlspecialchars()` in one place because of the issue detailed here: https://github.com/woocommerce/woocommerce/issues/9287

And much more briefly summarised as `esc_attr does not double encode` here: https://github.com/woocommerce/woocommerce/commit/cf54ea8fc055e079dc956889700af1c7cf614246

TukuToi:
The question is less what the use case is, and more that it is a valid escaping function inbuilt in PHP.

I would like to add my +1 to this "request".

If anyone is interested in use cases, this particular case involves using WPCS in a project that... is not built on WP.
Yes, it is clear to me that WPCS was made and will always be made for and with WP in mind, yet, this is my exact use case :)
(which might be argued to be "wrong", but still I would profit from htmlspecialchars being recognised, and agree with @thenbrent that it makes sense to include, unless concerns could be presented to not do so).

PR #709, opened 2016-10-22
Add `htmlspecialchars` to the list of valid escaping functions. See #708

schlessera:
When escaping output with `htmlspecialchars`, WPCS throws an error, as it does not recognize `htmlspecialchars` as an escaping function. This PR adds the function to avoid this false positive.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -54,6 +54,7 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 		'esc_url'              => true,
 		'filter_input'         => true,
 		'filter_var'           => true,
+		'htmlspecialchars'     => true,
 		'intval'               => true,
 		'json_encode'          => true,
 		'like_escape'          => true,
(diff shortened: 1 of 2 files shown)


westonruter:
Why shouldn't use of an `esc_*` function be required instead?

schlessera:
Why should using a WordPress-coupled function (with associated filter) be required?

I often have a combination of general-purpose code + WordPress wrapper to hook it up to WordPress.

I don't want to unnecessary couple my code to WordPress if this is not necessary.

schlessera:
Thought some more about this, and while I still think that WPCS sniffer should not throw an error if it encounters `htmlspecialchars()` instead of `esc_textarea()`, my specific issue in the code I'm currently working on is that I have some files on the wrong side of my PHP <=> WordPress separation. The general-purpose code is run through PSR-2 standard, and the WordPress through this standard.

westonruter:
@schlessera probably a better approach for you then is to register `htmlspecialchars` as an additional/custom escaping function in your own project ruleset. See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/0.10.0/WordPress/Sniffs/XSS/EscapeOutputSniff.php#L24-L31

PR #710, opened 2016-10-22
Fix I18n placeholder numbering and add error for mixed placeholders

### LINKED ISSUE #698: Wrong fix for multiple placeholders should be ordered
Given the following code:

```
__( '%s for %s item', 'my-slug' );
```

PHPCS will generate an error saying that `Multiple placeholders should be ordered. Expected '%1$s, %2$s', but got %s,%s.`. If I then run `phpcbf --standard=WordPress-Core example.php` to fix this error the resulting code is:

```
__( '%1$s for %1$d item', 'my-slug' );
```

Instead of:

```
__( '%1$s for %2$d item', 'my-slug' );
```

It seems to me that this is happening because when `str_replace()` is called (https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/WP/I18nSniff.php#L354) with the following arguments:

```
str_replace(array('%s', '%s'), array('%1$s', '%2$s'), "'%s for %s item'")
```

It is replacing all the occurrences of `%s` with `%1$s`.

jrfnl:
I've investigated #698 and discovered a couple of additional issues because of it.

This PR fixed both the original issue as well as the additionally discovered ones.

Fixes #698 

#### Issue #698 - Fix fixing of unordered placeholders.

If several of the same kind of unordered placeholders were found - `%s` -, the `str_replace()` would replace them all with the replacement value for the first, i.e. `%1$s`, resulting in something like `'my %1$s and my %1$s'`.

`preg_replace()` allows for limiting the amount of replacements per match so offers the ability to replace each placeholder with their corresponding replacement.

This does however create some other complexities as a `$` in a regex replacement value can be interpreted as a match replacement, so will need escaping to prevent potential replacement with a numbered match.

Example:
Placeholder found: `'%01d'`
Should become: `'%1$01d'`
But the `$01` within the replacement value in `preg_replace()` will be interpreted as the first submatch - which doesn't exist, so would be replaced by `''` resulting in `'%1d'`
So the dollar in the replacement value needs escaping before passing it on to `preg_replace()`.

#### Using dollars in double quoted strings with placeholders

Along the same lines as the above, if the original string was double quoted, the `$` would also need escaping in the result string to prevent it being interpreted as a PHP variable.

This was currently not done, which could lead to a replacement looking like `"my %1$d pennies"` and `$d` being interpreted as a PHP variable, resulting either in an undefined variable notice or an unintended replacement being made.

The check for double quoted strings I've build in now is quite simplistic, but should be sufficient in 95% of all cases.

#### Not all placeholders were being recognized.

The regexes used throughout the sniff (constants at the top of the file) did not account for all possible placeholders for `sprintf()`.
This has been fixed as well.

Ref: http://php.net/manual/en/function.sprintf.php

#### Mixed ordered and unordered placeholders.

Up to now, a string with mixed ordered and unordered placeholders - `%1$s for %d` - would be disregarded by the check for placeholder ordering.
As these do not lend themselves to automatic fixing, I've chosen to add an error notice whenever a string with mixed placeholders is found.

Includes additional unit tests + adjusted `.fixed` file.

### DIFF
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -331,27 +331,37 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 		$fixable_method = empty( $context['warning'] ) ? 'addFixableError' : 'addFixableWarning';
 
 		// UnorderedPlaceholders: Check for multiple unordered placeholders.
-		preg_match_all( self::UNORDERED_SPRINTF_PLACEHOLDER_REGEX, $content, $unordered_matches );
+		$unordered_matches_count = preg_match_all( self::UNORDERED_SPRINTF_PLACEHOLDER_REGEX, $content, $unordered_matches );
 		$unordered_matches       = $unordered_matches[0];
-		$unordered_matches_count = count( $unordered_matches );
 
 		if ( $unordered_matches_count >= 2 ) {
 			$code = 'UnorderedPlaceholders' . ucfirst( $arg_name );
 
-			$suggestions = array();
+			$suggestions     = array();
+			$replace_regexes = array();
+			$replacements    = array();
 			for ( $i = 0; $i < $unordered_matches_count; $i++ ) {
-				$suggestions[ $i ] = substr_replace( $unordered_matches[ $i ], ( $i + 1 ) . '$', 1, 0 );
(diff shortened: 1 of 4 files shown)


paulschreiber:
Has this fix made it in to a release?

jrfnl:
0.11.0 will be released later this week (or early next week).

PR #716, opened 2016-11-21
Added downloads count badge to readme

Rarst:
Was looking through stats and thought 74K and growing is quite a nice number of downloads to showcase with a badge. :)


jrfnl:
Just as a side-note: this only shows downloads through packagist AFAIKS, not direct downloads or git-based installs, so the real number of downloads will be higher.

PR #730, opened 2016-12-14
Document the functions whitelist option in AbstractFunctionRestrictionsSniff

jrfnl:
This feature was added in #626, but not properly documented so far.

### DIFF
--- a/WordPress/AbstractFunctionRestrictionsSniff.php
+++ b/WordPress/AbstractFunctionRestrictionsSniff.php
@@ -58,14 +58,19 @@ abstract class WordPress_AbstractFunctionRestrictionsSniff implements PHP_CodeSn
 	 * This method should be overridden in extending classes.
 	 *
 	 * Example: groups => array(
-	 * 	'lambda' => array(
-	 * 		'type'      => 'error' | 'warning',
-	 * 		'message'   => 'Use anonymous functions instead please!',
-	 * 		'functions' => array( 'eval', 'create_function' ),
-	 * 	)
+	 *     'lambda' => array(
+	 *         'type'      => 'error' | 'warning',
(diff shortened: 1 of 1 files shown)


WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -61,11 +61,16 @@
 	 * 	'lambda' => array(
 	 * 		'type'      => 'error' | 'warning',
 	 * 		'message'   => 'Use anonymous functions instead please!',
-	 * 		'functions' => array( 'eval', 'create_function' ),
+	 * 		'functions' => array( 'eval', 'create_function', 'mysql_*' ),
+	 *      // Only useful when using wildcards:
```
grappler: There is a mixed use of tabs and spaces.

jrfnl: Good catch, I hadn't noticed. It should be all spaces as it's in a comment, but the earlier code wasn't. Fixed now.

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -58,14 +58,19 @@
 	 * This method should be overridden in extending classes.
 	 *
 	 * Example: groups => array(
-	 * 	'lambda' => array(
-	 * 		'type'      => 'error' | 'warning',
-	 * 		'message'   => 'Use anonymous functions instead please!',
-	 * 		'functions' => array( 'eval', 'create_function' ),
-	 * 	)
+	 * 	   'lambda' => array(
+	 * 	       'type'      => 'error' | 'warning',
```
grappler: There is still a tab here

jrfnl: Sneaky little buggers...

PR #733, opened 2016-12-17
Increase PHPCS version to 2.7.1

grappler:
We should increase the PHPCS version in travis as composer installs this version on new installs.

The reason for this is that we do not have a `composer.lock` in the project and the version in `composer.json` is set for `^2.7` which will install any minor versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ php:
 
 env:
   - PHPCS_BRANCH=master
-  - PHPCS_BRANCH=2.7.0
+  - PHPCS_BRANCH=2.7.1
 
 matrix:
   fast_finish: true


jrfnl:
I don't agree. By far not all users use Composer to install this. The minimum PHPCS version we currently depend on is 2.7.0, so the tests should run against 2.7.0 and master.

AFAICS there is at this moment no reason to up the minimum required PHPCS version to 2.7.1, so why would we ?

grappler:
Ok, I did not think about people installing WPCS using a different method. Interesting enough the only places we define the minimum version is in composer and travis.

I am fine with leaving this open till it is need or closing it now.

jrfnl:
> Interesting enough the only places we define the minimum version is in composer and travis.

Good point. I suggest that the minimum PHPCS version WPCS requires should be made explicit in the readme.

jrfnl:
Related #1177

grappler:
I have updated the PR with the latest changes to the Readme. I have also removed `composer.lock` from `gitignore` so that only the version that is in `composer.lock` is defined is installed.

jrfnl:
Suggestion: leave the `.gitignore` file as it was and don't commit `composer.lock`, but update the composer config to require `2.7.1` as the minimum ?

JDGrimes:
I'm fine just updating the `.travis.yml` version, since the sniffs still work on PHPCS 2.7.0. The only potential downside I see is confusion for contributors trying to run the tests locally against 2.7.0. Do you think that will end up biting many people?

jrfnl:
> The only potential downside I see is confusion for contributors trying to run the tests locally against 2.7.0. Do you think that will end up biting many people?

Anyone who's watching the repo will have been notified about this thread and should/could know about it.
AFAICS, the only ones currently actively working on sniffs at this point in time are @grappler and me.

As for contributors to the Theme Review sniffs project, there is a bi-weekly meeting in Slack, so I suggest we mention it there as well just in case.

Only other thing I can think of is maybe add a note to the `contributing.md` file about this, though in that case, we'll have to remember to remove it when the PHPCS requirement would be upped to 2.7.1 or higher.

PR #735, opened 2016-12-19
Test XML file

grappler:
In PHPCS 2.7.1 the XML files are not being tested any more.

> XML rulesets were never intentionally tested. It was an accident that they were parsed. But because they were, it meant that tests were not running in isolation (they were running with all the XML ruleset rules and exclusions applied) so they were not accurate.

This mean that if there is syntax error in the XML file or a ruleset is misspelled there will be no error. This simple line of code will help test the XML for this issue and return an error if there are any issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,8 @@ matrix:
     # Run against HHVM and PHP nightly.
     - php: hhvm
       sudo: required
-      dist: trusty 
-      group: edge 
+      dist: trusty
+      group: edge
       env: PHPCS_BRANCH=master
     - php: nightly
       env: PHPCS_BRANCH=master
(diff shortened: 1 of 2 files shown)


GaryJones:
What are you actually trying to achieve here?

grappler:
Sorry, I was testing an idea. Meant to make a PR to my own repo first. As a PR is already created, I have cleaned it up and add a proper description.

grappler:
> First off, the rulesets do not need to be tested against every combi in the matrix, once is enough. if [[ "$SNIFF" == "1" ]]

Makes sense. I have changed it.

> Second, the WPCS sniff code itself is run against it's own standard in the command just below, so that will test the rulesets too - it runs against Extra (incl Core) + Docs. The only one which isn't tested that way is VIP.

It is sort of easier to run it on the whole `WordPress` ruleset. That also means any new rulesets that are added like `WordPress-Theme` are automatically tested too.


I had to create a dummy file that would not generate any issues.

grappler:
Closing as fixed by https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/914

PR #736, opened 2016-12-22
Fix a number of bugs in the global var override detection

### LINKED ISSUE #300: Issues with GlobalVariablesSniff
Ok, so I ran into a lot of warnings with the GlobalVariablesSniff where local function variables where flagged because they had the same name as a global variable. While this may not be good practice/could be confusing, the sniff is wrongly flagging these uses as errors.

Additionally complex variables are not being flagged while they should be.

So I created a number of extra unit tests. Haven't figured out how to fix the sniff yet, but I thought I'd put the tests out there for whomever wants to have a go at this. Should I send in a PR for this ? The build obviously wouldn't pass as is.

The extra tests can be found here: https://github.com/jrfnl/WordPress-Coding-Standards/commit/54a9352db45c6ef40c195a30aec1ad292c3dd8f4

### LINKED ISSUE #486: GlobalVariablesSniff seems to have too broad of a scope
Reduced test case:

```
<?php
/**
 * File
 * @package  WordPressTest
 */

/**
 * [test description].
 */
function get_a_post() {
	$post = get_post();
}

/**
 * [test description].
 */
function globalize_post() {
	global $post;
}
```

This fails wpcs:

```
FILE: .../Sites/vvv/www/hbs/htdocs/wp-content/themes/hbs-ck/test.php
 11 | ERROR | Overridding WordPress globals is prohibited
```

It seems to fail if there is a `global` reference anywhere in the file, not within the function's scope, AFAICT, this should error as it's perfectly fine to use the variable `$post` as long as you're not overriding the global, which it's not in `get_a_post()`.

jrfnl:
The `GlobalVariablesSniff` contained a number of bugs which will be fixed by this PR.
* Assignments to global variables using other assignment operators than the `=` (`T_EQUAL`) operator were not detected, so `$post .= 'something';` would skate by.
* If a `T_GLOBAL`, i.e. `global ...;`, was detected, the whole file would be checked for the variables which were made global, not just the code **_after_** the `global` statement.
* If a `T_GLOBAL` was detected, the whole file would be checked for the variables which were made global, **_including_** code contained within a function/closure/class scope where there is no access to the global variable (without it being set explicitly within the function as well).
* If a `T_GLOBAL` was detected within a function call or closure, the whole file would be checked for the variables which were made global, not just the code within the function or closure.
* If a `T_GLOBAL` was detected and an assignment was made to a static class variable using the same name as one of the variables made global, an error would be thrown.
* An override of a protected global via `$GLOBALS` in combination with simple string concatenation obfuscation was not being detected. Generally speaking this shouldn't ever occur anyway and probably never would be an issue, however, with the upcoming theme review ruleset, this is something which should be detected. /cc @grappler 

Additionally, this PR contains an enhancement [as requested](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/300#issuecomment-158778606) by @westonruter :
* Don't throw errors for this sniff if the global variable override is done from within a test method.

For this enhancement a new `is_token_in_test_method()` method has been introduced into the `WordPress_Sniff` class.

PR includes extensive unit tests.
PR includes minor extraneous optimizations for this sniff.

Note: This PR does not cover complex array key obfuscation or complex variables used for obfuscating a global variable override.

Fixes #300 (except for the complex obfuscation)
Fixes #486

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -572,6 +572,47 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	}
 
 	/**
+	 * Check if a token is used within a unit test.
+	 *
+	 * Unit test methods are identified as such:
+	 * - Method name starts with `test_`.
+	 * - Method is within a class which either extends WP_UnitTestCase or PHPUnit_Framework_TestCase.
+	 *
+	 * @since 0.11.0
+	 *
+	 * @param int $stackPtr The position of the token to be examined.
+	 *
+	 * @return bool True if the token is within a unit test, false otherwise.
+	 */
+	protected function is_token_in_test_method( $stackPtr ) {
+		// Is the token inside of a function definition ?
+		$functionToken = $this->phpcsFile->getCondition( $stackPtr, T_FUNCTION );
+		if ( false === $functionToken ) {
+			return false;
+		}
+
+		// Does the function name start with 'test_' ?
+		$functionName = $this->phpcsFile->getDeclarationName( $functionToken );
+		if ( 0 !== strpos( $functionName, 'test_' ) ) {
+			return false;
+		}
+
+		// Is method inside of a class ?
+		$classToken = $this->phpcsFile->getCondition( $functionToken, T_CLASS );
+		if ( false === $classToken ) {
+			return false;
+		}
+
+		// Does the class extend either of the start with 'test_' ?
+		$extendedClassName = $this->phpcsFile->findExtendedClassName( $classToken );
+		if ( in_array( $extendedClassName, array( 'WP_UnitTestCase', 'PHPUnit_Framework_TestCase' ), true ) ) {
+			return true;
+		}
+
+		return false;
+	}
+
+	/**
 	 * Check if this variable is being assigned a value.
 	 *
 	 * E.g., $var = 'foo';
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -630,10 +630,8 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	 */
 	protected function is_assignment( $stackPtr ) {
 
-		$tokens = $this->phpcsFile->getTokens();
-
 		// Must be a variable or closing square bracket (see below).
-		if ( ! in_array( $tokens[ $stackPtr ]['code'], array( T_VARIABLE, T_CLOSE_SQUARE_BRACKET ), true ) ) {
+		if ( ! in_array( $this->tokens[ $stackPtr ]['code'], array( T_VARIABLE, T_CLOSE_SQUARE_BRACKET ), true ) ) {
 			return false;
 		}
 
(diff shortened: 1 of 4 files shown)


grappler:
This looks really good. Will try to take time to look at in detail and try and break something 😛 

I am not sure if it fits in here but it would be good to add the following as OK code that should not cause any errors as this is this is a valid overriding of a core global.

```php
function _s_content_width() {
	$GLOBALS['content_width'] = apply_filters( '_s_content_width', 640 );
}
add_action( 'after_setup_theme', '_s_content_width', 0 );
```

jrfnl:
@grappler `content_width` was already previously excluded from the list of WP global variable for exactly that reason. See #276 and #331

WordPress/Tests/Variables/GlobalVariablesUnitTest.inc
```diff
@@ -1,9 +1,105 @@
...
+class Test_Class_B extends PHPUnit_Framework_TestCase {
+
+	public function test_something() {
+		global $cat_id;
+		$cat_id = 50; // Ok.
+	}
+
+	private function arbitrary_function() {
+		global $names;
+		$names = 50; // Bad - not a test function.
```
JDGrimes: I think that you are being too literal/strict here. Any method in a testcase could modify globals, maybe it is a helper method used in `setUp()` or across multiple of the tests methods.

WordPress/Tests/Variables/GlobalVariablesUnitTest.inc
```diff
@@ -1,9 +1,105 @@
...
+class Test_Class_A extends WP_UnitTestCase {
+
+	public function test_something() {
+		global $tabs;
+		$tabs = 50; // Ok.
+	}
+
+	private function arbitrary_function() {
+		global $post_ID;
+		$post_ID = 50; // Bad - not a test function.
```
JDGrimes: Same as below (line 84).

WordPress/Tests/Variables/GlobalVariablesUnitTest.inc
jrfnl: Ok, makes sense. In that case, the base test classes themselves should be treated as whitelisted too as they may contain generic methods used by tests which override globals. Agreed ?

WordPress/Sniff.php
```diff
@@ -572,6 +572,47 @@ protected function has_whitelist_comment( $comment, $stackPtr ) {
...
+
+		// Is this a method inside of a class ?
+		$classToken = $this->phpcsFile->getCondition( $functionToken, T_CLASS );
+		if ( false === $classToken ) {
+			return false;
+		}
+
+		// Does the class extend either of whitelisted test classes ?
+		$extendedClassName = $this->phpcsFile->findExtendedClassName( $classToken );
+		if ( in_array( $extendedClassName, array( 'WP_UnitTestCase', 'PHPUnit_Framework_TestCase' ), true ) ) {
```
jrfnl: Agreed. Implementation wise however, this is not as easy as it would seem.

The thing is, as the method is within the `Sniff.php` class, IMHO the class whitelist property should live there too.

However, AFAIK, the property setting from the ruleset needs to be done for each individual sniff. So in this case that would mean that the custom whitelist property would need to be set in the ruleset for any sniff using the `is_token_in_test_method()` method.
Not much of an issue at this moment what with only one sniff using it, but by the time there are three of four sniffs using it, it will start to feel weird having to add the same property for each of them in the ruleset.

jrfnl:
DON'T MERGE YET: the changing of the sniff property in the unit test is not working. I'm trying to fix this now.
N.B.: from a ruleset it works fine.

jrfnl:
Ok, fixed it, so if anyone feels inclined to merge this, go ahead ;-)

FYI: I've redone the last commit addressing the issue @JDGrimes raised. The renewed commit now also allows for traits which I originally intended to include in the checking for test classes, but forgot. Oops... fixed!
Interfaces are not addressed as they don't contain code.

grappler:
I ran the sniff on the [comment.php](https://themes.trac.wordpress.org/browser/monday/0.1.0/components/comment.php) of a theme I was looking at which contained `$GLOBALS['comment'] = $comment;` and the sniff did not trigger an error. Should there be an error or did understand something wrong.

jrfnl:
@grappler There should be an error. How did you run the sniff ? I just tried to reproduce this (downloaded the Monday theme and ran the sniff on it + ran all WP sniffs on just that one file) and can't.

theme-monday-just-sniff

theme-monday-file-all-sniffs

grappler:
@jrfnl Then there must be an issue on my end and not with the sniff.

jrfnl:
Anyone still wants to have a look at this ? If not, I propose we merge it tomorrow.

jrfnl:
Note to whomever creates the changelog: I know a lot of people have excluded this sniff in their rulesets because of the previous buggyness - quick search on GH yields 1000+ results -, so we may want to make very clear that the rule can - and should be - re-enabled once 0.11.0 has been released.

PR #742, opened 2016-12-25
Add checking for translators comments to the I18n sniff.

### LINKED ISSUE #423: Add a translator comment sniff
All strings with placeholders should have a translator comment to explain what the placeholders are replaced with. It would be very nice to have a sniff to automatically flag strings without a proper translator comment.

These would be flagged as incorrect:

``` php

__( 'A %1$s to be translated', 'text domain' );

/* translators: %2$s is replaced with the number of translations */
__( 'A %1$s to be translated', 'text domain' );
```

And these would be fine:

``` php
__( 'A string to be translated', 'text domain' );

/* translators: %1$s is replaced with "string" */
__( 'A %1$s to be translated', 'text domain' );
```

Strings can be checked for placeholders using the following regex: `%(\d+\$(?:\d+)?)?[bcdefgosuxEFGX]`.

jrfnl:
Implemented based on the following specs:
* Only one error should be thrown per gettext call.
* A translators comment is only required when (one of) the text string(s) contains placeholders.
* It's ok for there to be a translators comment if the text string(s) does not contain placeholders.
* The translators comment must be directly above the gettext call.
* If the comment is not on the *line* directly above the gettext call, there should only be whitespace lines between the comment and the line containing the gettext call.
* The translators comment must be in `//` or `/* */` format, docblocks are not allowed.
* A translators comment must start with `translators:`
* The check defaults to throwing a warning. If the `extra` ruleset is enabled, the error level will change to `error`.

Fixes #423 

I've added a sniff property `$check_translator_comments` to enable us to unit test the other checks within the I18n sniff without getting bogged down by errors thrown by the translators comment check.
This is a public property, but is only intended to be used in combination with the unit tests.
It *could* be used in custom rulesets too, but this shouldn't be needed as the translator comment checks can be excluded via their error codes.

#### Regarding the travis run failure:
The unit tests against PHPCS 2.7.0 are failing because the ruleset is erroneously taken into account when running the unit tests, which leads to PHPCS throwing an `error` instead of the expected `warning`.

This was fixed in PHPCS 2.7.1.
This *only* affects the unit tests however, not the functioning of the sniff itself.

Related: #733, #696 and https://github.com/squizlabs/PHP_CodeSniffer/issues/1177

### DIFF
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -515,7 +515,7 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 	 * @return bool
 	 */
 	private function is_translators_comment( $content ) {
-		if ( preg_match( '`^(?:(?://|/\*{1,2}) )?translators:`', $content, $matches ) === 1 ) {
+		if ( preg_match( '`^(?:(?://|/\*{1,2}) )?translators:`i', $content, $matches ) === 1 ) {
 			return true;
 		}
 		return false;
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -16,6 +16,7 @@
  * @package WPCS\WordPressCodingStandards
  *
  * @since   0.10.0
+ * @since   0.11.0 Now also checks for translators comments.
  */
 class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 
@@ -71,6 +72,16 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 		'_n_noop'                        => 'noopnumber',
 		'_nx_noop'                       => 'noopnumber_context',
 	);
+
+	/**
+	 * Toggle whether or not to check for translators comments for text string containing placeholders.
+	 *
+	 * Intended to make this part of the sniff unit testable, but can be used by end-users too,
+	 * though they can just as easily disable this via the sniff code.
+	 *
+	 * @var bool
+	 */
+	public $check_translator_comments = true;
 
 	/**
 	 * Returns an array of tokens this test wants to listen for.
@@ -213,6 +224,10 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 			$plural_context = $argument_assertions[1];
 
 			$this->compare_single_and_plural_arguments( $phpcs_file, $stack_ptr, $single_context, $plural_context );
+		}
+
+		if ( true === $this->check_translator_comments ) {
+			$this->check_for_translator_comment( $phpcs_file, $stack_ptr, $argument_assertions );
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -392,4 +407,118 @@ protected function check_text( PHP_CodeSniffer_File $phpcs_file, $context ) {
...
+					/*
+					 * Check that the comment starts with 'translators:'.
+					 */
+					if ( true === $correctly_placed ) {
+
+						if ( T_COMMENT === $tokens[ $previous_comment ]['code'] ) {
+							$comment_text = trim( $tokens[ $previous_comment ]['content'] );
+
+			  		   		// If it's multi-line /* */ comment, collect all the parts.
+			  		   		if ( '*/' === substr( $comment_text, -2 ) && '/*' !== substr( $comment_text, 0, 2 ) ) {
```
GaryJones: Could this "collect all parts from a multi-line comment" be abstracted into a re-usable method in the base class?

How about :

```
/*
 * This is a multiline comment,
 * But it also has * at the start
 of some lines ;-)
*/
```

WordPress/Tests/WP/I18nUnitTest.1.inc
```diff
@@ -0,0 +1,77 @@
...
+_n_noop( 'I have %d cat.', "I have %d cats.", 'my-slug' ); // OK.
+
+// translators: 1: number; 2: string.
+// Some other comment.
+esc_attr_e( 'Text to translate to %1$d languages. Another %2$s placeholder', 'my-slug' ); // Bad - translators comment has to be the first comment before the function call.
+
+printf(
+	/* translators: number of monkeys, location. */
+	__( 'There are %1$d monkeys in the %2$s', 'my-slug' ),
+	intval( $number ),
```
GaryJones: `intval` tends to be PHP 4 - `(int)` would be better.

GaryJones:
Code generally looks good - great work.

Pinging @ramiy, as I think he does a lot with translator comments, and can offer feedback on any missed cases / false positives.

WordPress/Sniffs/WP/I18nSniff.php
jrfnl: > Could this "collect all parts from a multi-line comment" be abstracted into a re-usable method in the base class?

It could, but I've not seen any other sniff within the WPCS which would need it. I suggest this can be done later if/when another sniff would need something similar.
Also - the snippet *only* handles multi-line /* */ type comments, no other type and (currently) does not care about how the lines are added together as for this sniff, only the start of the text is interesting, so if we would want it to be more generic, it will need some tweaking.

> how does it handle something like <code snippet>

It handles that fine, just tested it to be sure.

WordPress/Tests/WP/I18nUnitTest.1.inc
jrfnl: Fixed

jrfnl:
Rebased after #733 to allow unit tests to pass.

ramiy:
Looks good!!!

As for "**The translators comment must be directly above the gettext call.**",

And the examples on the testing file:

```
printf(
	/* translators: number of monkeys, location. */
	__( 'There are %1$d monkeys in the %2$s', 'my-slug' ),
	(int) $number,
	esc_html( $string )
); // Ok.

/* translators: number of monkeys, location. */
printf(
	__( 'There are %1$d monkeys in the %2$s', 'my-slug' ),
	(int) $number,
	esc_html( $string )
); // Bad - comment not directly before line containing the gettext call.
```

Although I prefer to add comments above the gettext call, WordPress core has tons of translators comments above the `printf()` and `sprintf()` functions. We will need to create a new ticket to move all those comments. Pinging @SergeyBiryukov and @ocean90.

jrfnl:
Hi @ramiy Thanks for looking this over!

> As for: The translators comment must be directly above the gettext call.

I've set that quite strict as in my personal experience with PoEdit, the comments don't get picked up otherwise. Not sure how well GlotPress/makepot does in picking up those comments if they are not on the line directly above (I expect just as badly as PoEdit uses makepot underneath the hood).

If it's on one line, it should be fine, see the last unit test:
```php
/* translators: number of monkeys, location. */
printf( __( 'There are %1$d monkeys in the %2$s', 'my-slug' ), intval( $number ), esc_html( $string ) ); // Ok - comment is directly before line containing the gettext call.
```

And if the gettext call is on the same line as the `(s)printf()`, but the parameters aren't, it should be fine as well:
```php
/* translators: number of monkeys, location. */
printf( __(
	'There are %1$d monkeys in the %2$s', 'my-slug' ),
	intval( $number ),
	esc_html( $string )
); // Ok - comment is directly before line containing the gettext call.
```

jrfnl:
I've just read through a number of trac tickets around this, starting from https://core.trac.wordpress.org/ticket/39116 and am left with one question: should `translators:` always be lowercase of should `Translators:` be allowed too ? (currently it's not)
IIRC PoEdit only accepts one or the other (it is a case-sensitive command line option you provide), not sure about other makepot tools.

grappler:
> as PoEdit uses makepot underneath the hood
I don't think so because PoEdit had to add this functionality.

> Poedit did support the "TRANSLATORS:" tag mentioned in gettext's own manual, but not the lowercase variant.

I found this in one of my emails with the PoEdit Developer.

ntwb:
> ... am left with one question: should translators: always be lowercase

Yes

_(I can't find the Trac ticket citing the above, when I do I'll update this comment)_

ntwb:
Via https://core.trac.wordpress.org/ticket/30603

Translator comments must be on a new line, and the line begin with either `/*` or `//`

GaryJones:
Comments with lowercase `t` may trigger other warnings about starting a comment with a capital, unless we have exclusions for them in place - I think we may even have that check turned off in a ruleset, because those exclusions don't exist yet.

I'd love to just allow capitalised `Translators`, but clearly prior art (namely, anyone who's ever added in these comments), means the lowercase versions needs to be supported too.

ocean90:
> > As for: The translators comment must be directly above the gettext call.
>
> I've set that quite strict as in my personal experience with PoEdit, the comments don't get picked up otherwise. Not sure how well GlotPress/makepot does in picking up those comments if they are not on the line directly above (I expect just as badly as PoEdit uses makepot underneath the hood).

Just a general FYI: GlotPress doesn't generate POT files from source files, that's why we have makepot.

makepot doesn't care where a translator comment is. If there is a comment it stores it and adds the comment to the next gettext call, no matter how many lines there are between the comment and the actual gettext call.

----

> We will need to create a new ticket to move all those comments. 

There is no need for that, see also https://make.wordpress.org/core/handbook/contribute/code-refactoring/.

jrfnl:
@grappler 

> I don't think so because PoEdit had to add this functionality.

It did for the automatic WP project setup which it now offers out of the box, but before that they already supported it *as long you configured it right*:
Preferences -> Extractors -> PHP ->"Command to extra translations' - if you added _"--add-comments=translators:"_ to this, it worked, even in (very) old versions of PoEdit.
That is the bit which I know only works in a case-sensitive way.

Oh and it also corrects me regarding what they use under the hood - not `makepot`, but `xgettext`.

jrfnl:
Based on the feedback I've made the checking for the `translators:` part case-insensitive and added unit tests to that effect.
I've also added the additional case from https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/742#issuecomment-269724979 to the unit tests (second example).

> makepot doesn't care where a translator comment is. If there is a comment it stores it and adds the comment to the next gettext call, no matter how many lines there are between the comment and the actual gettext call.

Based on this: should we loosen up the check for a translators comment on the line directly above the gettext call ?
I'd advocate against it for the following reasons:
* We know that if it's on the line directly above it will get picked up correctly by different tools, this might not be consistently so for all tools if the comment has other code between it and the gettext call.
* Similar to the _late escaping_ philosophy, having the translators comment tightly coupled with the actual call makes it easier for other devs to see that there is one in place.

Last question as asked before in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/423#issuecomment-269061710:
> Oh and input on whether or not some form of whitelisting for a gettext call without translators comment is needed and what form it should take would be welcome too !

I can imagine we should have a whitelist comment as - if I remember correctly - if the same phrase is used several times, it only needs a translators comment once.

Opinions ?

jrfnl:
@vslavik Thank you for your input & clarification! Very useful.

jrfnl:
Anyone still wants to have a look at this ? If not, I propose we merge it tomorrow.

PR #758, opened 2016-12-29
Add get_the_ID() & remove post_class() from the safe list

### LINKED ISSUE #746: Suggestions for addition to 'safe' function list / XSS sniff
I would suggest adding the following function to the safe list:
`get_the_id()` as `the_ID()` is already on the list and `get_the_id()` only echos `the_ID()` without any escaping
`get_post_class()` as `post_class()` is already on the list and `get_post_class()` echos the result from `post_class()` without escaping it.

grappler:
Fixes #746


jrfnl:
Related:
https://core.trac.wordpress.org/ticket/29259
https://core.trac.wordpress.org/ticket/20009 (discussed and closed without fixing it for fear of breaking funky plugin implementations)

JDGrimes:
I second the concerns that @jrfnl has raised in regard to `get_post_class()`. Of course, even `esc_html()` *et al.* apply filters *after* escaping the value. But in that case, anything hooked to the filter *knows* that the value is already escaped. It is entirely possible that folks could hook into the post class filter and not realize that escaping is needed on their part.

grappler:
I have update the PR and changed the title to match the code.

PR #759, opened 2016-12-30
Add file_get_contents() & file_get_contents()

grappler:
As I was looking at https://github.com/WPTRT/WordPress-Coding-Standards/pull/79 I realized I missed two functions.

This is also part of the Theme Check plugin and the VIP Scanner https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#filesystem-writes
https://github.com/WordPress/theme-check/blob/master/checks/malware.php#L9
https://github.com/Automattic/vip-scanner/blob/master/vip-scanner/checks/VIPRestrictedCommandsCheck.php#L143

Ref: #633

### DIFF
--- a/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
+++ b/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
@@ -74,6 +74,8 @@ class WordPress_Sniffs_WP_AlternativeFunctionsSniff extends WordPress_AbstractFu
 					'fclose',
 					'fread',
 					'fwrite',
+					'file_put_contents',
+					'file_get_contents',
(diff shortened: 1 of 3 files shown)


jrfnl:
You didn't miss `file_get_contents()` - it was already covered in the same file, in the group above it...

JDGrimes:
@jrfnl Yes, and I saw that as well, but it is for a different reason (fetching URLs). So it needs to be in both lists.

jrfnl:
> So it needs to be in both lists.

Good point. In that case the unit tests needs to be adjusted. `file_get_contents()` only needs to be in there once, but be set to expect two messages on that line.

PR #768, opened 2017-01-03
Implement ternary related extra rules

jrfnl:
> Clever Code
>
> In general, readability is more important than cleverness or brevity.
> ```php
> isset( $var ) || $var = some_function();
> ```
> Although the above line is clever, it takes a while to grok if you’re not familiar with it. So, just write it like this:
> ```php
> if ( ! isset( $var ) ) {
>     $var = some_function();
> }

See: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#clever-code

This part is so far not really covered in WPCS. Existing sniffs which come close - but still don't cover this completely - relate to inline if statements and comparison assignments.

This PR is intended to have a discussion on whether or not to add these sniffs to the rulesets.

### DIFF
--- a/WordPress/AbstractArrayAssignmentRestrictionsSniff.php
+++ b/WordPress/AbstractArrayAssignmentRestrictionsSniff.php
@@ -144,7 +144,10 @@ abstract class WordPress_AbstractArrayAssignmentRestrictionsSniff extends WordPr
 				continue;
 			}
 
-			$callback = ( isset( $group['callback'] ) && is_callable( $group['callback'] ) ) ? $group['callback'] : array( $this, 'callback' );
+			$callback = array( $this, 'callback' );
+			if ( isset( $group['callback'] ) && is_callable( $group['callback'] ) ) {
+				$callback = $group['callback'];
+			}
 
 			foreach ( $inst as $key => $assignments ) {
 				foreach ( $assignments as $occurance ) {
--- a/WordPress/AbstractVariableRestrictionsSniff.php
+++ b/WordPress/AbstractVariableRestrictionsSniff.php
@@ -139,7 +139,11 @@ abstract class WordPress_AbstractVariableRestrictionsSniff implements PHP_CodeSn
 
 			$patterns = array_map( array( $this, 'test_patterns' ), $patterns );
 			$pattern  = implode( '|', $patterns );
-			$delim    = ( T_OPEN_SQUARE_BRACKET !== $token['code'] ) ? '\b' : '';
+
+			$delim = '';
+			if ( T_OPEN_SQUARE_BRACKET !== $token['code'] ) {
+				$delim = '\b';
+			}
 
 			if ( T_DOUBLE_QUOTED_STRING === $token['code'] ) {
 				$var = $token['content'];
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -754,7 +754,10 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 		// If this is inside an isset(), check after it as well, all the way to the
 		// end of the scope.
 		if ( $in_isset ) {
-			$end = ( 0 === $start ) ? count( $tokens ) : $tokens[ $start ]['scope_closer'];
+			$end = $tokens[ $start ]['scope_closer'];
+			if ( 0 === $start ) {
+				$end = count( $tokens );
+			}
 		}
 
 		// Check if we've looked here before.
--- a/WordPress/Sniffs/Arrays/ArrayKeySpacingRestrictionsSniff.php
+++ b/WordPress/Sniffs/Arrays/ArrayKeySpacingRestrictionsSniff.php
@@ -56,8 +56,15 @@ class WordPress_Sniffs_Arrays_ArrayKeySpacingRestrictionsSniff implements PHP_Co
 			true
 		);
 
-		$spaced1 = ( T_WHITESPACE === $tokens[ ( $stackPtr + 1 ) ]['code'] );
-		$spaced2 = ( T_WHITESPACE === $tokens[ ( $token['bracket_closer'] - 1 ) ]['code'] );
+		$spaced1 = false;
+		if ( T_WHITESPACE === $tokens[ ( $stackPtr + 1 ) ]['code'] ) {
+			$spaced1 = true;
+		}
+
+		$spaced2 = false;
+		if ( T_WHITESPACE === $tokens[ ( $token['bracket_closer'] - 1 ) ]['code'] ) {
+			$spaced2 = true;
+		}
 
 		// It should have spaces only if it only has strings or numbers as the key.
 		if ( false !== $need_spaces && ! ( $spaced1 && $spaced2 ) ) {
--- a/WordPress/Sniffs/CSRF/NonceVerificationSniff.php
+++ b/WordPress/Sniffs/CSRF/NonceVerificationSniff.php
@@ -119,7 +119,10 @@ class WordPress_Sniffs_CSRF_NonceVerificationSniff extends WordPress_Sniff {
 		}
 
 		// If we're still here, no nonce-verification function was found.
-		$severity = ( in_array( $instance['content'], $this->errorForSuperGlobals, true ) ) ? 0 : 'warning';
+		$severity = 'warning';
+		if ( in_array( $instance['content'], $this->errorForSuperGlobals, true ) ) {
+			$severity = 0;
+		}
 
 		$phpcsFile->addError(
 			'Processing form data without nonce verification.'
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -242,8 +242,12 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 		$stack_ptr = $context['stack_ptr'];
 		$tokens    = $context['tokens'];
 		$arg_name  = $context['arg_name'];
-		$method    = empty( $context['warning'] ) ? 'addError' : 'addWarning';
 		$content   = $tokens[0]['content'];
+
+		$method    = 'addWarning';
+		if ( empty( $context['warning'] ) ) {
+			$method = 'addError';
+		}
 
 		if ( empty( $tokens ) || 0 === count( $tokens ) ) {
 			$code = 'MissingArg' . ucfirst( $arg_name );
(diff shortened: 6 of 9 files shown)


JDGrimes:
This really has nothing to do with ternaries. Actually, it is about using the `||` and `&&` operators *without* ternaries or an `if` statement. If you use a ternary or an `if` statement, then you are OK.

grappler:
I think `Squiz.PHP.DisallowInlineIf` is covered best by:

> Note that requiring the use of braces just means that single-statement inline control structures are prohibited.

This is documented here https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style

jrfnl:
@grappler _"single-statement inline control structures"_ are if/for/foreach/switch etc statements with only one "child" statement. PHP doesn't _require_ braces in that case, but it is best practice to use braces anyhow:
```php
// Basically, don't do this:
if ( false === $var ) $var++;

// Or this:
if ( false === $var )
	$var++;
else
	$var--;
```

GaryJones:
I'm definitely in favour of adding `DisallowInlineIf` to Core ruleset - the Handbook explicitly [covers](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style) that:

> Braces should always be used, even when they are not required

----

My concern was with the other bit - why something like:

```php
$method    = empty( $context['warning'] ) ? 'addError' : 'addWarning';
```

in our sniff needs changing to:

```php
$fixable_method = 'addFixableWarning';
if ( empty( $context['warning'] ) ) {
 	$fixable_method = 'addFixableError';
}
```

to comply with the `DisallowComparisonAssignment` in the Extra ruleset - this is the bit that suggests that no ternary conditions would be allowed.



> If you use a ternary or an if statement, then you are OK.

If @JDGrimes is right, why has some WPCS code been updated to comply with the new (potential) ruleset?

jrfnl:
Ok, looks like there's some confusion over the issue here.

One the one hand we have the handbook which disallows the use of control structures without braces.
That rule is actually already covered by the `Squiz.ControlStructures.ControlSignature` sniff and is not under discussion.

The current rule suggestions are about _Clever code_ - code which reduces readability/ease of comprehension.

The `Squiz.PHP.DisallowInlineIf` sniff is the one which is most often triggered in our code base and necessitated the code changes. That one effectively disallows ternaries in combination with assignments.
See: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Tests/PHP/DisallowInlineIfUnitTest.inc - line 8 triggers this sniff

The `Squiz.PHP.DisallowComparisonAssignment` sniff is the one which comes closer to matching the example code in the _Clever code_ section, even though it still doesn't match it. It also, like above, gets triggered by ternaries.
See: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Squiz/Tests/PHP/DisallowComparisonAssignmentUnitTest.inc - line 3, 5, 6, 7, 8, 10, 52, 53 trigger the sniff

To make it clearer which rule necessitated which change in the current codebase, I'll annotate the changes in the diff so it's clearer what we're talking about.

WordPress/AbstractArrayAssignmentRestrictionsSniff.php
```diff
@@ -144,7 +144,10 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 				continue;
 			}
 
-			$callback = ( isset( $group['callback'] ) && is_callable( $group['callback'] ) ) ? $group['callback'] : array( $this, 'callback' );
+			$callback = array( $this, 'callback' );
+			if ( isset( $group['callback'] ) && is_callable( $group['callback'] ) ) {
+				$callback = $group['callback'];
+			}
```
jrfnl: Squiz.PHP.DisallowInlineIf.Found

WordPress/AbstractVariableRestrictionsSniff.php
```diff
@@ -139,7 +139,11 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 
 			$patterns = array_map( array( $this, 'test_patterns' ), $patterns );
 			$pattern  = implode( '|', $patterns );
-			$delim    = ( T_OPEN_SQUARE_BRACKET !== $token['code'] ) ? '\b' : '';
+
+			$delim = '';
+			if ( T_OPEN_SQUARE_BRACKET !== $token['code'] ) {
+				$delim = '\b';
+			}
```
jrfnl: Triggers both sniffs

WordPress/Sniff.php
```diff
@@ -668,7 +668,10 @@ protected function has_nonce_check( $stackPtr ) {
 		// If this is inside an isset(), check after it as well, all the way to the
 		// end of the scope.
 		if ( $in_isset ) {
-			$end = ( 0 === $start ) ? count( $tokens ) : $tokens[ $start ]['scope_closer'];
+			$end = $tokens[ $start ]['scope_closer'];
+			if ( 0 === $start ) {
+				$end = count( $tokens );
+			}
```
jrfnl: Triggers both sniffs

WordPress/Sniffs/Arrays/ArrayKeySpacingRestrictionsSniff.php
```diff
@@ -56,8 +56,15 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
...
-		$spaced2 = ( T_WHITESPACE === $tokens[ ( $token['bracket_closer'] - 1 ) ]['code'] );
+		$spaced1 = false;
+		if ( T_WHITESPACE === $tokens[ ( $stackPtr + 1 ) ]['code'] ) {
+			$spaced1 = true;
+		}
+
+		$spaced2 = false;
+		if ( T_WHITESPACE === $tokens[ ( $token['bracket_closer'] - 1 ) ]['code'] ) {
+			$spaced2 = true;
+		}
```
jrfnl: Squiz.PHP.DisallowComparisonAssignment

WordPress/Sniffs/CSRF/NonceVerificationSniff.php
```diff
@@ -119,7 +119,10 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 		}
 
 		// If we're still here, no nonce-verification function was found.
-		$severity = ( in_array( $instance['content'], $this->errorForSuperGlobals, true ) ) ? 0 : 'warning';
+		$severity = 'warning';
+		if ( in_array( $instance['content'], $this->errorForSuperGlobals, true ) ) {
+			$severity = 0;
+		}
```
jrfnl: Triggers both sniffs

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -227,9 +227,13 @@ protected function check_argument_tokens( PHP_CodeSniffer_File $phpcs_file, $con
...
 		$tokens    = $context['tokens'];
 		$arg_name  = $context['arg_name'];
-		$method    = empty( $context['warning'] ) ? 'addError' : 'addWarning';
 		$content   = $tokens[0]['content'];
 
+		$method    = 'addWarning';
+		if ( empty( $context['warning'] ) ) {
+			$method = 'addError';
+		}
+
```
jrfnl: Triggers both sniffs

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -328,7 +332,11 @@ protected function check_text( PHP_CodeSniffer_File $phpcs_file, $context ) {
 		$stack_ptr      = $context['stack_ptr'];
 		$arg_name       = $context['arg_name'];
 		$content        = $context['tokens'][0]['content'];
-		$fixable_method = empty( $context['warning'] ) ? 'addFixableError' : 'addFixableWarning';
+
+		$fixable_method = 'addFixableWarning';
+		if ( empty( $context['warning'] ) ) {
+			$fixable_method = 'addFixableError';
+		}
```
jrfnl: Triggers both sniffs

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -351,8 +359,13 @@ protected function check_text( PHP_CodeSniffer_File $phpcs_file, $context ) {
...
 			$replacements    = array();
 			for ( $i = 0; $i < $unordered_matches_count; $i++ ) {
-				$to_insert         = ( $i + 1 );
-				$to_insert        .= ( '"' !== $content[0] ) ? '$' : '\$';
+				$to_insert = ( $i + 1 );
+				if ( '"' !== $content[0] ) {
+					$to_insert .= '$';
+				} else {
+					$to_insert .= '\$';
+				}
```
jrfnl: Squiz.PHP.DisallowInlineIf

WordPress/Tests/VIP/RestrictedFunctionsUnitTest.php
```diff
@@ -24,7 +24,7 @@ public function getErrorList() {
 		return array(
 			3  => 1,
 			17 => 1,
-			30 => ( PHP_VERSION_ID >= 50300 ) ? 0 : 1,
+			30 => (int) ( PHP_VERSION_ID < 50300 ),
```
jrfnl: Squiz.PHP.DisallowInlineIf

GaryJones:
Thanks for the explanation @jrfnl.

I think forbidding the use of `$var = ! $foo;`, and `$a = $b ? $c : $d;` is going to unnecessarily frustrate devs to the point that they'd want to turn this particular Extras check off. I wouldn't consider them _clever code_ or unreadable code.

JDGrimes:
> This really has nothing to do with ternaries. Actually, it is about using the || and && operators without ternaries or an if statement. If you use a ternary or an if statement, then you are OK.

This comment was in regard to the clever code section in the handbook, and I realize now that that may not have been clear.

The clever code section in the handbook has nothing to do with ternaries or assignments. It is about using `&&` and `||` *to directly chain expressions*, like function calls. It is not about using them in conditions, it is about using them *sans conditional structures* (like `if` statements and ternaries). At least this is the way that I have always taken it.

So a rule which would flag these would be something like: don't allow `||`, `&&`, etc., to be used outside of conditional structures.

WordPress/Tests/VIP/RestrictedFunctionsUnitTest.php
JDGrimes: IMHO, this is more clever than before. 😄

jrfnl:
Closing for now for lack of consensus. Leaving the branch available in case anyone wants to revisit this at some point.

PR #769, opened 2017-01-04
Add screenshot image of unit testing in terminal

GaryJones:
This links through to https://asciinema.org/a/98078 which is a replay of the unit tests being run.

If we ended up with a docs/ site on GitHub, we could embed the asciinema player :-)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,17 +23,7 @@ vendor/bin/phpunit --filter WordPress tests/AllTests.php
 
 Expected output:
 
-~~~sh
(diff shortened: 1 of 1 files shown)


jrfnl:
I like the idea though it does make presumptions on how PHPCS is installed (which is ok as long as we make those presumptions clear).

GaryJones:
> it does make presumptions on how PHPCS is installed

There's little different to the static code snippet which was already there.

jrfnl:
@GaryJones True, the remark was more about the section in general.

PR #770, opened 2017-01-05
Fix some undefined index notices in the `WordPress.XSS.EscapeOutput` sniff.

jrfnl:
These would be thrown when live coding/running the sniffs on code with parse errors.

Quick fix for https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/248#issuecomment-270635692

/cc @grappler

### DIFF
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -208,6 +208,11 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 
 			if ( T_OPEN_PARENTHESIS === $this->tokens[ $i ]['code'] ) {
 
+				if ( ! isset( $this->tokens[ $i ]['parenthesis_closer'] ) ) {
+					// Live coding or parse error.
+					break;
+				}
+
 				if ( $in_cast ) {
 
 					// Skip to the end of a function call if it has been casted to a safe value.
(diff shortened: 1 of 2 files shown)


WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -208,6 +208,11 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 
 			if ( T_OPEN_PARENTHESIS === $this->tokens[ $i ]['code'] ) {
 
+				if ( ! isset( $this->tokens[ $i ]['parenthesis_closer'] ) ) {
+					// Live coding or parse error.
+					break;
```
JDGrimes: Usually when PHPCS's core sniffs detect a parse error, they don't just skip out, they add a warning. At least they do in some cases. Should we consider doing the same here?

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -319,7 +324,12 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 						$i     = ( $function_opener + 1 );
 						$watch = true;
 					} else {
-						$i = $this->tokens[ $function_opener ]['parenthesis_closer'];
+						if ( isset( $this->tokens[ $function_opener ]['parenthesis_closer'] ) ) {
+							$i = $this->tokens[ $function_opener ]['parenthesis_closer'];
+						} else {
+							// Live coding or parse error.
+							break;
```
JDGrimes: Same as above.

WordPress/Sniffs/XSS/EscapeOutputSniff.php
jrfnl: IMHO that's what the `Generic.PHP.Syntax` sniff is for (included in `extra`)

JDGrimes: Yep, that makes sense. So I guess for consistency we should change [the one place](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/51eb33b15f3aad5d74e1c816b5d8edc42163de25/WordPress/Sniffs/PHP/StrictInArraySniff.php#L103) we give an error for syntax issues. But that can be a separate PR, I guess.

PR #772, opened 2017-01-05
Don't error for syntax errors in StrictInArray sniff

JDGrimes:
This should be left up to the `Generic.PHP.Syntax` sniff.

See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/770#discussion_r94809371


grappler:
@JDGrimes Did you see that you need to update the tests also?

JDGrimes:
@grappler Yep, fixed now.

PR #773, opened 2017-01-06
Add a check that associative arrays are multi-line - includes fixer

### LINKED ISSUE #638: "For associative arrays, values should start on a new line."
Ref: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation

This is currently not covered, but could be added to the `ArrayDeclaractionSniff` within the `processSingleLineArray()` method.

jrfnl:
> For associative arrays, values should start on a new line.
https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation

Detecting this was easy, creating the fixer.... not so much...
I got it working, but would not be surprised if I rewrite the fixer in a few months or so. I've tried to anticipate the most common situations, but the fixer will fail on arrays with very inconsistent spacing.
Think:
```php
array(
	'key' => array( 'key1' => 'value', 'key2' => array(
		'another_value'
	) ),
)
```

Anticipating a comment on line 43 of the `.fixed` file: the fixer will add a trailing comma to arrays which it is triggered on (if needed), but will *not* do so on arrays on which it is not triggered - that is covered by another sniff.

Fixes #638

* Allows for both long as well as short array syntax.
* Includes unit tests.
* Includes fixing the one file in the WPCS code base which did not comply with the rule.

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -33,13 +33,12 @@
 		<arg name="tab-width" value="4"/>
 		<rule ref="Generic.WhiteSpace.DisallowSpaceIndent"/>
 
-		<!-- Rule: For associative arrays, values should start on a new line.
-			 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/638 -->
-
 		<!-- Covers rule: Note the comma after the last array item: this is recommended. -->
 		<rule ref="WordPress.Arrays.ArrayDeclaration">
 			<exclude name="WordPress.Arrays.ArrayDeclaration.SingleLineNotAllowed" />
 		</rule>
+		<!-- Covers rule: For associative arrays, values should start on a new line.
+			 Also covers various single-line array whitespace issues. -->
 		<rule ref="WordPress.Arrays.ArrayDeclarationSpacing">
 			<!-- Exclude the upstream checks which are already thrown by the
 			     WordPress.Arrays.ArrayDeclaration sniff. -->
--- a/WordPress/Tests/Arrays/ArrayDeclarationSpacingUnitTest.inc
+++ b/WordPress/Tests/Arrays/ArrayDeclarationSpacingUnitTest.inc
@@ -2,7 +2,23 @@
 
 $good = array( 'value1', 'value2' ); // Ok.
 
-$query_vars = array('post_type' => 'food'); // Bad, no spaces after opening and before closing parenthesis.
+$query_vars = array('food'); // Bad, no spaces after opening and before closing parenthesis.
 
 // Test for fixing of extra whitespace.
 $test = array(   1, 2     );
+
+$bad = array( 'key' => 'value' ); // Bad, each value of an associative array should start on a new line.
+
+// Test for fixing nested associative arrays.
+$bad = array( array( 'key1' => 'value1', 'key2' => ['sub1' => 1, 'sub2' => 2] ), $key3 => 'value3', [ 'value4', 10 => 'value5', ] ); // Bad.
+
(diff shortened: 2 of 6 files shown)


WordPress/Tests/Arrays/ArrayDeclarationSpacingUnitTest.inc
```diff
@@ -2,7 +2,23 @@
 
 $good = array( 'value1', 'value2' ); // Ok.
 
-$query_vars = array('post_type' => 'food'); // Bad, no spaces after opening and before closing parenthesis.
+$query_vars = array('food'); // Bad, no spaces after opening and before closing parenthesis.
 
 // Test for fixing of extra whitespace.
 $test = array(   1, 2     );
+
+$bad = array( 'key' => 'value' ); // Bad, each value of an associative array should start on a new line.
```
JDGrimes: I have always thought that when there are *multiple* values in the array, it should be split to have one value per line. When there is just one key => value pair, I don't think that it has to be multiline (although that should still be permissible, in case of long values that need to be wrapped). I do see that the docs actually say:

>For associative arrays, values should start on a new line.

Though I'm a bit incredulous that that means for arrays with a single element.

WordPress-Core/ruleset.xml
```diff
@@ -33,13 +33,12 @@
...
 
-		<!-- Rule: For associative arrays, values should start on a new line.
-			 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/638 -->
-
 		<!-- Covers rule: Note the comma after the last array item: this is recommended. -->
 		<rule ref="WordPress.Arrays.ArrayDeclaration">
 			<exclude name="WordPress.Arrays.ArrayDeclaration.SingleLineNotAllowed" />
 		</rule>
+		<!-- Covers rule: For associative arrays, values should start on a new line.
+			 Also covers various array whitespace issues) -->
```
jrfnl: Fixed

WordPress/Tests/Arrays/ArrayDeclarationSpacingUnitTest.inc
jrfnl: The handbook doesn't say anywhere that multi-value arrays should be multi-line (though I totally agree that that would be good practice).
It only has the bit we both quoted and that bit doesn't specify that it doesn't apply to single-value associative arrays...

JDGrimes: Since this is what the handbook says, I guess lets go with it for now. If I feel the need in the future, I'll open a PR adding the option to ignore single-element arrays.

WordPress/Tests/Arrays/ArrayDeclarationSpacingUnitTest.inc.fixed
```diff
@@ -2,7 +2,53 @@
 
 $good = array( 'value1', 'value2' ); // Ok.
 
-$query_vars = array( 'post_type' => 'food' ); // Bad, no spaces after opening and before closing parenthesis.
+$query_vars = array( 'food' ); // Bad, no spaces after opening and before closing parenthesis.
```
GaryJones: Why is this bad?

JDGrimes: It is fine here, as this is the fixed version of the file. It was bad before, in the unfixed version.

And as to the reason for making this a non-associative array, that is because it could no longer be on a single line then, as associative arrays have to be multiline. So to check spacing, it had to be changed to a non-associative array now.

Yes, both of these initially confused me as well. 😄

PR #777, opened 2017-01-07
I18n Sniff: Only recognize the space as a padding character when accompanied by a width specifier.

### LINKED ISSUE #776: Don't require a translators comment for "foo 100% bar"
```bash
$ echo "<?php __( 'foo 100% bar', 'domain' );" | phpcs --standard=wordpress-core

FILE: STDIN
 1 | WARNING | A gettext call containing placeholders was found, but
   |         | was not accompanied by a "translators:" comment on the
   |         | line above to clarify the meaning of the placeholders.

Time: 62ms; Memory: 4Mb
```

Relates to #742.

jrfnl:
This PR implements the third option mentioned in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/776#issuecomment-271109986

> ... only recognize the optional padding character if there is also a width specifier in the placeholder. This could potentially mean that some valid placeholders would no longer be recognized as such.

Fixes #776

* Includes additional unit test.
* Includes detailed documentation of the regular expressions as they were getting quite hard to manage without.

### DIFF
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -22,13 +22,49 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 
 	/**
 	 * These Regexes copied from http://php.net/manual/en/function.sprintf.php#93552
+	 * and adjusted for better precision and updated specs.
 	 */
-	const SPRINTF_PLACEHOLDER_REGEX = '/(?:(?<!%)(%(?:[0-9]+\$)?[+-]?(?:[ 0]|\'.)?-?[0-9]*(?:\.[0-9]+)?[bcdeEufFgGosxX]))/';
+	const SPRINTF_PLACEHOLDER_REGEX = '/(?:
+		(?<!%)                     # Don\'t match a literal % (%%).
+		(
+			%                          # Start of placeholder.
+			(?:[0-9]+\$)?              # Optional ordering of the placeholders.
+			[+-]?                      # Optional sign specifier.
+			(?:
+				(?:0|\'.)?                 # Optional padding specifier - excluding the space.
+				-?                         # Optional alignment specifier.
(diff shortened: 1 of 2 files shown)


WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -22,13 +22,49 @@ class WordPress_Sniffs_WP_I18nSniff extends WordPress_Sniff {
 
 	/**
 	 * These Regexes copied from http://php.net/manual/en/function.sprintf.php#93552
+	 * and adjusted for better precision and updated specs.
 	 */
-	const SPRINTF_PLACEHOLDER_REGEX = '/(?:(?<!%)(%(?:[0-9]+\$)?[+-]?(?:[ 0]|\'.)?-?[0-9]*(?:\.[0-9]+)?[bcdeEufFgGosxX]))/';
+	const SPRINTF_PLACEHOLDER_REGEX = '/(?:
```
GaryJones: Is there any change to this regex, beyond breaking it up into multipe lines?

jrfnl: Yes, there is (in both regexes).

> only recognize the optional padding character if there is also a width specifier in the placeholder

The change is that the space as a padding character is now only recognized as such if accompanied by a width specifier. Other padding characters are still recognized as normal even when there is no width specifier.

In detail:
Line 34 - The space as a padding character has been removed.
Line 38 - 42 - this is the new part where the space is now only recognized as a padding character if there is also a width specifier
Line 37 + 42: the specs mention an optional padding character which can be added to the precision specifier, this was previously unaccounted for. Was `(?:\.[0-9]+)?`, changed to account for the padding option here.

jrfnl: Oh and just to be clear - the first two changes in the regex is what allows the new unit test / test case by @ocean90 to pass correctly.

Issue #780, opened 2017-01-12
Recognize sanitize_hex_color as a sanitization function


paulschreiber:
WPCS should recognize `sanitize_hex_color()` as a sanitization function.

jrfnl:
This was fixed in #781

jrfnl:
@DannyCooper That's not necessarily up to me, but I can see your point, though I do find the inconsistency of the return values of the function a bit troubling, but that's another matter.

PR #781, opened 2017-01-13
Add sanitize for hex color

dingo-d:
Added `sanitize_hex_color` and `sanitize_hex_color_no_hash` to the list of sanitizing functions.
As mentioned in #780 by paulschreiber.


WordPress/Sniff.php
```diff
@@ -265,6 +265,8 @@
 		'sanitize_title'             => true,
 		'sanitize_user_field'        => true,
 		'sanitize_user'              => true,
+		'sanitize_hex_color'	     => true,
+		'sanitize_hex_color_no_hash' => true,
```
grappler: The function names are sorted alphabetically so these functions should be added to line 254.

dingo-d: Sorry, didn't notice this. Will fix it now.

WordPress/Sniff.php
```diff
@@ -252,6 +252,8 @@
 		'sanitize_email'             => true,
 		'sanitize_file_name'         => true,
 		'sanitize_html_class'        => true,
+		'sanitize_hex_color_no_hash' => true,
```
grappler: It is a tiny thing but `ht` would come after `he`.

dingo-d: On it! I should have just pulled the array out in phptester and sort it alphabetically XD

PR #782, opened 2017-01-13
Add unit testing of the exclude property for the abstract sniffs

jrfnl:


WordPress/Tests/Arrays/ArrayAssignmentRestrictionsUnitTest.inc
```diff
@@ -5,3 +5,16 @@ $foo = array( 'bar' => 1 ); // Bad.
 $foo['bar'] = 1; // Bad.
 
 $foo = 'test' . 'bar=1&taz=5&bar=2'; // Bad.
+
+/**
```
JDGrimes: This isn't a docblock, shouldn't it start with just `/*` and not `/**`? (And the same for the other three files.)

jrfnl: Good point. Fixed.

PR #788, opened 2017-01-15
Add section about writing sniffs / whitelist comments to CONTRIBUTING.md

jrfnl:


CONTRIBUTING.md
```diff
@@ -10,6 +10,36 @@ To contribute an improvement to this project, fork the repo and open a pull requ
 
 Once a commit is made to `develop`, a PR should be opened from `develop` into `master` and named "Next release". This PR will then serve provide a second round of Travis CI checks (especially for any hotfixes pushed directly to the `develop` branch), and provide collaborators with a forum to discuss the upcoming stable release.
 
+# Considerations when writing sniffs
+
+## Whitelist comments
+
+Sometimes, a sniff will flag code which upon further inspection by a human turns out to be ok.
```
GaryJones: ok -> OK

CONTRIBUTING.md
```diff
@@ -10,6 +10,36 @@ To contribute an improvement to this project, fork the repo and open a pull requ
...
+		if ( $this->has_whitelist_comment( 'CSRF', $stackPtr ) ) {
+			return;
+		}
+		
+		$phpcsFile->addError( ... );
+	}
+}
+```
+
+When you introduce a new whitelist comment, please don't forget to update the [whitelisting code wiki page](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Whitelisting-code-which-flags-errors) with the relevant details.
```
JDGrimes: The wiki should only be updated once the changes have actually been merged. I think that this should be reworded to make that more clear.

CONTRIBUTING.md
jrfnl: I don't understand ? This is a normal sentence, changing it to _ okay_ I could get, but why the ALLCAPS ok ?

jrfnl:
Updated for feedback.

FYI: I'll be updating the wiki page with the `@since` versions of each whitelist comment. That way if it gets updated while the comment is in a version not yet released, i.e. `develop`, and a new whitelist comment is added to the wiki page, it will be clear from what version onwards it can be used.

CONTRIBUTING.md
GaryJones: Because capitals is the standard way of displaying that.

jrfnl: I defer to your judgement ;-) Fixed.

PR #791, opened 2017-01-16
 	Improve custom property handling in WordPress_Sniffs_NamingConventions_ValidVariableNameSniff

jrfnl:
Allow for incorrectly set custom properties.

The `customPropertiesWhitelist` and (deprecated) `customVariablesWhitelist` properties should be provided with `type="array"`, but it's an easy mistake to miss that and provide a comma-delimited list instead as the value of the property would be the same.

Along the same lines, setting the property inline for unit testing the property handling will fail as there is no way to pass the `type="array"` part via an inline setting change.

Without the fix in this commit, this would lead to errors that `array_merge()` expects the parameters to be in array format.

* For the `$customPropertiesWhitelist`, a warning will be thrown that the format of the custom property should be corrected.
* If the received value is a string, the format will be corrected. Otherwise the value will be disregarded.
* In the case of the deprecated `$customVariablesWhitelist` property, a warning will be thrown when an incorrect format is encountered and the contents of the property will be disregarded regardless of what was received.

Includes unit tests for the custom property.

### DIFF
--- a/WordPress/Tests/NamingConventions/ValidVariableNameUnitTest.php
+++ b/WordPress/Tests/NamingConventions/ValidVariableNameUnitTest.php
@@ -76,7 +76,9 @@ class WordPress_Tests_NamingConventions_ValidVariableNameUnitTest extends Abstra
 	 * @return array <int line number> => <int number of warnings>
 	 */
 	public function getWarningList() {
-		return array();
+		return array(
+			1 => 1, // Warning about property setting.
+		);
 
 	}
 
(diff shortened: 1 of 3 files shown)


WordPress/Tests/NamingConventions/ValidVariableNameUnitTest.php
```diff
@@ -76,7 +76,9 @@ public function getErrorList() {
 	 * @return array <int line number> => <int number of warnings>
 	 */
 	public function getWarningList() {
-		return array();
+		return array(
+			1   => 1, // Warning about property setting.
```
grappler: As it is only a single value I would have only have a single space between the key and `=>`

jrfnl: Fixed

PR #795, opened 2017-01-16
Refactor the WordPress_AbstractFunctionRestrictionsSniff

jrfnl:
This PR contains three distinct commits. The first is a refactor of the abstract function restrictions class, the second and third are changes to child classes taking advantage of the refactor.

-----

Refactor the WordPress_AbstractFunctionRestrictionsSniff

This refactor is intended to make the WordPress_AbstractFunctionRestrictionsSniff more extendable and reusable.
The refactor does not contain any functional changes.

* Extend the main `WordPress_Sniff` to be able to use the `init()` function which will save us having to pass around the `$phpcsFile` and `$tokens` variables and make the utility functions contained in that class available (and allow for moving some utility functions contained in the `WordPress_AbstractClassRestrictionsSniff` to the main `WordPress_Sniff ` class in a future PR).
* Split up the logic of the sniff into three distinct functions which each individually can be overloaded in a child class.

Additionally:
* Added since tags to properties and functions introduced after the initial version of the class.

-----
Simplify the WordPress_AbstractClassRestrictionsSniff

The refactor of the `WordPress_AbstractFunctionRestrictionsSniff` allows for some code simplification in the `WordPress_AbstractClassRestrictionsSniff`.

The simplification does not contain any functional changes.

* Defer to parent `process()` method for initial verification & setup.
* Defer to parent `process_matched_token()` method for the throwing of the error.
* Use the `$this->tokens` and `$this->phpcsFile` properties now available instead of passing the variables between functions.

Also:
* Removed a property which couldn't be used as is anyway (as the parent uses `self` and PHCPS code still has to be compatible with PHP 5.2, so we can't use `static`).

-----
Simplify the WordPress_Sniffs_WP_DeprecatedFunctionsSniff & make it more efficient

The refactor of the `WordPress_AbstractFunctionRestrictionsSniff` allows to make the `WordPress_Sniffs_WP_DeprecatedFunctionsSniff` more efficient.

* No need to loop through the array and create all the error messages and subgroups for the initial `getGroups()` setup.
* Instead, overload the parent's `process_matched_token()` method and move the logic for setting up the error message and such there, so we only execute that logic when the function has actually been matched.

This does mean that individual deprecated function checks can no longer be excluded using the `exclude` property.
However, the errorCode has been changed to a dynamic one which will still allow modular exclusion of one or more deprecated functions from a ruleset, though now it would need to be done using the errorCode.

### DIFF
--- a/WordPress/AbstractFunctionRestrictionsSniff.php
+++ b/WordPress/AbstractFunctionRestrictionsSniff.php
@@ -17,8 +17,9 @@
  *                 Moved the file and renamed the class from
  *                 `WordPress_Sniffs_Functions_FunctionRestrictionsSniff` to
  *                 `WordPress_AbstractFunctionRestrictionsSniff`.
+ * @since   0.11.0 Extends the WordPress_Sniff class.
  */
-abstract class WordPress_AbstractFunctionRestrictionsSniff implements PHP_CodeSniffer_Sniff {
+abstract class WordPress_AbstractFunctionRestrictionsSniff extends WordPress_Sniff {
 
 	/**
 	 * Exclude groups.
@@ -34,6 +35,8 @@ abstract class WordPress_AbstractFunctionRestrictionsSniff implements PHP_CodeSn
 	 * Don't use this in extended classes, override getGroups() instead.
 	 * This is only used for Unit tests.
 	 *
+	 * @since 0.10.0
+	 *
 	 * @var array
 	 */
 	public static $unittest_groups = array();
@@ -41,12 +44,16 @@ abstract class WordPress_AbstractFunctionRestrictionsSniff implements PHP_CodeSn
 	/**
 	 * Regex pattern with placeholder for the function names.
 	 *
+	 * @since 0.10.0
+	 *
 	 * @var string
 	 */
 	protected $regex_pattern = '`\b(?:%s)\b`i';
 
 	/**
 	 * Cache for the group information.
+	 *
+	 * @since 0.10.0
 	 *
 	 * @var array
 	 */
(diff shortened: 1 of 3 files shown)


WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -198,27 +224,40 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
...
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 *
+	 * @return void
+	 */
+	public function process_matched_token( $stackPtr, $group_name, $matched_content ) {
 
+		if ( 'warning' === $this->groups[ $group_name ]['type'] ) {
+			$addWhat = array( $this->phpcsFile, 'addWarning' );
```
grappler: I think it would be better to directly call the method `addWarning` like is done further done in [#diff-ede7f8be23ca928d9ba5e981dc4c08fbR1137](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/795/files#diff-ede7f8be23ca928d9ba5e981dc4c08fbR1137) instead of using `call_user_func()`

jrfnl: While I agree, that is outside the scope of this PR.
There are a number of places in the code where a error/warning switch is made and several different ways of doing this are being used. Typical one which lends itself to a utility function to remove duplicate code / create logic consistency.

jrfnl: @grappler FYI: In the mean time I've prepared a branch with that utility function & fixing this across files. Waiting for some other merges to PR it to prevent conflicts.

PR #796, opened 2017-01-17
Remove unnecessary checks for `$phpcsFile->Fixer`

jrfnl:
That's what the `$fix` return value is for.


JDGrimes:
In case anyone is wondering, this was originally written the way it was for back-compat with old versions of PHPCS before the fixer feature was introduced. We don't support those old versions of PHPCS anymore, so there is no need to have this added legacy cruft hanging around and adding unnecessary complexity.

Issue #642, opened 2016-07-28
Add additional file name checks


jrfnl:
> Class file names should be based on the class name with class- prepended and the underscores in the class name replaced with hyphens, for example WP_Error becomes:
> 
> `class-wp-error.php`
> 
> This file-naming standard is for all current and new files with classes. There is one exception for three files that contain code that got ported into BackPress: class.wp-dependencies.php, class.wp-scripts.php, class.wp-styles.php. Those files are prepended with class., a dot after the word class instead of a hyphen.
> 
> Files containing template tags in wp-includes should have -template appended to the end of the name so that they are obvious.
> 
> `general-template.php`

Ref: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#naming-conventions

This rule is currently not covered and would have been untestable, however, a recent change in PHPCS which will be included in PHPCS 2.7 will allow for this: https://github.com/squizlabs/PHP_CodeSniffer/issues/1078

Issue #470, opened 2015-10-05
WordPress.Files.FileName.UnderscoresNotAllowed and the template hierarchy


mboynes:
Currently, this standard will throw an error if a file is encountered whose filename contains an underscore:

```
 1 | ERROR | [ ] Filename "taxonomy-post_format-post-format-link.php"
   |       |     with underscores found; use
   |       |     taxonomy-post-format-post-format-link.php instead
   |       |     (WordPress.Files.FileName.UnderscoresNotAllowed)
```

Technically, this is correct. The [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/) states that "hyphens should separate words" in file names. However, this may conflict with the [template hierarchy's requirements](https://developer.wordpress.org/themes/basics/template-hierarchy/), which use post types and taxonomies (as well as page/post slugs, author nicenames, mime types, and more) as part of the file name. For instance, from the docs on the template hierarchy,

> In the case of post formats, the taxonomy is ‘post_format’ and the terms are ‘post-format-{format}. i.e. taxonomy-post_format-post-format-link.php for the link post format.

Therefore, the file _unavoidably_ named `taxonomy-post_format-post-format-link.php` will throw an error as in the example above. I understand that exclusions can be added to the ruleset, but I think the question should still be raised: **Should template hierarchy files be excepted from this sniff?**

JDGrimes:
Is it possible to craft a general set of exclusions that can reliably detect template files?

westonruter:
I think we could whitelist template files based on known patterns. For instance, before reporting an error we could check if it matches the file matches the pattern `/taxonomy-.+-post-format-link\.php/`, along with the other possible template names that have slugs in them (e.g. `single-restaurant_location.php` or `archive-club_member.php`).

GaryJones:
I've not checked, but is WP clever enough to apply `taxonomy-post-format-post-format-link.php` to the `post_format` post type? Not something I've seen before, but might be sanitized in that way.

GaryJones:
- `~single-.*\.php~`
- `~archive-.*\.php~`
- `~taxonomy-.*\.php~`
- `~embed-.*\.php~`

Which gives `~(archive|embed|single|taxonomy)-.*\.php~`.

[...]

PR #797, opened 2017-01-17
Fix one error message not using the PHPCS native sprintf functionality

jrfnl:


WordPress/Sniffs/Files/FileNameSniff.php
```diff
@@ -43,8 +43,12 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 
 		if ( false !== strpos( $fileName, '_' ) ) {
 			$expected = str_replace( '_', '-', $fileName );
-			$error    = 'Filename "' . $fileName . '" with underscores found; use ' . $expected . ' instead';
-			$phpcsFile->addError( $error, $stackPtr, 'UnderscoresNotAllowed' );
+			$error    = 'Filename "%s" with underscores found; use %s instead';
```
grappler: Could we not directly add the strings to the method instead of first defining a variable?

jrfnl: We could. I generally try to stick to the "minimal change" principle, i.e. "Don't change more than what you are fixing.", but agree that the variable definitions in this case aren't really needed (goes for all three, not just the variable I added).

jrfnl:
Closing this PR in favour of an upcoming one PR which will address #470, #805 and in part #642

PR #804, opened 2017-01-17
Utility: Add `strip_quotes()` utility function & implement use

jrfnl:
Add `strip_quotes()` utility method to consistently and correctly strip quotes from `T_CONSTANT_ENCAPSED_STRING` / `T_DOUBLE_QUOTED_STRING` strings.

Until now, most often `trim( $var, '\'"' )` was used which might incorrectly strip too much.
Example: `"'my' string"` would become `my' string` (note the extra `'` which was removed) instead of `'my' string`.

This utility function will strip the quotes off correctly.

The second commit implements use of the `strip_quotes()` method where appropriate in the current codebase.


jrfnl:
For anyone wondering: in PCRE a `$` match means end of string **or** just before a newline at the end of string. The `D` makes it so it will only match end of string.
If you know your subject string is not supposed to have a new line at the end, it's good practice to always use it when using a `$` in a regex as you might otherwise open yourself up to [some security issues](http://web.archive.org/web/20131226183006/http://blog.php-security.org/archives/76-Holes-in-most-preg_match-filters.html).

PR #806, opened 2017-01-17
FileNames: add template exceptions, class and template tags rules

### LINKED ISSUE #470: WordPress.Files.FileName.UnderscoresNotAllowed and the template hierarchy

### LINKED ISSUE #642: Add additional file name checks

### LINKED ISSUE #805: FileNameSniff triggers on taxonomy-{taxonomy_slug}.php files with underscores
The advice when registering a [custom taxonomy](https://codex.wordpress.org/Function_Reference/register_taxonomy) is "Name should only contain lowercase letters and the underscore character", yet if you do this and have a custom taxonomy with underscores such as `resource_category` then have a theme template called `taxonomy-resource_category` the FileNameSniff triggers.

According to the [template hierarchy](https://developer.wordpress.org/themes/basics/template-hierarchy/#custom-taxonomies) documentation this sniff would also trigger if a theme were to customise based on post formats, such as `taxonomy-post_format.php`.

Seems to be that the sniff needs to take into account these taxonomy template files and allow these instances.

jrfnl:
This PR takes care of the - up to now - uncovered parts of the file name rules as outlined in #642.
It also adds a `public $is_theme` property, which when set to `true` via the ruleset will gracefully deal with exceptions to the file name rules for dynamic template file names as discussed in #470.

Notes:
* The `Core` ruleset used to include the `Generic.Files.LowercasedFilename` and this sniff for the file name checks. This would potentially lead to two messages being thrown with different suggestions about the expected file name. The lowercase check has now been incorporated into this sniff and the upstream sniff has been removed from the ruleset.
* The error message and the errorcode for this original check `UnderscoresNotAllowed` has changed to `NotHyphenatedLowercase` to better cover the check currently being executed.
* Added check for `class-` prefix for files containing a class including verification that the rest of the file name reflects the name of the class.
* Added check for `-template` suffix for files which are in the `wp-includes` directory and contain a `@subpackage Template` tag. This check will only be done for non-class files. This is in line with the current file names as used in WP Core.
* Added exception for dynamic template files. This requires that the custom ruleset sets the `is_theme` property to `true` (documentation upcoming in the wiki).
The exception covers nearly everything which is needed. The only uncovered part at this moment is "sublevel"-only mimetype template files, like `plain.php`.

Includes extensive unit tests for the sniff (was until now not unit tested)

Fixes #470
Fixes #642 
Fixes #805


JDGrimes:
Nice work.

As a plugin developer that doesn't follow the (IMO, silly) naming convention for classes, I guess the only real difference for my plugin between this sniff and the `Generic.Files.LowercasedFilename`, will still just be that this sniff checks that hyphens are used in filenames instead of underscores (since the other checks won't run, and the class check will be disabled by my custom ruleset). In that case, I will probably exclude this sniff in my custom ruleset, for performance reasons, and use `Generic.Files.LowercasedFilename` instead. Because even if the errors from the class files are disabled in a custom ruleset, the check will still be run. There are ways that this could be mitigated, if you think that it is worth it, but personally I'll be just as happy with the generic sniff. I can foresee this being a fairly common circumstance though.

jrfnl:
@JDGrimes I could easily add a public property to enable/disable the class and template tag rules completely (i.e. bow out before doing those checks) or disable them each individually (though the template one won't run if not in a directory with `/wp-includes/` in the path anyway).

Would that help ?

For a property name, I'm thinking `strict_class_file_names` and `strict_templatetag_file_names` with a boolean value. Or just `strict` if we'd make it an overall property.

JDGrimes:
@jrfnl Yes, that would help. As you say, it is most important for the class filenames check, since the other only applies inside of the `wp-includes` directory. `strict_class_file_names` sounds good to me.

jrfnl:
@JDGrimes Ok, done.

Issue #614, opened 2016-07-17
Add wiki page with info on ruleset customizations available ?


jrfnl:
As #599 now offers a [custom property which people can manipulate from the ruleset](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/599#issuecomment-233066050), maybe we should make info about these kind of properties a bit more easy to find for people ?

The basics are covered in https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml

But what about having a wiki page listing the WPCS sniffs which offer adjustable properties ?
And maybe add some info on generic things which can be adjusted such as `<type>` too ?

GaryJones:
Go for it.

grappler:
Another example

```xml
<rule ref="WordPress.WP.AlternativeFunctions">
    <properties>
        <property name="exclude" value="curl,file_get_contents"/>
    </properties>
</rule>
```

jrfnl:
FYI: I've got the list ready. There are just a few more things related to custom properties in the codebase which need to be adjusted before I can publish - see #765 as well.

If anyone wants to review the list already, I could create a gist with the info for now (or add it to the wiki anyway, but with a large "DRAFT WAITING FOR SOME PRS" above it ;-) ).

JDGrimes:
I'd say you might go ahead and add it to the wiki as a "draft". Some folks will probably find it useful even if it isn't quite perfect yet.  But you know best what discrepancies still exist between the list and the codebase, and how these affect the list's value at present.

jrfnl:
I've got PRs ready for all the discrepancies, but what with merge conflicts and such, they can not be pulled all at the same time, so are waiting their turn.

Current discrepancies involve:
* Some properties which are still set to `public`, but **really** should be made `protected` as changing these from the ruleset would break the sniff or at the very least break the intention of the sniff.
* Some properties already in the list, but not yet added to the codebase, like the `remove_only` property which is being introduced in #817 
* Some properties, most notably the ones which merge custom arrays with preset arrays not being handled correctly internally in WPCS as discussed in #765 and #786

Altogether it adds up to quite a lot, which makes me feel a bit uncomfortable putting it in the wiki yet.

jrfnl:
I've put a draft version online in a gist. Be aware: The page already contains info on things which have not been merged - or even not been pulled yet -, which is why I'm not adding it to the wiki (yet).

[...]

PR #808, opened 2017-01-19
Extra: Make the unnecessary string concat check a little bit more lenient

jrfnl:
Allow for string concat if the string is broken up in several lines - for instance to avoid overly long lines.

Just came across this property and figured it was worth having a discussion on whether or not to include it.


jrfnl:
> I did not know about this property,

Nor did I. I'm compiling a list of the PHPCS public properties for https://github.com/squizlabs/PHP_CodeSniffer/issues/1278, similar to what I've prepared for #614 (waiting for some merges including some branches waiting in the wings before I can publish it). Came across it that way.

Turns out that making the public properties more easily discoverable yields results straight away :sunglasses: 

> I have the sniff disabled for this very reason.

Same here ;-)

PR #815, opened 2017-01-22
Add function call parameter related utility functions and abstract class

### LINKED ISSUE #745: New Abstract for restricting arguments in functions
I started working on something for theme sniffs but realized this would fit better ustream. https://github.com/WPTRT/WordPress-Coding-Standards/pull/80

As we have an abstract to restrict function it would make sense to have one to restrict arguments.

A few examples for this would be 
- `get_bloginfo( 'home' )` where `home` is deprecated
- making sure the correct argument is used e.g. `date_i18n( get_option( 'date_format' ) )` #739
- requiring that a function has a certain argument `get_bloginfo( 'name', 'display' )`
- if a certain hook should not be used e.g `add_filter( 'mime_types', '' );` as it is plugin territory
- unneeded checking for backwards compatibility with `function_exist( 'add_action' )`

Related issues https://github.com/WPTRT/WordPress-Coding-Standards/issues/93 https://github.com/WPTRT/WordPress-Coding-Standards/issues/42 https://github.com/WPTRT/WordPress-Coding-Standards/issues/27 https://github.com/WPTRT/WordPress-Coding-Standards/issues/1 https://github.com/WPTRT/WordPress-Coding-Standards/issues/96
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/611

The current array is as below.
```php
array( 
'get_bloginfo' => array(
	1 => array(
		'home' => array(
			'alt'     => 'home_url()',
			'version' => '2.2'
		),
		'siteurl' => array(
			'alt'     => 'home_url()',
			'version' => '2.2'
		),
	),
),
);
```

First we have the function name, the position of the argument, the argument it should match. Alt is the alternative function for the deprecated parameter and version is when it was deprecated.

**What I unsure about is how I can differentiate when a argument should not be used and when a argument is required.**

jrfnl:
Now #795 has been merged, it was possible to create an abstract function parameter class which builds upon the logic for determining function calls already contained in `WordPress_AbstractFunctionRestrictionsSniff`.

This PR introduces the abstract and a number of related utility functions.

### Utility functions

I initially introduced these utility functions in the [PHPCompatibility](https://github.com/wimg/PHPCompatibility) standard were they are also [individually unit tested](https://github.com/wimg/PHPCompatibility/tree/master/Tests/BaseClass).
They have been in use there for a number of months and proven useful & accurate.

The functions are:
* `does_function_call_have_parameters( $stackPtr )` - returns bool
* `get_function_call_parameter_count( $stackPtr )` - returns int
* `get_function_call_parameters( $stackPtr )`  - returns array, key: 1-indexed parameter position, value: array with `start` - int stack pointer of param start position, `end` - int stack pointer of param end position and `raw` - trimmed raw content for the passed parameter
* `get_function_call_parameter( $stackPtr, $param_offset )` - returns array or false if no parameter was found at the specified offset

The functions expect to be passed a `T_STRING` stack pointer to a function call.

And while the function name might not indicate this, each of these functions can also be used with array stack pointers (`T_ARRAY`, `T_OPEN_SHORT_ARRAY`). This is a feature which was introduced after the initial naming & merge of the functions and the function documentation reflects this.

It is my intention to introduce these functions upstream in PHPCS itself at some point in the future.

### New `WordPress_AbstractFunctionParameterSniff` class

The new `WordPress_AbstractFunctionParameterSniff` extends the `WordPress_AbstractFunctionRestrictionsSniff` class for the yes/no function call determination, retrieves the parameters used by the targetted function call using the new utility functions and then passes processing on to the child class.

The child class is expected to implement a `process_parameters()` method to handle the actual testing of parameters.

Optionally a `process_no_parameters()` method can be added to the child class to handle cases where parameters are expected, but not found.

The abstract expects a `$target_functions` array, but is without opinion about how that array should look, except for the requirement that the top level array keys are the function names.
Further requirements of the array are not needed as the processing of the parameters is done in the child class, allowing for great flexibility in potential use of this class.

### Start using the abstract

As a last step - and to prove that the same functionality is maintained, and in some cases improved by the new abstract - I've refactored three existing sniffs which were prime candidates (duplicate logic and such) to start using the new `WordPress_AbstractFunctionParameterSniff`.

Fixes #745

### DIFF
--- a/WordPress/AbstractFunctionParameterSniff.php
+++ b/WordPress/AbstractFunctionParameterSniff.php
@@ -29,8 +29,9 @@ abstract class WordPress_AbstractFunctionParameterSniff extends WordPress_Abstra
 	 * Functions this sniff is looking for. Should be defined in the child class.
 	 *
 	 * @var array The only requirement for this array is that the top level
-	 * array keys are the names of the functions you're looking for.
-	 * Other than that, the array can have arbitrary content depending on your needs.
+	 *            array keys are the names of the functions you're looking for.
+	 *            Other than that, the array can have arbitrary content
+	 *            depending on your needs.
 	 */
 	protected $target_functions = array();
 
--- a/WordPress/AbstractFunctionParameterSniff.php
+++ b/WordPress/AbstractFunctionParameterSniff.php
@@ -1,0 +1,104 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+/**
+ * Advises about parameters used in function calls.
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   0.11.0
+ */
+abstract class WordPress_AbstractFunctionParameterSniff extends WordPress_AbstractFunctionRestrictionsSniff {
+
+	/**
+	 * The group name for this group of functions.
+	 *
+	 * Intended to be overruled in the child class.
+	 *
+	 * @var string
+	 */
+	protected $group_name = 'restricted_parameters';
+
+	/**
+	 * Functions this sniff is looking for. Should be defined in the child class.
+	 *
+	 * @var array The only requirement for this array is that the top level
+	 * array keys are the names of the functions you're looking for.
+	 * Other than that, the array can have arbitrary content depending on your needs.
+	 */
+	protected $target_functions = array();
+
+	/**
+	 * Groups of function to restrict.
+	 *
+	 * @return array
+	 */
+	public function getGroups() {
+		if ( empty( $this->target_functions ) ) {
+			return array();
+		}
+
+		return array(
+			$this->group_name => array(
+				'functions' => array_keys( $this->target_functions ),
+			),
+		);
+	}
+
+	/**
+	 * Process a matched token.
+	 *
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 *
+	 * @return void
+	 */
+	public function process_matched_token( $stackPtr, $group_name, $matched_content ) {
+
+		$parameters = $this->get_function_call_parameters( $stackPtr );
+
+		if ( empty( $parameters ) ) {
+			$this->process_no_parameters( $stackPtr, $group_name, $matched_content );
+		} else {
+			$this->process_parameters( $stackPtr, $group_name, $matched_content, $parameters );
+		}
+	}
+
+	/**
+	 * Process the parameters of a matched function.
+	 *
+	 * This method has to be made concrete in child classes.
+	 *
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return void
+	 */
+	abstract public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters );
+
+	/**
+	 * Process the function if no parameters were found.
+	 *
+	 * Default to doing nothing. Can be overloaded in child classes to handle functions
+	 * were parameters are expected, but none found.
+	 *
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 *
+	 * @return void
+	 */
+	public function process_no_parameters( $stackPtr, $group_name, $matched_content ) {
+		return;
+	}
+
+} // End class.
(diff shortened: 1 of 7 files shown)


WordPress/AbstractFunctionParameterSniff.php
```diff
@@ -0,0 +1,104 @@
...
+	 * @var string
+	 */
+	protected $group_name = 'restricted_parameters';
+
+	/**
+	 * Functions this sniff is looking for. Should be defined in the child class.
+	 *
+	 * @var array The only requirement for this array is that the top level
+	 * array keys are the names of the functions you're looking for.
+	 * Other than that, the array can have arbitrary content depending on your needs.
```
grappler: Does the text not need to be all on the same height? e.g.
```php
	/**
	 * @var array The only requirement for this array is that the top level
	 *            array keys are the names of the functions you're looking for.
	 *            Other than that, the array can have arbitrary content depending
	 *            on your needs.
```

WordPress/Sniff.php
```diff
@@ -1291,4 +1291,224 @@ protected function get_interpolated_variables( $string ) {
...
+
+		$parameters  = array();
+		$next_comma  = $opener;
+		$param_start = ( $opener + 1 );
+		$cnt         = 1;
+		while ( $next_comma = $this->phpcsFile->findNext( array( T_COMMA, $this->tokens[ $closer ]['code'], T_OPEN_SHORT_ARRAY ), ( $next_comma + 1 ), ( $closer + 1 ) ) ) {
+			// Ignore anything within short array definition brackets.
+			if ( 'T_OPEN_SHORT_ARRAY' === $this->tokens[ $next_comma ]['type']
+				&& ( isset( $this->tokens[ $next_comma ]['bracket_opener'] )
+					&& $this->tokens[ $next_comma ]['bracket_opener'] === $next_comma )
```
grappler: Is there a reason why there is an extra tab before the code?

WordPress/Sniffs/PHP/StrictInArraySniff.php
```diff
@@ -34,98 +44,61 @@ class WordPress_Sniffs_PHP_StrictInArraySniff implements PHP_CodeSniffer_Sniff {
...
+	 * @since 0.11.0
+	 *
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 *
+	 * @return void
+	 */
+	public function process_no_parameters( $stackPtr, $group_name, $matched_content ) {
+		$this->phpcsFile->addError( 'Missing arguments to %s.', $stackPtr, 'MissingArguments', array( $matched_content ) );
```
grappler: I don't think we need this as if there are no parameters then there will be a PHP error. https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/770#discussion_r94807710

WordPress/Sniff.php
jrfnl: Yes: nested parenthesis.

WordPress/Sniffs/PHP/StrictInArraySniff.php
jrfnl: We do need this.
By far not all function calls have required parameters, so a function call without parameters can be perfectly valid code, no syntax errors, but might still need to throw an error under certain rules.

grappler: The three functions that we are checking have required parameters. `in_array`, `array_search` and `array_keys`

jrfnl: Oh sorry, I was thinking the comment was about the abstract.

And true, in this case "no parameters" would also result in a PHP error, though not one which would be picked up by the `Generic.PHP.Syntax` sniff (not a parse error).
I added the method and the error message in this case to maintain backward compatibility. The same message was already being thrown by this sniff and I saw no compelling reason to remove it.

jrfnl:
> One thought that came to me is what would happen if one of the parameters that we were checking was an anonymous function. A common place where anonymous functions are used are hooks.

That works just fine, you will receive the start and end parameter and the raw content, no problem ;-)

More in-depth: The functions check for parenthesis at the correct nesting level and will skip over everything which is not at the right nesting level, so an anonymous function as a parameter will be returned and treated as one parameter.

And before you ask: The methods don't deal with parameters passed to anonymous functions, but don't need to either as that is a function declaration and you can use the PHPCS native `getMethodParameters()` for that.

WordPress/Sniff.php
```diff
@@ -1291,4 +1291,224 @@ protected function get_interpolated_variables( $string ) {
...
+	 * If no parameters are found, will return an empty array.
+	 *
+	 * Extra feature: If passed an T_ARRAY or T_OPEN_SHORT_ARRAY stack pointer,
+	 * it will tokenize the values / key/value pairs contained in the array call.
+	 *
+	 * @since 0.11.0
+	 *
+	 * @param int $stackPtr The position of the function call token.
+	 *
+	 * @return array
```
grappler: Could we document the output too? Perhaps with something like:

```php
	/**
	 * @return array {
	 *      Array of parameters. Empty array if there are no parameters.
 	 *
 	 *     @type int $position Position of the parameter. {
	 *         @type int $start Start position of the call token.
	 *         @type int $end  End position of the call token.
	 *         @type int $raw  Raw parameter content.
	 *     }
	 * }
```

WordPress/AbstractFunctionParameterSniff.php
JDGrimes: I second this.

jrfnl: Done

WordPress/Sniff.php
jrfnl: It already was documented a bit further up in the function documentation, but I've adjusted the format to a variant of what you suggested.

Issue #632, opened 2016-07-25
Some regex best practices


jrfnl:
With the regex functions, there are some best practices we could advise on /often made "mistakes" which we could warn people about.

I'm thinking along the lines of:
- Use of `/` as delimiter -> advise to use other delimiter
- Use of variables in the regex -> use `preg_quote()`
- Use of `g` modifier -> Not used in PHP use `preg_*_all()` functions instead
- Use of `s` and `m` modifier in one regex -> they don't go together, review regex

Additionally the sniff could potentially check for:
- non-PCRE syntaxes used in the regex
- modifiers which are not implemented/available in PHP

This sniff would go into the `extra` ruleset.

Opinions ?

westonruter:
💯 brilliant!

PR #816, opened 2017-01-23
Accuracy: always pass the delimiter to `preg_quote()` if not using `/`.

jrfnl:
Reviewed usage of the `preg_quote()` function which should always be passed the delimiter if not using `/` to make sure the correct characters are being escaped.


JDGrimes:
Shouldn't there be a sniff for this? 😄

jrfnl:
@JDGrimes It has been suggested before ;-) #632

PR #817, opened 2017-01-23
AdminBarRemoval: More thorough checking for admin bar removal and hiding

jrfnl:
This sniff has been completely overhauled with the additional wish-list of the Theme Review Team in mind.

Originally the sniff only checked for `show_admin_bar( ... )` and `add_filter( 'show_admin_bar', ... )`.
* It didn't contain a further check for the actual value passed to these functions.
* It didn't check for usage of CSS to hide the admin bar

This PR fixes that.
Includes extensive additional unit tests.

Notes:
* The sniff now checks both PHP as well as CSS files.
* The class now extends the `WordPress_AbstractFunctionParameterSniff` for the PHP function parameter checks.
* It overloads the `register()` method to register additional tokens.
* It overloads the `process()` method of that class to only use the functions in the abstract class when passed a `T_STRING` token. For the other token types, the processing is handled within the sniff.
* The sniff now contains a public property `$remove_only` which defaults to `true`, to toggle whether to just check that the admin bar is being manipulated or to do a more in-depth check and only throw errors when removing or hiding of the admin bar is detected.
* The sniff also contains a new utility function to work around a bug in the PHP 5.2 tokenizer regarding text strings containing `<s`, like `<style>`.

Who should I ping from the VIP team to have a look as well ?

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -1514,4 +1514,65 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 		return $parameters[ $param_offset ];
 	}
 
+	/**
+	 * Check if a content string contains a specific html open tag.
+	 *
+	 * {@internal For PHP 5.3+ this is straightforward, just check if $content
+	 * contains the tag.
+	 * PHP 5.2 however, creates a separate token for `<s` when used in inline HTML,
+	 * so in that case we need to check that the next token starts with the rest
+	 * of the tag.
+	 * I.e. PHP 5.2 tokenizes the inline HTML `text <span>text</span> text` as:
+	 * - T_INLINE_HTML 'text'
+	 * - T_INLINE_HTML '<s'
+	 * - T_INLINE_HTML 'pan>text</span> text'
+	 *
+	 * We don't need to worry about checking the rest of the content of the next
+	 * token as sniffs using this function will be sniffing for all text string
+	 * tokens, so the next token will be passed to the sniff in the next iteration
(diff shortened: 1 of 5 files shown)


grappler:
Please can you have a look @sboisvert @david-binda

WordPress/Sniff.php
```diff
@@ -1514,4 +1514,63 @@ public function get_function_call_parameter( $stackPtr, $param_offset ) {
 		return $parameters[ $param_offset ];
 	}
 
+	/**
+	 * Check if a content string contains a specific html open tag.
+	 *
+	 * {@internal For PHP 5.3+ this is straightforward, just check if $content
+	 * contains the tag.
+	 * PHP 5.2 however, creates a seperate token for `<s` when used in inline HTML,
```
jrfnl: > Why did the tokenizer create a separate token in PHP 5.2?

See https://bugs.php.net/bug.php?id=48446:
> It stops at "<s" to avoid a long PHP opening tag (e.g. <script language="php">,
etc.) from being taken as inline HTML.

> The good news, however, is that a new scanner is used for PHP 5.3, and as of a few weeks ago (5.3.0 RC2), it now works as you'd expect.

jrfnl: P.S.: Fixed the typo and added the link to the documentation.

sboisvert:
I haven't read the code in detail but just the goal of what this is achieving is good from VIP's point of view, thanks!

jrfnl:
@sboisvert Thanks for looking through it.

PR #818, opened 2017-01-23
Improve cron schedule sniff

jrfnl:
* Deal correctly with WP time constants leading to more accurate error/warning messages.
* Takes advantage of the new function parameter utility functions for greater accuracy.
* Allow for short array syntax in callback for `add_filter`.
* Make the finding of the referenced function declaration more accurate and efficient by only looking at `T_FUNCTION` tokens, skipping past the rest of the function and breaking out if the right function is found. (Still could get confused over two classes in the same file with a same name method, but better than it was before)
* Minor other efficiency tweaks


jrfnl:
Any one has an opinion on this ? Maybe someone from the VIP team ? /cc @sboisvert @david-binda

jrfnl:
Thanks @JDGrimes

PR #820, opened 2017-01-24
Abstracts: Fix potential bug - methods should allow for optionally returning a stack pointer

jrfnl:
PHPCS allows for the `process()` method to optionally return a stack pointer. If so, it will not call the sniff for the current file again until the returned stack pointer has been reached.

I did not take that into account in the refactor of the AbstractFunctionRestrictionsSniff and the creation of the AbstractFunctionParameterSniff which basically blocks any child classes from using the ability to return a stack pointer.

This has now been fixed.

### DIFF
--- a/WordPress/AbstractFunctionRestrictionsSniff.php
+++ b/WordPress/AbstractFunctionRestrictionsSniff.php
@@ -154,7 +154,8 @@ abstract class WordPress_AbstractFunctionRestrictionsSniff extends WordPress_Sni
 	 * @param int                  $stackPtr  The position of the current token
 	 *                                        in the stack passed in $tokens.
 	 *
-	 * @return void
+	 * @return int|void Integer stack pointer to skip forward or void to continue
+	 *                  normal file processing.
 	 */
 	public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 
@@ -168,7 +169,7 @@ abstract class WordPress_AbstractFunctionRestrictionsSniff extends WordPress_Sni
 		// Make phpcsFile and tokens available as properties.
 		$this->init( $phpcsFile );
 
-		$this->is_targetted_token( $stackPtr );
+		return $this->is_targetted_token( $stackPtr );
 
 	} // End process().
 
@@ -179,7 +180,8 @@ abstract class WordPress_AbstractFunctionRestrictionsSniff extends WordPress_Sni
 	 *
 	 * @param int $stackPtr The position of the current token in the stack.
 	 *
-	 * @return void
+	 * @return int|void Integer stack pointer to skip forward or void to continue
+	 *                  normal file processing.
 	 */
 	public function is_targetted_token( $stackPtr ) {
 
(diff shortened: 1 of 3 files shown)


WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -224,10 +228,12 @@ public function is_targetted_token( $stackPtr ) {
 			}
 
 			if ( preg_match( $group['regex'], $token_content ) === 1 ) {
-				$this->process_matched_token( $stackPtr, $groupName, $token_content );
+				$skip_to = $this->process_matched_token( $stackPtr, $groupName, $token_content );
```
JDGrimes: Shouldn't this check the previous value of `$skip_to`, and only overwrite it if the new value is less? Assuming that different matches can return different values there. In practical terms, that would mean that the minimum number of tokens requested by the groups would be skipped, rather than just the last thing that happened to be returned there. And if any of the matches return null, then probably nothing should be skipped.

Though on further thought, perhaps different sniffs would want different behaviors here.

jrfnl: > Shouldn't this check the previous value of $skip_to, and only overwrite it if the new value is less?

Good question. Should the highest or the lowest value be returned though ? A case can be made for either one.
Code wise, the simplest way would be to collect all the returned values in an array and return it at the end with either a `min()` or `max()` around it.

In practice, none of the sniffs which use the function restrictions abstract or the child abstracts of this one, return a value at this moment, but it could become an issue if any of them would at some point, though only if within that same sniff they would sniff for the same function in two different groups.
The only sniff which at this moment does so, is the `WP.AlternativeFunctions` sniff, which sniffs for `file_get_contents()` in both the `file_get_contents` and the `file_system_read` group.

jrfnl: > perhaps different sniffs would want different behaviors here.

Oh and I forgot to respond to that part. If we want to allow for that, I'd suggest splitting the `is_targetted_token()` method up into two, where the looping through the groups is moved to a separate method allowing child classes to overload just that particular part of the class.

JDGrimes: > Should the highest or the lowest value be returned though ? A case can be made for either one.

Yeah, that's why I say different sniffs might want different behaviors here. But the reason I suggested lowest is because generally skipping isn't a requirement to the proper operation of the sniff, the worst that could happen is an error gets reported twice or something. Whereas if you skip extra beyond a particular check, the sniff might not end up reporting errors that it is supposed to.

But all things considered, I guess it really isn't important to resolve this right now. Maybe it would be best to just leave it as it is, and change the behavior only when we have an actual use-case that requires a change.

jrfnl: We might as well sort it out now as the potential issue has been identified and might be a lot harder to debug/recall in two years time.

So to that end, I've split up the `is_targetted_token()` method and changed how the value to be returned is determined. See the new commit I've just added

PR #822, opened 2017-01-27
Bug fix: `is_sanitized()` did not consistently take all sanitization functions into account for final result

jrfnl:
The `$sanitizingFunctions` and `$unslashingSanitizingFunctions` are two distinct lists and custom sanitizing / unslashsanitizing functions are only merged with one or the other.

The `$sanitizingFunctions` list currently contained an (incomplete) overlap with the `$unslashingSanitizingFunctions` which would become even more incomplete if/when custom functions would be merged in.

The `$unslashingSanitizingFunctions` was only considered to determine whether or not unslashing was needed.
Functions contained in that list which were not *also* contained in the `$sanitizingFunctions` list were not taken into account for the final result of the `is_sanitized()` function.

This has now been fixed.

* Removed overlap between the functions listed in the two properties
* `is_sanitized()` now also takes the `$unslashingSanitizingFunctions` into account when doing the final determination of whether something is sanitized or not.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -229,22 +229,24 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	/**
 	 * Functions that sanitize values.
 	 *
+	 * This list is complementary to the `$unslashingSanitizingFunctions`
+	 * list.
+	 * Sanitizing functions should be added to this list if they do *not*
+	 * implicitely unslash data and to the `$unslashingsanitizingFunctions`
+	 * list if they do.
+	 *
 	 * @since 0.5.0
 	 *
 	 * @var array
 	 */
 	public static $sanitizingFunctions = array(
 		'_wp_handle_upload'          => true,
-		'absint'                     => true,
 		'array_key_exists'           => true,
 		'esc_url_raw'                => true,
 		'filter_input'               => true,
 		'filter_var'                 => true,
-		'floatval'                   => true,
 		'hash_equals'                => true,
 		'in_array'                   => true,
-		'intval'                     => true,
-		'is_array'                   => true,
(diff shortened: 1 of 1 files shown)


jrfnl:
@JDGrimes It's fine - though unnecessary - for the lists to overlap/intersect as long as they are both checked in the `is_sanitized()` function. The problem was caused by the fact that only one of the lists was being checked and that functions which had been added to the lists since where not consistently added to both.
Feels more like the original properties should have had documentation that anything in `unslashsanitize` should *also* be in `sanitize` and that documentation could then have been removed with the current change.

I don't mind adding extra info to the docs, but this feels like it would create more confusion instead of reducing it.

JDGrimes:
@jrfnl I don't know how others think about the properties, but I am accustomed to thinking about these two properties as being independent, intersecting lists. The sanitizing functions list contained all of the functions, and the unslashing-sanitizing list contained just those that also implicitly unslashed the data. I just think that people will assume that a `$sanitizingFunctions` property will contain a list of all sanitizing functions. And if they notice the other property, they may think that it is a separate list. I just think that it would be better to be explicit about this in the documentation here, since now the `$sanitizingFunctions` property doesn't contain a list of all sanitizing functions on its own. I realize what you are saying about it no longer making a difference, in practice, whether a function is in both lists or not. But I still think that people could get tripped up in regard to the non-exhaustive nature of this list when they are just looking at the code.

jrfnl:
@JDGrimes Point taken. I've added documentation to this effect to both lists.
Making them explicitly complementary instead of intersecting will hopefully prevent further mishaps of sanitizing not being correctly recognized - as it was now by the lists not being in sync (quite apart from the lists always being out of sync if there would be custom unslashsanitize functions).

Issue #576, opened 2016-06-28
Add sniffs for deprecated WP features


westonruter:
Instead of finding out about a deprecated usage at runtime, we can use static analysis via PHPCS to list deprecated features. We should add sniffs for:

* Deprecated functions (calls to `_deprecated_function`)
* Deprecated constructors (calls to `_deprecated_constructor`)
* Deprecated files (calls to `_deprecated_file`)
* Deprecated hooks (calls to `_deprecated_hook`)

See also https://github.com/gedex/wpdc

JDGrimes:
The only concern that I have is maintainability. If we can't automatically generate the list of deprecated features, then it won't end up getting maintained.

Also, we'd need to discuss what version of WordPress we'd be building the list against. I guess it makes sense to use the latest version, even though some folks using the sniffs may support older versions and need to use the deprecated functions for back-compat. Something to consider is whether to use the latest stable version or trunk, though.

fklein-lu:
This would be very useful for #578 as well.

I consider that one issue with the approach that the WordPress Deprecated Checker uses is that not all functions that are deprecated in Core have been moved to the corresponding file. Additionally deprecated classes won't probably ever be moved, they stay in their own file, and get marked deprecated via Doc block.

On that note, the [PHP Doc Parser](https://github.com/WordPress/phpdoc-parser) used by the WordPress.org developer documentation site has an understanding of what code elements are deprecated, so one could possibly use this to make this code sniff maintainable.

JDGrimes:
> On that note, the PHP Doc Parser used by the WordPress.org developer documentation site has an understanding of what code elements are deprecated, so one could possibly use this to make this code sniff maintainable.

I've hacked on the PHP Doc Parser before, so maybe I'll take a look at that again and see what is possible. I think that the parser portion will probably provide us the information that we need, and then some. The main issue will probably be sifting through all the info that is returned and then getting it into a format that we can use in our sniffs.

As far as that goes, I wonder if we should let the list be stored in a file separate from the sniff itself, so that folks can easily switch to a different list (maybe for a different version of WordPress).

jrfnl:
> I wonder if we should let the list be stored in a file separate from the sniff itself, so that folks can easily switch to a different list (maybe for a different version of WordPress).

That sounds like a great idea!

grappler:
[...]

Issue #823, opened 2017-01-28
get_function_call_parameters() support for dynamic varaibles


grappler:
As I was working on a sniff for deprecated parameter values. I noticed that for strings the start position did not include any white spaces but when using other values like functions, constants, arrays, or properties the start position included whitespaces.

This is test code
```php
echo bloginfo( function_name() );
echo bloginfo( CONSTANT );
echo bloginfo( array( 'home') );
echo bloginfo( $this->value );
```

My code to check the output.
```php
$parameters[ $position ]['raw']
$this->tokens[ $parameters[ $position ]['start'] ]['type']
$this->tokens[ $parameters[ $position ]['start'] ]['content']
```

These are the results

Raw output | "function_name()" | "CONSTANT" | "array( 'home' )" | "$this->value"
------------ | ------------- | ------------- | ------------- | -------------
Code type for start position | "T_WHITESPACE" | "T_WHITESPACE" | "T_WHITESPACE" | "T_WHITESPACE"
Content for start position | " " | " " | " " | " "

What I expected 

Raw output | "function_name()" | "CONSTANT" | "array( 'home' )" | "$this->value"
------------ | ------------- | ------------- | ------------- | -------------
Code type for start position | "T_STRING" | "CONSTANT" | "T_ARRAY" | "T_VARIABLE"
Content for start position | "function_name" | "CONSTANT" | "array" | "$this"

jrfnl:
> I noticed that for strings the start position did not include any white spaces

@grappler It's unclear to me what your issue is.

AFAICS the function does exactly what it says: it gives the start and end position of the parameter and the *trimmed* raw content.
Whitespace is not disregarded when determining the start/end position as the output of the function should also be usable to do code style checks on parameters. All token types inside parameters are treated the same, string or no string token type.

grappler:
@jrfnl Sorry, I made a mistake, the white space is included if there is only a string too.

I have created a PR which should fix this for all types of data. I find that the start point should be from the point the parameter starts and not after the bracket opens or after the comma.

jrfnl:
@grappler I do not agree and strongly object to your PR.

If all you need to check is the content of a param, you have the `raw` value already.
If you need to check for specific token types or combinations, you should **always** walk the tokens using the `start` and `end` values. That's what they are for.
Having accurate start and end tokens available makes for more efficient and accurate results using this function than walking everything between brackets in various inconsistent ways across different sniffs.

[...]

PR #826, opened 2017-01-28
Check for deprecated parameters

grappler:
Solves part of #576 

I got the list of deprecated parameters by looking where [`_deprecated_argument()`](https://developer.wordpress.org/reference/functions/_deprecated_argument/) was being used.

This is the first of three sniffs. The other two that will follow are
- `WordPress.WP.DeprecatedParameterValues` e.g. `bloginfo( 'url' );` where the value `url` is deprecated
- `WordPress.WP.DeprecatedParameterArgValues` e.g. For `wp_dropdown_categories( $args ) ` `$args['type']` cannot be `links`.

There will be an notice if the parameter value is not what is used by default. The default value is important as in some cases the deprecated parameter maybe in the middle. e.g. `load_plugin_textdomain()`

If a dynamic parameter is used then there will be still an error. The values for the deprecated functions are normally empty strings, bool or null. It is best to write them as is then using a function or variable.

I have again included a property to define the minimum supported WordPress version. Perhaps for future discussion: Would it make more sense to define the minimum supported WordPress version in the `Sniff.php` instead so it would only need to defined once instead for every deprecated sniff.

There maybe a better way to add the ordinal suffix. I have not added support for 11th, 12th or 13th as that many parameters would be an edge case.

### DIFF
--- a/WordPress/Sniffs/WP/DeprecatedParametersSniff.php
+++ b/WordPress/Sniffs/WP/DeprecatedParametersSniff.php
@@ -1,0 +1,345 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+/**
+ * Check for usage of deprecated parameters in WP functions and suggest alternative based on the parameter passed.
+ *
+ * This sniff will throw an error when usage of deprecated parameters is
+ * detected if the parameter was deprecated before the minimum supported
+ * WP version; a warning otherwise.
+ * By default, it is set to presume that a project will support the current
+ * WP version and up to three releases before.
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   0.12.0
+ */
+class WordPress_Sniffs_WP_DeprecatedParametersSniff extends WordPress_AbstractFunctionParameterSniff {
+
+	/**
+	 * The group name for this group of functions.
+	 *
+	 * @since 0.12.0
+	 *
+	 * @var string
+	 */
+	protected $group_name = 'wp_deprecated_parameters';
+
+	/**
+	 * Minimum WordPress version.
+	 *
+	 * This variable allows changing the minimum supported WP version used by
+	 * this sniff by setting a property in a custom ruleset XML file.
+	 *
+	 * Example usage:
+	 * <rule ref="WordPress.WP.DeprecatedParameters">
+	 *  <properties>
+	 *   <property name="minimum_supported_version" value="4.5"/>
+	 *  </properties>
+	 * </rule>
+	 *
+	 * @since 0.12.0
+	 *
+	 * @var string WordPress version.
+	 */
+	public $minimum_supported_version = 4.5;
+
+	/**
+	 * Array of function, argument, and default value for deprecated argument.
+	 *
+	 * The functions are ordered alphabetically.
+	 * Last updated for WordPress 4.8.0.
+	 *
+	 * @since 0.12.0
+	 *
+	 * @var array Multidimensional array with parameter details.
+	 *    $target_functions = array(
+	 *        (string) Function name. => array(
+	 *            (int) Target parameter position, 1-based. => array(
+	 *                'value'   => (mixed) Expected default value for the
+	 *                              deprecated parameter. Currently the default
+	 *                              values: true, false, null, empty arrays and
+	 *                              both empty and non-empty strings can be
+	 *                              handled correctly by the process_parameters()
+	 *                              method. When an additional default value is
+	 *                              added, the relevant code in the
+	 *                              process_parameters() method will need to be
+	 *                              adjusted.
+	 *                'version' => (int) The WordPress version when deprecated.
+	 *            )
+	 *         )
+	 *    );
+	 */
+	protected $target_functions = array(
+
+		'add_option' => array(
+			3 => array(
+				'value'   => '',
+				'version' => '2.3.0',
+			),
+		),
+		'comments_link' => array(
+			1 => array(
+				'value'   => '',
+				'version' => '0.72',
+			),
+			2 => array(
+				'value'   => '',
+				'version' => '1.3.0',
+			),
+		),
+		'comments_number' => array(
+			4 => array(
+				'value'   => '',
+				'version' => '1.3.0',
+			),
+		),
+		'convert_chars' => array(
+			2 => array(
+				'value'   => '',
+				'version' => '0.71',
+			),
+		),
+		'discover_pingback_server_uri' => array(
+			2 => array(
+				'value'   => '',
+				'version' => '2.7.0',
+			),
+		),
+		'get_category_parents' => array(
+			5 => array(
+				'value'   => array(),
+				'version' => '4.8.0',
+			),
+		),
+		'get_delete_post_link' => array(
+			2 => array(
+				'value'   => '',
+				'version' => '3.0.0',
+			),
+		),
+		'get_last_updated' => array(
+			1 => array(
+				'value'   => '',
+				'version' => '3.0.0', // Was previously part of MU.
+			),
+		),
+		'get_the_author' => array(
+			1 => array(
+				'value'   => '',
+				'version' => '2.1.0',
+			),
+		),
+		'get_user_option' => array(
+			3 => array(
+				'value'   => '',
+				'version' => '2.3.0',
+			),
+		),
+		'get_wp_title_rss' => array(
+			1 => array(
+				'value'   => '&#8211;',
+				'version' => '4.4.0',
+			),
+		),
+		'is_email' => array(
+			2 => array(
+				'value'   => false,
+				'version' => '3.0.0',
+			),
+		),
+		'load_plugin_textdomain' => array(
+			2 => array(
+				'value'   => false,
+				'version' => '2.7.0',
+			),
+		),
+		'safecss_filter_attr' => array(
+			2 => array(
+				'value'   => '',
+				'version' => '2.8.1',
+			),
+		),
+		'the_attachment_link' => array(
+			3 => array(
+				'value'   => false,
+				'version' => '2.5.0',
+			),
+		),
+		'the_author' => array(
+			1 => array(
+				'value'   => '',
+				'version' => '2.1.0',
+			),
+			2 => array(
+				'value'   => true,
+				'version' => '1.5.0',
+			),
+		),
+		'the_author_posts_link' => array(
+			1 => array(
+				'value'   => '',
+				'version' => '2.1.0',
+			),
+		),
+		'trackback_rdf' => array(
+			1 => array(
+				'value'   => '',
+				'version' => '2.5.0',
+			),
+		),
+		'trackback_url' => array(
+			1 => array(
+				'value'   => true,
+				'version' => '2.5.0',
+			),
+		),
+		'update_blog_option' => array(
+			4 => array(
+				'value'   => null,
+				'version' => '3.1.0',
+			),
+		),
+		'update_blog_status' => array(
+			4 => array(
+				'value'   => null,
+				'version' => '3.1.0',
+			),
+		),
+		'update_user_status' => array(
+			4 => array(
+				'value'   => null,
+				'version' => '3.0.2',
+			),
+		),
+		'unregister_setting' => array(
+			4 => array(
+				'value'   => '',
+				'version' => '4.7.0',
+			),
+		),
+		'wp_get_http_headers' => array(
+			2 => array(
+				'value'   => false,
+				'version' => '2.7.0',
+			),
+		),
+		'wp_get_sidebars_widgets' => array(
+			1 => array(
+				'value'   => true,
+				'version' => '2.8.1',
+			),
+		),
+		'wp_install' => array(
+			5 => array(
+				'value'   => '',
+				'version' => '2.6.0',
+			),
+		),
+		'wp_new_user_notification' => array(
+			2 => array(
+				'value'   => null,
+				'version' => '4.3.1',
+			),
+		),
+		'wp_notify_postauthor' => array(
+			2 => array(
+				'value'   => null,
+				'version' => '3.8.0',
+			),
+		),
+		'wp_title_rss' => array(
+			1 => array(
+				'value'   => '&#8211;',
+				'version' => '4.4.0',
+			),
+		),
+		'wp_upload_bits' => array(
+			2 => array(
+				'value'   => '',
+				'version' => '2.0.0',
+			),
+		),
+		'xfn_check' => array(
+			3 => array(
+				'value'   => '',
+				'version' => '2.5.0',
+			),
+		),
+
+	); // End $target_functions.
+
+	/**
+	 * Process the parameters of a matched function.
+	 *
+	 * @since 0.12.0
+	 *
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		$paramCount = count( $parameters );
+		foreach ( $this->target_functions[ $matched_content ] as $position => $parameter_args ) {
+
+			// Check that number of parameters defined is not less than the position to check.
+			if ( $position > $paramCount ) {
+				break;
+			}
+
+			// The list will need to updated if the default value is not supported.
+			switch ( $parameters[ $position ]['raw'] ) {
+				case 'true':
+					$matched_parameter = true;
+					break;
+				case 'false':
+					$matched_parameter = false;
+					break;
+				case 'null':
+					$matched_parameter = null;
+					break;
+				case 'array()':
+				case '[]':
+					$matched_parameter = array();
+					break;
+				default:
+					$matched_parameter = $this->strip_quotes( $parameters[ $position ]['raw'] );
+					break;
+			}
+
+			if ( $parameter_args['value'] === $matched_parameter ) {
+				continue;
+			}
+
+			$message = 'The parameter "%s" at position #%s of %s() has been deprecated since WordPress version %s.';
+			$is_error = version_compare( $parameter_args['version'], $this->minimum_supported_version, '<' );
+			$code = $this->string_to_errorcode( ucfirst( $matched_content ) . 'Param' . $position . 'Found' );
+
+			$data = array(
+				$parameters[ $position ]['raw'],
+				$position,
+				$matched_content,
+				$parameter_args['version'],
+			);
+
+			if ( isset( $parameter_args['value'] ) && $position < $paramCount ) {
+				$message .= ' Use "%s" instead.';
+				$data[]   = (string) $parameter_args['value'];
+			} else {
+				$message .= ' Instead do not pass the parameter.';
+			}
+
+			$this->addMessage( $message, $stackPtr, $is_error, $code, $data, 0 );
+		}
+	}
+
+}
--- a/WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
+++ b/WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
@@ -1,0 +1,66 @@
+<?php
+
+// All will be OK as the default value is used.
+
+wp_title_rss( '&#8211;' ); // 1st.
+wp_get_sidebars_widgets( true ); // 1st.
+wp_new_user_notification( '', null, '' ); // 2nd.
+the_attachment_link( '', false, false, false ); // 3rd.
+update_blog_option( '', '', '', null ); // 4th.
+wp_install( '', '', '', '', '' ); // 5th.
+get_category_parents( '', '', '', '', array() ); // 5th.
+get_category_parents( '', '', '', '', [] ); // 5th.
+
+// Method names within a class should be fine.
+
+Theme_Object::wp_title_rss( 'home' ); // Ok.
+$this->wp_title_rss( 'siteurl' ); // Ok.
+$theme_object->wp_title_rss( 'text_direction' ); // Ok.
+
+// All will give an Error even though they have a dynamic variable.
+
+wp_new_user_notification( '', $variable );
+wp_new_user_notification( '', function_name() );
+wp_new_user_notification( '', $this->method_name() );
+
+// All will give an ERROR. The functions are ordered alphabetically.
+
+add_option( '', '', [] );
+add_option( '', '', 1.23 );
+add_option( '', '', 10 );
+add_option( '', '', false );
+add_option( '', '', 'deprecated' );
+comments_link( 'deprecated', 'deprecated' );
+comments_number( '', '', '', 'deprecated' );
+convert_chars( '', 'deprecated' );
+discover_pingback_server_uri( '', 'deprecated' );
+get_delete_post_link( '', 'deprecated' );
+get_last_updated( 'deprecated' );
+get_the_author( 'deprecated' );
+get_user_option( '', '', 'deprecated' );
+get_wp_title_rss( 'deprecated' );
+is_email( '', 'deprecated' );
+is_email( '', 'false' ); // False as a string not bool.
+load_plugin_textdomain( '', 'deprecated' );
+safecss_filter_attr( '', 'deprecated' );
+the_attachment_link( '', '', 'deprecated' );
+the_author( 'deprecated', 'deprecated' );
+the_author_posts_link( 'deprecated' );
+trackback_rdf( 'deprecated' );
+trackback_url( 'deprecated' );
+update_blog_option( '', '', '', 'deprecated' );
+update_user_status( '', '', '', 'deprecated' );
+wp_get_http_headers( '', 'deprecated' );
+wp_get_sidebars_widgets( 'deprecated' );
+wp_install( '', '', '', '', 'deprecated' );
+wp_new_user_notification( '', 'deprecated' );
+wp_notify_postauthor( '', 'deprecated' );
+wp_notify_postauthor( '', 'null' ); // Null as a string not null.
+wp_title_rss( 'deprecated' );
+wp_upload_bits( '', 'deprecated' );
+xfn_check( '', '', 'deprecated' );
+
+// All will give an WARNING as they have been deprecated after WP 4.5.
+
+get_category_parents( '', '', '', '', array( 'deprecated') );
+unregister_setting( '', '', '', 'deprecated' );
--- a/WordPress/Tests/WP/DeprecatedParametersUnitTest.php
+++ b/WordPress/Tests/WP/DeprecatedParametersUnitTest.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * Unit test class for WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+/**
+ * Unit test class for the DeprecatedParameters sniff.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @since   0.12.0
+ */
+class WordPress_Tests_WP_DeprecatedParametersUnitTest extends AbstractSniffUnitTest {
+
+	/**
+	 * Returns the lines where errors should occur.
+	 *
+	 * @return array <int line number> => <int number of errors>
+	 */
+	public function getErrorList() {
+		$errors = array_fill( 28, 34, 1 );
+
+		$errors[22] = 1;
+		$errors[23] = 1;
+		$errors[24] = 1;
+
+		// Override number of errors.
+		$errors[33] = 2;
+		$errors[47] = 2;
+
+		return $errors;
+	}
+
+	/**
+	 * Returns the lines where warnings should occur.
+	 *
+	 * @return array <int line number> => <int number of warnings>
+	 */
+	public function getWarningList() {
+		return array(
+			65 => 1,
+			66 => 1,
+		);
+	}
+
+} // End class.
--- a/WordPress-Extra/ruleset.xml
+++ b/WordPress-Extra/ruleset.xml
@@ -64,6 +64,7 @@
 	</rule>
 	<rule ref="WordPress.WP.DeprecatedFunctions"/>
 	<rule ref="WordPress.WP.DeprecatedClasses"/>
+	<rule ref="WordPress.WP.DeprecatedParameters"/>
 	<rule ref="WordPress.WP.AlternativeFunctions"/>
 	<rule ref="WordPress.WP.DiscouragedFunctions"/>
 


jrfnl:
> Would it make more sense to define the minimum supported WordPress version in the Sniff.php instead so it would only need to defined once instead for every deprecated sniff.

That would make no difference as properties have to be defined per sniff and cannot be defined for all sniffs in a standard in one go via a ruleset. The only difference it would make is that that property would then be available to be set for *all* sniffs, while only a few use it, so I think it could lead to confusion.

IMHO it would be best to keep the property in the sniff class which actually uses it, but use the same name for the property everywhere. In the wiki page for properties (#614), these would then all be grouped together and intuitive to use.
We could also consider allowing that particular property to also be set from the command-line, similar to what was implemented for the I18n sniff and allow the command line value to be valid for all sniffs which use the property.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+/**
+ * Check for usage of deprecated parameters in WP functions and provide alternative based on the parameter passed.
```
jrfnl: `provide` => `suggest`

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+	 * @var string
+	 */
+	protected $group_name = 'wp_deprecated_parameters';
+
+	/**
+	 * Minimum WordPress version.
+	 *
+	 * This sniff will throw an error when usage of deprecated parameters is
+	 * detected if the parameter was deprecated before the minimum supported
+	 * WP version; a warning otherwise.
```
jrfnl: This first line seems more like a "long description" for the sniff itself, so might be better placed in the class doc.
Possibly the second line too.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+	 * By default, it is set to presume that a project will support the current
+	 * WP version and up to three releases before.
+	 * This variable allows changing the minimum supported WP version used by
+	 * this sniff by setting a property in a custom phpcs.xml ruleset.
```
jrfnl: A ruleset does not need to be named `phpcs.xml`

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+	 * Example usage:
+	 * <rule ref="WordPress.WP.DeprecatedParameters">
+	 *  <properties>
+	 *   <property name="minimum_supported_version" value="4.4"/>
+	 *  </properties>
+	 * </rule>
+	 *
+	 * @since 0.11.0
+	 *
+	 * @var string WordPress versions.
```
jrfnl: `versions` => `version`

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+	 * @since 0.11.0
+	 *
+	 * @var array <string function name> => <array target parameter positions>
+	 */
+	protected $target_functions = array(
+
+		'get_last_updated' => array(
+			1 => array(
+				'value'   => '',
+				'version' => '0.0',
```
jrfnl: version 0.0 ?

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+	 * @var string WordPress versions.
+	 */
+	public $minimum_supported_version = 4.4;
+
+	/**
+	 * Array of function, argument, and replacement function for deprecated argument.
+	 *
+	 * @since 0.11.0
+	 *
+	 * @var array <string function name> => <array target parameter positions>
```
jrfnl: Just like you've suggested yourself elsewhere, you may want to consider extending the property array documentation to contain more specific information about the sub-array.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+	 */
+	protected $target_functions = array(
+
+		'get_last_updated' => array(
+			1 => array(
+				'value'   => '',
```
jrfnl: Does this mean the value should be a blank string or nothing ? or is the value not relevant in this case ? And if so, should the value be set in this array at all if that's the case ?

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+				'version' => '1.3',
+			),
+		),
+		'the_author' => array(
+			1 => array(
+				'value'   => '',
+				'version' => '2.1',
+			),
+			2 => array(
+				'value'   => 'true',
```
jrfnl: Is this the string `true` or boolean `true` ? If you do a string based compare for a boolean value, you may well create false negatives. The variable type is important. See below.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+	 *
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		foreach ( $this->target_functions[ $matched_content ] as $position => $paramerter_args ) {
```
jrfnl: Spelling: `$paramerter_args` => `$parameter_args` (also whenever the variable is used below)

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+
+			if ( ! isset( $parameters[ $position ] ) ) {
+				continue;
+			}
+
+			$is_dynamic_parameter = $this->phpcsFile->findNext( array( T_VARIABLE, T_STRING ), $parameters[ $position ]['start'], $parameters[ $position ]['end'], false, null, true );
```
jrfnl: You need to use `( $parameters[ $position ]['end'] + 1)` for the end position to prevent running into a bug in PHPCS. The `findNext()` function looks for tokens up to - but **not** including the `$end` token.
See https://github.com/squizlabs/PHP_CodeSniffer/issues/1319

jrfnl: What are the tokens for deciding whether something is a "dynamic variable" based on ?

Just thinking out loud, I can imagine `T_STRING_VARNAME`, `T_METHOD_C`, anything with an operator in it etc to be considered dynamic. Would a list of non-dynamic types not make more sense / be smaller ? i.e. string types, numeric types, boolean types and null

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+
+			$matched_parameter = $this->strip_quotes( $parameters[ $position ]['raw'] );
+
+			if ( ! $is_dynamic_parameter && $paramerter_args[ 'value' ] === $matched_parameter ) {
```
jrfnl: If you do a comparison on value, the type of the parameter passed is important.

If boolean `false` is expected and string `false` is passed, according to this check, everything would be fine, while in reality string `false` would translate to boolean `true` which would be an invalid value.

 Also, `null` should always be considered valid as it will just fall through to the parameter default.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+				continue;
+			}
+
+			// 11th, 12th and 13th are not supported as they are an edge case.
+			switch ( $position ) {
```
jrfnl: While nice from a user friendliness perspective, this is inconsequential to the sniff itself. I wouldn't bother with this and would rewrite the error message to either not mention the parameter position, just its name - though that would require adding the names of all the parameters being checked to the above array - òr to read like `parameter at position #`.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+			$message = 'The %s parameter [%s] of %s() has been deprecated since WordPress version %s.';
+			$data    = array(
+				$pos_text,
+				$parameters[ $position ]['raw'],
+				$matched_content,
+				$paramerter_args['version'],
+			);
+
+			if ( isset( $paramerter_args['value'] ) ) {
+				$message .= ' Use "%s" instead.';
```
jrfnl: Suggestion:
* If the parameter is the last parameter passed for the function, suggest not passing the parameter instead.
* In all other cases, suggest passing `null` instead as that will automatically fall back to the default used by the function and will potentially reduce sniff maintenance (and code maintenance for devs using the sniff) as it removes the need to sync the defaults with whatever is happening in core.

WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
```diff
@@ -0,0 +1,53 @@
+<?php
+
+// all will give an ERROR.
```
jrfnl: Comments should start with a capital.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,317 @@
...
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		foreach ( $this->target_functions[ $matched_content ] as $position => $paramerter_args ) {
+
+			if ( ! isset( $parameters[ $position ] ) ) {
```
jrfnl: It might be an idea to add a `$paramCount = count( $parameters );` before the loop and to add a `if ( $position > $paramCount) { break; }` above this line (or instead of this line) for efficiency.
After all, no need to continue looping through for parameters which aren't set anyway.

WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
```diff
@@ -0,0 +1,53 @@
...
+is_email( '', 'deprecated' );
+update_user_status( '', '', '', 'deprecated' );
+update_blog_option( '', '', '', 'deprecated' );
+wp_notify_postauthor( '', 'deprecated' );
+wp_new_user_notification( '', 'deprecated' );
+wp_new_user_notification( '', $variable );
+wp_new_user_notification( '', function_name() );
+
+
+// all will give an WARNING.
```
jrfnl: Dito

WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
```diff
@@ -0,0 +1,53 @@
...
+
+get_wp_title_rss( 'deprecated' );
+wp_title_rss( 'deprecated' );
+unregister_setting( '', '', '', 'deprecated' );
+
+// all will be OK.
```
jrfnl: Dito

jrfnl: Maybe add some info on why these are ok ? Is it because they have the default value ? or because the params aren't deprecated ? or ... ?

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: The parameter was already deprecated when it was merged into core. https://developer.wordpress.org/reference/functions/get_last_updated/

jrfnl: In that case, I suggest setting it to the version nr in which it was merged into core or to `MU` as may be the case.
Goes for the other function(s) with version `0.0` as well.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: > If boolean false is expected and string false is passed, according to this check, everything would be fine, while in reality string false would translate to boolean true which would be an invalid value.

This should be fixed now.

> Also, null should always be considered valid as it will just fall through to the parameter default.

I tested this and it did not work for me http://sandbox.onlinephpfunctions.com/code/157fe24a88a2bbf0ec7fc981a56e0a5f2f8f3e18

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: The default value is an empty string.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: Turns out `xfn_check()` was deprecated in version 2.5. The version in core was incorrect. https://core.trac.wordpress.org/ticket/39737

grappler:
Thank you @jrfnl for the review. The issues you mentioned should be fixed now.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,326 @@
...
+			);
+
+			if ( isset( $parameter_args['value'] ) && $position < $paramCount ) {
+				$message .= ' Use "%s" instead.';
+				$data[]   = (string) $parameter_args['value'];
+			} else {
+				$message .= ' Instead do not pass the parameter.';
+			}
+
+			if ( version_compare( $parameter_args['version'], $this->minimum_supported_version, '>=' ) ) {
```
jrfnl: Depending on which PR gets merged first, this toggle can be replaced with the `addMessage()` helper function as introduced in PR #827

jrfnl:
@grappler I'll try and have another look later this week. Quick question in the mean time: to which ruleset will this sniff be added ? And shouldn't that change be included with this PR ?

grappler:
We added the other deprecated sniffs just to `Extra`. I can add that to the PR.

jrfnl:
I wouldn't be adverse to both this sniff as well as the one for `WP.DeprecatedFunctions` being added to `Core` as - obviously - core shouldn't be using methods and parameters which it has deprecated, but others might have a different opinion.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
jrfnl: FYI: the `addMessage()` utility function is now available as part of `develop`.

grappler:
I have used the new `addMessage()` utility function. I have added the sniff to the extra ruleset. Will do a separate PR to add it to the core ruleset.

grappler:
Does anyone else have any thoughts on this PR? Don't worry it is not as big as my last one.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,340 @@
...
+
+			if ( $position > $paramCount ) {
+				break;
+			}
+
+			if ( ! isset( $parameters[ $position ] ) ) {
+				continue;
+			}
+
+			$is_dynamic_parameter = $this->phpcsFile->findNext( array( T_CONSTANT_ENCAPSED_STRING, T_ARRAY, T_FALSE, T_TRUE, T_NULL, T_LNUMBER, T_WHITESPACE ), $parameters[ $position ]['start'], ( $parameters[ $position ]['end'] + 1 ), true, null, true );
```
JDGrimes: This function call could really benefit from being multi-lined.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,340 @@
...
+
+			if ( $is_dynamic_parameter ) {
+				$this->phpcsFile->addWarning(
+					'The dynamic parameter [%s] at possition #%s of %s() could possible by deprecated.',
```
JDGrimes: `s/possible by/possibly be/`

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,347 @@
...
+				$parameters[ $position ]['start'],
+				( $parameters[ $position ]['end'] + 1 ),
+				true,
+				null,
+				true
+			);
+
+			if ( $is_dynamic_parameter ) {
+				$this->phpcsFile->addWarning(
+					'The dynamic parameter [%s] at possition #%s of %s() could possibly be deprecated.',
```
GaryJones: Typo: position

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,347 @@
...
+				default:
+					$matched_parameter = $this->strip_quotes( $parameters[ $position ]['raw'] );
+					break;
+			}
+
+			if ( $parameter_args['value'] === $matched_parameter ) {
+				continue;
+			}
+
+			$message = 'The parameter [%s] at possition #%s of %s() has been deprecated since WordPress version %s.';
```
GaryJones: Typo: position

JDGrimes:
@jrfnl @GaryJones It looks like @grappler has fixed all of the issues you pointed out, would you like to take one more look and approve before merging?

jrfnl:
@JDGrimes Yup, this one is high on my priority list, will get to it soon.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,347 @@
...
+	 * @var string WordPress version.
+	 */
+	public $minimum_supported_version = 4.4;
+
+	/**
+	 * Array of function, argument, and replacement function for deprecated argument.
+	 *
+	 * @since 0.11.0
+	 *
+	 * @var array Multi-dimentional array with parameter details.
```
GaryJones: Typo here.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,347 @@
...
+	 *
+	 *               @type string Function name. {
+	 *                   @type int target Parameter positions. {
+	 *                       @type string|boolean|null  Default value.
+	 *                       @type int The WordPress version when deprecated.
+	 *                   }
+	 *               }
+	 */
+	protected $target_functions = array(
```
GaryJones: Could we get these in alphabetical order of function name? Makes it easier to check in the future.

grappler: They are currently sorted in order of version deprecated. @jrfnl @jdgrimes What do you think?

JDGrimes: Hmmm... I generally prefer things in alphabetical order. Like @GaryJones says it makes it easier to check if a function is already there (though of course you could also search). Having them in the order deprecated is a nice feature, but does it actually provide benefit for users/developers? Only things I can think of, is that as we add new functions to the sniff over time, we'll always be adding them at the end (which isn't necessarily any better/easier than adding them in alphabetical order, I guess). But I guess it does make it easier for somebody who is using the sniff to see which functions are going to be flagged based on what WordPress version they support. I can see the potential benefit of that, so I'm fine leaving the list like it is. But I don't have a strong opinion either way.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: Oh, the typo came originally from https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniff.php#L1596

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
GaryJones: I'm fine with the current order then (I hadn't spotted it was in WP version order), but at least add a comment before the array saying what the order is to clarify for future readers and contributors.

grappler: Makes sense, I have done that.

JDGrimes: I think all things considered, we should probably just have this list be in alphabetical order. I don't really see a compelling case for putting it in order of version number, so for simplicity and consistency alphabetical order seems best.

jrfnl:
@grappler Would you mind rebasing the PR to get rid of the conflict ? Thanks!

grappler:
@jrfnl The conflict has been resolved.

jrfnl:
@grappler Thanks for solving the merge conflict. Travis is failing now though on a superfluous empty line (two empties after each other on line 340)

jrfnl:
@JDGrimes Is your deprecation check tool capable of checking the list of parameters in this PR ? If not, we'll leave that for a future update round.

JDGrimes:
> @JDGrimes Is your deprecation check tool capable of checking the list of parameters in this PR ? If not, we'll leave that for a future update round.

Not yet. I haven't yet tried to figure out how to determine which parameter is deprecated from a `_depregated_argument()` call. It isn't as simple as `_deprecated_function()`, because the paramter itself isn't specified, except possibly within the message. So I think we'll have to wait on that for the next round.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,352 @@
...
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+/**
+ * Check for usage of deprecated parameters in WP functions and sugests alternative based on the parameter passed.
```
jrfnl: `sugests` => `suggests`

WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
```diff
@@ -0,0 +1,60 @@
...
+wp_new_user_notification( '', 'deprecated' );
+get_wp_title_rss( 'deprecated' );
+wp_title_rss( 'deprecated' );
+
+// All will give an WARNING as they have been deprecated after WP 4.5.
+
+unregister_setting( '', '', '', 'deprecated' );
+get_category_parents( '', '', '', '', array( 'deprecated') );
+
+// All will give an WARNING as they have a dynamic variable.
```
jrfnl: For practical purposes, I would like to suggest that the tests from this line to the end, be moved to the top of the test file as these tests won't change.

The version based ones above, will then be below them which will make it easier to add new tests for newly deprecated parameters + to move tests up from the warning list to the error list when a new WP version has been released without having to make drastic changes to the unit test line number arrays.

WordPress/Tests/WP/DeprecatedParametersUnitTest.php
```diff
@@ -0,0 +1,73 @@
...
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+/**
+ * Unit test class for the DeprecatedParameters sniff.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @since   0.11.0
```
jrfnl: Dito: `0.12.0`

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,352 @@
...
+	 *
+	 * @var string WordPress version.
+	 */
+	public $minimum_supported_version = 4.5;
+
+	/**
+	 * Array of function, argument, and default value for deprecated argument.
+	 *
+	 * The functions are ordered by the WordPress version the parameter was
+	 * deprecated in. Was last updated in WordPress version 4.8.0.
```
jrfnl: This comment looks to be out of date.

jrfnl: > Was last updated in WordPress version 4.8.0.

Suggestion: _Was last updated _for_ WordPress version 4.8.0._ or _Last updated based on WordPress version 4.8.0._

`in` does not sound right here. /cc @GaryJones

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,352 @@
...
+	 *
+	 * @since 0.12.0
+	 *
+	 * @var array Multidimensional array with parameter details.
+	 *
+	 *               @type string Function name. {
+	 *                   @type int target Parameter positions. {
+	 *                       @type string|boolean|null  Default value.
```
jrfnl: Should these two mention the array key they should use ?

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,352 @@
...
+	 *
+	 * The functions are ordered by the WordPress version the parameter was
+	 * deprecated in. Was last updated in WordPress version 4.8.0.
+	 *
+	 * @since 0.12.0
+	 *
+	 * @var array Multidimensional array with parameter details.
+	 *
+	 *               @type string Function name. {
+	 *                   @type int target Parameter positions. {
```
jrfnl: `positions` => `position`

jrfnl: What does the `target` signify in this context ? Is that part of the short description ? If so, it should be capitalized and `Parameter` should not be.

jrfnl: Also: It might be helpful to mention that the parameter positions are 1-based (in contrast to the more common 0-based).

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,352 @@
...
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		$paramCount = count( $parameters );
+		foreach ( $this->target_functions[ $matched_content ] as $position => $parameter_args ) {
+
+			if ( $position > $paramCount ) {
+				break;
+			}
+
+			if ( ! isset( $parameters[ $position ] ) ) {
```
jrfnl: Is this `if(){}` still necessary ? Doesn't the `if` above it prevent it from ever being triggered ?
Also: if it *is* still necessary, should this be a `continue` or a `break` ?

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
jrfnl: Regarding the default value, I think the `type` should be `mixed` as there is also `array()`.

Secondly, would it be an idea to list what the function below accounts for (`true`, `false`, `null`, `array()` and strings) and to add a note that if a different type of default value is encountered and added to the list, the function below should be adjusted to account for it ?
Think int `1` ?

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,352 @@
...
+					$matched_parameter = $this->strip_quotes( $parameters[ $position ]['raw'] );
+					break;
+			}
+
+			if ( $parameter_args['value'] === $matched_parameter ) {
+				continue;
+			}
+
+			$is_dynamic_parameter = $this->phpcsFile->findNext(
+				array( T_CONSTANT_ENCAPSED_STRING, T_ARRAY, T_FALSE, T_TRUE, T_NULL, T_LNUMBER, T_WHITESPACE ),
```
jrfnl: As this is quite a long array and it doesn't change during the loops, would it make sense to add it as a class property instead of redeclaring it every time ?

jrfnl: What about `T_OPEN_SHORT_ARRAY`, `T_CLOSE_SHORT_ARRAY`, `T_DNUMBER` (floats), `T_START_NOWDOC`, `T_NOWDOC` and `T_END_NOWDOC` ?
And maybe the comment tokens ?

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,352 @@
...
+				$parameters[ $position ]['start'],
+				( $parameters[ $position ]['end'] + 1 ),
+				true,
+				null,
+				true
+			);
+
+			if ( $is_dynamic_parameter ) {
+				$message = 'The dynamic parameter [%s] at position #%s of %s() could possibly be deprecated.';
```
jrfnl: ~~I think the message text here is unclear. The dynamic parameter at that position **_is_** deprecated, but it couldn't be determined whether the correct default value was used.~~

~~The associated errorcode should also be adjusted to reflect this.~~

On second thought: does it really matter ? Shouldn't this always trigger an `error` ? - even independently of the minimum WP version - as deprecated parameters, if passed, should always have a hard-coded default value equal to the expected default in WP (or not be passed at all) ?

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,352 @@
...
+				$is_error = false;
+				$code = $this->string_to_errorcode( ucfirst( $matched_content ) . 'PossiblyDeprecated' );
+			} else {
+				$message = 'The parameter [%s] at position #%s of %s() has been deprecated since WordPress version %s.';
```
jrfnl: I'd like to suggest adding quotes around the parameter value as it could make for really weird sentences otherwise:
`The parameter "[%s]"...`

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,352 @@
...
+				$is_error = version_compare( $parameter_args['version'], $this->minimum_supported_version, '<' );
+				$code = $this->string_to_errorcode( ucfirst( $matched_content ) . 'Found' );
```
jrfnl: Should the errorcode maybe include a ref to the param position to make it unique to that param ? Something like `ucfirst( $matched_content ) . 'Param' . $position . 'Found'`.

Otherwise, excluding the errorcode for a function will invalidate messages for *all* deprecated parameters for that function, even ones deprecated after someone has added the exclusion to their ruleset.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,352 @@
...
+
+			if ( isset( $parameter_args['value'] ) && $position < $paramCount ) {
+				$message .= ' Use "%s" instead.';
+				$data[]   = (string) $parameter_args['value'];
+			} else {
+				$message .= ' Instead do not pass the parameter.';
+			}
+
+			$this->addMessage( $message, $stackPtr, $is_error, $code, $data, 0 );
+		} // End foreach().
```
jrfnl: The `// End foreach()` can be removed, the requirement for it has been removed from the Core standard.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
GaryJones:
```
The functions are ordered by the WordPress version the parameter was deprecated in. Last updated for WordPress 4.8.0.
```

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: The reason for doing it a warning is that we could not be sure of the dynamic value. If the default value was an empty array and `_return_empty_array()` was used instead of `array()`. This would be still be valid.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: I think we should use either the brackets or the quotation marks otherwise we get: `The parameter "[1.234]" at position #2 of wp_new_user_notification() has been deprecated since WordPress version 4.3.1. Instead do not pass the parameter.`

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: The error code would be now `WordPress.WP.DeprecatedParameters.Wp_new_user_notificationParam2Found`

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: I have corrected this to `Target parameter position starting with 1.`

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: I have mentioned the array keys but let me know what you think of the new format.

grappler: Good idea. I have also added a comment to the switch code.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: Yes, it is not necessary, I have removed it.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: What about `protected $static_token_types = array( T_CONSTANT_ENCAPSED_STRING, T_ARRAY, T_FALSE, T_TRUE, T_NULL, T_LNUMBER, T_WHITESPACE, T_OPEN_SHORT_ARRAY, T_CLOSE_SHORT_ARRAY, T_DNUMBER, T_START_NOWDOC, T_NOWDOC, T_END_NOWDOC );`?

grappler: I don't think comment tokens are needed as comments are not a valid parameter.

grappler:
Thank you @jrfnl for your review. I have updated most of the items. Once we have clarified the last few items and the new changes look good I will make another commit.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
jrfnl: > Target parameter position starting with 1

I think the new phrase could be confusing as it is now. If I'd read this (and didn't know better) I would expect all the arrays to start with the `1` target parameter position as being deprecated.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
jrfnl: > (mixed) Default value. Currently boolean, null, array and strings are supported. Support for additional types can be added to $this->process_parameters().

I'd suggest breaking this long comment up into several lines and maybe making it a little more specific. Something along the lines of:
> (mixed) Expected default value for the deprecated parameter.
> Currently the default values: `true`, `false`, `null`, empty arrays and both empty and non-empty strings can be handled correctly by the `process_parameters()` method.
> If/when an additional default value would be added, the relevant code in the `process_parameters()` method will need to be adjusted.

Note: this phrasing can be definitely still be improved - @GaryJones suggestions ?

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
jrfnl: Yes, that's what I mean, though I would declare it as a multi-line array for readability.

jrfnl: So, how would code like this be handled:
```php
add_option( 'my_option', 'my_value', /* deprecated */ '', false );
```

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
jrfnl: That doesn't really address my comment, as calling `_return_empty_array()` from within a function call to pass an empty array to a deprecated parameter seems bad practice to me anyway and the parameter **is** deprecated, in contrast to _could be_.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
jrfnl: Good point, I'd overlooked the brackets.
All the same, everywhere else in these kind of situations, quotes are used. I don't remember brackets ever being used like that in the error messages so maybe replace the brackets with quotes in that case ? Or was there a specific reason why you used brackets ?

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -53,19 +53,21 @@ class WordPress_Sniffs_WP_DeprecatedParametersSniff extends WordPress_AbstractFu
...
 	 * @var array Multidimensional array with parameter details.
+	 *    $target_functions = array(
+	 *        (string) Function name. => array(
+	 *            (int) Target parameter position starting with 1. => array(
+	 *                'value'   => (mixed) Default value. Currently boolean, null, array and strings are supported. Support for additional types can be added to $this->process_parameters().
+	 *                'version' => (int) The WordPress version when deprecated.
+	 *            )
+	 *         )
+	 *    );
 	 *
```
jrfnl: PHPCS throws a notice for a superfluous empty comment line at the end here.

WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
jrfnl: Nearly there for the test case file: the `All will give an WARNING as they have been deprecated after WP 4.5.` ones should be below the `All will give an ERROR` for future additions/moving between lists to have the least impact.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: > The dynamic parameter [/* deprecated */ 'deprecated'] at position #2 of wp_new_user_notification() could possibly be deprecated.

The warning would be this. I see what you mean.

grappler: Regardless for the check if the parameter is dynamic the code will fail if a comment in placed with the parameters. Do we want to support this code format?

```php
wp_new_user_notification( '', /* deprecated */ 'deprecated' );
wp_new_user_notification( '', null /* deprecated */ );
```
`$parameters[ $position ]['raw']` returns `/* deprecated */ 'deprecated'`

Like I raised in #823 parameters return everything between the two commas instead of just the parameter.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: I will remove the check for dynamic variables.

grappler:
@jrfnl I have made a few more changes and add some replies to your comments.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
jrfnl: And like I replied in #823 - that's because you are supposed to walk the token array using the parameter `start` and `end` values you receive and not to presume that a parameter is one token.

> Do we want to support this code format?

This can be added later, but yes, I do think supporting this would be a good thing as this type of sniff should be code-style independent and therefore not have an opinion on whether or not comments are allowed as part of the parameter.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,349 @@
...
+	 *
+	 * @var string
+	 */
+	protected $group_name = 'wp_deprecated_parameters';
+
+	/**
+	 * Minimum WordPress version.
+	 *
+	 * This variable allows changing the minimum supported WP version used by
+	 * this sniff by setting a property in a custom ruleset xml file.
```
GaryJones: xml => XML

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,349 @@
...
+				'version' => '0.71',
+			),
+		),
+		'discover_pingback_server_uri' => array(
+			2 => array(
+				'value'   => '',
+				'version' => '2.7.0',
+			),
+		),
+		'get_delete_post_link' => array(
```
GaryJones: If this list is meant to be in alphabetical order, then `get_delete_post_link` should come after `get_category_parents`.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,349 @@
...
+				'version' => '2.5.0',
+			),
+		),
+		'trackback_url' => array(
+			1 => array(
+				'value'   => true,
+				'version' => '2.5.0',
+			),
+		),
+		'update_user_status' => array(
```
GaryJones: The order of these three `update_*` keys is not alphabetical.

GaryJones:
3 minor fixes if possible, but otherwise good to go.

grappler:
Thanks for spotting those. I have updated the PR.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,349 @@
...
+		'unregister_setting' => array(
+			4 => array(
+				'value'   => '',
+				'version' => '4.7.0',
+			),
+		),
+		'wp_get_http_headers' => array(
+			2 => array(
+				'value'   => false,
+				'version' => '2.7',
```
GaryJones: Does this value require being exactly `2.7` (as opposed to `2.7.0`)? If so, maybe worth adding a comment so it's not accidentally "fixed" in the future?

grappler: Not that I think so, it should be `2.7.0`

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,349 @@
...
+	 *  <properties>
+	 *   <property name="minimum_supported_version" value="4.5"/>
+	 *  </properties>
+	 * </rule>
+	 *
+	 * @since 0.12.0
+	 *
+	 * @var string WordPress version.
+	 */
+	public $minimum_supported_version = 4.5;
```
GaryJones: Each new WP release will require that this is updated, if the "last four major versions" gap is to be maintained.

Do we have other code that is based on a value of "$release - X"?

WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
```diff
@@ -0,0 +1,64 @@
+<?php
+
```
GaryJones: Are these groups in order of release? i.e. functions with deprecated params in 4.9 can be added to the end of the relevant groups? If so, maybe add a comment here to remind future contributors.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: Yes, all of the deprecated Sniffs work with the calculation.

WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
grappler: Yes the idea would be for future functions to be added at the end.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
GaryJones: I wonder if there's value in having a centralised `$current_version`, and then having the properties for the relevant Sniffs be calculated from there (not needed for this PR - just thinking allowed).

WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
GaryJones: OK, so just a commented needed here then please.

(I was tempted to suggest putting these functions in alphabetical order, but then remembered seeing discussion about the better order - a comment would help tell others what the logic is.)

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,349 @@
...
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		$paramCount = count( $parameters );
+		foreach ( $this->target_functions[ $matched_content ] as $position => $parameter_args ) {
+
+			// Check that number of parameters defined is not less than the position to check.
+			if ( $position > $paramCount ) {
+				break;
+			}
+
+			// The list will need to updated if the default value is not supported.
```
JDGrimes: This comment should be moved down, right about the `switch` statement. It is confusing where it is.

jrfnl: @grappler Didn't you remove this condition earlier based on a previous comment I made ? It's superfluous. (And removing it would sort the comment out as well ;-) )

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: a centralised `$current_version` would a good idea.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: Yes, I removed it. It slipped in again. Sorry.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
grappler: @jrfnl I just realized why I used the brackets. It was so that we did not have a double bracket e.g. `The parameter "'deprecated'" at position #2 of`

WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
grappler: I have ordered the functions in the test in alphabetical order.

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
jrfnl: I see what you mean, but I still don't think the square brackets are the solution. Quotes like you have now should be fine. If needs be, would could potentially change them to backticks or something later.

grappler: I have changed it again. I suppose if we walked through the tokens then the single quotes would not be included

jrfnl:
@JDGrimes I think all concerns have been addressed. Would like to take another look or can I dismiss your earlier review ?

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
```diff
@@ -0,0 +1,345 @@
...
+				default:
+					$matched_parameter = $this->strip_quotes( $parameters[ $position ]['raw'] );
+					break;
+			}
+
+			if ( $parameter_args['value'] === $matched_parameter ) {
+				continue;
+			}
+
+			$message = 'The parameter "s%" at position #%s of %s() has been deprecated since WordPress version %s.';
```
jrfnl: `s%` should be `%s`

WordPress/Sniffs/WP/DeprecatedParametersSniff.php
jrfnl: Actually, the would be as they are part of the token content for T_CONSTANT_ENCAPSED_STRING tokens.

jrfnl:
:tada: :champagne: :taco:

WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
GaryJones: > I have ordered the functions in the test in alphabetical order.

That wasn't my intent here - future functions, unless they start with `z`, will not be at the end. The grouped nature means that new functions added in the first group will naturally means all of the error line numbers below need to be updated anyway though.

GaryJones:
Well done and thank you @grappler :-)

PR #827, opened 2017-01-29
New utility function: addMessage()

jrfnl:
This PR addresses duplicate code related to toggling between throwing errors and warnings.
Related to https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/795#discussion_r96204333

It adds two helper functions `addMessage()` and `addFixableMessage()` (and one private grunt function).
Both functions take the same arguments as the PHPCS native `addError()`/`addWarning()` methods with one difference: an extra boolean `$is_error` parameter between the `$stackPtr` and the `$errorcode`.

The PR, of course, also implements the use of the helper functions removing duplicate code.

To use these functions, sniffs have to extend the `WordPress_Sniff`. Three of the sniffs involved did not yet do so. They now do & I've leveraged the fact that they do by letting them defer to using the `$phpcsFile` and `$tokens` properties instead of setting these within each function and/or passing these between functions.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -550,6 +550,76 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	}
 
 	/**
+	 * Add a PHPCS message to the output stack as either a warning or an error.
+	 *
+	 * @since 0.11.0
+	 *
+	 * @param string $message   The message.
+	 * @param int    $stackPtr  The position of the token the message relates to.
+	 * @param bool   $is_error  Optional. Whether to report the message as an 'error' or 'warning'.
+	 *                          Defaults to true (error).
+	 * @param string $code      Optional error code for the message. Defaults to 'Found'.
+	 * @param array  $data      Optional input for the data replacements.
+	 * @param int    $severity  Optional. Severity level. Defaults to 0 which will translate to
+	 *                          the PHPCS default severity level.
+	 * @return bool
+	 */
+	protected function addMessage( $message, $stackPtr, $is_error = true, $code = 'Found', $data = array(), $severity = 0 ) {
+		return $this->throwMessage( $message, $stackPtr, $is_error, $code, $data, $severity, false );
+	}
+
+	/**
+	 * Add a fixable PHPCS message to the output stack as either a warning or an error.
+	 *
+	 * @since 0.11.0
+	 *
+	 * @param string $message   The message.
+	 * @param int    $stackPtr  The position of the token the message relates to.
+	 * @param bool   $is_error  Optional. Whether to report the message as an 'error' or 'warning'.
+	 *                          Defaults to true (error).
+	 * @param string $code      Optional error code for the message. Defaults to 'Found'.
+	 * @param array  $data      Optional input for the data replacements.
+	 * @param int    $severity  Optional. Severity level. Defaults to 0 which will translate to
+	 *                          the PHPCS default severity level.
+	 * @return bool
+	 */
+	protected function addFixableMessage( $message, $stackPtr, $is_error = true, $code = 'Found', $data = array(), $severity = 0 ) {
+		return $this->throwMessage( $message, $stackPtr, $is_error, $code, $data, $severity, true );
+	}
+
+	/**
+	 * Add a PHPCS message to the output stack as either a warning or an error.
+	 *
+	 * @since 0.11.0
+	 *
+	 * @param string $message   The message.
+	 * @param int    $stackPtr  The position of the token the message relates to.
+	 * @param bool   $is_error  Optional. Whether to report the message as an 'error' or 'warning'.
+	 *                          Defaults to true (error).
+	 * @param string $code      Optional error code for the message. Defaults to 'Found'.
+	 * @param array  $data      Optional input for the data replacements.
+	 * @param int    $severity  Optional. Severity level. Defaults to 0 which will translate to
+	 *                          the PHPCS default severity level.
+	 * @param bool   $fixable   Optional. Whether this is a fixable error. Defaults to false.
+	 * @return bool
+	 */
+	private function throwMessage( $message, $stackPtr, $is_error = true, $code = 'Found', $data = array(), $severity = 0, $fixable = false ) {
+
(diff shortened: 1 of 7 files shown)


WordPress/Sniff.php
```diff
@@ -550,6 +550,76 @@ public function strip_quotes( $string ) {
...
+			$method .= 'Fixable';
+		}
+
+		if ( true === $is_error ) {
+			$method .= 'Error';
+		} else {
+			$method .= 'Warning';
+		}
+
+		return call_user_func( array( $this->phpcsFile, $method ), $message, $stackPtr, $code, $data, $severity );
```
grappler: I would have directly called the method instead of using `call_user_func`.

```php
if ( true === $fixable ) {
	return $this->phpcsFile->addFixable( $message, $stackPtr, $code, $data, $severity );
}

if ( true === $is_error ) {
	return $this->phpcsFile->addError( $message, $stackPtr, $code, $data, $severity );
} else {
	return $this->phpcsFile->addWarning( $message, $stackPtr, $code, $data, $severity );
}
```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -163,16 +166,17 @@ public function process( PHP_CodeSniffer_File $phpcs_file, $stack_ptr ) {
...
 
 		$func_open_paren_token = $phpcs_file->findNext( T_WHITESPACE, ( $stack_ptr + 1 ), null, true );
-		if ( false === $func_open_paren_token || T_OPEN_PARENTHESIS !== $tokens[ $func_open_paren_token ]['code'] ) {
+		if ( false === $func_open_paren_token || T_OPEN_PARENTHESIS !== $this->tokens[ $func_open_paren_token ]['code'] ) {
 			 return;
 		}
 
 		$arguments_tokens = array();
 		$argument_tokens  = array();
+		$tokens           = $this->tokens;
```
grappler: Is this needed?

WordPress/Sniff.php
jrfnl: Except you're forgetting that there are also fixable warnings.... so the if/else logic in that case would be a lot more complicated than what is there now.

WordPress/Sniffs/WP/I18nSniff.php
jrfnl: Yes as line 202/206 overrides the local version of `$tokens`. See: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/827/files#diff-df68b36623d243f82931be87d069e1efL202

WordPress/Sniff.php
grappler: Ok, I did not realize that `addFixableError`existed.

jrfnl: No problem, the four functions available - and accounted for by this PR - are:
* `addWarning()`
* `addError()`
* `addFixableWarning()`
* `addFixableError()`

WordPress/Sniff.php
```diff
@@ -550,6 +550,76 @@ public function strip_quotes( $string ) {
...
+	 * @param int    $stackPtr  The position of the token the message relates to.
+	 * @param bool   $is_error  Optional. Whether to report the message as an 'error' or 'warning'.
+	 *                          Defaults to true (error).
+	 * @param string $code      Optional error code for the message. Defaults to 'Found'.
+	 * @param array  $data      Optional input for the data replacements.
+	 * @param int    $severity  Optional. Severity level. Defaults to 0 which will translate to
+	 *                          the PHPCS default severity level.
+	 * @return bool
+	 */
+	protected function addMessage( $message, $stackPtr, $is_error = true, $code = 'Found', $data = array(), $severity = 0 ) {
```
westonruter: Instead of `$is_error` being a boolean, should it perhaps be a string `error` or `warning`? Just thinking that in the future I would hope PHPCS would add an informational message that isn't error or warning. Also, passing `'error'` is more readable than passing `true`.

jrfnl: > Instead of $is_error being a boolean, should it perhaps be a string error or warning? Just thinking that in the future I would hope PHPCS would add an informational message that isn't error or warning. 

I considered that, but AFAICS PHPCS does not have any plans to add more message types in the (near) future. I didn't see anything about it in the changelogs or plans for v3 either if I remember correctly.

The drawback of changing `$is_error` to a string is that this would add some small overhead each time the function is called as you would either need a ternary `( true === $something ) ? 'error' : 'warning'` of a full if/else conditional in which you set an `$errortype` variable you would then pass to the function.
The boolean just keeps things simpler in that regards.

> Also, passing 'error' is more readable than passing true.

True and I know WP is a big fan of "wordy" variables. But - unlike WP with themes and plugins -, I don't expect non-programmers to start writing code for WPCS. The method is well documented, so for any developer, I would expect this to be quite intuïtive.
Or am I overlooking something here ?

westonruter: Over and above wordy variables, I suppose the larger problem is positional arguments as opposed to passing associative arguments. Compare:

```php
$this->addMessage( 'Hello World', $stackPtr, true, 'Found', array(), 0 );
```

vs:

```php
$this->addMessage( array(
    'message' => 'Hello World', 
    'stack_pointer' => $stackPtr,
    'is_error' => true, 
    'code' => 'Found',
    'data' = array(), 
    'severity' => 0 
) );
```

Sure this is more verbose, but each of the params then become optional whereas with positional arguments they are all required up to the last argument you need to supply.

Anyway, I don't feel strongly about this, but I did want to offer it up as a suggestion.

jrfnl: > I suppose the larger problem is positional arguments as opposed to passing associative arguments.

That can be said about each and every function out there and kind of defeats the purpose of being able to pass more than one argument.
In the same vain as passing everything in one array argument, you could pass a `stdClass` with some properties set or even create an ErrorObject interface.

> but each of the params then become optional whereas with positional arguments they are all required up to the last argument you need to supply.

Again, true, except for the fact that not all parameters *are* optional.

I guess it's a choice of style more than anything and the choice I made was to most closely mirror the PHPCS native functions.

PR #833, opened 2017-02-01
Add `process()` method to WordPress_Sniff …

jrfnl:
…and change to `process_token()` method in child classes.

This saves us having to call `$this->init()` from every child class, always sets the basic properties and makes them available to the child classes.

This PR addressed the third bullet from issue #765

~~P.S.: For the CronInterval sniff, the changes have been kept to a minimum to avoid large conflicts with #818.~~ As #818 has been merged, fixed up the CronInterval sniff to be inline with the rest of the PR.

PR #839, opened 2017-02-09
Fix custom properties whitelist in valid variable name sniff

JDGrimes:
The value of the `customPropertiesWhitelist` property of the `WordPress.NamingConventions.ValidVariableName` sniff was always being discarded. This is because the logic that was intended to only be dealing with a non-array value was actually being applied even when the value was properly set as an array. (See line 299.) I noticed this while running against a rulseset that utilizes this property, and so I fixed it, and now it is working properly.

Apparently this isn't covered by the tests, which only cover the sniffs themselves and not rulesets. Not sure if there is a way to remedy that right now or not.

### DIFF
--- a/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
@@ -280,15 +280,16 @@ class WordPress_Sniffs_NamingConventions_ValidVariableNameSniff extends PHP_Code
 			return;
 		}
 
-		if ( ! empty( $this->customPropertiesWhitelist ) ) {
-			if ( ! is_array( $this->customPropertiesWhitelist ) ) {
-				// Incorrectly set property.
-				$phpcs_file->addWarning(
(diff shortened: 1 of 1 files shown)


jrfnl:
@JDGrimes I've got a PR ready (waiting for #833 to be merged) which will actually fix this for *all* sniffs using custom lists which need to be merged. If you don't mind, I'd be inclined to close this one in favour of the more comprehensive one I've got waiting to be pulled. Oh and it includes unit tests ;-)
See: https://github.com/jrfnl/WordPress-Coding-Standards/commits/WAIT-833/WPCS/feature/better-custom-property-handling

jrfnl:
@JDGrimes Oh and it would be nice if #818 could be merged too as there might be some overlap otherwise ;-)

jrfnl:
Ok, the above mentioned branch has now been pulled as PR #841

jrfnl:
@JDGrimes Just doing some clean up. Can I delete this branch ?

PR #841, opened 2017-02-11
Full property visibility/staticness review & fixing of custom property merging

### LINKED ISSUE #765: The future of the `WordPress_Sniff` class
As since #696, #703 and #633 have all been addressed & merged, nearly all sniffs which extended an upstream sniff have been reverted to stand-alone sniffs, I'm wondering if it's time to review the `WordPress_Sniff` class.

That class currently contains mostly static properties so other sniffs - which couldn't extend it - could use these.
It also contains a number of non-static methods (and a few such properties).

I intend to start adding a number of additional utility methods to this class in the near future to be used by a plentitude of sniffs.
These utility methods have been extensively tested over the last few months within the PHPCompatibility standard (I wrote them, so there is no licencing issue with bringing them over).
They will address things like a standardized way to determine if a `T_STRING` is a function call, getting the parameters from the function call, safely stripping the quotes off a `T_CONSTANT_ENCAPSED_STRING`, checking if something is within a namespace, getting the fully qualified function/method name etc

I imagine there might be advantages in:
* turning a number of sniffs (quite a few actually) into extended classes based on the `WordPress_Sniff` class
* turning the static properties into normal class properties
* changing the `init()` method to `process()` and once it has stored the basic information, letting it hand off to a `processToken()` method which would be implemented in the child classes instead of the current `process()` method. This would save having to pass around the `$phpcsFile` and retrieving the `$tokens` in nearly every function as they would be available as class properties instead.

Opinions ?

jrfnl:
Primarily, this PR is a review of all sniff class properties and their handling.

I've broken up the PR into a number of isolated commits for easier review & having a clearer history for the various affected files.

### tl;dr
This PR fixes/addresses a number of issues:
* bullet 2 from issue #765 - static class properties in the `WP_Sniff` class
* a variety of buggyness in the merging of custom lists with pre-set lists - see #786, #839 and comments in the #765 thread
* a number of issues which need to be fixed for #614, i.e. some properties which are still set to `public`, but **really** should be made `protected` as changing these from the ruleset would break the sniff or at the very least break the intention of the sniff.
* missing unit tests for custom sniff property handling

### Pre-amble
Sniff properties which are `public` can be overruled via a custom ruleset.
Sniff properties should therefore only be `public` if that is the intended behaviour. Otherwise a more restrictive visibility should be used.

Changes to the visibility of properties *may* break backward compatibility for some users, but only if they were **_doing_it_wrong_** in the first place, so I'm not concerned about this.

Properties set via the ruleset will be handled by PHPCS as follows:
* Properties with the `type=array` attribute - will be converted to arrays. PHPCS has the ability to handle key/value arrays if the property is set using the following syntax: `key=>value,key2=>value2`
* All other properties will be treated as strings.
* If the string value of a property is `true` or `false`, the property value will be cast to a boolean value before it's set.
* Setting a property in a ruleset will overrule the previously set value. This also means that if two rulesets have conflicting values for a property, the value received from the last ruleset parsed by PHPCS will be the value used.

### What this PR changes/fixes
* I've reviewed every single sniff class property for visibility & static vs non-static and changed them where appropriate.
* For the properties where custom lists are being merged with pre-set lists, I've added a helper function to handle the merging.
This helper function is a little bit "magical".
   - It will correctly handle custom properties passed as comma delimited strings as well as arrays. Basically being forgiving to end-users who forget to set the `type=array` bit when adding a property in a custom ruleset.
   - If anything else is passed it will disregard the value of the custom property.
   - If a `base` array to merge the custom property with is passed, it expects the base array  to be set up as `value => true` to allow for using `isset()` instead of `in_array()`. Only one existing array needed to be changed to allow for this `$test_class_whitelist`.
   - When merging a custom array with a base array, the custom array will be flipped, so the values will be added as keys to be in line with the above (can be turned off by passing `false` as the third param).
   - Any custom keys added to the base array through the merge will be given the value `false`. This has no impact when using `isset()`, but allows us to filter out anything added for a previous custom property value if the custom property would be overruled inline in a file. This in turn allows us to unit test the custom properties.
   - For more detail, see the function documentation
* I've implemented use of this new utility function where appropriate.
* For the custom array property merging issues, most sniffs used to add the custom values to the static pre-set arrays and only added these once for efficiency, using a `$added_custom_properties` setting to keep track of whether the custom properties had been merged in. Some of these static pre-set properties however would be used by several sniffs and could therefore be using a "contaminated" value. Also, the fact that the custom properties were only merged in once, prevented changing these values inline using the `@codingStandardsChangeSetting` syntax, which prevented unit testing them.
    I've fixed this by:
    - Turning the `$added_custom_properties` property into an array to keep track of the previous merged in value(s) instead of being a true/false toggle.
    - This allows us to avoid merging the same values time & time again (efficiency), while still allowing the value to be changed on the fly. This minimizes the performance impact of the change, while allowing more flexibility and greater compliance with the default capabilities of PHPCS.

### Additional notes:
* The PHPCS 2.x minimum requirement is PHP 5.1. The new utility function however uses the `array_fill_keys()` function which was only introduced in PHP 5.2. This means that the minimum PHP requirement for WPCS has to go up to PHP 5.2. Considering that's an ancient version, I don't think this will cause any issues. Just to be sure, I have added this info to the requirements section of the readme.
* I've corrected the default value of one property - was float, should have been string (`$minimum_supported_version` in WP/DeprecatedFunctions). This property is used in a `version_compare()`, so string is the expected value type anyway and would be the type if the property is overruled from the ruleset as well.
* Reviewed the internal handling of properties.
    - ControlStructureSpacing / `blank_line_check` and `blank_line_after_check`: PHPCS handles the casting to boolean already. For the compares, made sure they are done against "our" defaults, so that even if incorrectly set, the additional boolean cast is not necessary.
    - ControlStructureSpacing / `spaces_before_closure_open_paren`: moved the cast to int up instead of doing it three times.
* Implemented use of the `exclude` property in the WPCS rulesets, includes correcting an invalid (outdated) exclusion in the `Extra` ruleset.
* Removed the `$supportedTokenizers` property in the `WordPress_Sniffs_WhiteSpace_ControlStructureSpacingSniff` as it was set to the default value used by PHPCS anyway.
* I've moved the custom list merging to a separate function in the relevant sniffs.
* Where appropriate, I've moved the merging of custom properties to a later point in the code flow to only do so if the properties are needed.
* CSRF/Nonce sniff: This sniff uses the `is_only_sanitized()` method which calls the `is_sanitized()` method which uses the sanitization function lists, but there was no way to enhance them. Added the relevant custom properties to enable enhancing them for that specific sniff.
* DirectDatabaseQuery sniff: This sniff did so far not extend the `WP_Sniff` class, but should do so within the new structure. Changed the sniff to leverage the properties of the `WP_Sniff` class as well.

The PR includes unit tests for all custom properties which weren't unit tested before now.

### Question:
* Does anyone remember whether there was a reason to make the CSRF/Nonce sniff `$errorForSuperGlobals` and `$warnForSuperGlobals` properties public ? That was the one set of properties I was in doubt about changing the visibility for.

### Open issue/question:
What should be the value for the `exclude` property in the overall `WordPress` ruleset for the `WordPress.PHP.DiscouragedPHPFunctions` sniff ? Or should nothing be excluded ?

The `Extra` ruleset excludes `create_function`, the `VIP` ruleset excludes `obfuscation`. Based on that, I would suggest adding them both to the `WordPress` ruleset to maintain the same behaviour as before.

#### Technical details:
Previously, the excludes were done via the `<exclude..` syntax, not by using the `exclude` property. As the `WordPress` ruleset includes both these ruleset, both excludes where set for the overall ruleset.
With this PR, the excludes will be done via the `exclude` property instead. The properties in ruleset are not cumulative, but overrule each other. That means that in effect the `WordPress` ruleset would now only exclude the `obfuscation` group as that's the value set in the last ruleset included (VIP).
Setting the `exclude` property to both from within the `WordPress` ruleset will maintain the old behaviour.
However, a case could be made to exclude nothing if the overall `WordPress` ruleset is used, which is why I'm making a point of putting this up for discussion.

Closes #765
Closes #839


WordPress/Tests/XSS/EscapeOutputUnitTest.inc
```diff
@@ -182,3 +182,27 @@ _deprecated_hook( 'some_filter', '1.3.0', esc_html__( 'The $arg is deprecated.'
 _deprecated_hook( "filter_{$context}", '1.3.0', __( 'The $arg is deprecated.' ), sprintf( __( 'Some parsed message %s', $variable ) ) ); // Bad.
 
 echo add_filter( get_the_excerpt( get_the_ID() ) ; // Bad, but ignored as code contains a parse error.
+
+
```
JDGrimes: Extra blank line here.

WordPress/Sniffs/CSRF/NonceVerificationSniff.php
```diff
@@ -22,21 +22,23 @@ class WordPress_Sniffs_CSRF_NonceVerificationSniff extends WordPress_Sniff {
 	 * Superglobals to give an error for when not accompanied by an nonce check.
 	 *
 	 * @since 0.5.0
+	 * @since 0.11.0 Changed visibility from public to protected.
 	 *
 	 * @var array
 	 */
-	public $errorForSuperGlobals = array( '$_POST', '$_FILE' );
+	protected $errorForSuperGlobals = array( '$_POST', '$_FILE' );
```
JDGrimes: I originally made these public so that they could be modified in custom rulesets. I don't know if anybody is using that feature, but it was intentional. The reason for letting them be completely overridden rather than just merging custom settings with the default lists is so that values could be removed and not just added. I haven't actually used that feature though, and I doubt anybody else has, so it is fine with me if we change these to `protected`.

grappler:
Why do we need to change from `<exclude..` syntax to the exclude property?

jrfnl:
> Why do we need to change from `<exclude..` syntax to the `exclude` property?

I should probably have changed this already in #814. The errorcodes for these specific messages have changed, so the "old" codes won't work anymore.
I could have changes it over to the "new" codes, but especially for `obfuscation` that would mean adding a lot more codes.
I also figured that _leading by example_ with using the property would be a good thing.

WordPress/Sniffs/CSRF/NonceVerificationSniff.php
jrfnl: Got it. Thanks for explaining the history. I'm open to leaving these `public` and adding them to the upcoming wiki page if you prefer.

WordPress/ruleset.xml
```diff
@@ -6,4 +6,10 @@
 	<rule ref="WordPress-Docs"/>
 	<rule ref="WordPress-Extra"/>
 	<rule ref="WordPress-VIP"/>
+
+	<rule ref="WordPress.PHP.DiscouragedPHPFunctions">
```
grappler: I would add a comment here explaining why we are adding an exception or referencing to that they have been excluded from their respective rulesets.

WordPress/Tests/XSS/EscapeOutputUnitTest.inc
jrfnl: Fixed

WordPress/ruleset.xml
jrfnl: Good point. Done.

jrfnl:
@GaryJones Have you got an opinion on this PR or should I remove you as a reviewer ?

GaryJones:
> Have you got an opinion on this PR or should I remove you as a reviewer ?

Reviewed and approved.

jrfnl:
@GaryJones Great! Thanks.

PR #843, opened 2017-02-12
Checking callbacks for restricted functions

### LINKED ISSUE #611: Restricted functions used in callbacks
While debugging the `AbstractFunctionRestrictionsSniff` class, I realized that any restricted functions used as callbacks, i.e. within `add_action()`, `call_user_func()`, `array_map()` are completely ignored.

That makes this kind of sniff extremely easy to bypass and while this is not so much an issue for people who *elect* to use the WPCS, this *will* be an issue for the Theme Review Theme Check to sniffs project as in that case, theme authors won't have a choice and bypassing checks that way is something we'll need to guard against.

I'm investigating how we can solve this. /cc @grappler

grappler:
fixes #611

I removed `preg_replace_callback_array()` from the list as the callback is not a parameter but part of the value. In the following example where `foobar` is a restricted function.
```php
$subject = 'Aaaaaa Bbb';
preg_replace_callback_array(
    [
        '~[a]+~i' => 'foobar',
    ],
    $subject
);
```

We could make a separate issue for this.

I was not sure where to place the tests for the abstract so I ended up using the old `FunctionRestrictionsSniff`. If there is a better place to add the tests I can move them there.


grappler:
@jrfnl @GaryJones Can this be merged?

jrfnl:
@grappler Reviewing this now.

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -229,7 +281,7 @@ public function is_targetted_token( $stackPtr ) {
 	 *                  normal file processing.
 	 */
 	public function check_for_matches( $stackPtr ) {
-		$token_content = strtolower( $this->tokens[ $stackPtr ]['content'] );
+		$token_content = strtolower( $this->strip_quotes( $this->tokens[ $stackPtr ]['content'] ) );
```
jrfnl: Please add a comment here about the use of `strip_quotes()` as it is not needed for the "normal" function check.

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -256,6 +308,64 @@ public function check_for_matches( $stackPtr ) {
 	} // End check_for_matches().
 
 	/**
+	 * Verify if the current token is one of the targetted functions as callback.
+	 *
+	 * @since 0.11.0
+	 *
+	 * @param int $stackPtr The position of the current token in the stack.
+	 *
+	 * @return int|false
```
jrfnl: Please document the return value: `stackPtr if for where to skip to if a callback was found, false if the function call is not a callback.`

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -170,7 +215,14 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 		$this->init( $phpcsFile );
 
 		if ( true === $this->is_targetted_token( $stackPtr ) ) {
+
+			$callback_matches = $this->check_for_callback_matches( $stackPtr );
+			if ( $callback_matches ) {
```
jrfnl: Please be explicit in what you are checking here.

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -256,6 +308,64 @@ public function check_for_matches( $stackPtr ) {
...
+
+		foreach ( $positions as $position ) {
+
+			// Calculate the last argument if the position is negative.
+			if ( $position < 0 ) {
+				$position = count( $parameters ) + 1 + $position;
+			}
+
+			if ( ! isset( $parameters[ $position ] ) ) {
+				return false;
```
jrfnl: Should the function return `false` or just return ?

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -256,6 +308,64 @@ public function check_for_matches( $stackPtr ) {
...
+				array( T_CONSTANT_ENCAPSED_STRING ),
+				$parameters[ $position ]['start'],
+				null,
+				false,
+				null,
+				true
+			);
+
+			if ( ! $callback ) {
+				return false;
```
jrfnl: Should the function return `false` or just return ?

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -256,6 +308,64 @@ public function check_for_matches( $stackPtr ) {
...
+			}
+
+			$skip_to[] = $this->check_for_matches( $callback );
+		}
+
+		if ( empty( $skip_to ) || min( $skip_to ) === 0 ) {
+			return false;
```
jrfnl: Should the function return `false` or just return ?

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -256,6 +308,64 @@ public function check_for_matches( $stackPtr ) {
...
+				$position = count( $parameters ) + 1 + $position;
+			}
+
+			if ( ! isset( $parameters[ $position ] ) ) {
+				return false;
+			}
+
+			// Only get function name not anonymous funtions.
+			$callback = $this->phpcsFile->findNext(
+				array( T_CONSTANT_ENCAPSED_STRING ),
```
jrfnl: No need for the `array()` wrapper here.

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -256,6 +308,64 @@ public function check_for_matches( $stackPtr ) {
...
+				$parameters[ $position ]['start'],
+				null,
```
jrfnl: Please pass the end position too to prevent incorrect results.

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -256,6 +308,64 @@ public function check_for_matches( $stackPtr ) {
...
+
+			// Calculate the last argument if the position is negative.
+			if ( $position < 0 ) {
+				$position = count( $parameters ) + 1 + $position;
+			}
+
+			if ( ! isset( $parameters[ $position ] ) ) {
+				return false;
+			}
+
```
jrfnl: A callback can be an array. In that case, you want to pass the stack pointer of the `T_ARRAY` or `T_SHORT_ARRAY_OPEN` to the `get_function_call_parameters()` method again.

jrfnl: You could bow out early for closures.

WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -256,6 +308,64 @@ public function check_for_matches( $stackPtr ) {
...
+	 *
+	 * @return int|false
+	 */
+	public function check_for_callback_matches( $stackPtr ) {
+
+		$token_content = strtolower( $this->tokens[ $stackPtr ]['content'] );
+
+		// Check if the function is used as a callback.
+		if ( ! isset( $this->callback_functions[ $token_content ] ) ) {
+			return false;
```
jrfnl: Should the function return false or just return ?

WordPress/AbstractFunctionRestrictionsSniff.php
grappler: The reason is that callback parameters have quotes around the function name. I have move it where it is needed.

grappler:
Thank you @jrfnl for your comments. I think an extra level of abstraction is good.

When we move `check_for_callback_matches()`/`is_callback_method()` to `WordPress_Sniff` then we would need to move `check_for_matches()` too as it is being used by `check_for_callback_matches()`.

My concern is that with the additional utility the functions `WordPress_Sniff` will become a really big class. My suggestions would be to create a `Functions_Sniff` class that extends `WordPress_Sniff` but will be extended by `WordPress_AbstractFunctionRestrictionsSniff`.

I have fixed a few of the things mentioned.

jrfnl:
@grappler 

> When we move check_for_callback_matches()/is_callback_method() to WordPress_Sniff then we would need to move check_for_matches() too as it is being used by check_for_callback_matches().

Not if you split the logic for determining whether something is a callback off from the `check_for_callback_matched` function the way I suggested above.

The sniff specific logic - checking for excluded groups etc - should stay in the sniff.

grappler:
@jrfnl It would be great if you could have a look at this PR.

The "Preliminary check" only checks for functions in the list but not for the callback functions which was causing the code to fail. So I had to disable it for the callback functions but I am not sure if that is the best way. You can find more info in the commits.

What is still missing is:
> We may even want to split the list in two (or three) - there are callback functions which directly execute (the PHP ones) and callback functions which are used in specific circumstances and even functions which take a callback, but don't execute it.

jrfnl:
Thank you @grappler for working on this and I'm sorry that it took us so long to get back to it.

As I mentioned in the [comment in the original ticket](https://github.com/WordPress/WordPress-Coding-Standards/issues/611#issuecomment-1311514080) which I left today, I think it would be better if we have some utility function with regards to callbacks in PHPCSUtils before we try to address this for WPCS, especially as the code which would be needed will need additional changes to be able to include handling PHP 8.0 named parameters etc.

With that in mind, I'm closing this PR. I'm really sorry about this and hope you may want to work with me on getting the utility functions sorted in PHPCSUtils.

PR #844, opened 2017-02-13
Updated $autoEscapedFunctions list

dingo-d:
Added `get_the_category_list()` to the $autoEscapedFunctions list

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -140,6 +140,7 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 		'get_template_part'         => true,
 		'get_the_author_link'       => true,
 		'get_the_author'            => true,
+		'get_the_category_list'     => true,
 		'get_the_date'              => true,
 		'get_the_ID'                => true,
 		'get_the_post_thumbnail'    => true,


grappler:
> While the output could be unsafe due to the `the_category` filter, (as is true of other functions in this list), in practice it would be highly unusual.

Also to be able to use the filter the malicious code would need to run on the server and if you could run any code then there are easier ways of hacking the site then trying to manipulate a filter.

> No, `$separator` is not escaped by the function.

If the separator is hardcoded, then there is no need to escape it. This is an example from `$categories_list = get_the_category_list( esc_html__( ', ', '_s' ) );`. As the output contains HTML it makes difficult to escape. On a sidenote WordPress core does not escape it in `the_category()`

On a general note it would be good to create a definition of what a function needs to fulfill to be placed on the `$autoEscapedFunctions`.

JDGrimes:
> Also to be able to use the filter the malicious code would need to run on the server and if you could run any code then there are easier ways of hacking the site then trying to manipulate a filter.

What if it is just a filter that is doing something stupid? Simplified example just to give the general idea:

```php
add_filter( 'the_category', function ( $text ) {
    return $text . $_POST['something'];
});
```

Not good, but yeah, a lot of those functions already in the list have filters, so...



> On a sidenote WordPress core does not escape it in the_category()

WordPress does not follow the late-escaping philosophy entirely. And this does sometimes lead to security issues. But yes, it is unlikely that there is a vuln from that in core, it is just plugins that might do unsafe things without even realizing it.



> On a general note it would be good to create a definition of what a function needs to fulfill to be placed on the $autoEscapedFunctions.

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/282#issuecomment-66624566

Especially this from the next comment there:

>Any non-pluggable function that returns a known limited set of safe static values not directly based on the input arguments

Basically, this fails because the output is based on the input arguments.

dingo-d:
> Basically, this fails because the output is based on the input arguments.

This seems a bit too broad to me. `wp_tag_cloud` also accepts input arguments, but is on the safe list, no?

JDGrimes:
@dingo-d It is OK, so long as the arguments are properly escaped before they are incorporated into the output. It looks like `wp_generate_tag_cloud()` does a relatively good job on that, but a close look reveals that it also does not escape `$args['separator']`, when it is passed. We could introduce more specific checking for these functions, and only flag them when particular args are passed. But they also incorporate other data from elsewhere, like the database, not all of which is always escaped. 

So, there might be some functions on that list that shouldn't really be there. The issue is that different people want to have different levels of strictness in regard to what is on that list. Possibly we should really do a review of the whole list, and consider splitting it in two. Keep only the functions there that escape all of their output, and anything else should go in a second list. Then that second list could be enabled in some people's rulesets if they want to be less strict. Alternatively, just remove those functions from the list completely and let people who want to be less strict add the list of functions via their custom ruleset. But the sniff itself is designed to force late escaping, and although that sometimes clashes with WordPress (which doesn't always late-escape right before output), it's the safest strategy.

IMHO, many of these controversial functions just shouldn't exist in WordPress in the way that they do anyway. They should not return the output string, they should echo it directly, and escape everything incorporated into that output that isn't hard-coded. But it is basically a legacy pattern that is now stuck in WordPress, and which we have to deal with (which is a pain). Especially for theme developers. I know that I personally get a break because I don't need to use these functions in my plugins. 😄

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/748 is related. I guess that at some point we are going to need two sniffs here. One for XSS low-hanging fruit, and one that forces late-escaping everywhere like the XSS sniff does now. But the point is, strictly speaking, the output of  `get_the_category_list()` is *not* auto-escaped. And neither is `wp_tag_cloud()` completely, so it should be moved from that list as well.

Any thoughts on introducing a second list of "dangerous whitelisted unescaped functions"? 😄

grappler:
> The issue is that different people want to have different levels of strictness in regard to what is on that list. Possibly we should really do a review of the whole list, and consider splitting it in two. Keep only the functions there that escape all of their output, and anything else should go in a second list.

I think this is a good solution. The second list would be good for the theme review team.

> I guess that at some point we are going to need two sniffs here. One for XSS low-hanging fruit, and one that forces late-escaping everywhere like the XSS sniff does now.

Yes, the jump from no checks to no echoing without escaping is too big the way functions are written in WordPress.

I think it would be good to open a new issue for this. @JDGrimes do you want to do it or shall I?

dingo-d:
>  It looks like wp_generate_tag_cloud() does a relatively good job on that, but a close look reveals that it also does not escape $args['separator'], when it is passed. 

This was exactly the argument that I tried to target to see if I can somehow inject some harmful stuff to my database, but I didn't have much time to play with it on my workplace :D

I also think that two lists would be a good solution.

I can close the PR, or you guys can close it once a new issue for this is created.

JDGrimes:
@grappler See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/845, and please comment there with any thoughts about what you would like to see from a less strict sniff.

@dingo-d I'll open a PR/issue for the two lists next chance I get, unless somebody else takes the initiative and beats me to it. 😄

dingo-d:
I started going through the list and there are some deprecated functions in the `$autoEscapedFunctions` list. `comments_popup_script` and `comments_rss_link` for instance. 

Also majority of the functions on the list have `apply_filters` at the end. That kinda makes them unsafe, no?

grappler:
@dingo-d I have already started on separating the lists. You are welcome to help me. https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/845#issuecomment-279834686

dingo-d:
Will comment there :)

PR #846, opened 2017-02-17
Prevent erroneous dropping of customPropertiesWhitelist array on ValidVariableName sniff

westonruter:
My PHPCS ruleset included:

```xml
<rule ref="WordPress.NamingConventions.ValidVariableName">
	<properties>
		<property name="customPropertiesWhitelist" type="array" value="textContent,nodeName,nodeValue" />
	</properties>
</rule>
```

Nevertheless, I was still getting flagged for code like:

```php
'label' => $link->textContent,
```

I found that the conditions surrounding the retrieval of the `customPropertiesWhitelist` property were inadvertently dropping the value when it was defined properly as an `array`.


grappler:
@westonruter I think #841 will fix your issue once it is merged.

jrfnl:
Marking this as a duplicate of #839 and partial duplicate to the more comprehensive #841.
@westonruter Could you please test with #841 ? That PR should fix these kind of issues for **all** sniffs merging custom lists.

westonruter:
👍 #841 fixes it.

PR #848, opened 2017-02-20
Whitelist invalid snake_case WordPress globals

barryceelen:
Probably not often-used, but the global $is_IIS it is giving me a warning in an otherwise pristine coding-standards-adhering file :-(

### DIFF
--- a/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
@@ -54,6 +54,7 @@ class WordPress_Sniffs_NamingConventions_ValidVariableNameSniff extends PHP_Code
 	 */
 	protected $wordpress_mixed_case_vars = array(
 		'EZSQL_ERROR' => true,
+		'$is_IIS'     => true,
 		'PHP_SELF'    => true,
 	);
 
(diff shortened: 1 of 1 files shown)


WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php
```diff
@@ -54,6 +54,7 @@ class WordPress_Sniffs_NamingConventions_ValidVariableNameSniff extends PHP_Code
 	 */
 	protected $wordpress_mixed_case_vars = array(
 		'EZSQL_ERROR' => true,
+		'$is_IIS'     => true,
```
JDGrimes: I think you need to omit the `$` here, like the other variable names in the list?

jrfnl:
Inspired by this PR, I just looked at the [Global Variable Codex page](https://codex.wordpress.org/Global_Variables) and see a number of other variables which you might want to add to the PR:

* `$is_NS4`
* `$is_macIE`
* `$is_winIE`
* `$is_IE`

Alternatively, I can open a new PR to add these to the list.

barryceelen:
@JDGrimes @jrfnl Thanks for pointing that out, removed the `$` and added the other variables to this pull request.

jrfnl:
@barryceelen Thanks for adding these & fixing the `$`. The PR looks good to go IMO!

JDGrimes:
Thanks @barryceelen. 👍

PR #855, opened 2017-02-27
Support phpcodesniffer-composer-installer for easy integrate code style

evilebottnawi:
https://github.com/DealerDirect/phpcodesniffer-composer-installer

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ The WordPress Coding Standards are currently [not compatible with the upcoming P
 
 ### Composer
 
-Standards can be installed with [Composer](https://getcomposer.org/) dependency manager:
+Standards can be installed with the [Composer](https://getcomposer.org/) dependency manager:
 
     composer create-project wp-coding-standards/wpcs --no-dev
 
@@ -59,6 +59,16 @@ Running this command will:
 4. Make `phpcs` command available from `wpcs/vendor/bin`.
 
 For convenience of using `phpcs` as global command you might want to add path to `wpcs/vendor/bin` directory to a `PATH` environment of your operating system.
+
+#### Installing WPCS as a dependency
+
+When installing the WordPress Coding Standards as a dependency in a larger project, the above mentioned step 3 will not be executed automatically.
+
+There are two actively maintained Composer plugins which can handle the registration of standards with PHP_CodeSniffer for you:
+* [composer-phpcodesniffer-standards-plugin](https://github.com/higidi/composer-phpcodesniffer-standards-plugin)
+* [phpcodesniffer-composer-installer](https://github.com/DealerDirect/phpcodesniffer-composer-installer)
+
+It is strongly suggested to `require` one of these plugins in your project to handle the registration of external standards with PHPCS for you.
 
 ### Standalone
 
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,9 @@
 	"require"    : {
 		"squizlabs/php_codesniffer": "^2.9.0"
 	},
+	"suggest" : {
+		"dealerdirect/phpcodesniffer-composer-installer": "*"
+	},
 	"minimum-stability" : "RC",
 	"support"    : {
 		"issues": "https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues",
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
 		"wiki"  : "https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki",
 		"source": "https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards"
 	},
+	"type"       : "phpcodesniffer-standard",
 	"scripts"    : {
 		"post-install-cmd": "\"vendor/bin/phpcs\" --config-set installed_paths ../../..",
 		"post-update-cmd" : "\"vendor/bin/phpcs\" --config-set installed_paths ../../.."


JDGrimes:
Related: #546 

\cc @schlessera

jrfnl:
@Rarst Would you have an opinion on this ?

Rarst:
Oops, sorry my notifier was broken for some days and I missed bunch of notifications.

I am not sure what the installer _does_ in this case. I remember seeing one that copied stuff around and I don't like that approach. In my opinion Composer packages should never be nested and specific case of standards should be handled with configuration rather than location.

evilebottnawi:
@Rarst this only set `installed_paths` config right

Rarst:
Ah, good to hear. :) I will try to have a proper look at it and play with later then.

Potherca:
As this _litterally_ only adds `"type"       : "phpcodesniffer-standard",` to the `composer.json`, could this PR be looked at?

(There is further documentation outlining [the rational behind this change](https://github.com/DealerDirect/phpcodesniffer-composer-installer/wiki/Change-%60composer.json%60-%22type%22-to-%60phpcodesniffer-standard%60) if that helps...)

JDGrimes:
I think this makes sense, I can see no downside. I do want to confirm that this will work properly though, since WPCS is actually a bundle of several standards (`WordPress`, `WordPress-Core`, etc.) ??

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/546 could then be closed, as I'd expect that if PHPCS ever does add native support for something like this (which seems unlikely at this point), it would follow the same conventions already being employed by existing plugins.

Rarst:
I am getting to looking at it properly this week, will report. :)

Rarst:
Ok, finally got to it.

So given this test `compose.json`

```json
{
	"require" : {
		"squizlabs/php_codesniffer" : "^2.0",
		"dealerdirect/phpcodesniffer-composer-installer" : "0.3.2",
		"wp-coding-standards/wpcs":"dev-patch-1"
	},
	"repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/evilebottnawi/WordPress-Coding-Standards"
        }
    ]
}
```

Running `composer install --no-dev` will net us following `vendor/squizlabs/CodeSniffer.conf`

```php
<?php
 $phpCodeSnifferConfig = array (
  'installed_paths' => 'C:\\server\\phpcs-installer-test\\vendor/wp-coding-standards/wpcs',
)
?>
```

I like this approach:

 - just works (no scripting in `composer.json` needed)
 - doesn't standards files around
 - produces reliable absolute path to standards in config

One note on the PR itself — it sets type, but doesn't require installer plugin itself. So it would need either for plugin to be otherwise required or installed globally. If we want to go with this then we should probably require plugin ourselves as well.

Notably this will not work on checkout of _this_ repository itself (as opposed to be pulled in as dependency) since installer doesn't seem to currently handle the case (see https://github.com/DealerDirect/phpcodesniffer-composer-installer/issues/20 ) so we would have to keep `scripts` setup in `composer.json` we already have.

Potherca:
@Rarst Should we fix https://github.com/DealerDirect/phpcodesniffer-composer-installer/issues/20 before this PR (possibly including the `require`) gets merged?

Rarst:
That would be cleaner overall, but I don't think it's critical since we currently already handle root case ourselves with `composer.json` script instructions https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/5dbac380f4c0b8c61640b42825cc91034ce2a77b/composer.json#L21-L22

jrfnl:
Possibly superfluous, but this issue provides some more information about the cases the DealerDirect composer plugin handles: https://github.com/DealerDirect/phpcodesniffer-composer-installer/issues/24

frenck:
@Rarst DealerDirect/phpcodesniffer-composer-installer#20 has been closed and a new release is out.

The plugin is now able to handle the root case.

Potherca:
@GaryJones As mentioned in [the rational](https://github.com/DealerDirect/phpcodesniffer-composer-installer/wiki/Change-%60composer.json%60-%22type%22-to-%60phpcodesniffer-standard%60) linked [in my earlier comment](https://github.com/DealerDirect/phpcodesniffer-composer-installer/wiki/Change-%60composer.json%60-%22type%22-to-%60phpcodesniffer-standard%60): The `type` should not cause any problems as the specified behaviour for [custom-installers](https://getcomposer.org/doc/articles/custom-installers.md) is to ignore any type that composer (or any composer plugin) does not recognize and default to "library".

JDGrimes:
So the PR should now be updated, as @Rarst requested, to:

- `require` `dealerdirect/phpcodesniffer-composer-installer`. 
- Remove the `scripts` section as it is then no longer needed at all.

My only question is:

I was under the impression that there are several different installers for PHPCS standards. This one seems to be well maintained, so I'm fine with using/depending on it. But I would just like to be clear about what affects, if any, this might have on somebody who uses a different installer. I have a feeling if somebody installed WPCS with a different installer, us requiring this installer might cause issues/conflicts—am I correct?

Rarst:
>  I have a feeling if somebody installed WPCS with a different installer, us requiring this installer might cause issues/conflicts—am I correct?

Given that there is no official way (I think) for installers to claim `type` for themselves, two installers trying to target same type will likely conflict to some degree.

We could _not_ add the specific plugin to require, more compatible less convenient.

frenck:
@Rarst Correct, there is no official way to claim it.

My suggestion in this case would be:
- Do not require the installer, for compatibility sake.
- Add plugin package suggestion in the `composer.json`
- Keep old script instructions (should not be a problem if the plugin is used).
- Update documentation

You can still decide (over time), what to do.

JDGrimes:
> We could *not* add the specific plugin to require, more compatible less convenient.

I think being more convenient makes sense here, but I just wanted to be clear that it would potentially have compatibility issues for some people at some point. If it ever were to become a significant issue, we could revisit it.

JDGrimes:
Posted my last comment at the same time as @frenck. On second thought I'd say he is probably correct, and we might want to just suggest the use of the plugin in the README for now, to ease installation.

So this PR would be merged as-is, and then the install instructions would be updated to mention this option.

Rarst:
> So this PR would be merged as-is

I'd like to see `suggest` added, since other plugin I've seen is way more messy IMO than this one. Won't hurt to include docs in PR as well.

PS oh dear, we just keep picking on this. :) if @evilebottnawi sick of it I can finish this up [next week].

Potherca:
I would also suggest not adding a `require` just yet. The `type` (and possibly a `suggest`) should be fine.

A require can always be added later. That would also give us time to look at the "conflicting installers" scenario and ways of dealing with it in the plugin, as that is _bound_ to show up anyway...

jrfnl:
I've rebased & updated the PR. @Rarst @GaryJones Would you mind having another look ?

jrfnl:
Closes #546

composer.json
```diff
@@ -13,12 +13,16 @@
 	"require"    : {
 		"squizlabs/php_codesniffer": "^2.9.0"
 	},
+	"suggest" : {
+		"dealerdirect/phpcodesniffer-composer-installer": "*"
```
GaryJones: Worth suggesting the alternative as well?

evilebottnawi:
Guys, sorry, i don't have time to this now, feel free to close this PR and create new if something wrong

composer.json
jrfnl: Wouldn't that create the risk of people blindly installing everything that's suggested and then having two plugins doing the same thing (and possibly conflicting with each other) ?

That was the reason for me to just suggest one anyway.

jrfnl:
@evilebottnawi No worries, we appreciate you taking the initiative for this and getting the discussion going. I've updated your PR (so you'll get whuffies for your work) and hope to get this merged soon.

README.md
```diff
@@ -60,6 +60,16 @@ Running this command will:
 
 For convenience of using `phpcs` as global command you might want to add path to `wpcs/vendor/bin` directory to a `PATH` environment of your operating system.
 
+#### Installing WPCS as a dependency
+
+When installing the WordPress Coding Standards as a dependency in a larger project, the above mentioned step 3 will not be executed automatically.
+
+There are two actively maintained Composer plugins which can handle the registration of standards with PHP_CodeSniffer for you:
+* [composer-phpcodesniffer-standards-plugin](https://github.com/higidi/composer-phpcodesniffer-standards-plugin)
+* [phpcodesniffer-composer-installer](https://github.com/DealerDirect/phpcodesniffer-composer-installer)
```
GaryJones: Do these two both work in roughly the same way? How would someone choose between the two?

Worth WPCS picking one as the recommendation, and leaving the other as a link for an alternative, instead of presenting both equally (in this doc)?

jrfnl: > Do these two both work in roughly the same way?

AFAIK yes, they do.

> How would someone choose between the two?

Probably the same way as when choosing between two different WP plugins to use:
* See if it's still actively maintained, issues are being responded to, has good reviews, how many users it has on packagist.
* Install each, test it, see if you like it and it does what you want.

> Worth WPCS picking one as the recommendation, and leaving the other as a link for an alternative, instead of presenting both equally (in this doc)?

We could, but I figured most Composer users would just `require` WPCS and see the `suggest` and follow that without reading the readme. I'd be happy to adjust it though.

The reason I choose the @DealerDirect version as the recommendation is that the team of that plugin has been very active trying to help external standards get this issue sorted, created their Composer plugin in response to an issue with another external standard I'm involved with and has been very responsive to queries related to the plugin.

GaryJones:
Good to merge.

PR #858, opened 2017-02-28
Adding missing escaping check for `<?=`

### LINKED ISSUE #857: Report missing escaping for PHP <?= tag
While the WordPress ruleset.xml is including the `Generic.PHP.DisallowShortOpenTag` sniff, we still should report on missing escaping in case, the `<?=` is used without escaping function

david-binda:
This commit registers `T_OPEN_TAG_WITH_ECHO` in EscapeOutputSniff and adds new test for those cases

Fixes #857 

cc @westonruter

### DIFF
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -116,6 +116,7 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 			T_PRINT,
 			T_EXIT,
 			T_STRING,
+			T_OPEN_TAG_WITH_ECHO,
 		);
 
 	}
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -124,7 +124,6 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 		 * which is how open short tags are being handled in that case.
 		 */
 		if ( false === $this->is_short_open_tag_enabled() ) {
-			var_export( 'Short open tag is disabled!' );
 			$tokens[] = T_INLINE_HTML;
 		}
 		return $tokens;
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -131,7 +131,7 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 			T_OPEN_TAG_WITH_ECHO,
 		);
 
-		/**
+		/*
 		 * In case open_short_tag is turned off, we can attempt to regex T_INLINE_HTML
 		 * which is how open short tags are being handled in that case.
 		 */
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -181,7 +181,7 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 				return;
 			}
 
-			// Report on what very likely is a PHP short open tag outputing variable.
+			// Report on what very likely is a PHP short open tag outputting variable.
 			if ( preg_match( '/\<\?\=[\s]*(\$[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*(?:(?:->\S+|\[[^\]]+\]))*)[\s]*;?[\s]*\?\>/', $this->tokens[ $stackPtr ]['content'], $matches ) ) {
 				$this->phpcsFile->addError( 'Expected next thing to be an escaping function, not %s.', $stackPtr, 'OutputNotEscaped', $matches[1] );
 				return;
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -106,11 +106,23 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 	);
 
 	/**
+	 * Status of short_open_tag feature
+	 *
+	 * @var bool
+	 */
+	private $short_open_tag_enabled = true;
+
+	/**
 	 * Returns an array of tokens this test wants to listen for.
 	 *
 	 * @return array
 	 */
 	public function register() {
+		// Check whether short_open_tag is disabled on PHP version < 5.4 (it''s enabled by default in later versions).
+		if ( true === version_compare(phpversion(), '5.4', '<' ) && false === (bool) ini_get( 'short_open_tag' ) ) {
+			$this->short_open_tag_enabled = false;
+		}
+
 		$tokens =  array(
 			T_ECHO,
 			T_PRINT,
@@ -123,7 +135,7 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 		 * In case open_short_tag is turned off, we can attempt to regex T_INLINE_HTML
 		 * which is how open short tags are being handled in that case.
 		 */
-		if ( false === $this->is_short_open_tag_enabled() ) {
+		if ( false === $this->short_open_tag_enabled ) {
 			$tokens[] = T_INLINE_HTML;
 		}
 		return $tokens;
@@ -161,7 +173,7 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 			if ( in_array( $function, array( 'trigger_error', 'user_error' ), true ) ) {
 				$end_of_statement = $this->phpcsFile->findEndOfStatement( $open_paren + 1 );
 			}
-		} else if ( false === $this->is_short_open_tag_enabled() && T_INLINE_HTML === $this->tokens[ $stackPtr ]['code'] ) {
+		} else if ( false === $this->short_open_tag_enabled && T_INLINE_HTML === $this->tokens[ $stackPtr ]['code'] ) {
 			// Skip if no PHP short_open_tag is in the string.
 			if ( false === strpos( $this->tokens[ $stackPtr ]['content'], '<?=' ) ) {
 				return;
@@ -438,19 +450,5 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 			$this->addedCustomFunctions['print'] = $this->customPrintingFunctions;
 		}
 	}
-
-	/**
-	 * Checks whether short_open_tag is enabled.
-	 *
-	 * Since PHP 5.4, <?= is always available.
-	 *
-	 * @return bool False if short_open_tag is not enabled, true otherwise
-	 */
-	public function is_short_open_tag_enabled() {
-		if ( true === version_compare(phpversion(), '5.4', '<' ) && false === (bool) ini_get( 'short_open_tag' ) ) {
-			return false;
-		}
-		return true;
-	} 
 
 } // End class.
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -120,7 +120,7 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 	 * @return array
 	 */
 	public function register() {
-		// Check whether short_open_tag is disabled on PHP version < 5.4 (it''s enabled by default in later versions).
+		// Check whether short_open_tag is disabled on PHP version < 5.4 (it's enabled by default in later versions).
 		if ( PHP_VERSION_ID < 50400 && false === (bool) ini_get( 'short_open_tag' ) ) {
 			$this->short_open_tag_enabled = false;
 		}
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -182,7 +182,7 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 			}
 
 			// Report on what very likely is a PHP short open tag outputing variable.
-			if ( preg_match( '/\<\?\=[\s]*(\$[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)[\s]*;?[\s]*\?\>/', $this->tokens[ $stackPtr ]['content'], $matches ) ) {
+			if ( preg_match( '//\<\?\=[\s]*(\$[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*(?:(?:->\S+|\[[^\]]+\]))*)[\s]*;?[\s]*\?\>//', $this->tokens[ $stackPtr ]['content'], $matches ) ) {
 				$this->phpcsFile->addError( 'Expected next thing to be an escaping function, not %s.', $stackPtr, 'OutputNotEscaped', $matches[1] );
 				return;
 			}
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -132,8 +132,8 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 		);
 
 		/*
-		 * In case open_short_tag is turned off, we can attempt to regex T_INLINE_HTML
-		 * which is how open short tags are being handled in that case.
+		 * In case short_open_tag is turned off, we can attempt to regex T_INLINE_HTML
+		 * which is how short open tags are being handled in that case.
 		 */
 		if ( false === $this->short_open_tag_enabled ) {
 			$tokens[] = T_INLINE_HTML;
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -179,16 +179,16 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 				return;
 			}
 
-			// Report on what very likely is a PHP short open tag outputing variable
+			// Report on what very likely is a PHP short open tag outputing variable.
 			if ( preg_match( '/\<\?\=[\s]*(\$[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)[\s]*;?[\s]*\?\>/', $this->tokens[ $stackPtr ]['content'], $matches ) ) {
-				$this->phpcsFile->addError( "Expected next thing to be an escaping function, not %s.", $stackPtr, 'OutputNotEscaped', $matches[1] );
+				$this->phpcsFile->addError( 'Expected next thing to be an escaping function, not %s.', $stackPtr, 'OutputNotEscaped', $matches[1] );
 				return;
 			}
 
-			// Throw warning in case the T_INLINE_HTML looks like a open_short_tag
-			if ( false !== strpos( $this->tokens[$stackPtr]['content'], '<?=' ) ) {
+			// Throw warning in case the T_INLINE_HTML looks like a open_short_tag.
+			if ( false !== strpos( $this->tokens[ $stackPtr ]['content'], '<?=' ) ) {
 				$this->phpcsFile->addWarning( 'Possible use of PHP short open tag ( "<?=" ) detected. Needs manual inspection.', $stackPtr, 'PossibleShortOpenTag' );
-				return;	
+				return;
 			}
 			return;
 		}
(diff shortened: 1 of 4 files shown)


westonruter:
Travis CI may not be responding for awhile due to the AWS outage.

WordPress/Tests/XSS/EscapeOutputUnitTest.inc
```diff
@@ -205,3 +205,13 @@ to_screen( esc_form_field( $var1), esc_attr( $var2 ) ); // Ok.
...
 echo post_info( $post_id, 'field' ); // Bad.
 echo cpt_info( $post_type, 'query' ); // Bad.
+
+// Bad.
+?>
+<?= $var ?>
+<?php
+
+// Ok.
+<?= esc_html( $var ); ?>
```
JDGrimes: I think a PHP closing tag is missing here?

westonruter: Good catch. The immediately-preceding `<?php` can be removed.

david-binda: Good catch!

jrfnl: Seeing the build failure: you could just change the comment to an HTML comment instead ?

jrfnl:
@david-binda Just checking: do you know how to fix the PHP 5.3 build failure or do you need a hand with that ?

david-binda:
@jrfnl thanks for checking! I'll need to fire up some 5.3 environment and run the phpunit tests in there. My guess is that the shorttag is not being properly parsed in that version. I'll figure that out.

jrfnl:
@david-binda I suspect you may need to look at the `short_open_tag` ini  setting which is quick likely the culprit.

david-binda:
@jrfnl right. It actually is the culprit. I'm looking into whether there might be some way to report on the short open tags even when that `short_open_tag` ini is not set to `true|"1"`

jrfnl:
@david-binda You may want to have a look here for inspiration: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Generic/Sniffs/PHP/DisallowAlternativePHPTagsSniff.php

As for the unit tests, you can provide an ini settings file to travis to use. See here for an example: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/.travis.yml
And here for the docs on this: https://docs.travis-ci.com/user/languages/php#Custom-PHP-configuration

david-binda:
> @david-binda You may want to have a look here for inspiration: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/CodeSniffer/Standards/Generic/Sniffs/PHP/DisallowAlternativePHPTagsSniff.php

Thanks, that's helpful. I have something similar in place.

> As for the unit tests, you can provide an ini settings file to travis to use. See here for an example: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/.travis.yml
And here for the docs on this: https://docs.travis-ci.com/user/languages/php#Custom-PHP-configuration

@jrfnl  I don't think that the failing test is what should be fixed. Or at least not the only thing which should be fixed.

The sniff, imho, should definitely report all `<?=` occurrences as Warnings with prompt for manual inspection in case the `ini_get( 'short_open_tag' )` returns `false|''` and I'm also playing with catching some obvious direct variable outputs as errors, but it involves regex and I'm not quite sure whether that is the best approach. Any thoughts on this?

jrfnl:
@david-binda 

> @jrfnl I don't think that the failing test is what should be fixed. Or at least not the only thing which should be fixed.

Very true, but both situations - `short_open_tags` `on|off` - should probably be tested, that's why I pointed to the ini file option which can be used to do just that. You could use an `ENV` variable to set which build you'd want to test against the ini file.

> I'm also playing with catching some obvious direct variable outputs as errors, but it involves regex and I'm not quite sure whether that is the best approach. Any thoughts on this?

I'd need to see some example code before I can give an opinion on the approach, but in case you decide to go with regex and need help with that, you're welcome to ping me.

david-binda:
Whoops. Looks like I have broken the Sniff outside the PHP 5.3. I'll look into it.

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -111,14 +111,24 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
...
+		$tokens =  array(
 			T_ECHO,
 			T_PRINT,
 			T_EXIT,
 			T_STRING,
-			T_OPEN_TAG_WITH_ECHO,
 		);
 
+		// We can successfully report on unescaped open_short_tags only if they are enabled.
+		if ( true === (bool) ini_get( 'short_open_tag' ) ) {
```
jrfnl: @david-binda This is the issue for PHP 5.4+ as the `T_OPEN_TAG_WITH_ECHO` is supported independently of the `short_open_tag` setting from PHP 5.4+ upwards, so basically, leave the token in the base array and only use the `else` clause of this part with the reversed condition and you should be fine.

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -116,16 +116,15 @@ public function register() {
...
-			/**
-			 * In case open_short_tag are turned off, we can attempt to regex T_INLINE_HTML
-			 * which is how open_short_tags are being handled in that case.
-			 */
+		/**
+		 * In case open_short_tag are turned off, we can attempt to regex T_INLINE_HTML
+		 * which is how open_short_tags are being handled in that case.
+		 */
+		if ( false === $this->is_short_open_tag_enabled() ) {
+			var_export( 'Short open tag is disabled!' );
```
jrfnl: Please remove the debug code.

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -441,4 +440,18 @@ protected function mergeFunctionLists() {
...
 	}
 
+	/**
+	 * Checks whether short_open_tag is enabled.
+	 *
+	 * Since PHP 5.4, <?= is always available.
+	 *
+	 * @return bool False if short_open_tag is not enabled, true otherwise
+	 */
+	public function is_short_open_tag_enabled() {
```
jrfnl: The returned value of this function is not going to change during the PHPCS run, so I'd like to suggest adding a class (private) property instead, assigning the value to the property at the start of the `register()` function and referring to the property later on for efficiency.

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -441,4 +440,18 @@ protected function mergeFunctionLists() {
...
+		if ( true === version_compare(phpversion(), '5.4', '<' ) && false === (bool) ini_get( 'short_open_tag' ) ) {
```
jrfnl: Please use the `PHP_VERSION_ID` constant instead for the comparison (which is backfilled in PHPCS).
`PHP_VERSION_ID < 504000` is two less functions calls for the same result.

WordPress/Tests/XSS/EscapeOutputUnitTest.php
```diff
@@ -77,7 +77,7 @@ public function getWarningList() {
 		$list = array();
 
 		//Adding Warning which is triggerred in case open_short_tag is set to Off
-		if ( false === (bool) ini_get( 'short_open_tag' ) ) {
+		if ( true === version_compare(phpversion(), '5.4', '<' ) && false === (bool) ini_get( 'short_open_tag' ) ) {
```
jrfnl: Please use the `PHP_VERSION_ID` constant instead (see above)

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -116,16 +116,15 @@ public function register() {
...
-		// We can successfully report on unescaped open_short_tags only if they are enabled.
-		if ( true === (bool) ini_get( 'short_open_tag' ) ) {
-			$tokens[] = T_OPEN_TAG_WITH_ECHO;
-		} else {
-			/**
-			 * In case open_short_tag are turned off, we can attempt to regex T_INLINE_HTML
-			 * which is how open_short_tags are being handled in that case.
-			 */
+		/**
+		 * In case open_short_tag are turned off, we can attempt to regex T_INLINE_HTML
```
jrfnl: `are` => `is`

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -116,16 +116,15 @@ public function register() {
...
+		 * which is how open_short_tags are being handled in that case.
```
jrfnl: `open_short_tags` => `short open tags`

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -152,6 +162,24 @@ public function process_token( $stackPtr ) {
...
+			}
+
+			// Report on what very likely is a PHP short open tag outputing variable
+			if ( preg_match( '/\<\?\=[\s\t]*(\$[a-zA-Z_][[:alnum:]_]+)[\s\t]*\;?[\s\t]*\?\>/', $this->tokens[ $stackPtr ]['content'], $matches ) ) {
+				$this->phpcsFile->addError( "Expected next thing to be an escaping function, not %s.", $stackPtr, 'OutputNotEscaped', $matches[1] );
+				return;
+			}
+
+			// Throw warning in case the T_INLINE_HTML looks like a open_short_tag
+			if ( false !== strpos( $this->tokens[$stackPtr]['content'], '<?=' ) ) {
```
jrfnl: There is a separate sniff for this, so no need to throw that warning here.

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -152,6 +162,24 @@ public function process_token( $stackPtr ) {
...
 				$end_of_statement = $this->phpcsFile->findEndOfStatement( $open_paren + 1 );
 			}
+		} else if ( false === $this->is_short_open_tag_enabled() && T_INLINE_HTML === $this->tokens[ $stackPtr ]['code'] ) {
+			// Skip if no PHP short_open_tag is in the string.
+			if ( false === strpos( $this->tokens[ $stackPtr ]['content'], '<?=' ) ) {
+				return;
+			}
+
+			// Report on what very likely is a PHP short open tag outputing variable
+			if ( preg_match( '/\<\?\=[\s\t]*(\$[a-zA-Z_][[:alnum:]_]+)[\s\t]*\;?[\s\t]*\?\>/', $this->tokens[ $stackPtr ]['content'], $matches ) ) {
```
jrfnl: * The regex for a variable name in PHP is: `[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*` Ref: http://php.net/manual/en/language.variables.basics.php
* `'\t` is already included in `\s`, so no need to make it explicit in the character class.
* `;` does not need to be escaped in regex

WordPress/Sniffs/XSS/EscapeOutputSniff.php
david-binda: @jrfnl what Sniff do you have in mind, please?

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -106,18 +106,39 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
...
 			T_PRINT,
 			T_EXIT,
 			T_STRING,
+			T_OPEN_TAG_WITH_ECHO,
 		);
 
+		/**
+		 * In case open_short_tag is turned off, we can attempt to regex T_INLINE_HTML
+		 * which is how open short tags are being handled in that case.
+		 */
```
jrfnl: `open_short_tag` => `short_open_tag`
`open short tags` => `short open tags`

WordPress/Sniffs/XSS/EscapeOutputSniff.php
david-binda: Very good call on the array based values. Outputting object properties might suffer from the same issue. I'll look into those cases later in the day. Thanks for pointing that out!

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -106,11 +106,23 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
 	);
 
 	/**
+	 * Status of short_open_tag feature
+	 *
+	 * @var bool
+	 */
+	private $short_open_tag_enabled = true;
```
jrfnl: Please add a `@since` tag

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -106,11 +106,23 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
...
+
+	/**
 	 * Returns an array of tokens this test wants to listen for.
 	 *
 	 * @return array
 	 */
 	public function register() {
+		// Check whether short_open_tag is disabled on PHP version < 5.4 (it''s enabled by default in later versions).
```
jrfnl: `it''s` => `it's`

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -152,6 +173,24 @@ public function process_token( $stackPtr ) {
...
 				$end_of_statement = $this->phpcsFile->findEndOfStatement( $open_paren + 1 );
 			}
+		} else if ( false === $this->short_open_tag_enabled && T_INLINE_HTML === $this->tokens[ $stackPtr ]['code'] ) {
+			// Skip if no PHP short_open_tag is in the string.
+			if ( false === strpos( $this->tokens[ $stackPtr ]['content'], '<?=' ) ) {
+				return;
+			}
+
+			// Report on what very likely is a PHP short open tag outputing variable.
+			if ( preg_match( '/\<\?\=[\s]*(\$[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)[\s]*;?[\s]*\?\>/', $this->tokens[ $stackPtr ]['content'], $matches ) ) {
```
jrfnl: Re: your remark that the regex still needs improving for array and object based variables:

What about changing the regex to:
`/\<\?\=[\s]*(\$[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*(?:(?:->\S+|\[[^\]]+\]))*)[\s]*;?[\s]*\?\>/`

WordPress/Sniffs/XSS/EscapeOutputSniff.php
jrfnl: Both short open tags as well as alternative open tags are sniffed for in the `WordPress-Core` ruleset:
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress-Core/ruleset.xml#L104-L105

WordPress/Sniffs/XSS/EscapeOutputSniff.php
jrfnl: See my suggestion below

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -106,18 +106,39 @@ class WordPress_Sniffs_XSS_EscapeOutputSniff extends WordPress_Sniff {
...
+
+		$tokens = array(
 			T_ECHO,
 			T_PRINT,
 			T_EXIT,
 			T_STRING,
+			T_OPEN_TAG_WITH_ECHO,
 		);
 
+		/**
```
JDGrimes: This isn't a docblock, so it should only have one `*`.

WordPress/Sniffs/XSS/EscapeOutputSniff.php
JDGrimes: Just thinking out loud, I wonder if there is a way to replace each `<?=` with `<?php echo` and then reparse that part of the string as PHP.  That would avoid regex entirely. But perhaps it really isn't such a good idea, or isn't really worth it even if it is possible.

jrfnl: That is definitely possible, but might be more complicated and I wonder if it's really worth it as it would only be necessary for PHP <= 5.3 with `short_open_tag` off.

What would make it more complicated is that the `T_INLINE_HTML` token is split at line-end and you may need to join several T_INLINE_HTML tokens together if the php code is split over several lines as otherwise the tokenizer might fail.

Think:
```php
<p>some text<?=
$variable;
?></p>
```

N.B.: the _php code split over several T_INLINE_HTML tokens issue_ is not being addressed with the regex either, but as it's an edge-case, I don't have a problem with leaving that as is.

JDGrimes: Yeah, it sounds like it isn't worth it.

WordPress/Sniffs/XSS/EscapeOutputSniff.php
david-binda: @jrfnl I'm happy to get rid of his Warning, but I have noticed that in PHP 5.3 with disabled short_open_tag the referenced sniff is not producing any warnings (as it's hitting the same issue with `T_OPEN_TAG_WITH_ECHO` as we are ).

So as long as we already have something in place for checking on `T_INLINE_HTML`, it might be good to take an advantage of that and report on those as well.

david-binda:
I'll look into the failing regex, tho `phpunit` is processing that w/o issues on the PHP 5.3 vargrant I'm using :-/

david-binda:
Gah, turned out to be a bad copy-pasting 😊

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -152,6 +175,24 @@ public function process_token( $stackPtr ) {
 			if ( in_array( $function, array( 'trigger_error', 'user_error' ), true ) ) {
 				$end_of_statement = $this->phpcsFile->findEndOfStatement( $open_paren + 1 );
 			}
+		} else if ( false === $this->short_open_tag_enabled && T_INLINE_HTML === $this->tokens[ $stackPtr ]['code'] ) {
+			// Skip if no PHP short_open_tag is in the string.
+			if ( false === strpos( $this->tokens[ $stackPtr ]['content'], '<?=' ) ) {
+				return;
+			}
+
+			// Report on what very likely is a PHP short open tag outputing variable.
```
jrfnl: Darn.. missed this before: `outputing` => `outputting`

WordPress/Sniffs/XSS/EscapeOutputSniff.php
jrfnl: Just checked and you're right. I thought I'd pull that upstream, but seems I only included that logic when I pulled the `Generic.PHP.DisallowAlternativePHPTags` sniff.

So, yes. let's leave the warning in place for now until the upstream sniff has been improved.
You may want to open an issue for this upstream.

jrfnl:
@david-binda Thanks for all your efforts on this. This PR can now be merged.

As we're about to release 0.11.0, however, I suggest waiting to merge this PR until after the release so it can be thoroughly tested before being included in the next release.

JDGrimes:
> As we're about to release 0.11.0, however, I suggest waiting to merge this PR until after the release so it can be thoroughly tested before being included in the next release.

+1

Great work @david-binda.

jrfnl:
@david-binda Could you please rebase the PR on the current `develop` branch ? I expect the unit tests to pass without problem after that and will merge the PR.

WordPress/Sniffs/XSS/EscapeOutputSniff.php
jrfnl: Upstream issue opened: https://github.com/squizlabs/PHP_CodeSniffer/issues/1398

WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -152,6 +175,24 @@ public function process_token( $stackPtr ) {
 			if ( in_array( $function, array( 'trigger_error', 'user_error' ), true ) ) {
 				$end_of_statement = $this->phpcsFile->findEndOfStatement( $open_paren + 1 );
 			}
+		} else if ( false === $this->short_open_tag_enabled && T_INLINE_HTML === $this->tokens[ $stackPtr ]['code'] ) {
```
jrfnl: Just noticed one last minor thingie: the first part of this `if` condition is not necessary as `T_INLINE_HTML` is only registered to this sniff if `short_open_tag_enabled` is `true`.

Having said that, that also means that the property can be removed and turned into a local variable within the `register()` method.

WordPress/Sniffs/XSS/EscapeOutputSniff.php
jrfnl: I've opened an upstream PR to fix the `Generic.PHP.DisallowAlternativePHPTags` sniff https://github.com/squizlabs/PHP_CodeSniffer/pull/1400

jrfnl:
Hi @david-binda, just wondering if you saw the last few messages. Would be great if you could rebase the PR and fix the tiny last note. I look forward to having this in WPCS :heart_eyes_cat:

WordPress/Sniffs/XSS/EscapeOutputSniff.php
jrfnl: Oh and before I forget: the upstream PR was merged and will be included in the next version of PHPCS, so the additional message can be removed once we up the minimum PHPCS version.

jrfnl:
PR #1010 supersedes this PR and takes care of the issues which were still open.

PR #861, opened 2017-03-02
Hotfix for master: Up the minimum PHPCS version to 2.8.1

jrfnl:
PHPCS 2.8.1 has just been released and contains a fix for security vulnerability which affects the `WordPress-Extra` ruleset. See PR #860 

I'd like to suggest releasing v 0.10.1 with this fix as soon as this has been merged.

(Can we still merge to master from another branch than `develop` ?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ php:
 
 env:
   - PHPCS_BRANCH=master
-  - PHPCS_BRANCH=2.6.0
+  - PHPCS_BRANCH=2.8.1
 
 matrix:
   fast_finish: true
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
 		}
 	],
 	"require"    : {
-		"squizlabs/php_codesniffer": "^2.6"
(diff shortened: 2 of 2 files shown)


jrfnl:
Looks like the build will fail unless we include the fix for #696 (#703)

JDGrimes:
That seems like a lot to backport. Would it be easier just to release 0.11.0, or is that a bad idea/not ready yet?

GaryJones:
> Would it be easier just to release 0.11.0

We're well overdue for it, so I'd be +1 for 0.11.0, even if 0.12.0 comes along in a few weeks with the almost-ready stuff.

jrfnl:
> Would it be easier just to release 0.11.0, or is that a bad idea/not ready yet?

I was thinking along those same lines.

The changelog for 0.11.0 will need some careful crafting as there are a number of BC breaks in this release - but for end-users using custom rulesets as well as for developers building sniffs with WPCS as base -, but other than that, I see no reason not to release 0.11.0 over the next few days.

jrfnl:
Closing as superseded by the release of version 0.11.0.

PR #866, opened 2017-03-06
Changelog for version 0.11.0

jrfnl:
Ok, finished... & ready for review :tada: 

Looking at the changelog now, I kind of think we may need to release new versions more often...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,128 @@ This projects adheres to [Semantic Versioning](http://semver.org/) and [Keep a C
 
 _Nothing yet._
 
+## [0.11.0] - 2017-03-08
+
+### Important notes for end-users:
+
+If you use the WordPress Coding Standards with a custom ruleset, please be aware that some of the checks have been moved between sniffs and that the naming of a number of error codes has changed.
+If you exclude some sniffs or error codes, you may have to update your custom ruleset to be compatible with WPCS 0.11.0.
+
+Additionally, to make it easier for you to customize your ruleset, two new wiki pages have been published with information on the properties you can adjust from your ruleset:
+* [WPCS customizable sniff properties](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties)
+* [PHPCS customizable sniff properties](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Customisable-Sniff-Properties)
+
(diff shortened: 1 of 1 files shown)


GaryJones:
> I kind of think we may need to release new versions more often

We originally said release should be on a monthly basis.

CHANGELOG.md
```diff
@@ -8,6 +8,128 @@ This projects adheres to [Semantic Versioning](http://semver.org/) and [Keep a C
...
+- The `WordPress.Arrays.ArrayDeclaration` sniff has been split into two sniffs: `WordPress.Arrays.ArrayDeclaration` and `WordPress.Arrays.ArrayDeclarationSpacing` for better compatibility with PHPCS upstream.
+- The `WordPress.Arrays.ArrayDeclaration` sniff has been synced with the PHPCS upstream version to get the benefit of some bug fixes and improvements which had been made upstream since the sniff was originally copied over.
+- The `WordPress.VIP.FileSystemWritesDisallow`, `WordPress.VIP.TimezoneChange` and `WordPress.VIP.SessionFunctionsUsage` sniffs now extend the `WordPress_AbstractFunctionRestrictionsSniff`.
+- Property handling of custom properties set via a custom ruleset where the property is expected to be set in array format (`type=""array"`) has been made more lenient and will now also handle properties passed as a comma delimited lists correctly. This affects all customizable properties which expect array format.
```
JDGrimes: Extra quote in `type=""array"`

JDGrimes:
> We originally said release should be on a monthly basis.

Monthly might be too quick sometimes, depending on how active contribution is. But as long as there are a few significant changes at the end of the month, I think we should go ahead and drop a release. I just need to be better about actually setting aside the time to do it (sometimes easy, sometimes not).

CHANGELOG.md
jrfnl: Thanks for spotting that one. Fixed now.

jrfnl:
Releases based on time often become quite forced. What about something along the lines of: if there are twenty (arbitrary number) or more merged PRs and no on-going PR-chains (like the recent one with all the internal changes building onto each other and having to be pulled one by one as they would conflict) ?

jrfnl:
> I've not checked for what might be missing

FYI: I went through all the merged PRs, but I've grouped the entries for numerous PRs together where appropriate and - logically - not made separate entries for bugs in features which were freshly introduced in this release.
I've glossed over a lot of the internal changes, while still trying to give enough pointers for sniff devs building onto WPCS.

PR #870, opened 2017-03-09
Make the impact of the errorcode change slightly less.

jrfnl:
A number of sniffs use capitalization in the errorcodes to improve readability. This was unnecessarily undone by the `string_to_errorcode()` implementation.

For the sniffs were the errorcodes have not necessarily changed while using this function, this maintains BC instead of breaking it.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -686,7 +686,7 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	 * @return string
 	 */
 	protected function string_to_errorcode( $base_string ) {
-		return preg_replace( '`[^a-z0-9_]`i', '_', strtolower( $base_string ) );
+		return preg_replace( '`[^a-z0-9_]`i', '_', $base_string );
 	}
 
 	/**


jrfnl:
P.S.: build failure has nothing to do with this PR, but with a problem with PHPUnit on Travis for PHP7+

JDGrimes:
Even though the build failure is unrelated, I won't be able to merge this (from GitHub's UI) until it is fixed. I guess we probably want to fix that before 0.11.0 anyway? I assume the issue is just that the PHPUnit version has been upped and we may need to manually install a lower version, or something like that?

jrfnl:
> I guess we probably want to fix that before 0.11.0 anyway?

We kind of need to as otherwise we won't be able to merge the changelog or `develop` into `master`.

> I assume the issue is just that the PHPUnit version has been upped and we may need to manually install a lower version, or something like that?

We currently use the PHPUnit version included by default with the Travis PHP environments and don't install PHPUnit ourselves. So this is a Travis environment issue.
See: https://twitter.com/jrf_nl/status/839845035214524418

We could start requiring PHPUnit ourselves via composer, but it will (at least) double the build time.

westonruter:
We could just grab the `phpunit` phar instead. That should be fast. Alternatively we could install it via composer and then make sure that `vendor` is cached by Travis. Thoughts?

jrfnl:
@westonruter It's not as simple as that as we'd need different versions of PHPUnit for the different PHP versions we're testing against. The latest PHPUnit version has a minimum PHP requirement of PHP 7.

JDGrimes:
Here is what WordPress core is doing: https://core.trac.wordpress.org/ticket/40086

Basically, installing the PHPUnit version manually.

jrfnl:
As this affects PHPCS and all other PHPCS standards as well, I'm working on a fix to pull upstream for this.

jrfnl:
Upstream fix has been pulled: https://github.com/squizlabs/PHP_CodeSniffer/pull/1384

jrfnl:
As upstream is taking its time, I've pulled #873 as a temporary fix in the mean time to get the builds passing again.

PR #873, opened 2017-03-18
:rocket: Temporary solution for build failure.

jrfnl:
As PHPCS upstream is taking its time to decide whether or not to make the test suite compatible with PHPUnit 6 or not, we might as well get the builds passing again in the mean time, so we can get the 0.11.0 release out the door.

This is intended (fingers crossed) as a temporary solution to be reverted once the testsuite for PHPCS is cross-version compatible with PHPUnit.

I've chosen to *only* download PHPUnit when we need a different version than the one installed by default in the Travis images to keep the build time down.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,13 +41,18 @@ matrix:
 
 before_script:
     - export PHPCS_DIR=/tmp/phpcs
+    - export PHPUNIT_DIR=/tmp/phpunit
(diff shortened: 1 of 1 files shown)


jrfnl:
Thanks. Rebasing the other PRs for the 0.11.0 build now & will merge once they have passed (all already have been approved).

jrfnl:
FYI: The upstream PR is (finally) getting some traction. If & when this gets merged, I will revert this commit & make any necessary changes (= include the upstream bootstrap in our phpunit bootstrap).

See: squizlabs/PHP_CodeSniffer/pull/1384

PR #883, opened 2017-03-21
Exclude test files from class name rule

### LINKED ISSUE #882: WordPress.Files.FileName.InvalidClassFileName for tests/test-sample.php
See https://travis-ci.org/wp-cli/sample-plugin/jobs/213413061

jrfnl:
It is common practice to name test file classes `test-something.php`. This conflicts with the strict checking for `class-classname.php` filenames.

This PR changes the behaviour of the `FilenameSniff` to exclude classes which extend (or are) well-known unit test classes from the class name check.

The `custom_test_class_whitelist` property should now also be available to `FilenameSniff` to allow for whitelisting additional test classes.
See: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#global-variable-overloads-in-unit-tests

Notes:
* Split the `is_token_in_test_method()` method into two methods to be able to re-use the logic.
* Added the PHPUnit6 base test class to the whitelist.
* The `FileNameSniff` class now extends the `WordPress_Sniff` class to use the new `is_test_class()` method.
* Leveraged use of the class.
* Don't apply strict class name rules for test files.

Includes unit tests.

Fixes #882

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -487,6 +487,7 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	protected $test_class_whitelist = array(
 		'WP_UnitTestCase'            => true,
 		'PHPUnit_Framework_TestCase' => true,
+		'PHPUnit\Framework\TestCase' => true,
 	);
 
 	/**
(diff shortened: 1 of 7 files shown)


westonruter:
YES! Oh yes. I've had to turn this sniff off for test files: https://github.com/WP-API/wp-api-customize-endpoints/blob/f17ebdbf26437aca3e4e6dca09286c83b846374e/phpcs.ruleset.xml#L26-L28

jrfnl:
@westonruter You only had to ask ;-) This wasn't difficult to implement at all.

PR #885, opened 2017-03-21
XSS EscapeOutput: various tweaks

jrfnl:
Reduce false positives and minor efficiency & speed improvements.

Notes:
* Add the one magic constant `__DIR__` which was still missing from the list. _Actually, there is one more, but that one `ClassName::class` does not have it's own token so is a lot harder to sort out_
* Move safe cast tokens to a class property.
* Add the `unset` cast (which is rather useless, but is very safe when used in an echo statement ;-)
* Move safe component tokens to a class property.
* Add the `T_DNUMBER` (floats) token to the safe list.
* Add the Nowdocs related tokens to the safe list.
* Add missing arithmetic operator tokens to the safe list.

Includes unit tests.


WordPress/Tests/XSS/EscapeOutputUnitTest.php
```diff
@@ -63,6 +63,7 @@ public function getErrorList() {
 			205 => 1,
 			206 => 1,
 			207 => 1,
+			212 => ( PHP_VERSION_ID < 50300 ) ? 1 : 0, // PHPCS on PHP 5.2 does not recognize T_NOWDOC.
```
GaryJones: I see this version compare being used quite a bit. Would it make sense to have a constant defined for this magic number? Or have a `is_php_at_least_53()` function which returns a boolean?

jrfnl: A straight compare with a constant is one of the fastest operations in PHP.

That's why I backported the `PHP_VERSION_ID` constant in PHPCS so it could be used here instead of having to make two function calls `version_compare( phpversion(), '5.3', '>' ) )`.

PHPCS min requirement is PHP 5.1 and `PHP_VERSION_ID` wasn't introduced until PHP 5.2.7 ;-)

PR #889, opened 2017-03-21
Fix persistent bug in VariableRestrictions abstract

jrfnl:
While unit testing something else I ran into this bug.

Basically, the sniff allows for adding several arrays with things to be sniffed for, however when examining a `T_DOUBLE_QUOTED_STRING` or - when I add this (waiting in the backwings) - a `T_HEREDOC`, only the first set of patterns is taken into account instead of all possible patterns.

This in effect meant that the sniff suffered from false negatives.

The original intend seems quite clear what with the $patterns being merged in every part of the control structure. The bug has been in the sniff from the very beginning. De-coupling the `if` statements solves this.

Additionally, within the last `if` structure, `$token['bracket_closer']` is referenced, while this array key will not be available when a `T_DOUBLE_QUOTED_STRING` or `T_HEREDOC` is encountered.
The `$var` setting for this is already corrected further along in the code, so making this part conditional fixes the `Undefined index` notice being thrown.

### DIFF
--- a/WordPress/AbstractVariableRestrictionsSniff.php
+++ b/WordPress/AbstractVariableRestrictionsSniff.php
@@ -130,7 +130,9 @@ abstract class WordPress_AbstractVariableRestrictionsSniff extends WordPress_Sni
 				$patterns = array_merge( $patterns, $group['variables'] );
 				$var      = $token['content'];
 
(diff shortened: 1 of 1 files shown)


jrfnl:
@JDGrimes Yes, I have those ready in a follow up PR which addresses heredoc, but that one would conflict with this one, which is why I haven't pulled it yet.

PR #891, opened 2017-03-22
I18n: Add limited support for heredoc and nowdoc recognition.

jrfnl:
More importantly, add unit tests for these syntaxes.


JDGrimes:
I doubt that most parsers that extract translations strings from PHP support heredoc and nowdoc though. So we might want to look into that and if so, give a warning when heredoc and nowdoc are used in this context.

jrfnl:
@JDGrimes Tht's what the "limited" support is about. The sniff will throw a 'not single literal string' error when it encountes a heredoc or nowdoc, so we should be fine.

PR #897, opened 2017-03-26
Add 'slow query' whitelist flag

### LINKED ISSUE #894: Possible slow query warnings for meta queries have no whitelist flags
Possible slow query warnings for meta queries have no whitelist flags, 
Should these be added?

barryceelen:
Adds the `slow query` whitelist flag to better cover the cases the `SlowDBQuerySniff` applies to.
The `tax_query` flag it aims to replace is still respected.

Closes #894

### DIFF
--- a/WordPress/Sniffs/VIP/SlowDBQuerySniff.php
+++ b/WordPress/Sniffs/VIP/SlowDBQuerySniff.php
@@ -42,7 +42,7 @@ class WordPress_Sniffs_VIP_SlowDBQuerySniff extends WordPress_AbstractArrayAssig
 	 * Processes this test, when one of its tokens is encountered.
 	 *
 	 * @since 0.10.0
-	 * @since 0.12.0 Introduced new 'slow_query' whitelist comment.
+	 * @since 0.12.0 Introduced new 'slow query' whitelist comment.
 	 *
 	 * @param int $stackPtr The position of the current token in the stack.
 	 *
--- a/WordPress/Sniffs/VIP/SlowDBQuerySniff.php
+++ b/WordPress/Sniffs/VIP/SlowDBQuerySniff.php
@@ -42,7 +42,7 @@ class WordPress_Sniffs_VIP_SlowDBQuerySniff extends WordPress_AbstractArrayAssig
 	 * Processes this test, when one of its tokens is encountered.
 	 *
 	 * @since 0.10.0
-	 * @since 0.12.0 Added the 'slow query' comment.
+	 * @since 0.12.0 Introduced new 'slow_query' whitelist comment.
 	 *
 	 * @param int $stackPtr The position of the current token in the stack.
 	 *
--- a/WordPress/Sniffs/VIP/SlowDBQuerySniff.php
+++ b/WordPress/Sniffs/VIP/SlowDBQuerySniff.php
@@ -42,6 +42,8 @@ class WordPress_Sniffs_VIP_SlowDBQuerySniff extends WordPress_AbstractArrayAssig
 	 * Processes this test, when one of its tokens is encountered.
 	 *
 	 * @since 0.10.0
+	 * @since 0.12.0 Added the 'slow query' comment.
+	 *
 	 *
 	 * @param int $stackPtr The position of the current token in the stack.
 	 *
@@ -49,6 +51,10 @@ class WordPress_Sniffs_VIP_SlowDBQuerySniff extends WordPress_AbstractArrayAssig
 	 *                  normal file processing.
 	 */
 	public function process_token( $stackPtr ) {
+
+		if ( $this->has_whitelist_comment( 'slow query', $stackPtr ) ) {
+			return;
+		}
 
 		if ( $this->has_whitelist_comment( 'tax_query', $stackPtr ) ) {
 			return;
--- a/WordPress/Sniffs/VIP/SlowDBQuerySniff.php
+++ b/WordPress/Sniffs/VIP/SlowDBQuerySniff.php
@@ -51,11 +51,12 @@ class WordPress_Sniffs_VIP_SlowDBQuerySniff extends WordPress_AbstractArrayAssig
 	 */
 	public function process_token( $stackPtr ) {
 
-		if ( $this->has_whitelist_comment( 'slow query', $stackPtr ) ) {
-			return;
-		}
+		if (
+			$this->has_whitelist_comment( 'slow query', $stackPtr )
+			||
+			$this->has_whitelist_comment( 'tax_query', $stackPtr )
+		) {
 
-		if ( $this->has_whitelist_comment( 'tax_query', $stackPtr ) ) {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


WordPress/Sniffs/VIP/SlowDBQuerySniff.php
```diff
@@ -42,6 +42,7 @@ public function getGroups() {
 	 * Processes this test, when one of its tokens is encountered.
 	 *
 	 * @since 0.10.0
+	 * @since 0.12.0 Added the 'slow query' comment.
```
jrfnl: This sounds a little non-descript. What about something like _Introduced new `slow_query` whitelist comment_ ?

WordPress/Sniffs/VIP/SlowDBQuerySniff.php
```diff
@@ -50,6 +51,10 @@ public function getGroups() {
 	 */
 	public function process_token( $stackPtr ) {
 
+		if ( $this->has_whitelist_comment( 'slow query', $stackPtr ) ) {
+			return;
+		}
+
 		if ( $this->has_whitelist_comment( 'tax_query', $stackPtr ) ) {
 			return;
```
jrfnl: Should a warning be thrown here that indicating that the `tax_query` whitelist comment is deprecated ?

(if not, you may want to fold the two blocks into one)

WordPress/Tests/VIP/SlowDBQueryUnitTest.inc
```diff
@@ -28,10 +28,10 @@ $test = array(
...
 	// Single-line statements.
 	'tax_query' => array(), // Bad.
-	'tax_query' => array(), // WPCS: tax_query ok.
+	'tax_query' => array(), // WPCS: slow query ok.
 
 	// Multi-line statement.
-	'tax_query' => array( // WPCS: tax_query ok.
+	'tax_query' => array( // WPCS: slow query ok.
 		array(
 			'taxonomy' => 'foo',
```
jrfnl: I'd like to suggest having at least one example with the `tax_query` whitelist comment so that the fact that it's still respected is covered by the tests.
In case a deprecation warning would be thrown for it, this would still result in a warning on that line (in contrast to `slow_query` which would have no warning).

WordPress/Sniffs/VIP/SlowDBQuerySniff.php
barryceelen: Should the `tax_query` flag be respected indefinitely, throwing a warning seems like an unnecessary nag for people using the current flag to me.

If throwing a warning is considered best practice in a case like this (I haven’t found a precedent for this in this repo) I’d be happy to add that.

(For the time being I just combined the two ifs into one).

jrfnl: @barryceelen Let's see what the others say about this. /cc @JDGrimes @westonruter @GaryJones 

There are some instances in WPCS where we have deprecated properties before (not whitelist comments AFAIK) and those generally do throw a warning.
For an example, see: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/NamingConventions/ValidVariableNameSniff.php#L313

JDGrimes: The only reason to issue a warning is if we want to remove support for the `tax_query` whitelist comment at some point. It is saying, "Hey, this is deprecated and is going to stop working soon."

Since checking for a whitelist comment does add some overhead (maybe negligible, but still), I'd say that we should probably plan on removing it eventually. Assuming that it is rarely used anyway, the effort to update to the new comment shouldn't be too much for folks. If that assumption is wrong though, then I'd be OK keeping it, I guess. But otherwise, we should add a warning.

jrfnl: If it helps, I can run a quick search of the plugin repo to see how often the whitelist comment is used within those 50.000+ plugins to get some indication. (won't be able to run this until Wednesday though, still in Rome at the moment)

jrfnl: A quick search on public repos on GH, shows only 13 files using it: https://github.com/search?p=1&q=%22tax_query+ok%22&type=Code&utf8=%E2%9C%93

barryceelen: I’m new to this stuff so bear with me: tried just dropping a warning inside the if statement but it gets returned twice. Can’t seem to figure out why that is happening :-(

```
if ( $this->has_whitelist_comment( 'tax_query', $stackPtr ) ) {
	$message = 'The \'tax_query\' whitelist flag is deprecated in favor of \'slow query\'.';
	$this->addMessage( $message, $stackPtr, false );
	return;
}
```

jrfnl: @barryceelen I'll have a look when I'm back home if you like.

```php
$this->addMessage( $message, $stackPtr, false );
```

You can just use a straight `$this->phpcsFile->addWarning()` for this.
Also: please add a distinct error code to the message, something like `DeprecatedWhitelistFlagFound`

barryceelen: @jrfnl Thanks for your help. I’ll go ahead and include the warning in this pull request, it would be great if you could take a look whenever you find the time.

jrfnl: @barryceelen Hang on... just thinking.. the sniff probably gets triggered twice on that same snippet of code, causing the double warning.
To get round that, we'd need to return the token to the whitelist comment to skip past the other tokens on that line.
As it's within an array assignment, you can't use the `$phpcs->findEndOfStatement()` method, so we may need to look into changing the whitelist check method to return `false` when no whitelist comment is found (like it already does) and the `int` token pointer to the whitelist comment in case a whitelist comment is found which could then be passed via return.

Mind: this is just some random free thinking about it without looking at the code, so there may well be a simpler way to do this.

jrfnl: Regarding whether or not we'd be ok to deprecate the "old" whitelist tag in the future:

I just ran a search for `tax_query ok` on a mirror of both the theme repository as well as the plugin repository.

Results:
* themes: no usage found in 4.000+ themes
* plugins: used in 1 file in 1 plugin out of 55.000+ plugins

Of course, this is excluding any themes and plugins not hosted on wordpress.org, but I think it gives a clear indication of how little this whitelist comment is currently used :sunglasses:

WordPress/Sniffs/VIP/SlowDBQuerySniff.php
```diff
@@ -50,7 +51,12 @@ public function getGroups() {
 	 */
 	public function process_token( $stackPtr ) {
 
-		if ( $this->has_whitelist_comment( 'tax_query', $stackPtr ) ) {
+		if (
+			$this->has_whitelist_comment( 'slow query', $stackPtr )
+			||
+			$this->has_whitelist_comment( 'tax_query', $stackPtr )
```
JDGrimes: This `if` statement needs to be split back up and a warning added when the `tax_query` comment is found.

jrfnl:
Superseded by #966

Issue #900, opened 2017-03-31
Add sniff to check function prefixes and such


jrfnl:
I'm thinking of developing a sniff to verify that all global functions, classes, global variables and global constants are prefixed.
_[Edit]: and hook names ;-)_

Obviously this sniff would be irrelevant for core, but I believe it would be a good addition for `Extra`.

For the sniff to work, a property would need to be set with the expected prefix. I imagine allowing this property to be set both from a custom ruleset as well as from the command-line, similar to how we currently do so for the `text_domain` in the I18n sniff.

I imagine this sniff would be interesting for both theme as well as plugin devs + the theme review team. /cc @grappler 

Opinions ?

dingo-d:
For the theme review team, the prefixing could be set as the text domain of the theme. In that case you could pull that information out from `style.css` or even `functions.php` file (`load_theme_textdomain`).

In that case we would have to make that a rule in the handbook about prefixing things.

jrfnl:
> the prefixing could be set as the text domain of the theme

@dingo-d I'd been thinking about that and while this may be possible to "demand" from a theme review perspective, the two don't necessarily need to be the same.

Just to use a popular plugin as an example: the `text_domain` for Yoast SEO is `wordpress-seo` (also note the dash which would not be allowed in a prefix) and the function prefixes they mostly use are `yoast` and `wpseo`.

So for now, I'd like to keep these properties separate.

> In that case you could pull that information out from style.css or even functions.php file (load_theme_textdomain).

That's something for the theme check plugin / theme review command line tool, not for the sniff.

> In that case we would have to make that a rule in the handbook about prefixing things.

I think this wouldn't be a bad thing. Maybe put it on the agenda for the next theme review board meeting ?

dingo-d:
@jrfnl I know that text domains don't have to be made from the single line, but I was thinking that maybe, for the themes specifically, we could request that they are the same. That would make things easier for new theme check plugin to use.

In the case of a text domain with hyphens, these could all be converted to underscores (`_`) and used as a prefix.

I'll ask in the slack channel and see if this can be added to the agenda for next week :)

grappler:
I have been thinking about this previously.

As a default we can use the root directory name and convert it to underscores. For example `my-theme` would be `my_theme`. It would be good to be able to overwrite it (or perhaps extend) with multiple prefixes as the theme or plugin may be using libraries.

[...]

PR #904, opened 2017-04-01
Improve WP global vars list

jrfnl:
* Removed one variable which seemed out of place `$var` (separate commit).
* Ordered the WP global variable list alphabetically to make future maintenance easier.
* Changed the array to `key => true` style to allow for using `isset()` instead of `in_array()`.
* Changed the property name to `$wp_globals` to be more descriptive.


GaryJones:
I think a comment explaining what the `=> true` is for (or not) would be handy.

GaryJones:
> This is the same pattern which has by now been applied to most of these kind of arrays. The value actually doesn't matter. It's just so the in_array() checks can be replaced by the faster isset() for the same result.

I get that, but someone who hasn't come across this pattern before will be needlessly searching the code trying to work out what the boolean value is for.

jrfnl:
Ok, step back - this PR was created as a stepping stone for two additional commits:
* Moving the list to `WordPress_Sniff` as I intend to also use it in the "Prefix all things" sniff (#900). The move would be part of that PR.
* Updating the list in a separate PR.

I can see that I should have probably have mentioned that before ;-)


> I get that, but someone who hasn't come across this pattern before will be needlessly searching the code trying to work out what the boolean value is for.

It's not that I'm not willing to do so, however it would be totally inconsistent with the documentation of all the other lists.
Currently, `true` is used everywhere as a value just to allow for using `isset()`. This value is never documented **_unless_** there is some "hidden" meaning.
See the following two places where the value *does* have meaning (and is therefore documented):
* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniff.php#L463
* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/CSRF/NonceVerificationSniff.php#L24

If I would document the value for this list, I believe we should document it for *all* these type of lists and in that case, it would be better IMHO to do that in a separate commit/PR for all lists using this kind of undocumented value in one go.

Alternatively, we could add documentation to that effect to the class doc block of the `WordPress_Sniff` class, explaining the reasoning for setting up all those lists in this way and not document it with the individual properties.

Opinions ?

> Probably we should go ahead and update the list, no?

> Edit: The list was created by searching for all occurrences of global $ and $GLOBALS[' in core, and then extracting the variable name(s) from each line. Possibly could have missed some from multi-line global declarations, I guess, probably best to double-check each removal (if we remove any).

I agree the list should be updated, but this needs some careful checking throughout the core code which is why I wanted to leave it for a later point in time.

While searching for `global` and `$GLOBALS` gives us a good starting point (Thanks for that @JDGrimes !), any variable created in the global namespace and not unset should be on the list.
Luckily core has become slightly better over the years at not creating globals for everything, but there will be old variables around all over the place.

All the same, for the original purpose of the list, there is at least one I know of that is not on the list with reason:
* `content_width` was removed as it is *intended* to be overridden by themes /cc @grappler 

So in that case, if we want to have a complete list in the `WordPress_Sniff` class, we should probably have a small `$overrule_allowed` list in the Global Var Override sniff and do an `array_diff()` to prime the list for usage in that sniff.

Altogether, updating the list to be complete will be a mountainous task which I'd rather not take on at this time.

Some more references about WP global vars (which may well be out of date):
* https://codex.wordpress.org/Global_Variables
* https://codex.wordpress.org/WPMU_Global_Variables

JDGrimes:
> - Updating the list in a separate PR.
> 
> I can see that I should have probably have mentioned that before ;-)

Yep, but that's OK. It definitely makes sense to do it in a separate PR, so we can discuss things further there once it is opened.



> Alternatively, we could add documentation to that effect to the class doc block of the WordPress_Sniff class, explaining the reasoning for setting up all those lists in this way and not document it with the individual properties.

I think this is all that is necessary. At least for now, unless or until we get complaints of people being confused by each property not documenting this.

jrfnl:
> > Alternatively, we could add documentation to that effect to the class doc block of the WordPress_Sniff class, explaining the reasoning for setting up all those lists in this way and not document it with the individual properties.
>
> I think this is all that is necessary. At least for now, unless or until we get complaints of people being confused by each property not documenting this.

This has now been addressed by PR #906

PR #907, opened 2017-04-01
:sparkles: New sniff to verify that everything in the global namespace is prefixed.

### LINKED ISSUE #900: Add sniff to check function prefixes and such

jrfnl:
_**Important**: This sniff will not do anything **unless** the `prefixes` property has been added to the custom ruleset or provided via the command line. (See below for usage examples)_

This new sniff intends to verify that all functions, classes, interfaces, traits, variables, constants and hook names which are declared/defined in the global namespace are prefixed with one of the `prefixes` provided via a custom property or via the command line.

Prefixing is best practice to prevent naming collisions what with there being so many plugins and themes out there in the wider WordPress arena.

The sniff has been added to the `WordPress-Extra` ruleset.
I can imagine that the VIP team might be interested in this sniff as well. Please let me know if it should also be added to the `WordPress-VIP` ruleset. /cc @sboisvert @david-binda
For the Theme review team, this will/should be sorted out in their fork. /cc @grappler.

This sniff has the following features:
* Allows for more than one prefix to be set as some plugins/themes use libraries and those may use their own prefix.
* It will validate the passed prefixes against the PHP naming rule for characters allowed to be used.
* It will throw an error if someone tries to use `wp` or `_` as a prefix. This list can be extended with other prefixes which should not be used.
* Prefix vs _name_ comparisons are always done in a case-insensitive manner as the typical case for variables (lowercase), constants (uppercase), class names (Camelcase), function names (lowercase) are all different.
* It is allowed to have a _name_ which is just and only the prefix.
* If the _name_ is used as a prefix, it is expected to be separated from the rest of the _name_ by an underscore `_`
* There are some valid use-cases for using non-prefixed _names_.  Such as back-fills for PHP functions later introduced, calling a WP core hook etc. To this end, use of non-prefixed _names_ can be whitelisted via a whitelist comment. For some examples of use-cases, see the unit test file.
* Anything in classes extending known unit test classes is exempted from this sniff as this is, generally speaking, code which will not be executed during a WP run and therefore will not cause conflicts.
    Additional unit test classes can be whitelisted using the `custom_test_class_whitelist` property.
* The PHP superglobals and all known WP globals are whitelisted from prefixing of the variable name.
* Hook names clearly indicated as deprecated through usage of the `do_action_deprecated()` or `apply_filters_deprecated()` functions will be disregarded.
* Anything outside of the global namespace will be disregarded by this sniff.

To do if/when this PR is merged:
- [ ] Add information on the use of the `custom_test_class_whitelist` property with this sniff to the wiki.
- [ ] Add information about the new whitelist comment to the wiki.

Fixes #900

----
#### Method 1: via a custom ruleset

Add the following to a custom ruleset replacing the values for the property with your own.
```xml
<rule ref="WordPress.NamingConventions.PrefixAllGlobals">
	<properties>
		<property name="prefixes" type="array" value="my_prefix,tgmpa" />
	</properties>
</rule>
```

#### Method 2: via the command line

Pass the prefixes you want to check against via the command line, like so:
```bash
phpcs -p . --standard=WordPress --runtime-set prefixes my_prefix,tgmpa
```


WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -0,0 +1,503 @@
...
+					$condition = $this->phpcsFile->getCondition( $stackPtr, T_CLOSURE );
+				}
+
+				$has_global = $this->phpcsFile->findPrevious( T_GLOBAL, ( $stackPtr - 1 ), $this->tokens[ $condition ]['scope_opener'] );
+				if ( false === $has_global ) {
+					// No variable import happening.
+					return;
+				}
+
+				// Ok, this may be an imported global variable.
```
grappler: The alignment of the comment seems off.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -0,0 +1,503 @@
...
+		        }
+
+		        for ( $ptr = ( $has_global + 1 ); $ptr <= $end_of_statement; $ptr++ ) {
+		            $imported = $this->phpcsFile->findNext( T_VARIABLE, $ptr, $end_of_statement, false, null, true );
+
+		            if ( false === $imported ) {
+						// Reached the end of the global statement without finding the variable,
+						// so this must be a local variable.
+						return;
+					}
```
grappler: There seem to be a few extra indent tabs here.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -0,0 +1,503 @@
...
+
+		            if ( substr( $this->tokens[ $imported ]['content'], 1 ) === $variable_name ) {
+						break;
+					}
+
+					// Move the stack pointer forward for the next loop.
+					$ptr = $imported;
```
grappler: Dito as above.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -0,0 +1,503 @@
...
+		        }
+
+		        unset( $condition, $has_global, $end_of_statement, $ptr, $imported );
+
+			} // End if().
+
+			// Properties in a class do not need to be prefixed.
```
grappler: Dito...

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -0,0 +1,503 @@
...
+	        $conditions = array_keys( $this->tokens[ $stackPtr ]['conditions'] );
+	        $ptr        = array_pop( $conditions );
+	        if ( isset( $this->tokens[ $ptr ] )
+	            && in_array( $this->tokens[ $ptr ]['code'], array( T_CLASS, T_ANON_CLASS, T_TRAIT ), true )
+	        ) {
+				return;
+	        }
+
+	        // Function parameters do not need to be prefixed.
+	        if ( isset( $this->tokens[ $stackPtr ]['nested_parenthesis'] ) ) {
```
grappler: Spaces used instead of tabs.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
jrfnl: Thanks for catching those tabs vs spaces issues. Should be all fixed now.

I though there was a sniff which was supposed to check this ?

jrfnl: > I though there was a sniff which was supposed to check this ?

Checked & there is, but turns out that sniff had some bugs in it when there was a mix of tabs & spaces on a line.

Fixed & pulled upstream: https://github.com/squizlabs/PHP_CodeSniffer/pull/1404, so hopefully these type of issues will soon be caught by the travis builds ;-)

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -0,0 +1,503 @@
...
+	 * Prevents having to do the same prefix validation over and over again.
+	 *
+	 * @since 0.12.0
+	 *
+	 * @var string[]
+	 */
+	private $previous_prefixes = array();
+
+	/**
+	 * A list of superglobals that incorporate user input.
```
JDGrimes: This description appears to by copy-pasted, and need to be updated I think.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -0,0 +1,503 @@
...
+				foreach ( $this->tokens[ $stackPtr ]['nested_parenthesis'] as $opener => $closer ) {
+					if ( isset( $this->tokens[ $opener ]['parenthesis_owner'] ) && T_FUNCTION === $this->tokens[ $this->tokens[ $opener ]['parenthesis_owner'] ]['code'] ) {
+						return;
+					}
+				}
+			}
+		} // End if().
+
+		// Ignore superglobals and WP global variables.
+		if ( isset( $this->superglobals[ $variable_name ] ) || isset( $this->wp_globals[ $variable_name ] ) ) {
```
JDGrimes: This condition can be moved up higher in the function (but may need to be repeated?), which would improve performance by bypassing above logic that in this case is unnecessary. The `$variable_name` is not modified in the above `else`, only in the `if`. So probably the best way to do this would be to split the `if and `else` and place this condition between them.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -0,0 +1,503 @@
...
+			return;
+		}
+
+		if ( $this->is_prefixed( $variable_name ) === true ) {
```
JDGrimes: Ditto for this condition.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -0,0 +1,503 @@
...
+		if ( $this->previous_prefixes === $this->prefixes ) {
+			return;
+		}
+
+		// Set the cache *before* validation so as to not break the above compare.
+		$this->previous_prefixes = $this->prefixes;
+
+		// Validate the passed prefix(es).
+		foreach ( $this->prefixes as $key => $prefix ) {
+			$prefixLC = strtolower( $prefix );
```
JDGrimes: Ironic variable naming, there. 😄

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -0,0 +1,503 @@
...
+			$this->prefixes[ $key ] = $prefixLC;
+		}
+
+		// Set the validated prefixes cache.
+		$this->validated_prefixes = $this->prefixes;
+
+	} // End validate_prefixes().
+
+	/**
+	 * Check if a function/class/constant/variable name is prefixed with one of the expected prefixed.
```
JDGrimes: `s/expected prefixed./expected prefixes./`

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
jrfnl: Good point. I've done some minor refactoring to allow for this.

jrfnl:
@JDGrimes Thanks for the feedback. I've made the fixes in a separate commit to make them easier to review.

If all is ok, ping me and I'll squash the last commit into the previous one before this PR is merged.

JDGrimes:
Looks good @jrfnl. [The changes give a small performance increase](https://blackfire.io/profiles/compare/c30f8c96-aa73-4755-aa90-cd69d3c84d6a/graph).

jrfnl:
@JDGrimes Thanks, I've squashed the last commits into one, so ready for merge when you are ;-)

PR #909, opened 2017-04-04
:sparkles: New CapitalPDangitSniff

### LINKED ISSUE #876: Core: Capital P dangit sniff
Just a thought: should we add a sniff which checks whether `WordPress` is spelled correctly wherever used (comments, text strings, class/function names etc) ? :sunglasses:

jrfnl:
This PR adds a new sniff which checks whether `WordPress` is spelled correctly in text strings, comment text and class/interface/trait names and can catch a wide range of typical misspellings.

For this sniff, I've erred on the side of caution as this is **_very_** prone to false positives.
This means that some misspellings will likely be missed by this sniff, but that seemed the better choice than to have everyone be terribly annoyed by the sniff being really buggy.

Notes:
* The sniff has been added to the `WordPress-Core` ruleset.
* The sniff will throw a `warning` when a misspelling is found.
* The sniff will only throw one warning per token, even when more than one misspelling is found in the same phrase. The warning will in that case contain a count of the number of misspellings found + list them.
* For misspellings in text strings and comments, the sniff includes a fixer. For class names and the like, no fixer is included as that will always need attention from a developer.
* The fixer has been primed to be able to deal with text strings which contain exceptions as well as misspellings which should be corrected, and will deal with this properly.
* The sniff adds a new whitelist comment `spelling`, though this can obviously not be used to whitelist misspellings in comments.

The sniff takes the following situations into account and will ignore these when encountered:
- URLs for wordpress.org/com/net/tv.
- `@...` usernames starting with `wordpress`
- email addresses with a domain starting with `wordpress`
- email addresses with a user name ending with `wordpress` (typical default email address set up by core)
- (most) variable names (as used in docblock @param comments).
- directory paths containing a folder starting or ending with `wordpress`.
- file names containing `wordpress` for a limited set of extensions.
- `wordpress` prefixed or suffixed with dashes as those are indicators that the
   term is probably used as part of a CSS class, such as `fa-wordpress`
   or filename/path like `class-wordpress-importer.php`.
- back-tick quoted `wordpress` (to help with deliberate misspellings in comments)
- URL query strings containing `wordpress`
- Any docblock `@link` tag content (if not previously ignored already)
- HTML attributes containing `wordpress`
- escape quoted `\'wordpress\'` as used in xpath queries.

The sniff will ignore *anything* within an array, as well as string array keys like `$var['wordpress']` as these will generate false positives in 80% of the cases.
The sniff will skip forward to the end of the array when one is encountered. This also prevents the sniff "hanging" when processing large configuration arrays.

The sniff is accompanied by extensive unit tests and has been run over both WP core as well as some 180+ recently updated themes (7500+ php files) to try and eliminate most false positives.

The only "common" (1 instance in core) false positive I've not eliminated is when the term is used in a parameter in a function call. Excluding that would also exclude examining text strings passed to translation functions which would be counter-productive.

I expect we'll see some more false positives reports over the next few months, but am hopeful this will be a good base to start from.

Fixes #876

To do if/when this PR is merged:
- [ ] Add information about the new whitelist comment to the wiki.

### DIFF
--- a/WordPress/Tests/WP/CapitalPDangitUnitTest.inc
+++ b/WordPress/Tests/WP/CapitalPDangitUnitTest.inc
@@ -1,0 +1,173 @@
+<?php
+
+// Bad: This is a comment with WORDPRESS spelled incorrectly.
+
+/* Bad: This is a comment with wordpress spelled incorrectly. */
+
+/*
+ * Bad: This is a comment with Wordpress spelled incorrectly.
+ */
+
+/**
+ * Function comment
+ *
+ * @param string $wordpress OK: Here the incorrect spelling is OK to comply with the variable name rules.
+ */
+function something( $wordpress ) {} // OK.
+
+/**
+ * Function comment
+ *
+ * @param string $my_wordpress_test OK: Here the incorrect spelling is OK to comply with the variable name rules.
+ */
+function something( $my_wordpress_test ) {} // OK.
+
+/**
+ * Bad: In this comment wordPresss should be fixed.
+ *
+ * @param string $test Bad: In this comment word press should be fixed.
+ */
+function something( $test ) {} // OK.
+
+function wordpress_function() {} // OK - comply with the function name rules.
+
+class Wordpress_Something {} // Bad.
+class Something_Word_press_Something {} // Bad.
+class Something_wordpressss {} // Bad.
+class WordPress_Something {} // OK.
+class Something_WordPress {} // OK.
+
+echo 'This is an explanation about wordpress.'; // Bad.
+echo "This is an {$explanation} about wordpress."; // Bad.
+
+// Bad.
+echo <<<EOD
+This is an {$explanation} about wordpress.
+EOD;
+
+echo 'http://wordpress.org/something'; // OK - part of a URL.
+?>
+
+<div class="copyright"><?php printf( wp_kses_post( __( 'Powered by <a href="%s">WordPress</a>', 'theme-slug' ) ), 'https://wordpress.org/' ); ?></div><!-- OK. -->
+
+<p>Here we have an inline HTML tag with wordpress spelled incorrectly.</p><!-- Bad. -->
+<p>Here we have an inline HTML tag with WordPress spelled correctly.</p><!-- OK. -->
+
+<p>Here we have an inline HTML tag with a URL, of course this should be coded differently, but even when it isn't, it should be ignored, so here goes: http://wordpress.org/ spelled incorrectly.</p><!-- OK. -->
+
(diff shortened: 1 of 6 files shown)


WordPress/Tests/WP/CapitalPDangitUnitTest.inc
```diff
@@ -0,0 +1,173 @@
...
+?>
+
+<div class="copyright"><?php printf( wp_kses_post( __( 'Powered by <a href="%s">WordPress</a>', 'theme-slug' ) ), 'https://wordpress.org/' ); ?></div><!-- OK. -->
+
+<p>Here we have an inline HTML tag with wordpress spelled incorrectly.</p><!-- Bad. -->
+<p>Here we have an inline HTML tag with WordPress spelled correctly.</p><!-- OK. -->
+
+<p>Here we have an inline HTML tag with a URL, of course this should be coded differently, but even when it isn't, it should be ignored, so here goes: http://wordpress.org/ spelled incorrectly.</p><!-- OK. -->
+
+<p>Here we have an inline HTML tag with wordpressers spelled incorrectly.</p><!-- OK, part of another word. -->
```
GaryJones: I'd like to think that this one would still be picked up, and correctable to WordPressers.

grappler: What about WordPressing? 😄

jrfnl: The checking is current bound to complete words to prevent false positives, so `wordpress` as part of a larger word will not match.

I don't think we could blanket fix any larger words, though we could possible create a target list of common words which contain `wordpress` which could and should be fixed.

This target list will need some careful consideration, but suggestions of larger words which could be on there are welcome.

ntwb: What about `WordPress'`, I'm often writing that whereas many others write `WordPress's`?

JDGrimes: @ntwb Technically, `WordPress's` is correct. The trailing `s` after the apostrophe is only eliminated [if the word ends in `s` already *and is plural*](http://www.dailywritingtips.com/the-possessive-apostrophe/). `WordPress`, of course, is singular.

jrfnl: I would like to suggest we open a new issue for this.
The issue can be used to gather a list of other typical misspellings which should be warned about/potentially fixed. This information gathered there could then be used as a basis to enhance this sniff at a later point in time.

Opinions ?

PR #911, opened 2017-04-04
Update the `project.ruleset.xml` example file.

jrfnl:
* Recommend using the `WordPress-Extra` + `WordPress-Docs` ruleset rather than `WordPress-Core`.
* Add example to exclude the Composer vendor directory.
* Remove some of the "extra" included rules which are already included by the WP rulesets (and move the rest of them down).
* Add info and example on how to include the `PHPCompatibility` standard. We already recommend its use in the readme, so we might as well show how.
* Add a number of the recommended custom properties to the example ruleset.
* Add/Clarify some documentation.

### DIFF
--- a/project.ruleset.xml.example
+++ b/project.ruleset.xml.example
@@ -2,6 +2,7 @@
 <ruleset name="Example Project">
 	<description>A custom set of rules to check for a WPized WordPress project</description>
 
+	<!-- Exclude select folders and files from being checked. -->
 	<exclude-pattern>/docroot/wp-admin/*</exclude-pattern>
 	<exclude-pattern>/docroot/wp-includes/*</exclude-pattern>
 	<exclude-pattern>/docroot/wp-*.php</exclude-pattern>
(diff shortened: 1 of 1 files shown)


project.ruleset.xml.example
```diff
@@ -10,25 +11,74 @@
...
+	<rule ref="Generic.CodeAnalysis.UnusedFunctionParameter"/>
+	<rule ref="Generic.Commenting.Todo"/>
+
+	<!-- Check for PHP cross-version compatibility. -->
+	<!--
+	To enable this, you need to make sure that the PHPCompatibility
+	standard is installed.
+	See the readme for installation instructions:
+	https://github.com/wimg/PHPCompatibility
+	<!--
```
JDGrimes: Shouldn't a `-->` come before that?

project.ruleset.xml.example
```diff
@@ -10,25 +11,74 @@
...
+		the different Sniffs, as their rules are broken. From here, we can opt to
 		exclude problematic sniffs like so.
 		-->
 
 		<exclude name="WordPress.WhiteSpace.ControlStructureSpacing" />
 		<exclude name="WordPress.XSS.EscapeOutput" />
 	</rule>
+
+	<!-- Let's also check that everything is properly documented. -->
+	<rule ref="WordPress-Docs" >
```
JDGrimes: It isn't important, but I don't think the extra space before `>` is intentional here.

I think we need a code-style checker for our rulesets. 😄

jrfnl: :sunglasses: PHPCS does not have a schema attached to rulesets, so we wouldn't be able to check against that. All the same, if someone wants to have a go, I suppose setting up a travis command to run `xmllint` on the rulesets, should be possible.
http://xmlsoft.org/xmllint.html

project.ruleset.xml.example
jrfnl: Good catch! Fixed.

project.ruleset.xml.example
jrfnl: Made the endings consistent across the file now.

project.ruleset.xml.example
```diff
@@ -10,25 +11,75 @@
...
+		the different Sniffs, as their rules are broken. From here, we can opt to
 		exclude problematic sniffs like so.
 		-->
 
 		<exclude name="WordPress.WhiteSpace.ControlStructureSpacing" />
 		<exclude name="WordPress.XSS.EscapeOutput" />
 	</rule>
+
+	<!-- Let's also check that everything is properly documented. -->
+	<rule ref="WordPress-Docs">
```
GaryJones: This needs a closing `/` to stop the file being not well-formed.

project.ruleset.xml.example
GaryJones: The extra space was intentional - but it was a `/` that was missing from it.

project.ruleset.xml.example
jrfnl: Thanks for catching this. I'm clearly working on too many different things at the same time ;-)

jrfnl:
The build isn't passing due to other reasons - see #912

GaryJones:
#912 merged, so this PR needs rebasing so Travis will pass.

jrfnl:
Let's merge #914 first and I'll rebase after that as that one will conflict anyway.

I've already locally made the necessary (extra) adjustments to pass the linter check and made a minor extra change to the new documentation. Other documentation uses "we", while I wrote it using "you", so I've adjusted the new documentation to be more inline with the existing.

I'll rebase those changes & push after #914

jrfnl:
That was quick ;-) Ok, rebased & pushed. Please have a quick look at the adjusted documentation to see if it's ok as those adjustments were made after @GaryJones and @JDGrimes gave their approval for this PR.

PR #914, opened 2017-04-05
Lint the XML ruleset files and check them for consistent code style

jrfnl:
Add checking of the XML files to the build testing as discussed here: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/911#discussion_r109780017

Notes:
* Runs only once per build (`SNIFF=1`)
* Checks for valid XML
* Checks for consistent codestyle of the XML by doing an auto-format and comparing that against the original file.

I've tested this extensively with deliberately malformed XML and incorrect formatting and am confident this will work as intended.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ matrix:
     - env: PHPCS_BRANCH=3.0
 
 before_install:
+    - export XMLLINT_INDENT="	"
     - export PHPCS_DIR=/tmp/phpcs
     - export PHPUNIT_DIR=/tmp/phpunit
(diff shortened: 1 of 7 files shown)


JDGrimes:
For clarity, the code style is as follows:

- Tabs must be used for indenting.
- Self-closing tags must not have a space before the `/>`.

jrfnl:
@JDGrimes Yes, thanks, I should have mentioned that.

For prosperity:
* Tabs as the rest of the project uses tabs everywhere as well, so that is most inline with the WP codestyle.
* No space in self-closing tag: that's just how the auto-formatter is set up and not configurable ;-)
* Indentation has to be 1 level for each nesting. 

From the second commit msg regarding rule 3 and changes to comply with this codestyle:
> The Core ruleset used additional indentation to show which rules belonged with each handbook paragraph, making the ruleset more easily scannable.
> 
> As this extra indentation violates the formatting rules, the handbook paragraph headers have been given "fencing" to maintain the scannability of the document.

PR #918, opened 2017-04-07
Add Squiz.Classes.SelfMemberReference to WordPress-Core

westonruter:
This will flag the following as errors:

```php
Self::do_something();
self:: do_something();
self ::do_something();
self :: do_something();
```

Note that it excludes the `NotUsed` error code which would flag the following:

```php
MyClassName::do_something();
```

Since there are instances of this in the core codebase, for better or worse.

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -189,6 +189,13 @@
 	<!-- Covers rule: ... array items, only include a space around the index if it is a variable. -->
 	<rule ref="WordPress.Arrays.ArrayKeySpacingRestrictions"/>
 
+	<!-- References to self in a class should be lower-case and not have extraneous spaces,
+		 per implicit conventions in the core codebase; the NotUsed code refers to using the
+		 fully-qualified class name instead of self, for which there are instances in core. -->
+	<rule ref="Squiz.Classes.SelfMemberReference" />
+	<rule ref="Squiz.Classes.SelfMemberReference.NotUsed">
+		<severity>0</severity>
+	</rule>
 
 	<!--
 	#############################################################################
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -189,14 +189,6 @@
 	<!-- Covers rule: ... array items, only include a space around the index if it is a variable. -->
 	<rule ref="WordPress.Arrays.ArrayKeySpacingRestrictions"/>
 
-	<!-- References to self in a class should be lower-case and not have extraneous spaces,
-		 per implicit conventions in the core codebase; the NotUsed code refers to using the
-		 fully-qualified class name instead of self, for which there are instances in core. -->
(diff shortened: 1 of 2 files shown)


WordPress-Core/ruleset.xml
```diff
@@ -189,6 +189,13 @@
 	<!-- Covers rule: ... array items, only include a space around the index if it is a variable. -->
 	<rule ref="WordPress.Arrays.ArrayKeySpacingRestrictions"/>
 
+	<!-- References to self in a class should be lower-case and not have extraneous spaces,
+		 per implicit conventions in the core codebase; the NotUsed code refers to using the
+		 fully-qualified class name instead of self, for which there are instances in core. -->
+	<rule ref="Squiz.Classes.SelfMemberReference" />
```
jrfnl: Code style: `" />` => `"/>` (no space)

WordPress-Core/ruleset.xml
```diff
@@ -189,7 +189,6 @@
 	<!-- Covers rule: ... array items, only include a space around the index if it is a variable. -->
 	<rule ref="WordPress.Arrays.ArrayKeySpacingRestrictions"/>
 
-
```
jrfnl: I don't think you meant to remove this line here or did you ?

WordPress-Core/ruleset.xml
```diff
@@ -356,6 +356,13 @@
 	<!-- Class opening braces should be on the same line as the statement. -->
 	<rule ref="Generic.Classes.OpeningBraceSameLine"/>
 
+	<!-- References to self in a class should be lower-case and not have extraneous spaces,
+		 per implicit conventions in the core codebase; the NotUsed code refers to using the
+		 fully-qualified class name instead of self, for which there are instances in core. -->
+	<rule ref="Squiz.Classes.SelfMemberReference"/>
+	<rule ref="Squiz.Classes.SelfMemberReference.NotUsed">
+		<severity>0</severity>
```
GaryJones: We know WP doesn't follow half of what's in this main ruleset, so why should this exception cater just for WP Core? It would mean that every single other plugin and theme project who wanted to enable the check would have to explicitly set it, having looked up what severity it should be.

Could it not be left enabled, and let a fictional http://develop.svn.wordpress.org/trunk/phpcs.xml.dist disable it?

What are the chances of getting a `phpcs.xml.dist` added to core, that could handle these "WordPress doesn't do it right yet" cases, and let this `ruleset.xml` be the showcase example of how every plugin and theme in the community should be doing it?

jrfnl: @GaryJones The parts of the sniff which have been enabled for `Core` deal with whitespace only, which is quite clear-cut.

Whether or not to use `self` vs a `\Fully\Qualified\Classname` is something else altogether.

So independently of whether Core currently complies with it's own rules (in which neither of the above are covered), I do believe the `NotUsed` code is better placed in `Extra` than in `Core` as it is opinionated/best practice.

PR #926, opened 2017-04-09
AbstractFunctionRestrictions: Efficiency tweak.

jrfnl:
As there are by now quite a number of sniffs which use the `AbstractFunctionRestrictions` class as the basis and `T_STRING` tokens are quite common, the "_checking whether a T_STRING is a function call_" part might be done ten or more times for each `T_STRING` token.

By adding a preliminary check whether the token content matches one of the targeted functions, we can bow out early and prevent a lot of this overhead.

It would be even better if we could use a straight array with `isset()`, but unfortunately that won't work here as the list of functions can contain function names with wildcards.

Having this extra check *does* mean we have to decouple the `AbstractClassRestrictionsSniff::process_token()` method from the parent as for the class sniffs we don't know the class name yet at that point in time.

@JDGrimes Can you work your magic with the speed test on this PR ? I'm definitely curious to see how much this helps (presuming it does...).

### DIFF
--- a/WordPress/AbstractClassRestrictionsSniff.php
+++ b/WordPress/AbstractClassRestrictionsSniff.php
@@ -17,7 +17,7 @@
 abstract class WordPress_AbstractClassRestrictionsSniff extends WordPress_AbstractFunctionRestrictionsSniff {
 
 	/**
-	 * Regex pattern with placeholder for the function names.
+	 * Regex pattern with placeholder for the class names.
 	 *
 	 * @var string
 	 */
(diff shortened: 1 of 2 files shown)


JDGrimes:
🚀 This improves performance by as much as 15%: https://blackfire.io/profiles/compare/d8843cc9-c09a-4f0b-868f-318db3f0f90c/graph

comparison_from_untitled_to_untitled_-_blackfire_-_2017-04-10_09 51 02

jrfnl:
@JDGrimes Thanks for that! I was hoping it would ;-)

Issue #927, opened 2017-04-11
Quotes Remain on File (string) When Passed from IDE


kopepasah:
When an IDE (such as Atom) uses `--stdin-path=` to pass the parameter to `phpcs` the quotes remain on the beginning and end of the string, causing errors within the IDE for `WordPress_Sniffs_Files_FileNameSniff`.

Quotes (single and double), should not be on the beginning or end of the string, especially since the `substr( $fileName, 0, -4 )` conditional is meant to remove `.php` from the `$fileName`.

PR #928, opened 2017-04-11
Trim Quotes (single and double) from $file.

kopepasah:
See #927.

### DIFF
--- a/WordPress/Sniffs/Files/FileNameSniff.php
+++ b/WordPress/Sniffs/Files/FileNameSniff.php
@@ -127,7 +127,7 @@ class WordPress_Sniffs_Files_FileNameSniff extends WordPress_Sniff {
 	 */
 	public function process_token( $stackPtr ) {
 
-		$file     = $this->phpcsFile->getFileName();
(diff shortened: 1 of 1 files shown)


WordPress/Sniffs/Files/FileNameSniff.php
```diff
@@ -127,7 +127,7 @@ public function register() {
 	 */
 	public function process_token( $stackPtr ) {
 
-		$file     = $this->phpcsFile->getFileName();
+		$file     = trim( $this->phpcsFile->getFileName(), "\x22\x27" );
```
jrfnl: WPCS actually contains a utility function to do quote stripping properly & consistently, you can call it using `$this->strip_quotes( $file )`.

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniff.php#L865-L867

kopepasah:
@jrfnl updated to use the `strip_quotes` method and added an inline comment.

jrfnl:
@kopepasah Thanks for the update. Looks good to me.

JDGrimes:
Thanks @kopepasah !

PR #934, opened 2017-04-18
Merge non validated and non sanitized errors

david-binda:
The ValidatedSanitizedInput Sniff is currently reporting two different errors for the same code, in case the global variable is not sanitized nor validated. This commit is merging the two errors into one, in caes both conditions are met. It's preserving the individual errors for cases when only one of them is true.

We have been using this improvement on the WordPress.com VIP for a while already (introduced in https://github.com/Automattic/VIP-Coding-Standards/pull/12 )

cc. @grappler

### DIFF
--- a/WordPress/Sniffs/VIP/ValidatedSanitizedInputSniff.php
+++ b/WordPress/Sniffs/VIP/ValidatedSanitizedInputSniff.php
@@ -82,6 +82,8 @@ class WordPress_Sniffs_VIP_ValidatedSanitizedInputSniff extends WordPress_Sniff 
 	 */
 	public function process_token( $stackPtr ) {
 
+		$notValidated = $notSanitized = false;
+
 		$superglobals = $this->input_superglobals;
 
 		// Handling string interpolation.
@@ -122,8 +124,7 @@ class WordPress_Sniffs_VIP_ValidatedSanitizedInputSniff extends WordPress_Sniff 
 
 		// Check for validation first.
 		if ( ! $this->is_validated( $stackPtr, $array_key, $this->check_validation_in_scope_only ) ) {
-			$this->phpcsFile->addError( 'Detected usage of a non-validated input variable: %s', $stackPtr, 'InputNotValidated', $error_data );
-			// return; // Should we just return and not look for sanitizing functions ?
+			$notValidated = true;
 		}
 
 		if ( $this->has_whitelist_comment( 'sanitization', $stackPtr ) ) {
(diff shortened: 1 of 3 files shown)


WordPress/Sniffs/VIP/ValidatedSanitizedInputSniff.php
```diff
@@ -122,8 +125,7 @@ public function process_token( $stackPtr ) {
 
 		// Check for validation first.
 		if ( ! $this->is_validated( $stackPtr, $array_key, $this->check_validation_in_scope_only ) ) {
-			$this->phpcsFile->addError( 'Detected usage of a non-validated input variable: %s', $stackPtr, 'InputNotValidated', $error_data );
-			// return; // Should we just return and not look for sanitizing functions ?
+			$notValidated = true;
 		}
 
 		if ( $this->has_whitelist_comment( 'sanitization', $stackPtr ) ) {
```
JDGrimes: Because of this if condition and the one below it, if sanitization isn't needed the validation error would no longer be given either. This logic needs to be updated to fix that. (And also the unit tests, so that they'd catch this.)

david-binda: Oh, good catch! I'll look into it, thanks!

WordPress/Sniffs/VIP/ValidatedSanitizedInputSniff.php
```diff
@@ -82,6 +82,9 @@ public function register() {
 	 */
 	public function process_token( $stackPtr ) {
 
+		$notValidated = false;
+		$notSanitized = false;
```
jrfnl: Just wondering if the code would be easier to read if these variables would be reversed in meaning, i.e. 
```php
$validated = true;
$sanitized = true;
```

Also wondering whether the default should be `true` (or `false` as the current code stands), though changing that might mean more extensive changes to the logic of the sniff are needed.

GaryJones:
Has merge conflicts.

jrfnl:
I've had another look at this and IMO this change will make the sniff **_less_** easy to work with for end-users and should not be merged.

The error codes, as they are, address separate problems which can currently be ignored via the ruleset or via inline comments separately using the appropriate error code.
By combining the error codes in certain cases, it makes it more fiddly for people to ignore issues.

For example, say someone has non-validated, non-sanitized code, but has in another method already validated and sanitized all superglobal keys they use. In that case, they could ignore the new `InputNotValidatedNotSanitized` code.

Now, in a refactor, they move the sanitization to the places where the superglobals are used. As the error code will now change, all the ignores will need to be updated, while in the original situation, this would not be necessary and the ignores (covering both original codes) would still work as expected.
   
With that in mind, I'm going to suggest closing this PR.

Aside from that:
1. No response was ever received on the last feedback round.
2. It's been in a merge conflict state for a pretty long time.
3. This a breaking change which would have to go into a major release as it changes the error code for this sniff.

jrfnl:
Closing for lack of response/progress.

PR #935, opened 2017-04-24
OperatorSpacing: Ignore operators in declare statements

jrfnl:
Declare statements are typically written without spaces in them.

This commit makes the `OperatorSpacing` sniff ignore them, similarly as is done upstream - https://github.com/squizlabs/PHP_CodeSniffer/pull/1411

It would be a good idea to do a complete compare and sync with upstream for this sniff at some point, but that's another matter.

### DIFF
--- a/WordPress/Sniffs/WhiteSpace/OperatorSpacingSniff.php
+++ b/WordPress/Sniffs/WhiteSpace/OperatorSpacingSniff.php
@@ -71,12 +71,14 @@ class WordPress_Sniffs_WhiteSpace_OperatorSpacingSniff implements PHP_CodeSniffe
 			}
 
 			// Skip default values in function declarations.
+			// Skip declare statements.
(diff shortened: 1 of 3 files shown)


GaryJones:
> Declare statements are typically written without spaces in them.

Since that's yet another hole in the Handbook standards, allowing `declare(strict_types=1);` and `declare( strict_types = 1 );` is a good thing.

JDGrimes:
> Declare statements are typically written without spaces in them.

Because the style is copied and pasted from php.net, and they don't use a spacey code style. IMHO, this isn't really a gap in the handbook. "When in doubt, space it out" takes in everything like this. 😈  

But if we don't want to be dogmatic about this right now, I'm fine with that.  I can imagine that WordPress might adopt spaceless codestyle for `declare` just because that is especially pervasive.

PR #936, opened 2017-05-01
Update sanitizing functions list

dingo-d:
Added `sanitize_textarea_field` to the list of the sanitizing functions.


JDGrimes:
This function wasn't on the list yet because it was added to WordPress in 4.7.0.

dingo-d:
Yeah, I didn't know about it, until djrmom mentioned it on slack :)

PR #939, opened 2017-05-04
Use latest PHP_CodeSniffer, fixes #938

### LINKED ISSUE #938: Be compatible with PHP_CodeSniffer v3
Dependency on the old version is expressed here:

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/master/composer.json#L14

However a new version was just released here:

https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.0.0

Therefore we are now depending on an old version.

fulldecent:


jrfnl:
This will not work. Suggest: close.

PR #941, opened 2017-05-06
:sparkles: New sniff to verify correct array indentation

jrfnl:
* Multi-line arrays should be indented by one tab from the beginning of the line containing the array opener.
* The array closer should be level with the indentation of the line containing the array opener.

While the `ScopeIndent` sniff checks this type of alignment for most structures, it turns out it doesn't do so for arrays (I suspect this is because PHPCS itself prefers a different array style).

This PR fixes that by adding a new sniff. Both the errors this sniff throws have been made fixable.

The new sniff has been added to the `WordPress-Core` ruleset.

Violations against this sniff in the WPCS codebase have been fixed.

N.B.: I consider this sniff a candidate for pulling upstream to `Generic` in due time.

### DIFF
--- a/WordPress/Sniffs/Arrays/ArrayIndentationSniff.php
+++ b/WordPress/Sniffs/Arrays/ArrayIndentationSniff.php
@@ -1,0 +1,178 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+/**
+ * Enforces WordPress array indentation for multi-line arrays.
+ *
+ * @link    https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   0.12.0
+ */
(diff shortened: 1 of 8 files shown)


jrfnl:
Hmm.. something weird going on with tabs vs spaces. Trying to fix it now, but can't reproduce the travis results locally.

WordPress/Sniffs/Arrays/ArrayIndentationSniff.php
```diff
@@ -0,0 +1,189 @@
...
+		if ( $this->tokens[ $opener ]['line'] === $this->tokens[ $closer ]['line'] ) {
+			// Not interested in single line arrays.
+			return;
+		}
+
+		/*
+		 * Determine the indentation of the line containing the array opener.
+		 */
+		$indentation = '';
+		$column 	 = 1;
```
grappler: The alignment is a bit off here.

JDGrimes: Looks to be an issue of tabs instead of spaces.

JDGrimes: Note that some assignments below are also aligned via tabs, but should just use spaces, IMO.

jrfnl: It was & is now fixes.

Yet another thing for which I'm wondering why there is no sniff to auto detect & fix this ;-)

jrfnl: Right, sorted: #942

PR #942, opened 2017-05-06
:sparkles: New sniff to verify that spaces are used for mid-line alignment

jrfnl:
> Tabs should be used at the beginning of the line for indentation, while spaces can be used mid-line for alignment.

This new sniff covers the second part of this rule. (The first part is already covered by the `Generic.WhiteSpace.DisallowSpaceIndent` sniff).

The error which is thrown is auto-fixable.

The new sniff has been added to the `WordPress-Core` ruleset.

Violations against this sniff in the WPCS codebase have been fixed.
Mostly involved `getGroups()` function example comments. Some of these have also been adjusted slightly to be in line with an earlier change (#812) in which these were missed and/or with the comment from the abstract parent class.

N.B.: I consider this sniff a candidate for pulling upstream to `Generic` in due time.

N.B.2: This sniff could probably do with some additional unit tests, but I couldn't think of any more, so ideas/code snippets welcome. (The run over the WPCS code base did not yield any incorrect fixes or false positives, so I'm not too worried)

### DIFF
--- a/WordPress/Sniffs/Functions/FunctionRestrictionsSniff.php
+++ b/WordPress/Sniffs/Functions/FunctionRestrictionsSniff.php
@@ -26,11 +26,11 @@ class WordPress_Sniffs_Functions_FunctionRestrictionsSniff extends WordPress_Abs
 	 * Groups of functions to restrict.
 	 *
 	 * Example: groups => array(
-	 * 	'lambda' => array(
-	 * 		'type'      => 'error' | 'warning',
-	 * 		'message'   => 'Use anonymous functions instead please!',
-	 * 		'functions' => array( 'eval', 'create_function' ),
-	 * 	)
+	 *  'lambda' => array(
+	 *      'type'      => 'error' | 'warning',
+	 *      'message'   => 'Use anonymous functions instead please!',
+	 *      'functions' => array( 'file_get_contents', 'create_function' ),
+	 *  )
 	 * )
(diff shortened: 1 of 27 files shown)


WordPress/Sniffs/Functions/FunctionRestrictionsSniff.php
```diff
@@ -26,11 +26,11 @@ class WordPress_Sniffs_Functions_FunctionRestrictionsSniff extends WordPress_Abs
...
 	 * Example: groups => array(
-	 * 	'lambda' => array(
-	 * 		'type'      => 'error' | 'warning',
-	 * 		'message'   => 'Use anonymous functions instead please!',
-	 * 		'functions' => array( 'eval', 'create_function' ),
-	 * 	)
+	 *  'lambda' => array(
+	 *      'type'      => 'error' | 'warning',
+	 *      'message'   => 'Use anonymous functions instead please!',
+	 *      'functions' => array( 'file_get_contents', 'create_function' ),
```
GaryJones: This seems to be an unrelated change?

jrfnl: See above: 
> Mostly involved getGroups() function example comments. Some of these have also been adjusted slightly to be in line with an earlier change (#812) in which these were missed and/or with the comment from the abstract parent class.

PR #949, opened 2017-05-15
Composer: add links to the wiki and source in the `support` section

jrfnl:


jrfnl:
@GaryJones Thanks! Fixed it. Sorry about that, but will probably keep happening regularly while I'm working on two standards with different whitespace requirements at the same time.

PR #950, opened 2017-05-15
Fix the builds for HHVM

jrfnl:
The builds against HHVM have been failing for quite a while. This fixes it for now.

Generally speaking there is no official support for PHPUnit on HHVM. Select versions do appear to work, but this is hit & miss.
The version used to fix it now, is based on going back through Travis build logs to find a previous HHVM build which did pass and checking the PHPUnit version used by Travis at the time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,16 +48,17 @@ before_install:
     - export PHPCS_BIN=$(if [[ $PHPCS_BRANCH == master ]]; then echo $PHPCS_DIR/bin/phpcs; else echo $PHPCS_DIR/scripts/phpcs; fi)
     - mkdir -p $PHPCS_DIR && git clone --depth 1 https://github.com/squizlabs/PHP_CodeSniffer.git -b $PHPCS_BRANCH $PHPCS_DIR
     - $PHPCS_BIN --config-set installed_paths $(pwd)
-    # Download PHPUnit 5.x for builds on PHP 7, nightly and HHVM as
-    # PHPCS test suite is currently not compatible with PHPUnit 6.x.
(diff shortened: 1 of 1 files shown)


ntwb:
I think WPCS should remove HHVM altogether 😄 

We removed it from core ~3 weeks ago https://core.trac.wordpress.org/ticket/40548:

> Maintaining the HHVM test infrastructure on Travis takes a disproportionately high amount of effort. Particularly because the HHVM build:
>
> • Is in the allowed failures list
> • Has never passed
> • Often errors
> Given the exceptionally low usage numbers of HHVM, support in the Travis test suite should be removed.
>
> The PHP world owes a lot to HHVM for helping it push it forward. Without HHVM, maybe we wouldn't have seen such incredible performance gains in PHP 7. Thank you, HHVM.

jrfnl:
> I think WPCS should remove HHVM altogether

WPCS !== WP. Some people may use HHVM as a dev environment - including for non-WP projects - and use the WP coding standards because they are used to them/like them etc.

HHVM builds *are* allowed to fail, so if they do, it's not a real issue.
And as far as I know, we've put no extra effort in to make things work on HHVM, so if it works, we may as well show it does for those who prefer to work on HHVM.

I suggest we revisit dropping HHVM support, if and when at some point it would start to be "work" to maintain compatibility with HHVM

jrfnl:
> So something broke HHVM in 5.7.x after 5.7.17?

Yes and in PHPUnit 5.6 after 5.6.6 I believe and .... It's kind of hit & miss, but Sebastian has made it very clear that HHVM is not officially supported, though he has made patches to accommodate now & again.

johnbillion:
It's worth noting that the primary reason core dropped support for HHVM in its test infrastructure is due to the disproportionately high amount of time needed to maintain them. Stephen Edgar and myself literally lost days trying to backport HHVM fixes. You might find that going forward you run into a similar problem.

jrfnl:
@johnbillion Thanks for that clarification. For the snifs HHVM is set to be allowed to _fail_ (won't break builds) and - so far - support hasn't costs any extra time (except for the few minutes I spend on fixing the travis build for this PR).

PR #958, opened 2017-06-07
Fix issue 957: allow for template files using content-{slug}.php format

### LINKED ISSUE #957: WordPress Files_FileNameSniff should ignore slugs in content-{slug}.php template files
We should have an exception for `content-{slug}.php` since the slug can contain '_' and format not allowed by the sniff.

The same way we have an exception for:
- `archive-{slug}.php` 
- ` single-{slug}.php`
- etc. 

See: https://developer.wordpress.org/themes/template-files-section/partial-and-miscellaneous-template-files/#content-slug-php

jrfnl:
Allow for the newly added `content-{$slug}.php` template file format.

Fixes #957

### DIFF
--- a/WordPress/Sniffs/Files/FileNameSniff.php
+++ b/WordPress/Sniffs/Files/FileNameSniff.php
@@ -38,6 +38,7 @@ class WordPress_Sniffs_Files_FileNameSniff extends WordPress_Sniff {
 	 * @link https://developer.wordpress.org/themes/basics/template-hierarchy/#custom-post-types
 	 * @link https://developer.wordpress.org/themes/basics/template-hierarchy/#embeds
 	 * @link https://developer.wordpress.org/themes/basics/template-hierarchy/#attachment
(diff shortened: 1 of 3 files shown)


guillaumemolter:
@jrfnl I would suggest:
- Adding `@link https://developer.wordpress.org/themes/template-files-section/partial-and-miscellaneous-template-files/#content-slug-php` in the docblock to be coherent with other exceptions.
- Adding the exception in `WordPress/Tests/Files/FileNameUnitTest.php` to be coherent with other exceptions.

Hope this helps!

jrfnl:
@guillaumemolter 

> Adding ... in the docblock

Good point. Done.

> Adding the exception in WordPress/Tests/Files/FileNameUnitTest.php to be coherent with other exceptions.

The `WordPress/Tests/Files/FileNameUnitTest.php` file only contains a list of the files within the `FileNameUnitTests` subdirectory for which an error should be thrown.
Both additional unit tests I've added, should now *not* throw an error (and this is tested!), they therefore should not be listed in the `WordPress/Tests/Files/FileNameUnitTest.php` file.

PR #969, opened 2017-06-16
Stop using wpcom_vip_get_term_by() instead of get_term_by()

alexiskulash:
As of WordPress 4.8, the core function `get_term_by()` is cached so we're going to stop recommending the use of `wpcom_vip_get_term_by()` instead. The function `wpcom_vip_get_term_by()`will be deprecated.

### DIFF
--- a/WordPress/Tests/DB/RestrictedFunctionsUnitTest.php
+++ b/WordPress/Tests/DB/RestrictedFunctionsUnitTest.php
@@ -46,6 +46,7 @@ class WordPress_Tests_DB_RestrictedFunctionsUnitTest extends AbstractSniffUnitTe
 			48 => 1,
 			49 => 1,
 			50 => 1,
+			51 => 1,
 
 			54 => 1,
 			55 => 1,
--- a/WordPress/Tests/DB/RestrictedFunctionsUnitTest.php
+++ b/WordPress/Tests/DB/RestrictedFunctionsUnitTest.php
@@ -38,7 +38,7 @@ class WordPress_Tests_DB_RestrictedFunctionsUnitTest extends AbstractSniffUnitTe
 			39 => 1,
 			40 => 1,
 			41 => 1,
-			42 => 1,
+			
 			43 => 1,
 			44 => 1,
 
@@ -46,7 +46,6 @@ class WordPress_Tests_DB_RestrictedFunctionsUnitTest extends AbstractSniffUnitTe
 			48 => 1,
 			49 => 1,
 			50 => 1,
-			51 => 1,
 
 			54 => 1,
 			55 => 1,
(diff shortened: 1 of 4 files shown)


david-binda:
@alexiskulash would you, please, look into updating the unit tests as well? Related code:

* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Tests/VIP/RestrictedFunctionsUnitTest.inc#L51
* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Tests/VIP/RestrictedFunctionsUnitTest.inc#L40

and

* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/a92b51b2afad95c31e7d9b8a0d78286d5434907b/WordPress/Tests/VIP/RestrictedFunctionsUnitTest.php#L38
* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/a92b51b2afad95c31e7d9b8a0d78286d5434907b/WordPress/Tests/VIP/RestrictedFunctionsUnitTest.php#L32

Further, what about `get_category_by_slug` ? It's using `get_term_by` as well internally:

https://developer.wordpress.org/reference/functions/get_category_by_slug/

We should adjust the restricted functions sniff accordingly as well.

Related code:

* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/c0d7578a80314ddbaf9d188553c1366d63f5180a/WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php#L96

alexiskulash:
@david-binda Thank you for the suggested fixes! I've implemented them and believe the changeset is now ready for a review.

WordPress/Tests/DB/RestrictedFunctionsUnitTest.php
```diff
@@ -46,7 +46,6 @@ public function getErrorList() {
 			48 => 1,
 			49 => 1,
 			50 => 1,
-			51 => 1,
```
jrfnl: This change was made in the wrong file (which is why the unit tests are still failing).
You need to change the `WordPress/Tests/VIP/RestrictedFunctionsUnitTest.php` file. This is the database restricted unit test file.

alexiskulash:
Thank you for catching that @jrfnl - fixed in the most recent commit.

jrfnl:
Side-note: if/when this PR is merged, the "[Customizable sniff properties](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#excluding-a-group-of-checks)" wiki page will need to be updated for the changed group names (`exclude` property).

- [ ] Update wiki page

JDGrimes:
This looks good to me. @david-binda would you like to review before merge?

mboynes:
Related, I believe you can kill `WordPress.VIP.RestrictedFunctions.get_term_link_get_term_link`

jrfnl:
FYI: customizable sniff properties wiki page has been updated to reflect this change

PR #972, opened 2017-06-17
:sparkles: New sniff: FunctionCallSignatureNoParams

### LINKED ISSUE #970: Faulty Core fixes: spaces between function call parenthesis
#### Summary:
A function call with no parameters should not have any spaces between the parenthesis.
The current fixer seems to remove some spaces, but still leaves one.

Most likely cause: `PEAR.Functions.FunctionCallSignature`

#### Example:

`./src/wp-settings.php`

Source:
```php
ms_cookie_constants(  ); // Line 282
wp_templating_constants(  ); // Line 393
```

Fixed as:
```php
ms_cookie_constants( );
wp_templating_constants( );
```

Expected fix:
```php
ms_cookie_constants();
wp_templating_constants();
```

jrfnl:
Report on and remove whitespace between function call parenthesis when the function call has no parameters.

This sniff *will* result in contradictory messages when run with PHPCS, but that can't really be helped.

```
 5 | ERROR | [x] Expected 1 spaces after opening bracket; 6 found
   |       |     (PEAR.Functions.FunctionCallSignature.SpaceAfterOpenBracket)
 5 | ERROR | [x] Expected 1 spaces before closing bracket; 6 found
   |       |     (PEAR.Functions.FunctionCallSignature.SpaceBeforeCloseBracket)
 5 | ERROR | [x] Function calls without parameters should have no spaces
   |       |     between the parenthesis.
   |       |     (WordPress.Functions.FunctionCallSignatureNoParams.WhitespaceFound)
```

I've chosen to put this sniff in the `Functions` category to be consistent with the related sniff in the `PEAR` standard.

Includes an auto-fixer.

This auto-fixer has been tested to not conflict with the `PEAR.Functions.FunctionCallSignature` sniff (at this time).

Fixes #970

N.B.: I've also opened upstream issue https://github.com/squizlabs/PHP_CodeSniffer/issues/1512 about this, but there are two reasons to pull this PR anyway:
1. Chances are that a fix for this won't make it into PHPCS 2.x which is where we currently need it.
2. The inconsistency I reported upstream can be fixed in two different manners. The most _consistent_ way to fix it would be to end up with 1 space between the parenthesis, which is not what we want.

As it is currently unclear how, when and in which branch this will be fixed upstream, pulling this as an interim solution seems appropriate. Depending on if & when this will be addressed upstream and in which branch, this sniff may be removed at a later date or may need adjusting to not conflict with the updated version of the PEAR sniff.

### DIFF
--- a/WordPress/Sniffs/Functions/FunctionCallSignatureNoParamsSniff.php
+++ b/WordPress/Sniffs/Functions/FunctionCallSignatureNoParamsSniff.php
@@ -1,0 +1,86 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+/**
+ * Enforces no whitespace between the parenthesis of a function call without parameters.
+ *
+ * @link    https://make.wordpress.org/core/handbook/coding-standards/php/#space-usage
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   0.12.0
+ */
+class WordPress_Sniffs_Functions_FunctionCallSignatureNoParamsSniff extends WordPress_Sniff {
+
+	/**
+	 * Returns an array of tokens this test wants to listen for.
+	 *
+	 * @return array
+	 */
+	public function register() {
+		return PHP_CodeSniffer_Tokens::$functionNameTokens;
+	}
+
+	/**
+	 * Processes this test, when one of its tokens is encountered.
+	 *
+	 * @param int $stackPtr The position of the current token in the stack.
+	 *
+	 * @return int Integer stack pointer to skip the rest of the file.
+	 */
+	public function process_token( $stackPtr ) {
+
+		// Find the next non-empty token.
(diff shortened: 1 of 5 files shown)


WordPress/Sniffs/Functions/FunctionCallSignatureNoParamsSniff.php
```diff
@@ -0,0 +1,86 @@
...
+			// Not a function call.
+			return;
+		}
+
+		// Find the previous non-empty token.
+		$search   = PHP_CodeSniffer_Tokens::$emptyTokens;
+		$search[] = T_BITWISE_AND;
+		$previous = $this->phpcsFile->findPrevious( $search, ( $stackPtr - 1 ), null, true );
+		if ( T_FUNCTION === $this->tokens[ $previous ]['code'] ) {
+			// It's a function definition, not a function call.
```
JDGrimes: Why skip if it is a function definition? Shouldn't they also follow this rule?

I'd think that we'd want to catch this as well:

```php
function some_func(   ) {}
```

Or am I missing a technical consideration?

jrfnl: For a function declaration, this is already reported on via the `Squiz.Functions.FunctionDeclarationArgumentSpacing.SpacingBetween` sniff/error code.

Excluding function declarations from the sniff in this pull is intended to prevent duplicate messages from being thrown.

Issue #731, opened 2016-12-15
Support checking for methods in addition to functions


grappler:
> As these are methods, not functions, I can imagine we might need a separate AbstractMethodsSniff / DeprecatedMethodsSniff to deal with this or alternatively, the existing AbstractFunctionRestrictionsSniff could be adjusted to make the distinction between methods and functions.
By @jrfnl https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/633#discussion_r91864236

These methods are deprecated and would need to be checked.

```php
WP_Filesystem_Base::find_base_dir(); // Bad, use WP_Filesystem::abspath instead.
WP_Filesystem_Base::get_base_dir(); // Bad, use WP_Filesystem::abspath instead.
WP_Widget_Recent_Comments::flush_widget_cache();
```

PR #979, opened 2017-06-17
DeprecatedFunctions sniff: update minimum WP version

jrfnl:
The version number in this property is supposed to be 3 versions before the current WP release and what with WP 4.8 having been released, should now be updated.

I'm not aware of any deprecated functions in WP 4.8, but have to admit that I haven't followed this very closely.

According to the official changelog, there are none: https://codex.wordpress.org/Version_4.8
And I can't seem to get a one-page diff view using trac: https://core.trac.wordpress.org/changeset?old_path=%2Ftrunk&old=40888&new_path=%2Fbranches%2F4.7&new=39524&sfp_email=&sfph_mail= (aside: would be great if there was a way to only get the diff of PHP files at that)

Does anyone know of a clever way to check for newly deprecated functions between releases ?


JDGrimes:
A search for `@deprecated 4.8` reveals `wp_dashboard_plugins_output()` was deprecated. Appears to be the only one.

jrfnl:
@JDGrimes Thanks for that. I've updated the PR with an additional commit adding the newly deprecated function & updating the unit tests. Feel free to squash the commits on merge.

jrfnl:
Oh dear... looks like - contrary to the earlier assertion as annotated in the code - there were actually deprecated functions in WP 4.7 as well.
Added these now in - yet another - separate commit.

I'm by no means claiming completeness, just did a quick search on `@deprecated 4.7` which would not have caught anything with more spaces between the `@deprecated` and `4.7`. So if anyone feels like doing a more extensive search, please do! and let me know what else you turn up :sunglasses:

JDGrimes:
I've created a tool to scan a WordPress codebase for all deprecated elements: https://github.com/JDGrimes/wp-deprecated-code-scanner

Currently just detects classes and methods based on the use of `_deprecated_function()`. Results seem to indicate that we may have missed some other deprecated functions as well.

Currently it just prints the results to the command line in markdown format. I'll see about generating some output with it that is easier to compare with what we have here in this sniff, which will make it easier to compare results and see what the differences are.

It is bleeding edge, so has some bugs still.

jrfnl:
@JDGrimes Wow! That sounds really useful - if only for this snifff ;-)

Might be an idea to add the other `_deprecated_...()` functions at some point as well. `_deprecated_argument()` will be useful for #826

For easiest compare, I think it would be helpful to have a consistent array order for each of this sniffs and maybe see if your tool can display the lists in the same order ?

I propose:
* Deprecated function - Ordered by WP version in which deprecation occurred & order alphabetically on function name within each subset. (I'll fix up this PR and see if anything in the original array needs changing too)
* Deprecated function arguments - As noted previously in the ticket, those should be ordered alphabetically to prevent a later deprecated second argument overruling a previous array entry for the same function.

JDGrimes:
I've created an issue for deprecated arguments: https://github.com/JDGrimes/wp-deprecated-code-scanner/issues/2. I do plan to add checks for hooks as well, etc.

>Deprecated function - Ordered by WP version in which deprecation occurred & order alphabetically on function name within each subset.

Exactly what I was thinking.

jrfnl:
> Exactly what I was thinking.

Re-ordering the array in the sniff as we speak ;-)

JDGrimes:
It should now be able to generate the output as needed:

```
bin/wpdcs run --format wpcs /path/to/wp/src
```

Oh, it will include methods though, not just regular functions.

jrfnl:
I've pushed a new commit which sorts out the re-ordering for the functions we already accounted for.
I've also added delimiters between each set within the test case file to make maintenance easier and adjusted the test functions accordingly.

Now we "just" need to do a compare between the output of your tool and the current list in the sniff. @JDGrimes Would you like to have a go at that ? You can add a new commit to this branch with the changes.

> Oh, it will include methods though, not just regular functions.

That's for another sniff - see https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/731

JDGrimes:
Here is an initial diff: https://www.diffchecker.com/BmXjilmq

It revealed several inconsistencies not related to different functions:

- For one, there are a few functions that don't have the versions specified correctly in the `_deprecated_function()` call, see https://core.trac.wordpress.org/ticket/41094.
- The versions in WordPress are almost always specified in full, like `4.8.0`, but only `4.8` in the sniff. I can change this in the generator, or we can change how we specify the versions in the sniff.
- Sometimes the alternative is specified differently than in the sniff now. For example, the sniff is currently inconsistent as to whether parentheses are included; in core they almost always are.
- There may be more methods in the sniff's list than we realized, these aren't included in the generated output.

jrfnl:
I also noticed some issue with the ordering of the functions - I did the ordering case-insensitively and the functions are declared in the array in lowercase (and compared as such as well).
Is it possible to adjust the ordering in your tool to be in line with that ?

> The versions in WordPress are almost always specified in full, like 4.8.0, but only 4.8 in the sniff. I can change this in the generator, or we can change how we specify the versions in the sniff.

I'm in favour of using complete version numbers in the sniff. When I have time, I'll update the sniff if nobody objects.

> Sometimes the alternative is specified differently than in the sniff now. For example, the sniff is currently inconsistent as to whether parentheses are included; in core they almost always are.

I'm in favour of using the parentheses in the sniff as well. I checked to see if this could be done generically via the error message, but with some alternatives passing arguments `add_theme_support( \'custom-background\', $args )`, it can't.
Again, if nobody objects and when I have some time, I'll update the sniff.

@JDGrimes Once I've done the sniff update, could you run a new compare to see how much is left ?

JDGrimes:
> I also noticed an issue with the ordering of the functions - I did the ordering case-insensitively and the functions are declared in the array in lowercase (and compared as such as well).
> Is it possible to adjust the ordering in your tool to be in line with that ?

I noticed this as well but forgot to mention it. I'll update the tool to lowercase all of the function names before ordering in this case.

> I'm in favour of using complete version numbers in the sniff.

+1

> I'm in favour of using the parentheses in the sniff as well.

+1 as well.



> @JDGrimes Once I've done the sniff update, could you run a new compare to see how much is left ?

Sure thing.

jrfnl:
Okidoki, updates to the version nrs + parentheses/alternatives all sorted.

JDGrimes:
Great, here is a new diff: https://www.diffchecker.com/E4xhROBg

There are still some differences that aren't just a result of functions appearing in one list and not in the other, particularly in regard to alternative functions:

- When a function has an alternative that isn't a function (like "the Loop", for `start_wp()`), WordPress doesn't specify that as the third parameter of the `_deprecated_function()` call. So in the sniff it will currently have a value, but the generator will not detect the alternative.
- There are some cases where the alternative function in the sniff is incorrect. For example, it should be `wp_editor()` for `tinymce_include()`, but is listed as `wp_tiny_mce()`. I haven't checked other cases to be sure that they are correct.
- In some cases the spacing for the alternative functions is different. `get_the_author_meta( \'description\' )` in the sniff vs `get_the_author_meta(\'description\')` from the generator.

It also appears that not all deprecated functions call `_deprecated_function()`, so the generator currently doesn't pick them up. I'll see about updating it so that it will check for the `@deprecated` dockblock annotation as well.

jrfnl:
>    When a function has an alternative that isn't a function (like "the Loop", for start_wp()), WordPress doesn't specify that as the third parameter of the _deprecated_function() call. So in the sniff it will currently have a value, but the generator will not detect the alternative.

This should be fine I think, though we might want to verify those instances & annotate in the sniff with a comment that it's ok.

>    There are some cases where the alternative function in the sniff is incorrect. For example, it should be wp_editor() for tinymce_include(), but is listed as wp_tiny_mce(). I haven't checked other cases to be sure that they are correct.

I'm wondering if some might have a different alternative function mentioned in the `@deprecated` tag compared to the `_deprecated_function()` call. Or the alternative may have been deprecated itself ?

Maybe @grappler who originally created this PR can shed some light on this ? There might be a good reason for the difference.

>    In some cases the spacing for the alternative functions is different. get_the_author_meta( \'description\' ) in the sniff vs get_the_author_meta(\'description\') from the generator.

I'll adjust that one - it had a different quote style from what was used in the WP code previously, so I already adjusted it, just hadn't noticed the spaces. Will also check & adjust others for this if there are more.

> It also appears that not all deprecated functions call _deprecated_function(), so the generator currently doesn't pick them up. I'll see about updating it so that it will check for the @deprecated dockblock annotation as well.

That would be awesome!

On that note and for future consideration if we want to start sniffing deprecated hooks as well:
I know for a fact that most deprecated hooks do have a `@deprecated` tag, but hardly any have a `_deprecated_hook()` call. A patch for this has been untouched for nearly a year, see https://core.trac.wordpress.org/ticket/10441

jrfnl:
Ok, so what started out as a simple property value change is quickly becoming quite a large PR with wide-reaching ripples.

I've gone through the complete list of diffs and these are my findings:
* One diff in a function name - `the_category_ID` - is caused by the sniff list being all lowercase (which it doesn't need to be as the sniff accounts for a case-insensitive compare). I will fix that in the sniff.
* One function+ it's alternative was missing an `_` at the start - `_nc` not `nc`. I will fix that in the sniff.
* One function is not deprecated at all - `wp_dashboard_quick_press`, but the alternative for a very similarly named function `wp_dashboard_quick_press_output` which *is* deprecated in the version listed. I will fix that in the sniff.
* **_Nine_** functions were listed as deprecated under the wrong WP version (according to the docs - I haven't checked the commit history). I will fix that in the sniff.
* A few diffs in alternatives are due to spacing in the text string. I will fix that in the sniff.
* A few diffs in alternatives are due to WP not listing the alternative in the `_deprecated_function()` call. I will be opening a trac ticket for those.
* There are four diffs in alternatives due to the function call in WP missing the braces. I will open an issue in trac for those.
* Four alternatives are in actual fact a filter, not a function. The availability of this alternative is not listed in the `_deprecated_function()` call, nor in the sniff array. I will open an issue in trac to get this fixed up and will apply the same change here in the sniff.
* A few diffs in alternatives are due to the sniff array not listing an alternative, while there is one. Or listing an outdated alternative - i.e. the alternative function itself has been deprecated since -. Or just plainly the wrong alternative. I will fix these in the sniff.
* There is one class listed in the function array. The class constructor contains a `_deprecated_function()` call.
    * I've created a new sniff called `WP.DeprecatedClasses` to deal with complete classes having been deprecated and will pull that later.
    * I'll be opening a trac ticket for core to add a new `_deprecated_class()` function to be used in class constructors for deprecated classes.
    * None of the methods of the deprecated class currently had `_deprecated_function()` calls. I will also open a trac ticket to get those added.
* There are **_eight_** methods listed in the function array - on top of the three which had been previously identified as such. I will add a comment to these referring to #731 which is the related open issue.
* There are a number of functions missing from the sniff list. I will add these later.
* On top of the three @JDGrimes previously identified in https://core.trac.wordpress.org/ticket/4109 , there are **_sixteen_** more deprecated functions (based on the diff) in Core which are missing a `_deprecated_function()` call. I will be opening trac tickets for these.
* Lastly, there are **_twelve_** deprecated methods in Core which are missing a `_deprecated_function()` call. I will be opening trac tickets for these too.

A new diff after I've pushed the changes to the PR would be appreciated, but things should get shiny once the patches for the trac tickets are merged into core.

Once I've opened the trac tickets, I will edit this comment to add links to them.

jrfnl:
Just thinking.... maybe we should create a sniff that checks for calls to any of the `_deprecated_....()` functions and ensures that there is a `@deprecated` tag and visa versa ?

jrfnl:
After discussing this with @pento I've opened just the one ticket containing all the patches: https://core.trac.wordpress.org/ticket/41121

GaryJones:
Great investigation @jrfnl - improving Core, from improving the Code Standards. Who says they aren't valuable?! :-)

jrfnl:
Secondary ticket opened to propose a new `_deprecated_class()` function: https://core.trac.wordpress.org/ticket/41125

jrfnl:
Okidoki, updates to the sniff have been pushed, so the compare should be a **lot** cleaner now ;-) @JDGrimes want to do your thing ?

This PR may now conflict with PR #994. As that one is the easier/smaller one, I suggest I rebase this PR after #994 has been merged.

WordPress/Sniffs/WP/DeprecatedFunctionsSniff.php
```diff
@@ -334,7 +349,7 @@ class WordPress_Sniffs_WP_DeprecatedFunctionsSniff extends WordPress_AbstractFun
 			'version' => '2.8.0',
 		),
 		'get_the_author_description' => array(
-			'alt'     => 'get_the_author_meta( \'description\' )',
+			'alt'     => 'get_the_author_meta(\'description\')',
```
GaryJones: I realise this is only going to be displayed as a message to the debugging developer but is there any rationale for preferring no-spaces over spaces? Is this how WP core defines the alternatives?

WordPress/Sniffs/WP/DeprecatedFunctionsSniff.php
```diff
@@ -809,7 +834,7 @@ class WordPress_Sniffs_WP_DeprecatedFunctionsSniff extends WordPress_AbstractFun
 			'version' => '3.5.0',
 		),
 		'get_default_page_to_edit' => array(
-			'alt'     => 'get_default_post_to_edit()',
+			'alt'     => 'get_default_post_to_edit( \'page\' )',
```
GaryJones: As a difference to the above comment, this has spaces around the string.

WordPress/Sniffs/WP/DeprecatedFunctionsSniff.php
jrfnl: Yes, the spacing of the various alternatives is currently inconsistent and inconsistent with the WP coding standard.
Consistency with WP core has been given preference for easier comparisons.
Also see: https://core.trac.wordpress.org/ticket/41121#comment:10

WordPress/Sniffs/WP/DeprecatedFunctionsSniff.php
jrfnl: See above ;-)

GaryJones:
Probably worth waiting on the outcome of https://core.trac.wordpress.org/ticket/41121 before this is approved for merge?

jrfnl:
> Probably worth waiting on the outcome of https://core.trac.wordpress.org/ticket/41121 before this is approved for merge?

We don't really need to wait, though I would like to see another compare just to be sure I haven't missed anything.

There is some discussion about replacement string styles in core ticket 41121, but - if needed - we can adjust those minor details later. That is, after all, just a minor part of this PR.

JDGrimes:
New diff: https://www.diffchecker.com/fBGrpxWW

jrfnl:
Rebased after the merge of #994 & squashed a number of commits together to tidy up the branch. Also fixed one more alternative function missing braces. @JDGrimes Care to rerun the comparison one more time ?

JDGrimes:
https://www.diffchecker.com/YiqiNtww

jrfnl:
Thanks @JDGrimes. As far as I can see, the remaining differences have to be solved in core and have appropriate tickets & patches, so as far as I'm concerned, this PR is good to go.

PR #981, opened 2017-06-18
ControlStructureSpacing sniff: fix blank line after control structure / disappearing comment

### LINKED ISSUE #976: Faulty core fixes: the case of the disappearing comment
#### Summary:
Comment disappears when the above inline control structure gets fixed.

Most likely cause: `Generic.ControlStructures.InlineControlStructure`

#### Example:

`./src/wp-admin/menu-header.php` around line 201

Source:
```php
if ( isset( $submenu_file ) ) {
	if ( $submenu_file == $sub_item[2] )
		$class[] = 'current';
// If plugin_page is set the parent must either match the current page or not physically exist.
// This allows plugin pages with the same hook to exist under different parents.
} elseif (
	( ! isset( $plugin_page ) && $self == $sub_item[2] ) ||
	( isset( $plugin_page ) && $plugin_page == $sub_item[2] && ( $item[2] == $self_type || $item[2] == $self || file_exists($menu_file) === false ) )
) {
	$class[] = 'current';
}
```

Fixed as:
```php
if ( isset( $submenu_file ) ) {
	if ( $submenu_file == $sub_item[2] ) {
		$class[] = 'current';
	}
} elseif (
	( ! isset( $plugin_page ) && $self == $sub_item[2] ) ||
	( isset( $plugin_page ) && $plugin_page == $sub_item[2] && ( $item[2] == $self_type || $item[2] == $self || file_exists( $menu_file ) === false ) )
) {
	$class[] = 'current';
}
```

Expected fix:
```php
if ( isset( $submenu_file ) ) {
	if ( $submenu_file == $sub_item[2] ) {
		$class[] = 'current';
	}
// If plugin_page is set the parent must either match the current page or not physically exist.
// This allows plugin pages with the same hook to exist under different parents.
} elseif (
	( ! isset( $plugin_page ) && $self == $sub_item[2] ) ||
	( isset( $plugin_page ) && $plugin_page == $sub_item[2] && ( $item[2] == $self_type || $item[2] == $self || file_exists( $menu_file ) === false ) )
) {
	$class[] = 'current';
}
```

jrfnl:
This bug was introduced in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/789/commits/786c9bb5855b3bdeb6c33852dd92f089b160a5fa but had so far gone unreported. This PR effectively reverts the offending part of that commit.

Fixes #976

### DIFF
--- a/WordPress/Sniffs/WhiteSpace/ControlStructureSpacingSniff.php
+++ b/WordPress/Sniffs/WhiteSpace/ControlStructureSpacingSniff.php
@@ -496,7 +496,8 @@ class WordPress_Sniffs_WhiteSpace_ControlStructureSpacingSniff extends WordPress
 			return;
 		}
 
-		$trailingContent = $this->phpcsFile->findNext( PHP_CodeSniffer_Tokens::$emptyTokens, ( $scopeCloser + 1 ), null, true );
+		// {@internal This is just for the blank line check. Only whitespace should be considered.}}
+		$trailingContent = $this->phpcsFile->findNext( T_WHITESPACE, ( $scopeCloser + 1 ), null, true );
 		if ( false === $trailingContent ) {
 			return;
 		}
@@ -504,6 +505,16 @@ class WordPress_Sniffs_WhiteSpace_ControlStructureSpacingSniff extends WordPress
 		if ( T_ELSE === $this->tokens[ $trailingContent ]['code'] && T_IF === $this->tokens[ $stackPtr ]['code'] ) {
 			// IF with ELSE.
 			return;
+		}
+
+		if ( T_COMMENT === $this->tokens[ $trailingContent ]['code'] ) {
+			if ( $this->tokens[ $trailingContent ]['line'] === $this->tokens[ $scopeCloser ]['line'] ) {
+				if ( '//end' === substr( $this->tokens[ $trailingContent ]['content'], 0, 5 ) ) {
+					// There is an end comment, so we have to get the next piece
+					// of content.
+					$trailingContent = $this->phpcsFile->findNext( T_WHITESPACE, ( $trailingContent + 1), null, true );
+				}
+			}
 		}
 
 		if ( T_BREAK === $this->tokens[ $trailingContent ]['code'] ) {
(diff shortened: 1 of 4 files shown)


WordPress/Sniffs/WhiteSpace/ControlStructureSpacingSniff.php
```diff
@@ -496,7 +496,8 @@ public function process_token( $stackPtr ) {
 			return;
 		}
 
-		$trailingContent = $this->phpcsFile->findNext( PHP_CodeSniffer_Tokens::$emptyTokens, ( $scopeCloser + 1 ), null, true );
+		// {@internal This is just for the blank line check. Only whitespace should be considered.}}
```
GaryJones: Is the internal annotation required? This is not a DocBlock.

jrfnl: Not required, but felt like the appropriate annotation. It's mostly to prevent me from changing it back when doing spot checks whether the appropriate token is used :innocent:

JDGrimes: Could add "not other empty tokens" to make it more clear.

WordPress/Sniffs/WhiteSpace/ControlStructureSpacingSniff.php
```diff
@@ -506,6 +507,16 @@ public function process_token( $stackPtr ) {
 			return;
 		}
 
+		if ( T_COMMENT === $this->tokens[ $trailingContent ]['code'] ) {
+			if ( $this->tokens[ $trailingContent ]['line'] === $this->tokens[ $scopeCloser ]['line'] ) {
+				if ( '//end' === substr( $this->tokens[ $trailingContent ]['content'], 0, 5 ) ) {
```
JDGrimes: What about `// end`, `// End`, etc. ? Does it really even matter what the comment looks like here? Would be good to see some tests that show what this part does and that it works correctly.

jrfnl: Just ran some tests and interestingly enough, that causes yet another "case of the disappearing comment" similar to #976 

Looking at the history, this sniff was original cloned from the Squiz standard and the fixers were build in afterwards in 2014. Since the original cloning, the upstream sniff also gained fixers.
I know I've synced a number of relevant changes from upstream in before, but don't think that part has changed that much. Will have a look and see what we can learn from the upstream version.

jrfnl:
I've sync in a number of snippets from upstream & fixed the second disappearing comment bug. Fingers crossed, this should be good now.

PR #989, opened 2017-06-20
OperatorSpacing sniff: Defer to upstream sniff for improved results.

### LINKED ISSUE #988: Ternary whitespace check ?
Looks like whitespace around ternary operators is currently not checked.

Should it be ?

Upstream this has been added to the `WhiteSpace.OperatorSpacing` sniff. The WPCS version of this sniff has not been synced with upstream for quite a long time.
Might be worth checking if the WP version can be removed and defer to the upstream version instead.

```php
$a = ($a > 10)?$a:10;
```

will yield no errors.

I would expect this to error on:
* no whitespace before "inline then" / ? - expected 1 space
* no whitespace after "inline then" / ? - expected 1 space
* no whitespace before "inline else" / : - expected 1 space
* no whitespace after "inline else" / : - expected 1 space

As a side-note, I'm also missing these notifications:
* no whitespace after the open parenthesis
* no whitespace before the closing parenthesis

jrfnl:
Up to now, this sniff was based on a copy of a very old and outdated version of the upstream sniff.
The upstream and downstream sniffs had diverged quite widely code-wise, but not intention-wise.

I've run both the `WordPress.WhiteSpace.OperatorSpacing` sniff as well as its upstream version `Squiz.WhiteSpace.OperatorSpacing` over both the WPCS unit tests for this sniff as well as the Squiz unit tests for this sniff.

While there were differences in the results, they were for the better in favour of the upstream version of the sniff.

The unit test file has been adjusted to only test for the WPCS additional token.

Fixes #988

### 1. "Boolean not" was not handled by the upstream sniff:
This was added to the WPCS version in #67
```php
if (!$var ) {}
```
**Fixed**: by overloading the `register()` method and adding the token to the sniff.

### 2. Operator spacing after `-` (minus) was not verified in the WPCS version
This was [previously disabled](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/commit/d6c64115ec48889176f030b931c92829a8efbf2a) to prevent the sniff from triggering on code like:
```php
$a = -1;
return -1
```
However, as a side effect, this also prevented the sniff from triggering & fixing this:
```php
$result = 1  -        2;
```

The `+` operator can be used in a similar vain, but was unaccounted for in WPCS, so the below code *would* trigger an error with the WPCS version of the sniff:
```php
$a = +1;
```
**Fixed**: the current upstream sniff already accounts for these situations and will not throw errors for it (example 1 and 3), but will correctly throw an error for example 2.

### 3. Checking for spacing around the `&` (bitwise and) was removed from the WPCS version
I suspect this was done to prevent false positives for `&` being used as a reference.
This also meant however that spacing around *real* "bitwise and" was not being checked, so code like this would not trigger any errors nor be fixed:
```php
if ($result&4 && $result  &  4) {}
```
**Fixed**: the upstream sniff correctly accounts for and bow out if `&` is being used as a reference operator, so there should be no problem with that check.

### 4. New lines should always be acceptable instead of a space
Code like the below should not trigger the sniff:
```php
$y = 1
   + 2  
   - 3;
```
**Fixed**: since PHPCS 2.2.0 / early 2015 the upstream sniff has had a `$ignoreNewLines` property which can be set to `true` to handle this. Fixed by setting that property to `true` in our child class.

### DIFF
--- a/WordPress/Sniffs/WhiteSpace/OperatorSpacingSniff.php
+++ b/WordPress/Sniffs/WhiteSpace/OperatorSpacingSniff.php
@@ -7,8 +7,12 @@
  * @license https://opensource.org/licenses/MIT MIT
  */
 
+if ( ! class_exists( 'Squiz_Sniffs_WhiteSpace_OperatorSpacingSniff', true ) ) {
+	throw new PHP_CodeSniffer_Exception( 'Class Squiz_Sniffs_WhiteSpace_OperatorSpacingSniff not found' );
+}
+
 /**
- * Verify operator spacing, based upon Squiz code.
+ * Verify operator spacing, uses the Squiz sniff, but additionally also sniffs for the `!` (boolean not) operator.
  *
  * "Always put spaces after commas, and on both sides of logical, comparison, string and assignment operators."
  *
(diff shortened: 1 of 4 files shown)


WordPress/Sniffs/WhiteSpace/OperatorSpacingSniff.php
```diff
@@ -17,194 +21,39 @@
...
- * @since   0.3.0 This sniff now has the ability to fix the issues it flags.
- * @since   0.12.0 This class now extends WordPress_Sniff.
+ * @since   0.3.0  This sniff now has the ability to fix the issues it flags.
+ * @since   0.12.0 This sniff used to be a copy of a very old and outdated version of the
+ *                 upstream sniff.
+ *                 Now, the sniff defers completely to the upstream sniff, adding just one
+ *                 additional token - T_BOOLEAN_NOT - via the registration method
+ *                 and changing the value of the customizable $ignoreNewlines property.
  *
- * Last synced with base class December 2008 at commit f01746fd1c89e98174b16c76efd325825eb58bf1.
```
GaryJones: 2008?! Impressive!

jrfnl: LOL, yup, that was when the original was cloned and never synced again ;-)

Until now that is ;-)

PR #992, opened 2017-06-20
Core: Set the default encoding in the ruleset

### LINKED ISSUE #990: Set encoding to utf-8
From https://core.trac.wordpress.org/ticket/41057#comment:33

> The default encoding PHPCS 2.x uses for interpreting files is
 `ISO-8859-1`. This can be changed by setting `--encoding=utf-8` on the
 command line or using `<config name="encoding" value="utf-8"/>` in a
 custom ruleset.
>
> AFAIK, this can not be changed for an individual file, so will affect how
 all files are read.
>
>  If so desired, we can add this directive to the `WordPress-Core` ruleset
 in WPCS.

I'd definitely be +1 to this. My understanding is, files previously treated as (and only have characters from) `ISO-8859-1` should continue to be the same when treated as `utf-8`.

jrfnl:
Also moved the `tab-width` up as this is not a sniff specific, but a generic setting.

Fixes #990

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -2,6 +2,12 @@
 <ruleset name="WordPress Core">
 	<description>Non-controversial generally-agreed upon WordPress Coding Standards</description>
 
+	<!-- Treat all files as UTF-8. -->
+	<config name="encoding" value="utf-8"/>
+
+	<!-- Default tab width for indentation fixes and such. -->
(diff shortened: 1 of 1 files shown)


GaryJones:
@jrfnl Travis failing here.

jrfnl:
@GaryJones I've just pushed the fix. Trailing whitespace removed.

PR #1004, opened 2017-06-24
ArrayDeclarationSpacing sniff: various bug fixes

### LINKED ISSUE #1002: PHP Notice when running phpcbf on a set of Core files
When running `phpcbf` on a set of Core files, I get this notice a lot:

`PHP Notice:  Undefined offset: 2843 in /srv/www/phpcs/CodeSniffer/Standards/WordPress/WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php on line 118
`

Unfortunately, when running `phpcbf` on the file it was processing when the notice occurred, the notice goes away.

I've been able to get it down to running against these two files:

`phpcbf src/wp-admin/credits.php src/wp-admin/custom-background.php`

jrfnl:
The "associative arrays should always be multi-line" check was added a while ago.

Since then two important additions were made to this repo:
* The `get_function_call_parameters()` utility method was added to the `WordPress_Sniff` which makes retrieving individual items in an array a breeze.
* The `WordPress.Arrays.ArrayIndentation` sniff was created and added.

The fixer for the "associative arrays should always be multi-line" check was *also* trying to sort out the needed indentation. This is no longer needed as that will be sorted out by the `WordPress.Arrays.ArrayIndentation` sniff.

This PR essentially removes the indentation fixing part from the `WordPress.Arrays.ArrayDeclaration` sniff.

Notes:
* To be able to use the `get_function_call_parameters()` utility method, the sniff needs to extend the `WordPress_Sniff` while it extended the `Squiz_Sniffs_Arrays_ArrayDeclarationSniff` sniff up to now.
* A new utility method `find_array_open_closer()` to determine an array's opener & closer has been added to the `WordPress_Sniff`.
    This method is used by this sniff, but the `WordPress.Arrays.ArrayIndentation` sniff used the same logic, so that has been switched out in favour of using the new method too.
* While working on this sniff, I realized that a non-associative single-line array which contained a nested associative array would also be auto-fixed by this sniff, while only the nested associated array *should* be fixed. This has now been corrected.
* Additionally, when an associative single-line array is detected, the "space after array opener" and "space before array closer" checks become irrelevant as the array has to be changed to become multi-line anyway which automatically creates the needed spacing. With that in mind, I've changed the order of the checks in the sniff so only non-associative single-line arrays will throw errors about the spacing.

For more detail & easier review, please see the individual commits.

Fixes #1002 (at least it should as the offending code has been removed ;-) )

This *may* fix several other array related issues which have been recently reported, but that will need to be checked for each issue individually.


pento:
Confirmed that this fixes #1002, but exposes an exciting new bug that I'm tracking down. :-)

PR #1024, opened 2017-07-13
Improve performance of the function call/array analysis sniffs.

jrfnl:
This PR caches the results of the function call parameter / array parsing as this can be quite a "heavy" operation. The trade-off of doing this will be that the memory usage of the PHPCS run will go up.

@JDGrimes Could you run your performance test magic to see if this has any effect & if so, if it is significant enough to merge this ?

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -810,6 +810,18 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	protected $tokens;
 
 	/**
+	 * Internal cache of the results of the `get_function_call_parameters()` method.
+	 *
+	 * The results are stored in a multi-dimensional array by filename and $stackPtr of
+	 * the function call/array token.
+	 *
+	 * @since 0.12.0
+	 *
+	 * @var array
+	 */
+	private $get_function_call_parameters_cache = array();
+
+	/**
 	 * Set sniff properties and hand off to child class for processing of the token.
 	 *
 	 * @since 0.11.0
@@ -1901,6 +1913,7 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	 * it will tokenize the values / key/value pairs contained in the array call.
 	 *
 	 * @since 0.11.0
+	 * @since 0.12.0 The results of this function are now cached for improved performance.
 	 *
 	 * @param int $stackPtr The position of the function call token.
 	 *
@@ -1914,7 +1927,15 @@ abstract class WordPress_Sniff implements PHP_CodeSniffer_Sniff {
 	 *               }
 	 */
 	public function get_function_call_parameters( $stackPtr ) {
+		$current_file = $this->phpcsFile->getFilename();
+
+		// Check if this function call/array has been parsed before and if so, use the cached results.
+		if ( isset( $this->get_function_call_parameters_cache[ $current_file ][ $stackPtr ] ) ) {
+			return $this->get_function_call_parameters_cache[ $current_file ][ $stackPtr ];
+		}
+
 		if ( false === $this->does_function_call_have_parameters( $stackPtr ) ) {
+			$this->get_function_call_parameters_cache[ $current_file ][ $stackPtr ] = array();
 			return array();
 		}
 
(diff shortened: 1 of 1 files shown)


JDGrimes:
https://blackfire.io/profiles/compare/dca8cfbe-f83c-4f92-8117-6b37e0859f47/graph

The increase in memory usage is 26%, while the increase in performance is negligible. I'm not sure why there is so little effect on performance.

FYI, when caching like this, I think that it is unnecessary to keep the cache for any file other than the current one being processed. Discarding the cache from the previous file decreases the total memory being used at once.

jrfnl:
@JDGrimes Good point about the cache removing. I was thinking about that after I pulled it, will have a look to add it.

Regarding the increase in performance: the performance check is based on a run against what code ? The unit tests ? WP core ? (Blackfire does not let me see the command used)

If purely based on a unit test run, I would not expect to see any difference as each sniff is tested in isolation and therefore each function call/array is only evaluated once by PHPCS.

Similarly, the `Core` ruleset only contains two sniffs I can think of which would benefit from this change.

Running with the `Extra` ruleset however should show some more improvement when either run against WP core or against the `/Tests` directory without it being in a unit test run, and that improvement should increase more once the `DeprecatedFunctionParameter` sniff has been merged.

JDGrimes:
I run it over [the WordPoints plugin](https://github.com/WordPoints/wordpoints), just because that is what I have handy. I could certainly run it over core if you'd like, but I'm guessing that would probably take a bit longer. WordPoints represents a reasonably large and diverse codebase and the run only takes 2 minutes.

The ruleset is a custom one that includes `Core`, `Extra`, `VIP`, and `Docs`, with some exclusions.

Running the performance check is very simple if you have Blackfire set up. You just prepend `blackfire run` to your phpcs command, and it will spit out the URL for the results. Then checkout the other branch of WPCS to test and repeat. (I'm perfectly happy to run it myself, I'm just offering the info if anyone else is interested. I'm using the free version of Blackfire, so anyone can be set up to run checks like this on WPCS or other code in a matter of the few minutes it takes to install it (at least install was pretty easy for me using homebrew)).

I'll try running `Extra` over core and see what that results in.

jrfnl:
Just for the record: these are the sniffs which **_should_** benefit from the caching:
* Arrays/ArrayDeclarationSpacing
* Arrays/ArrayIndentation
* Arrays/CommaAfterArrayItem

Additionally, these *could* benefit if any of these would sniff for the same function calls:
* VIP/CronInterval
* NamingConventions/PrefixAllGlobals
* NamingConventions/ValidHookName
* PHP/StrictInArray
* VIP/AdminBarRemoval
* VIP/PluginMenuSlug
* _WP/DeprecatedFunctionParameters_

jrfnl:
> FYI, when caching like this, I think that it is unnecessary to keep the cache for any file other than the current one being processed. Discarding the cache from the previous file decreases the total memory being used at once.

I've been thinking about this some more and I don't think we should discard the cache. While it would have some benefit for PHPCS 2.x, PHPCS 3.x brings the ability to process files in parallel - and while I haven't looked in detail at how this is implemented - I suspect that would mean we'd have to remove it again to allow for WPCS to be used in conjunction with parallel processing of files

Ref: https://github.com/squizlabs/PHP_CodeSniffer/issues/421

JDGrimes:
I tried running over core, but it hit the memory limit. I tried just running `WordPress-Extra` over that plugin, but the results showed no performance difference relating to those sniffs. Really, there were some other issues in the results, and I'm thinking that maybe the Blackfire probe is not working properly for me right now, because I have PHP 7.0 ZTS installed. Blackfire is still considered only beta on ZTS. I'll see about switching to non-ZTS PHP sometime soon (I had only installed for debugging purposes at one point), but I might not have a chance to right away.

As far as parallel processing goes, I wonder if there is another way to detect that a file is finished processing, and discard the cache for that file?

jrfnl:
> As far as parallel processing goes, I wonder if there is another way to detect that a file is finished processing, and discard the cache for that file?

I honestly don't know. I'd need to study the PHPCS 3.x code in more detail for that and would prefer to wait until we actually upgrade to do that.

Anyway, I'm ok with closing this PR - it was worth to try & see if this would make a difference - especially with large arrays - and I appreciate you testing it @JDGrimes .

Alternatively we could leave it open for now to revisit once WPCS is compatible with PHPCS 3.x.

GaryJones:
> Alternatively we could leave it open for now to revisit once WPCS is compatible with PHPCS 3.x.

Still planning to revisit this?

jrfnl:
Let's close this for now.

PR #1025, opened 2017-07-13
ArrayIndentation sniff: various improvements to the reporting & fixing

### LINKED ISSUE #973: Faulty Core fixes: Array indentation / value on new line
#### Summary:
When in an associative array, the value is on the next line (long lines), it is normally indented one tab from the key. If the key was indented incorrectly, then after the fixer has run, the value will no longer be indented one tab in from the key.

Most likely cause: `WordPress.Arrays.ArrayIndentation` only accounts for key indentation, not for values.

#### Question:
AFAIK, there is no clear guideline for how this type of indentation "should" be done. I can try & adjust the fixer to indent the value one in from the key if the value starts on a new line, but as there is no clear guideline, should this be auto-fixed or not ?

#### Example:

`./src/wp-admin/edit-comments.php` around line 174

Source:
```php
get_current_screen()->add_help_tab( array(
'id'		=> 'overview',
'title'		=> __('Overview'),
'content'	=>
	'<p>' . __( 'You can manage comments made on your site similar to the way you manage posts and other content. This screen is customizable in the same ways as other management screens, and you can act on comments using the on-hover action links or the Bulk Actions.' ) . '</p>'
) );
```

Fixed as:
```php
get_current_screen()->add_help_tab( array(
	'id'        => 'overview',
	'title'     => __( 'Overview' ),
	'content'   =>
	'<p>' . __( 'You can manage comments made on your site similar to the way you manage posts and other content. This screen is customizable in the same ways as other management screens, and you can act on comments using the on-hover action links or the Bulk Actions.' ) . '</p>',
) );
```

Expected fix:
```php
get_current_screen()->add_help_tab( array(
	'id'        => 'overview',
	'title'     => __( 'Overview' ),
	'content'   =>
		'<p>' . __( 'You can manage comments made on your site similar to the way you manage posts and other content. This screen is customizable in the same ways as other management screens, and you can act on comments using the on-hover action links or the Bulk Actions.' ) . '</p>',
) );
```

### LINKED ISSUE #985: Potentially faulty core fixes: array indentation / array interspersed with comments
#### Summary:
While looking at the array indentation sniff, I realized that if there are comments between array items, the indentation of these would not be corrected.

Related #973

#### Example:

N/A - I have no doubt, we will actually run into this, but haven't come across it yet in the limited group of fixed files I've inspected so far.

Source:
```php
get_current_screen()->add_help_tab( array(
'id'		=> 'overview',
// Here we have a comment.
'title'		=> __('Overview'),
// Here we have a comment.
'content'	=> 'value',
) );
```

Fixed as:
```php
get_current_screen()->add_help_tab( array(
	'id'		=> 'overview',
// Here we have a comment.
	'title'		=> __('Overview'),
// Here we have a comment.
	'content'	=> 'value',
) );
```

Expected fix:
```php
get_current_screen()->add_help_tab( array(
	'id'		=> 'overview',
	// Here we have a comment.
	'title'		=> __('Overview'),
	// Here we have a comment.
	'content'	=> 'value',
) );
```

jrfnl:
Finally managed to get the unit tests for this one working with the gracious help of @gsherwood.

The `WordPress.Arrays.ArrayIndentation` sniff did not account for multi-line array items.
Up to now, the first line would be handled, the subsequent lines, however, were disregarded.
This has now been fixed.

The fix takes into account that subsequent lines may have varying indentation, for instance to indicate HTML layout.
 > Subsequent lines may be indented more or less than the mimimum expected indent,  but the "first line after" should be indented - at least - as much as the very first line of the array item.
Indentation correction for subsequent lines will be based on that diff.
(inline comment in sniff)

Fixes #973

------

While working on the above, I realized that heredoc/nowdoc tokens - except for the opener - should be disregarded from indentation fixes by this sniff as it could break the closer and distort the text layout in a heredoc/nowdoc.
The same goes for inline HTML, even thought that token should never be encountered in an array item.

This has now been sorted as well.

------

Lastly, arrays interspersed with comments, both for single line items as well as multi-line items, were not being aligned properly.

This has also been fixed.

Fixes #985

------

Additional notes:
* Includes an extensive range of additional unit tests.
* I've added a public `$tabIndent` property to allow for this sniff to behave in a similar manner as the `Generic.WhiteSpace.ScopeIndent` sniff with regards to space vs tab indents. This will allow also for - at some point - potentially upstreaming the sniff to PHPCS itself. The space indent variation is unit tested in a second unit test case file using the same cases. To that end, the first unit test file has been renamed to `ArrayIndentationUnitTest.1.inc` (was `ArrayIndentationUnitTest.inc`).
* I've renamed a number of internally used variables to make it easier to distinguish between them.
* I've also split off parts of the code into separate methods to prevent/reduce code duplication.
    The `get_indentation_size()` and `get_indentation_string()` methods - and possibly even the `fix_alignment_error()` method - can be regarded as utility methods and could/should in the future be moved to the `WordPress_Sniff` class if & when other sniffs would need them.


GaryJones:
Despite the large increase in tests, I'm only seeing `phpunit --filter WordPress` report that `Tests generated 521 unique error codes; 42 were fixable (8.06%)` for this branch, compared to the `Tests generated 519 unique error codes; 40 were fixable (7.71%)` before it was merged into `develop`. 

Is that what others see?

jrfnl:
@GaryJones And that is correct. I introduced two new error**codes**, both fixable.

The message is not about the nr of notices, but about the nr of unique error codes encountered.

PR #1028, opened 2017-07-14
Example project.ruleset.xml: improve exclusion pattern suggestions

jrfnl:
P.S.: it would be nice if WP-core would actually comply with the `.min.js` naming conventions for minified JS ...

Some examples of files which are minified without the `.min.js` extensions:
* `html5.js` as shipped with most default themes
* `wp-includes/js/jquery/jquery.js`
* `wp-includes/js/jquery/jquery.query.js`
* `wp-includes/js/jquery/jquery.ui.touch-punch.js`
* `wp-includes/js/swfobject.js`

---- 

PR has been updated and now covers:
* Add exclusion of minified JS files.
* Add exclusion of Node modules directory
* Remove exclusion of `*.twig` files as that's not an extension that is included by PHPCS by default, so would need to be added via `--extensions=...` for those files to be picked up in the first place.

### DIFF
--- a/project.ruleset.xml.example
+++ b/project.ruleset.xml.example
@@ -2,17 +2,22 @@
 <ruleset name="Example Project">
 	<description>A custom set of rules to check for a WPized WordPress project</description>
 
-	<!-- Exclude select folders and files from being checked. -->
+	<!-- Exclude WP Core folders and files from being checked. -->
 	<exclude-pattern>/docroot/wp-admin/*</exclude-pattern>
 	<exclude-pattern>/docroot/wp-includes/*</exclude-pattern>
 	<exclude-pattern>/docroot/wp-*.php</exclude-pattern>
 	<exclude-pattern>/docroot/index.php</exclude-pattern>
 	<exclude-pattern>/docroot/xmlrpc.php</exclude-pattern>
 	<exclude-pattern>/docroot/wp-content/plugins/*</exclude-pattern>
-	<exclude-pattern>*.twig</exclude-pattern>
 
 	<!-- Exclude the Composer Vendor directory. -->
 	<exclude-pattern>/vendor/*</exclude-pattern>
+
+	<!-- Exclude the Node Modules directory. -->
+	<exclude-pattern>/node_modules/*</exclude-pattern>
+
+	<!-- Exclude minified Javascript files. -->
+	<exclude-pattern>*.min.js</exclude-pattern>
 
 	<!-- Include the WordPress-Extra standard. -->
 	<rule ref="WordPress-Extra">


ntwb:
The reason why those files do not ship with `.min.js` extension is because they are 3rd party libraries and that is the format that they were added with a long long time ago, as plugins and themes have come to depend on these files being included changing the extension of the files has been avoided so as to not cause any breakage.

With that out of the way, why only exclude minified JavaScript files, why not exclude all JavaScript files?

And further, shouldn't only `.php` files be included in the first place?

jrfnl:
> With that out of the way, why only exclude minified JavaScript files, why not exclude all JavaScript files?
> And further, shouldn't only .php files be included in the first place?

PHPCS is capable of sniffing PHP, JS and CSS files and has natively 70 sniffs available which can detect & fix issues in more than just PHP files.

WPCS uses a quite a few of these sniffs.

WPCS also contains a few sniffs which have specific features targetting JS and/or CSS ( `VIP.AdminBarRemoval` and the upcoming `Classes.ClassInstantiation`).

So, for the example ruleset to suggest to people to only target PHP files, would cause them to miss out on some of the features and benefits WPCS can bring.

ntwb:
> PHPCS is capable of sniffing PHP, JS and CSS files and has natively 70 sniffs available which can detect & fix issues in more than just PHP files.

Oh, I did not know this 👍 

> So, for the example ruleset to suggest to people to only target PHP files, would cause them to miss out on some of the features and benefits WPCS can bring.

Neat 👌 

----

Should this list of files then be added to the exclusion in this PR also then?
> html5.js as shipped with most default themes
> wp-includes/js/jquery/jquery.js
> wp-includes/js/jquery/jquery.query.js
> wp-includes/js/jquery/jquery.ui.touch-punch.js
> wp-includes/js/swfobject.js

project.ruleset.xml.example
```diff
@@ -14,6 +14,9 @@
 	<!-- Exclude the Composer Vendor directory. -->
 	<exclude-pattern>/vendor/*</exclude-pattern>
 
+	<!-- Exclude minified Javascript files. -->
+	<exclude-pattern>*.min.js</exclude-pattern>
+
```
westonruter: For this matter, should we also exclude `/node_modules/*` since we're also excluding `vendor`?

jrfnl:
> Should this list of files then be added to the exclusion in this PR also then?

The example ruleset already contains a group of patterns to exclude WP Core completely from a run.
See: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/project.ruleset.xml.example#L5-L12

project.ruleset.xml.example
jrfnl: Good idea. Is `/node_modules/*` the pattern ? I have never used node, so no idea what to exclude.

westonruter: Yes, the same as for `vendor` above.

westonruter:
Should this file get renamed to `phpcs.xml.sample`? Cf. https://github.com/wp-cli/scaffold-command/pull/19

project.ruleset.xml.example
jrfnl: Done & updated the PR description to reflect the changes.
(re-did the comit and removed the twig exclusion as well while I was at it)

jrfnl:
> Should this file get renamed to phpcs.xml.sample?

In that case I would suggest to name it `phpcs.xml.dist.sample` instead.

JDGrimes:
A separate PR can be opened to rename the file.

PR #1030, opened 2017-07-14
Rename project.ruleset.xml.example to phpcs.xml.dist.sample

GaryJones:
See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1028#issuecomment-315335479.


jrfnl:
@GaryJones Thanks for this. The filename will also need to be adjusted in the part of the travis script which verifies that the XML is valid + I think the file is referenced in the readme (and maybe elsewhere too), so that link would need to be adjusted as well.

PR #1032, opened 2017-07-18
Update `.travis.yml` to use Ubuntu Trusty by default and Precise for PHP 5.2/5.3

ntwb:
Via https://core.trac.wordpress.org/ticket/41292

Travis CI have announced they are going to switch to Trusty as the default distro come July 18th
> https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming

----

The current availability of PHP 5.2 and 5.3 in the Travis CI build environment is the following:

1.  `dist: precise` with `sudo: false` 
2.    `dist: trusty` with `sudo: false`
3.    `dist: precise` with `sudo: false`
4.    `dist: precise` with `sudo: required`

• WPCS' current Travis CI config uses option `1` above

• Travis CI will override this configuration with option `2` come July 18th 2017

• The workaround for this issue for PHP 5.2 & 5.3 until September 2017 is to use option `3` for the PHP 5.2 and 5.3 Travis CI jobs

• Come September 2017 Travis CI will overwrite option `3` above with option `4`

WordPress Core have just committed the above change just now via [r41072](https://core.trac.wordpress.org/changeset/41072)

----

Travis CI doesn't play well when calculating the `php`/`env` build matrix that includes a `dist:` section, as such the PHP 5.2 and 5.3 jobs are moved to the `include` section of the `matrix` 🤖

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 sudo: false
 
+dist: trusty
+
 language:
     - php
 
 php:
-    - 5.2
-    - 5.3
     - 5.4
     - 5.5
     - 5.6
@@ -38,6 +38,19 @@ matrix:
     # Test PHP 5.3 with short_open_tags set to On (is Off by default)
     - php: 5.3
       env: PHPCS_BRANCH=2.9 SHORT_OPEN_TAGS=true
+      dist: precise
+    - php: 5.3
+      env: PHPCS_BRANCH=2.9
+      dist: precise
+    - php: 5.3
+      env: PHPCS_BRANCH=2.9.0
+      dist: precise
+    - php: 5.2
+      env: PHPCS_BRANCH=2.9
+      dist: precise
+    - php: 5.2
+      env: PHPCS_BRANCH=2.9.0
+      dist: precise
   allow_failures:
     # Allow failures for unstable builds.
     - php: nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ sudo: false
 
 dist: trusty
 
+cache:
+  apt: true
+     
 language:
     - php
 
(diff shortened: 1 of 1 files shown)


jrfnl:
Just FYI: WPCS is expected to drop support for PHP 5.2 before long.
If we can get a cross-version PHPCS version working, PHP 5.2 support will be dropped (namespaces needed). If cross-version support will not work and we start to only support PHPCS 3.x, the minimum required version of PHP for WPCS will go up to 5.4 - in line with PHPCS itself -.

jrfnl:
@grappler Would you mind elaborating on the thumbs down ?

grappler:
@jrfnl Must have been by fat finger. Have removed it.

ntwb:
@jrfnl Indeed, it says they will be dropped but they are not actually dropped

You can confirm this by looking at the PHP 5.2 & 5.3 jobs, they are in fact running `precise` and the other jobs are running `trusty`

I created an upstream issue for this 6days ago here https://github.com/travis-ci/travis-yaml/issues/118

(I've also just restarted the two failing jobs of the build)

jrfnl:
@ntwb Just ran into similar build failures as you're seeing on this PR with another repo.
To fix the build, adding this near the top of the travis script worked for me:
```
cache:
  apt: true

addons:
  apt:
    packages:
      - libxml2-utils
```

ntwb:
@jrfnl Thanks, I've added that for the specific PHP 5.5 job rather than all the jobs 👍

ntwb:
Confirming the affected PHP 5.5 job is now passing: https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/255125652

jrfnl:
> I've added that for the specific PHP 5.5 job rather than all the jobs

Makes sense, though with the app cache on it shouldn't really matter much if I understand the caching thing correctly.

ntwb:
Hmmm, not sure as there are no actual caches right now, there should be 2 after this was merged

https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/caches

Will investigate later 🤔 

Thanks for merging 😄

PR #1033, opened 2017-07-18
New `ClassInstantiation` sniff to `WordPress-Extra`.

### LINKED ISSUE #919: Require parentheses when constructing classes
In WordPress's JSHint config there is an error that gets raised when you try doing this:

```
x = new Thing
```

> Missing '()' invoking a constructor.

This is likewise a coding convention in WordPress's PHP to always use the parentheses when constructing a class, however this is currently not getting enforced by PHPCS. I actually haven't been able to find an existing sniff that checks for that issue, but it seems like something that should be in `Squiz`.

jrfnl:
New sniff to verify that objects being instantiated comply with some basic rules:
* Always use parenthesis.
* Don't have spaces (or comments) between the class name and the parenthesis.
* Don't assign new by reference (PHP only).

Sniffs both PHP as well as JS code.

Auto-fixes _missing parentheses_ and _space between class name and parentheses_.

Includes unit tests.

Currently doesn't handle cases of object instantiations where the class name is held in a variable variable.

Fixes #919

### DIFF
--- a/WordPress-Extra/ruleset.xml
+++ b/WordPress-Extra/ruleset.xml
@@ -109,4 +109,8 @@
 	<!-- Verify that everything in the global namespace is prefixed. -->
 	<rule ref="WordPress.NamingConventions.PrefixAllGlobals"/>
 
+	<!-- Check that object instantiations always have braces & are not assigned by reference.
+		 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/919 -->
+	<rule ref="WordPress.Classes.ClassInstantiation"/>
+
 </ruleset>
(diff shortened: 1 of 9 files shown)


GaryJones:
Looks good. Were you planning to send this upstream?

JDGrimes:
@jrfnl For the record, is there a particular reason why variable variable class names are not supported?

jrfnl:
> Were you planning to send this upstream?

@GaryJones Potentially this could be pulled upstream at some point, though without the `assigning by reference` check.

> For the record, is there a particular reason why variable variable class names are not supported?

@JDGrimes Yes there is. Variable variables can take a lot of forms and it would take a **_lot_** of code to attempt to reliably detect whether the class is instantiated with or without parenthesis in that case. It also would most likely still lead to false positives with the sniff getting confused over whether parentheses are part of the variable variable or for the object instantiation.
See https://github.com/wimg/PHPCompatibility/blob/master/PHPCompatibility/Tests/sniff-examples/variable_variables.php for some examples of what would need to be covered.

PR #1037, opened 2017-07-19
Improve detection of condition beginning in Yoda conditions sniff

### LINKED ISSUE #1036: Incorrect Yoda condition error
Hi,

With WPCS 0.11, I get a Yoda condition error on the line: 
```php
$sample = false !== strpos( $link, '%pagename%' );
```
I believe that it is a false positive and looks to be the same issue as #509
I don't get any error if I modify the code to:
```php
$sample = ( false !== strpos( $link, '%pagename%' ) );
```

JDGrimes:
This sniff is supposed to loop over the left-hand side of the condition,
checking for variables. If the left-hand side of the condition is not a
variable, then it should not give an error about it.

However, there were some issues with how we were detecting the beginning
of the left-hand side of the condition. Take the following example code:

```php
$sample = false !== strpos( $link, '%pagename%' ); // Ok.
```

This should not cause an issue, because the left-hand side of the
condition is just `false`. However, we weren't detecting the beginning
of the condition correctly; we'd end up checking right through the `=`
assignment operator and seeing the `$sample` variable. As a result, an
error would be given for this snippet.

To correct this, we just need to take assignment operators into account
when detecting the beginning of the condition.

With this correction, we can also simplify some other code that was
checking for a function call (or certain other tokens) within the
condition, that would indicate it didn't need to be checked:

- The `T_RETURN` token is actually an indicator of the beginning of the
condition, so it should be moved to that check instead.
- The `T_CONSTANT_ENCAPSED_STRING` token is no longer needed to be checked
at all, because it was added to circumvent the bug that is now fixed by
better detection of the beginning of the condition.
- The `T_OPEN_PARENTHESIS` really shouldn't be needed, because the closing
parenthesis will be detected first anyway (we are looping backward
here).

The `T_CLOSE_PARENTHESIS` is the only one that needs to stay, because of
cases like this:

```php
$sample = foo( $var ) !== strpos( $link, '%pagename%' ); // Ok.
```

Without the check for the closing parenthesis token, the `$var`
parameter in the call to the `foo()` function would cause the condition
to be flagged as needing to be a Yoda condition.

Fixes #1036

### DIFF
--- a/WordPress/Sniffs/PHP/YodaConditionsSniff.php
+++ b/WordPress/Sniffs/PHP/YodaConditionsSniff.php
@@ -35,12 +35,14 @@ class WordPress_Sniffs_PHP_YodaConditionsSniff extends WordPress_Sniff {
 	 */
 	public function register() {
 
-		$this->condition_start_tokens  = PHP_CodeSniffer_Tokens::$booleanOperators;
-		$this->condition_start_tokens += PHP_CodeSniffer_Tokens::$assignmentTokens;
-		$this->condition_start_tokens[ T_CASE ] = T_CASE;
-		$this->condition_start_tokens[ T_RETURN ] = T_RETURN;
-		$this->condition_start_tokens[ T_SEMICOLON ] = T_SEMICOLON;
-		$this->condition_start_tokens[ T_OPEN_PARENTHESIS ] = T_OPEN_PARENTHESIS;
+		$starters                       = PHP_CodeSniffer_Tokens::$booleanOperators;
+		$starters                      += PHP_CodeSniffer_Tokens::$assignmentTokens;
+		$starters[ T_CASE ]             = T_CASE;
+		$starters[ T_RETURN ]           = T_RETURN;
+		$starters[ T_SEMICOLON ]        = T_SEMICOLON;
+		$starters[ T_OPEN_PARENTHESIS ] = T_OPEN_PARENTHESIS;
+
+		$this->condition_start_tokens = $starters;
 
 		return array(
 			T_IS_EQUAL,
--- a/WordPress/Sniffs/PHP/YodaConditionsSniff.php
+++ b/WordPress/Sniffs/PHP/YodaConditionsSniff.php
@@ -46,6 +46,8 @@ class WordPress_Sniffs_PHP_YodaConditionsSniff extends WordPress_Sniff {
 		$beginners   = PHP_CodeSniffer_Tokens::$booleanOperators;
 		$beginners[] = T_IF;
 		$beginners[] = T_ELSEIF;
+		$beginners[] = T_RETURN;
+		$beginners = array_merge( $beginners, PHP_CodeSniffer_Tokens::$assignmentTokens );
 
 		$beginning = $this->phpcsFile->findPrevious( $beginners, $stackPtr, null, false, null, true );
 
@@ -68,7 +70,7 @@ class WordPress_Sniffs_PHP_YodaConditionsSniff extends WordPress_Sniff {
 			}
 
 			// If this is a function call or something, we are OK.
-			if ( in_array( $this->tokens[ $i ]['code'], array( T_CONSTANT_ENCAPSED_STRING, T_CLOSE_PARENTHESIS, T_OPEN_PARENTHESIS, T_RETURN ), true ) ) {
+			if ( T_CLOSE_PARENTHESIS === $this->tokens[ $i ]['code'] ) {
 				return;
 			}
 		}
--- a/WordPress/Sniffs/PHP/YodaConditionsSniff.php
+++ b/WordPress/Sniffs/PHP/YodaConditionsSniff.php
@@ -37,9 +37,10 @@ class WordPress_Sniffs_PHP_YodaConditionsSniff extends WordPress_Sniff {
 
 		$this->condition_start_tokens  = PHP_CodeSniffer_Tokens::$booleanOperators;
 		$this->condition_start_tokens += PHP_CodeSniffer_Tokens::$assignmentTokens;
-		$this->condition_start_tokens[ T_IF ] = T_IF;
-		$this->condition_start_tokens[ T_ELSEIF ] = T_ELSEIF;
+		$this->condition_start_tokens[ T_CASE ] = T_CASE;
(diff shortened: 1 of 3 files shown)


WordPress/Sniffs/PHP/YodaConditionsSniff.php
```diff
@@ -46,6 +46,8 @@ public function process_token( $stackPtr ) {
 		$beginners   = PHP_CodeSniffer_Tokens::$booleanOperators;
 		$beginners[] = T_IF;
 		$beginners[] = T_ELSEIF;
+		$beginners[] = T_RETURN;
+		$beginners = array_merge( $beginners, PHP_CodeSniffer_Tokens::$assignmentTokens );
```
JDGrimes: It is a separate issue, but I wonder if it would make more sense to create this list only once, and save it in a property.

jrfnl: Definitely! This list could be created & assigned to a property in the `register()` method, similar like we already do in several other sniffs.

jrfnl: Oh.. also, all the `PHP_CodeSniffer_Tokens` arrays have the same constant as both key as value. If that pattern was applied here for the three additional tokens, the `array_merge()` can be replaced with the faster `+` union operator.

WordPress/Sniffs/PHP/YodaConditionsSniff.php
```diff
@@ -46,6 +46,8 @@ public function process_token( $stackPtr ) {
 		$beginners   = PHP_CodeSniffer_Tokens::$booleanOperators;
 		$beginners[] = T_IF;
 		$beginners[] = T_ELSEIF;
+		$beginners[] = T_RETURN;
```
jrfnl: Some additional tokens which could be added here to prevent unnecessary token walking:
* T_SEMICOLON
* T_OPEN_PARENTHESIS
* T_CASE

Adding the `T_OPEN_PARENTHESIS` allows for removing the `T_IF` and `T_ELSEIF` by recognizing that the condition is within parenthesis, i.e. `$sample = ( true === $something )`, but also `if ( $something === true )`.

jrfnl:
Additional unit tests I ran which all passed correctly:
```php
if ( $sample = false !== strpos( $link, '%pagename%' ) ) {} // Ok.
if ( $sample = ( false !== strpos( $link, '%pagename%' ) ) ) {} // Ok.
if ( ( $sample = false ) !== strpos( $link, '%pagename%' ) ) {} // Ok.

switch ( true ) {
	case $sample === 'something': // Bad.
		// Do something.
		break;

	case 'something' === $sample: // OK.
		// Do something.
		break;
}

for ( $i = 0; $i !== 100; $i++ ) {} // Bad.
for ( $i = 0; 100 != $i; $i++ ) {} // OK.

do {
	// Something.
} while ( $sample === false ); // Bad.

do {
	// Something.
} while ( CONSTANT_A === $sample ); // OK.

while ( $sample === false ) {} // Bad.
while ( false != $sample ) {} // OK.

$a = ( $sample ) === 'yes';
```

The last unit test here is the only one I'm not sure of what the expected behaviour should be. Should this throw an error or not ?

JDGrimes:
> The last unit test here is the only one I'm not sure of what the expected behaviour should be. Should this throw an error or not ?

I think that we don't need to throw an error there, because the point of the sniff is mostly to avoid accidental assignment, which can't occur when the variable is enclosed in parentheses. A pattern like this would be especially useful when both sides of the comparison are variables, although we don't currently throw an error at all in that case.

jrfnl:
> I think that we don't need to throw an error there

In that case the sniff behaves as expected, so no need for any changes to accommodate this.

WordPress/Sniffs/PHP/YodaConditionsSniff.php
```diff
@@ -20,11 +20,28 @@
...
 	 * @return array
 	 */
 	public function register() {
+
+		$this->condition_start_tokens  = PHP_CodeSniffer_Tokens::$booleanOperators;
+		$this->condition_start_tokens += PHP_CodeSniffer_Tokens::$assignmentTokens;
+		$this->condition_start_tokens[ T_CASE ] = T_CASE;
+		$this->condition_start_tokens[ T_RETURN ] = T_RETURN;
+		$this->condition_start_tokens[ T_SEMICOLON ] = T_SEMICOLON;
+		$this->condition_start_tokens[ T_OPEN_PARENTHESIS ] = T_OPEN_PARENTHESIS;
```
jrfnl: Maybe align the `=`'s ?

JDGrimes: Like this?

```php
		$this->condition_start_tokens  = PHP_CodeSniffer_Tokens::$booleanOperators;
		$this->condition_start_tokens += PHP_CodeSniffer_Tokens::$assignmentTokens;
		$this->condition_start_tokens[ T_CASE ]             = T_CASE;
		$this->condition_start_tokens[ T_RETURN ]           = T_RETURN;
		$this->condition_start_tokens[ T_SEMICOLON ]        = T_SEMICOLON;
		$this->condition_start_tokens[ T_OPEN_PARENTHESIS ] = T_OPEN_PARENTHESIS;
```

Or the top two as well?

jrfnl: I'd personally align the whole list, but also initialize it as a local variable only to assign it to the property at the end so the lines don't get as long.

```php
		// Create the condition starters array only once.
		$starters                       = PHP_CodeSniffer_Tokens::$booleanOperators;
		$starters                      += PHP_CodeSniffer_Tokens::$assignmentTokens;
		$starters[ T_CASE ]             = T_CASE;
		$starters[ T_RETURN ]           = T_RETURN;
		$starters[ T_SEMICOLON ]        = T_SEMICOLON;
		$starters[ T_OPEN_PARENTHESIS ] = T_OPEN_PARENTHESIS;

		$this->condition_start_tokens = $starters
```

jrfnl:
@JDGrimes You'll need to rebase the PR to get the build to pass considering the recent changes in the Travis infrastructure.

PR #1042, opened 2017-07-20
Unit tests: Prevent bleed through between unit tests.

jrfnl:
Unit tests which have a `setUp()` method and use this to set the value of properties or command line parameters, should always have a `tearDown()` method in which they clean up after themselves.

This PR is part of the preparation for PHPCS 3.x compatibility.

### DIFF
--- a/WordPress/Tests/Arrays/ArrayAssignmentRestrictionsUnitTest.php
+++ b/WordPress/Tests/Arrays/ArrayAssignmentRestrictionsUnitTest.php
@@ -34,6 +34,14 @@ class WordPress_Tests_Arrays_ArrayAssignmentRestrictionsUnitTest extends Abstrac
 	}
 
 	/**
+	 * Reset the $groups property.
+	 */
+	protected function tearDown() {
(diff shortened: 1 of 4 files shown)


GaryJones:
For the very slim chance that ordering of things matter, since it's parent setUp THEN setting array value, shouldn't the tearDown be reset array value THEN parent tearDown?

Practically, it may not matter here, but treating the calls to parent methods as either the outer (or inner) calls, and not the first (or second) calls satisfy my coding senses.

jrfnl:
> For the very slim chance that ordering of things matter, since it's parent setUp THEN setting array value, shouldn't the tearDown be reset array value THEN parent tearDown?
>
> Practically, it may not matter here, but treating the calls to parent methods as either the outer (or inner) calls, and not the first (or second) calls satisfy my coding senses.

For all practical purposes, it doesn't matter as the parent does not have a `tearDown()` method, so effectively falls through to the empty `tearDown()` in the `PHPUnit_Framework_TestCase` class.

The call is only made on the off-chance that PHPCS at some point would add a `tearDown()` method.

All the same, happy to oblige to appease your coding senses ;-)

PR #1047, opened 2017-07-24
Make WPCS compatible with PHPCS 3.x

### LINKED ISSUE #718: PHPCS 3.0.0 compatibility.
The new version of PHP CodeSniffer has some nice features, but introduces breaking changes which mean that the WordPress coding standards are not compatible. 

I had a quick look, and I think the minimum required is to namespace the WPCS codebase and import dependencies to ensure everything is autoloaded.

For reference:
- [Alpha release notes with more info on new features](https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.0.0a1 )
- [Upgrade guide](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Version-3.0-Upgrade-Guide)
- [RC1 release notes](https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.0.0RC1)

jrfnl:
This PR provides cross-version compatibility for WPCS with PHPCS 2.x as well as 3.x.

The minimum PHPCS 3.x requirement will be PHPCS 3.0.2 due to various bugs in earlier versions of the PHPCS 3.x autoloader.

Please see the individual commits for more detailed information. The commits have been set up to facilitate easier code review.

For end users, this PR will allow them to upgrade to PHPCS 3.x. Aside from the path to the `phpcs` command changing in PHPCS 3.x (upstream change), there is no backward-compatibility break for them.

For sniff developers who contribute to WPCS, the way to run the unit tests changes and - unfortunately - will be different for running these in combination with PHPCS 2.x vs 3.x.
The information regarding running the unit tests in the `CONTRIBUTING.md` file has been updated to reflect this change.

The PR has been set up with eventually dropping PHPCS 2.x support in mind. Where possible I've made choices which will make the update to WPCS to drop PHPCS 2.x support easier and less involved.
I will open a separate issue documenting how to drop PHPCS 2.x support.

Fixes #718

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,11 +24,15 @@ When you introduce new `public` sniff properties, or your sniff extends a class 
 Sometimes, a sniff will flag code which upon further inspection by a human turns out to be OK.
 If the sniff you are writing is susceptible to this, please consider adding the ability to [whitelist lines of code](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Whitelisting-code-which-flags-errors).
 
-To this end, the `WordPress_Sniff::has_whitelist_comment()` method was introduced.
+To this end, the `WordPress\Sniff::has_whitelist_comment()` method was introduced.
 
 Example usage:
 ```php
-class WordPress_Sniffs_CSRF_NonceVerificationSniff extends WordPress_Sniff {
+namespace WordPress\Sniffs\CSRF;
+
+use WordPress\Sniff;
+
+class NonceVerificationSniff extends Sniff {
 
 	public function process_token( $stackPtr ) {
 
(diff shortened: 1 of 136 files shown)


CONTRIBUTING.md
```diff
@@ -48,92 +52,142 @@ When you introduce a new whitelist comment, please don't forget to update the [w
...
-composer install
-vendor/bin/phpunit --filter WordPress tests/AllTests.php
+Presuming you have installed PHP CodeSniffer and the WordPress-Coding-Standards as [noted in the README](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards#how-to-use-this), all you need now is `PHPUnit`.
+
+If you already have PHPUnit installed on your system: Congrats, you're all set.
+
+If not, you can navigate to the directory where the `PHP_CodeSniffer` repo is checked out and do `composer install` to install the `dev` dependencies.
+Alternatively, you can [install PHPUnit](https://phpunit.de/manual/5.7/en/installation.html) as a PHAR file.
+
+**Pro-tip**: PHPUnit is easiest to work with if you add the installation directory to your `PATH` environment variable.
```
jrfnl: I'll remove the "Pro-tip".

> This, and line 64, seems to be encouraging global install of PHPUnit, which isn't the best practice.

Isn't it (anymore) ?
It's kind of what we are encouraging in our installation instructions for WPCS and PHPCS.

I've always thought that PHPUnit falls under the same umbrella of global tools, though I imagine that version requirements may differ more widely for projects.

GaryJones: OK - I should clarify - it's not best practice according to my understanding of https://12factor.net/dependencies (or https://roots.io/twelve-factor-02-dependencies/ if you prefer).

For each of my projects, I include WPCS as a `require-dev`. I only use the global git checkout of PHPCS and WPCS for Editors which don't have project-specific settings.

jrfnl: @GaryJones I've already remove the pro-tip part.

Other than that, I think how to handle dependencies and what dependencies are, is a different discussion. IMO there is a difference between application dependencies and dev tools, but that's for another day (while sitting in the :sun_with_face:  somewhere with a glass of :beer: or :wine_glass: )

PR #1050, opened 2017-07-25
PHPCS 3.x compat: remove php 5.2 work arounds and PHP 5.2 specific code

jrfnl:
This PR is part of the PHPCS 3.x compatibility PR series.

See the individual commits for more details. These have been set up for easy review.


WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.php
```diff
@@ -67,18 +61,7 @@ public function getErrorList( $testFile = 'PrefixAllGlobalsUnitTest.inc' ) {
 
 			case 'PrefixAllGlobalsUnitTest.1.inc':
```
GaryJones: This case could simply fall through to the `default`. No need for a separate `return array()`.

jrfnl: Good point. Fixed

jrfnl:
FYI: now this PR is merged, the builds for PHP nightly will start to pass (were failing on a _`create_function` is deprecated since PHP 7.2_ notice)

PR #1051, opened 2017-07-25
PHPCS 3.x compat: Implement recognition of the `--ignore-annotations` setting

### LINKED ISSUE #862: Respect `--ignore-annotations`
In PHPCS 3.0 a new  `--ignore-annotations` command line option has been added to ignore the PHPCS native `@codingStandards` annotations.

I'd like to suggest that we respect that setting when checking for whitelist comments, i.e. ignore whitelist comments when that option has been enabled.

See https://github.com/squizlabs/PHP_CodeSniffer/issues/811#issuecomment-283538235

jrfnl:
This PR is part of the PHPCS 3.x compatibility PR series.

PHPCS 3.x introduces a `--ignore-annotations` command line option to ignore any and all inline file ignore / line ignore / coding standard setting changes etc.

This PR ensures that the WPCS specific whitelist comments will also be ignored if the `--ignore-annotations` command line option is passed.

Fixes #862

### DIFF
--- a/WordPress/PHPCSHelper.php
+++ b/WordPress/PHPCSHelper.php
@@ -112,4 +112,30 @@ class PHPCSHelper {
 		return $tab_width;
 	}
 
+	/**
+	 * Check whether the `--ignore-annotations` option has been used.
+	 *
+	 * @since 0.13.0
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcsFile Optional. The current file being processed.
+	 *
+	 * @return bool True if annotations should be ignored, false otherwise.
+	 */
+	public static function ignore_annotations( File $phpcsFile = null ) {
+		if ( class_exists( '\PHP_CodeSniffer\Config' ) ) {
+			// PHPCS 3.x.
+			if ( isset( $phpcsFile, $phpcsFile->config->annotations ) ) {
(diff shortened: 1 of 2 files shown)


WordPress/PHPCSHelper.php
```diff
@@ -112,4 +112,31 @@ public static function get_tab_width( File $phpcsFile ) {
...
+	 * Check whether the `--ignore-annotations` option has been used.
+	 *
+	 * @since 0.13.0
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcsFile Optional. The current file being processed.
+	 *
+	 * @return bool True if annotations should be ignored, false otherwise.
+	 */
+	public static function ignore_annotations( File $phpcsFile = null ) {
+		// @TODO THIS NEEDS SOME CAREFUL & THOROUGH TESTING!
```
JDGrimes: I'm assuming that this comment is intended to be left in the code for now?

jrfnl: Oh darn! No, that comment should be removed. Was a WIP comment to myself & I have tested it since and confirmed it works as expected.

jrfnl: @JDGrimes I've removed the comment now.

PR #1052, opened 2017-07-25
Build: improve our own codestyle checking

jrfnl:
* Run the sniffs using a higher PHP and PHPCS version.
* Minor simplification in the command to call phpcs.
* Show warnings, but do not break the build for it. Previously warnings wouldn't be shown at all.
* Move a number of command line options to the ruleset.
* Exclude the Composer `vendor` directory if it exists.
* Update the section in the `Contributing.md` regarding code style

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,9 @@ env:
 matrix:
   fast_finish: true
   include:
-    # Run PHPCS against WPCS. I just picked to run it against 5.5.
-    - php: 5.5
-      env: PHPCS_BRANCH=2.9 SNIFF=1
+    # Run PHPCS against WPCS. I just picked to run it against 7.0.
+    - php: 7.0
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -25,9 +25,9 @@ env:
 matrix:
   fast_finish: true
   include:
-    # Run PHPCS against WPCS. I just picked to run it against 5.5.
-    - php: 5.5
-      env: PHPCS_BRANCH=2.9 SNIFF=1
+    # Run PHPCS against WPCS. I just picked to run it against 7.0.
```
GaryJones: What's the benefit here?

jrfnl: Higher PHPCS branch: benefit from the latest fixes upstream
Higher PHP version: benefit from more tokens being natively recognized (as still not everything is backfilled in PHPCS).

Issue #1061, opened 2017-07-25
StrictInArraySniff should use strtolower in check


websupporter:
If you define your own ruleset in which you want to enforce constants in uppercase
( `$boolean = TRUE;`) but you want to use the `StrictInArray`-rule, you run into a problem, because this check relies on lowercase:
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/PHP/StrictInArraySniff.php#L83

This is not a problem, if you enforce lowercase anyway, but I think the `StrictInArraySniff` should not implicitly rely on this rule

jrfnl:
Related to #763

PR #1062, opened 2017-07-25
make comparision in StrictInArraySniff unaware of uppercase (Solves #1061)

websupporter:
With using `strtolower` the check does not implicitly rely on constants being lowercase.

Solves #1061

### DIFF
--- a/WordPress/Tests/PHP/StrictInArrayUnitTest.inc
+++ b/WordPress/Tests/PHP/StrictInArrayUnitTest.inc
@@ -35,3 +35,4 @@ array_keys( $testing, 'my_key' ); // Warning.
 array_keys( array( '1', 1, true ), 'my_key' ); // Warning.
 array_keys( $testing, 'my_key', false ); // Warning.
 array_keys( array( '1', 1, true ), 'my_key', false ); // Warning.
+in_array( 1, array( '1', 1 ), TRUE ); // Ok.
--- a/WordPress/Tests/PHP/StrictInArrayUnitTest.inc
+++ b/WordPress/Tests/PHP/StrictInArrayUnitTest.inc
@@ -34,5 +34,6 @@ array_keys( array( '1', 1, true ), 'my_key', true ); // Ok.
 array_keys( $testing, 'my_key' ); // Warning.
 array_keys( array( '1', 1, true ), 'my_key' ); // Warning.
 array_keys( $testing, 'my_key', false ); // Warning.
-array_keys( array( '1', 1, true ), 'my_key', false ); // Warning.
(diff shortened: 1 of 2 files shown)


GaryJones:
Thanks - I feel this needs a unit test entry though, to show that we're covered for upper and lowercase.

jrfnl:
Thank you for your PR. Looks good to me!

FYI: It will be a little while before your PR gets merged as we're working on finishing a dedicated release which will make WPCS compatible with PHPCS 3.x.

Ref: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1040

websupporter:
Hi,
thanks for the quick feedback. I hope, this is what is needed for the test.

emgk:
@websupporter You will need to append the line number and no. of error in `StrictInArrayUnitTest.php` file. 

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Tests/PHP/StrictInArrayUnitTest.php#L40-L54

websupporter:
Hi @emgk,
I want ` in_array( 1, array( '1', 1 ), TRUE ); // Ok. ` not to throw any warnings, like e.g. https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Tests/PHP/StrictInArrayUnitTest.inc#L3

jrfnl:
> You will need to append the line number and no. of error in StrictInArrayUnitTest.php file.

@emgk In this case that is actually not needed as the unit test is for a code sample which *used* to give an error where there shouldn't have been one.

WordPress/Tests/PHP/StrictInArrayUnitTest.inc
```diff
@@ -35,3 +35,4 @@ array_keys( $testing, 'my_key' ); // Warning.
 array_keys( array( '1', 1, true ), 'my_key' ); // Warning.
 array_keys( $testing, 'my_key', false ); // Warning.
 array_keys( array( '1', 1, true ), 'my_key', false ); // Warning.
+in_array( 1, array( '1', 1 ), TRUE ); // Ok.
```
GaryJones: There should be a clear line at the end of the file, please.

jrfnl: @websupporter That should be a new line **below** your code sample, not above it like you've now done. :smirk_cat:

websupporter: @jrfnl :man_facepalming: the next commit should do the trick.

PR #1068, opened 2017-07-26
PHPCS 3.x compat: Update the pre-commit hook code

jrfnl:
This update presumes that devs using this pre-commit hook will be using PHPCS 3.x.

N.B. Can someone who actually uses the pre-commit hook test this ? I'm on Windows and AFAIK this won't work there. /cc @westonruter

### DIFF
--- a/bin/pre-commit
+++ b/bin/pre-commit
@@ -11,4 +11,4 @@ fi
 
 find . \( -name '*.php' \) -exec php -lf {} \;
 
-phpunit --filter WordPress "$phpcs_root/tests/AllTests.php"
+phpunit --bootstrap="./../Test/phpcs3-bootstrap.php" --filter WordPress "$phpcs_root/tests/AllTests.php"
--- a/bin/pre-commit
+++ b/bin/pre-commit
@@ -11,4 +11,4 @@ fi
 
 find . \( -name '*.php' \) -exec php -lf {} \;
 
-phpunit --bootstrap="./../Test/phpcs3-bootstrap.php" --filter WordPress "$phpcs_root/tests/AllTests.php"
+PHPCS_DIR=$phpcs_root phpunit --bootstrap="./Test/phpcs3-bootstrap.php" --filter WordPress "$phpcs_root/tests/AllTests.php"


bin/pre-commit
```diff
@@ -11,4 +11,4 @@ fi
 
 find . \( -name '*.php' \) -exec php -lf {} \;
 
-phpunit --filter WordPress "$phpcs_root/tests/AllTests.php"
+PHPCS_DIR=$phpcs_root phpunit --bootstrap="./Test/phpcs3-bootstrap.php" --filter WordPress "$phpcs_root/tests/AllTests.php"
```
westonruter: The WPCS repo root has already been traversed into via `cd "$(git rev-parse --show-toplevel)"`, so no need for `../Test`.

Without the `PHPCS_DIR` environment var, I get:

```
Uh oh... can't find PHPCS. Are you sure you are using PHPCS 3.x ?

If you use Composer, please run `composer install`.
Otherwise, make sure you set a `PHPCS_DIR` environment variable in your phpunit.xml file
pointing to the PHPCS directory.

Please read the contributors guidelines for more information:
https://is.gd/WPCScontributing
```

jrfnl: @westonruter Thanks for testing & correcting this.

The `PHPCS_DIR` environment variable should be only be needed when not using Composer, but having it, even when Composer was used should not cause an issue.

Or did you also get the notice while using a Composer based install ?
(and did you find the notice useful ? :smile: )

westonruter: I wasn't actually using PHPCS via Composer. I had cloned it from GitHub. So I was confused.

jrfnl: Just making sure: did the updated instructions in the `Contributing.md  document help clear up the confusion ? If not, what can be improved both in this message and in the contributing document ?

westonruter: I actually didn't refer to it. I just naïvely switched to to 3.0 and hacked to get it to work.

jrfnl: *grin* I suppose that could work too ;-)

Issue #1066, opened 2017-07-26
Issue template


dingo-d:
Maybe it would be a good idea to have issue template. I noticed that some repositories have this - you need to describe your issue in details, and provide information like version of the software, OS this happens on, steps to reproduce etc.

Thoughts?

GaryJones:
Go for it. 

Using the raw versions of the files in https://github.com/GaryJones/plugin-boilerplate/tree/a2cca0aefc4df5585a7959851177fad9a12bd160/.github (has extra HTML comments only shown in raw) would be a good starting point.

GaryJones:
Should probably provide a `.github/SUPPORT.md` as well.
https://github.com/blog/2400-support-file-support

mikaelz:
Could be something like:
https://github.com/facebook/react/blob/master/docs/community/support.md
https://github.com/piwik/developer-documentation/blob/master/docs/support.md
https://github.com/jquery/learn.jquery.com/blob/master/page/about-jquery/additional-support.md

jrfnl:
I personally am not necessarily a fan of issue templates.

For most people there is a barrier to report something anyway (am I doing it wrong or is the software wrong?), especially if it is someone who doesn't "live" on GitHub but might even need to register for GH to report something in the first place.

An issue template which asks for more than the bare minimum, will throw up an insurmountable barrier for people and often cause them to not report something at all.

mikaelz:
But when they report it should be helpful for fixing. Saves some commenting. Sure it shouldn't be long, but some bare minimum as you mentioned.

SUPPORT.md is great to separate dev from support, of course if there is already an existing support channel.

GaryJones:
I like them, and if there were insurmountable evidence that contributors didn't use them, GitHub wouldn't have bothered to introduce them.

So long as the questions are guiding, then providing a starting point for teasing out the information that needed for it to be helpful, is going to be more successful than a blank box to fill, and then that first round of "Please provide information ..." back and forth, which this issue template seeks to avoid.

jrfnl:
> if there were insurmountable evidence that contributors didn't use them, GitHub wouldn't have bothered to introduce them.

Chicken/egg. There was no evidence at all as they didn't before GH introduced them...

PR #1072, opened 2017-07-27
Add issue template

dingo-d:
Added the issue template #1066

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,23 +1,11 @@
-<!--- Provide a general summary of the issue in the Title above -->
+<!--
+If you're submitting a bug report, it is helpful if you include the following information:
 
-## Current Behavior
-<!--- If reporting a bug, describe the behavior you're experiencing -->
-<!--- If suggesting a change/improvement, explain the difference from current behavior -->
+- A minimal snippet of example code that demonstrates the issue.
+- The error code for the sniff that is being triggered (you can see the sniff error codes by running `phpcs` with the `-s` flag).
+- The versions of PHPCS, WPCS and PHP you are using.
 
-## Expected Behavior
-<!--- If you're reporting a bug, describe what should happen -->
-<!--- If you're suggesting a change/improvement, tell us how it should work -->
+If you are submitting a feature describe what that feature would be and why it should be useful.
 
-## Steps to Reproduce (for bugs)
-<!--- Include the code that triggered the bug -->
-
-## Possible Solution
-<!--- Not obligatory, but suggest a fix/reason for the bug, -->
-<!--- or ideas how to implement the addition or change -->
-
-## Your Environment
-<!--- Include as many relevant details about the environment you experienced the bug in -->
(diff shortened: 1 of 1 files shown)


JDGrimes:
I agree with @jrfnl. I find that issue templates can be annoying and overwhelming at times. I think that keeping it simple is good.

In this case, I think it is important to remember that we're dealing with developers here, not non-techie users of a piece of software. We probably don't need to walk them through how to submit a bug report or request a feature. They likely deal with software features and bugs every day. 😄 The structure for reports that is provided by the different sections is nice, but I don't think it's needed/worth it in this case.

I think that really all we need to do is mention the extra bits of information that are helpful specifically for bug reports for sniffs. Something like:
```
<!--
If you're submitting a bug report, it is helpful if you include the following information:

- A minimal snippet of example code that demonstrates the issue.
- The error code for the sniff that is being triggered (you can see the sniff error codes by running `phpcs` with the `-s` flag).
- The versions of PHPCS and PHP you are using.

Thanks!
-->
```

jrfnl:
> I think it is important to remember that we're dealing with developers here, not non-techie users of a piece of software

Not necessarily so once the theme review team switches over to their new process. For a lot of theme devs WPCS is _just a piece of software_ and well outside their wheelhouse.

JDGrimes:
@jrfnl  True, something like that went through my mind, but I still contend that we probably don't need to give them this much generic instruction about how to write a feature request or bug report. I don't think it is as useful in our case. Mainly just the information specific to WPCS.

jrfnl:
@JDGrimes I'm totally with you on that and I completely agree that something along the lines of your example + maybe PHPCS/WPCS/PHP version would suffice. I just wanted to add a little nuance about our users.

dingo-d:
I will edit this when I get home :) Thanks for the feedback

dingo-d:
Will this be ok?

~~~md
<!--
If you're submitting a bug report, it is helpful if you include the following information:

- A minimal snippet of example code that demonstrates the issue.
- The error code for the sniff that is being triggered (you can see the sniff error codes by running `phpcs` with the `-s` flag).
- The versions of PHPCS, WPCS and PHP you are using.

If you are submitting a feature describe what that feature would be and why it should be useful.

Thanks!
-->
~~~

.github/issue_template.md
```diff
@@ -0,0 +1,11 @@
+<!--
+If you're submitting a bug report, it is helpful if you include the following information:
+
+- A minimal snippet of example code that demonstrates the issue.
+- The error code for the sniff that is being triggered (you can see the sniff error codes by running `phpcs` with the `-s` flag).
+- The versions of PHPCS, WPCS and PHP you are using.
+
+If you are submitting a feature describe what that feature would be and why it should be useful.
```
jrfnl: This sentence feels a bit off. Maybe something like:
_If you are submitting a feature request, please describe the feature in detail and why it would be useful._

/cc @GaryJones

PR #1075, opened 2017-07-27
Add a .gitattributes file

jrfnl:
This PR adds a `.gitattributes` file to keep the archives GH creates of the repo clean of development related files.
People using Composer can still get these files in their setup if they really want to, by using `--prefer-source`.

Refs:
* [Reddit: I don't need your tests in my production](https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production)
* [Blog: I don't need your tests in my production](https://blog.madewithlove.be/post/gitattributes/)

As the 0.13.0 release adds yet more files (`/Test` directory) which are not relevant to the end-user, I suggest this to be added to the 0.13.0 release.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,25 @@
+#
+# Exclude these files from release archives.
+# This will also make them unavailable when using Composer with `--prefer-dist`.
+# If you develop for WPCS using Composer, use `--prefer-source`.
+# https://blog.madewithlove.be/post/gitattributes/
+#
+/.travis.yml export-ignore
+/phpunit.xml.dist export-ignore
+/.github export-ignore
+/bin export-ignore
+/Test export-ignore
+/WordPress/Tests export-ignore
+
+#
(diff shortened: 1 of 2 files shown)


jrfnl:
> This change needs documenting in CONTRIBUTING.md, so that someone doesn't end up just pulling down half of the repo when trying to contribute.

@GaryJones Good point. I just had a look though and I believe the "Contributing patches" instructions in the `Contributing` are set up in such a way that this should not be a problem.

> To contribute an improvement to this project, fork the repo and open a pull request to the `develop` branch. 

As the unit test section _does_ refer to using Composer, I'll see if I can add it to that text somewhere.

jrfnl:
@GaryJones I've added a note in the Unit Test section of `CONTRIBUTING`.

GaryJones:
Needs a rebase, but good to go otherwise.

jrfnl:
Rebase done.

Issue #547, opened 2016-03-28
Spacing for Return Types


GaryJones:
For PHP 7, the format for adding return types appears to be of the format:

`function foo( $arg ): array {`

This format of characters and whitespace between `)` and `{` can be seen in examples on:
- [PHP Manual](http://php.net/manual/en/functions.returning-values.php#functions.returning-values.type-declaration)
- [PHP Return Type RFC](https://wiki.php.net/rfc/return_types)
- Draft [PSR-12](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#45-method-and-function-arguments)
- [SitePoint](http://www.sitepoint.com/php-7-revolution-return-types-removed-artifacts/) example.
- [This blog post](https://blog.engineyard.com/2015/what-to-expect-php-7)
- [And this blog post](https://blog.feryn.eu/php-7-is-now-available-new-features-improvements/) (which has no space after the return type.)

Just for a counter-argument, [this post](https://www.digitalocean.com/company/blog/getting-ready-for-php-7/) puts a space before the `:`.

Regardless, the current WPCS flags up `):` as `Expected one space after closing parenthesis`. It doesn't flag up no space between the return type and `{`.

We, as WPCS, either need to make a decision that `): array {` is correct for the WordPress community, or that particular combination of characters / tokens should be excluded from whitespace checks until TPTB have updated the coding standards.

JDGrimes:
It seems pretty clear-cut to me that there should be a space before the opening bracket. I think it wold be really inconsistent for WP not to adopt that (54 years from now :-).

As far as the space after the closing parenthesis, I think we should just allow it to be configured either way, but not flag it by default.

JDGrimes:
This error partly comes from an upstream sniff:

```
 3 | ERROR | [x] Expected 1 space after closing parenthesis; found 0
   |       |     (Generic.Functions.OpeningFunctionBraceKernighanRitchie.SpaceAfterBracket)
```

So partly this is because PHPCS doesn't fully support PHP 7 yet, I think. Although I also get this from our sniffs:

```
 3 | ERROR | [x] Space between opening control structure and closing
   |       |     parenthesis is required
   |       |     (WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceAfterCloseParenthesis)
```

Also, it looks like `WordPress.Arrays.ArrayDeclaration` isn't compatible with PHP 7:

```
[...]

Issue #1071, opened 2017-07-26
"No space after opening parenthesis is prohibited" when using traits


felixarntz:
I'm using traits in my code, and obviously WordPress doesn't. :)

The sniffer rules give give me the following error:
`No space after opening parenthesis is prohibited`

This happens for every line with a `use` statement for a trait, where I need to resolve a method conflict. For example:

    use Meta_Submodule_Trait, Settings_Submodule_Trait {
        Meta_Submodule_Trait::get_meta_fields as protected _get_meta_fields;
    }

(in the above example the error is always thrown for the first line)

This is valid PHP, but causes an issue, probably because the codesniffer hasn't yet considered such cases. Not sure whether I should just disable this, or whether the sniff could be adjusted.

jrfnl:
Related to #764.

@felixarntz Could you give me a more complete code sample ? The current one gives a syntax error so is not a valid test case.

felixarntz:
This is basically what the class looks like:

    class Access_Control extends Submodule implements Meta_Submodule_Interface, Settings_Submodule_Interface {
        use Meta_Submodule_Trait, Settings_Submodule_Trait {
            Meta_Submodule_Trait::get_meta_fields as protected _get_meta_fields;
        }

        [functions etc]
    }

jrfnl:
@felixarntz  Ok, thanks. I've been able to reproduce the issue.

For the short term, I'd suggest disabling that particular error message using:
```xml
<!-- Disable it selectively for a few files: -->
<rule ref="WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceAfterOpenParenthesis">
	<exclude-pattern>*/something*.php</exclude-pattern>
</rule>

<!-- Or disable it completely: -->
<exclude name="WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceAfterOpenParenthesis"/>
```

For the longer term, I think we need to create a dedicated sniff for `use` statement whitespace as there are so many variations of how use statements can be laid out because of its different uses, that the `WhiteSpace.ControlStructure` sniff which currently handles this, is not really suited for handling this.

As input for the dedicated sniff - it should handle all cases, i.e.:
1. Import/alias classes, traits, constants, etc. in namespaces,
2. Insert traits in classes,
3. Inherit variables in closures. 

And let's not forget the variations within that with `use function` / `use const` / group use statements etc.

Refs:
* http://php.net/manual/en/language.namespaces.importing.php
[...]

PR #1084, opened 2017-08-04
Add ReturnType Sniff

GaryJones:
Only slightly adapted from the [work](https://github.com/zendframework/zend-coding-standard/blob/52b435c714879609262aa36b004c4075cd967acc/src/ZendCodingStandard/Sniffs/Formatting/ReturnTypeSniff.php) Michał Bundyra (@webimpress) did for the Zend Coding Standard.

Required format for WordPress is the same as most others in the PHP community:

 - no space before the colon
 - exactly one space after colon before return type
 - no space after nullable operator (hasn't been discussed yet, but I see no reason to vary)
 - simple types should be given as lowercase

End result:

```php
function foo(): bool { // Simple return type.
};

function foo(): ?string { // Nullable simple return type.
};

function foo( $a ): \MyClass { // Return type of a class.
};
```

❗️ Note: Not yet added to `WordPress-Core/ruleset.xml`.

See #547.


GaryJones:
My own notes:

**To get the line numbers that have errors in tests:**

```sh
# In WPCS directory.
$ phpcs --standard=WordPress -s WordPress/Tests/Functions/ReturnTypeUnitTest.inc --sniffs=WordPress.Functions.ReturnType
```

**To run tests:**

```sh
# In WPCS directory.
$ ../phpcs/vendor/bin/phpunit --bootstrap="./Test/phpcs3-bootstrap.php" --filter WordPress ../phpcs/tests/AllTests.php
```

**To check code standards for this new code:**

```sh
# In WPCS directory.
$ phpcs WordPress/Sniffs/Functions --standard=bin/phpcs.xml
$ phpcs WordPress/Tests/Functions --standard=bin/phpcs.xml
```

jrfnl:
There is an open PR for return type declarations upstream: https://github.com/squizlabs/PHP_CodeSniffer/pull/1219/
Would that implementation work for us ? and if so, why not use that as basis to make it easier to use the upstream version eventually ?

GaryJones:
The [one I based mine on](https://github.com/squizlabs/PHP_CodeSniffer/issues/1527) is referenced in an issue labelled with 3.2.0, whereas the one you've highlighted isn't :-)

I'm happy to defer to upstream if/as/when it gets there, but in the meantime, we could be using this today.

GaryJones:
Rebased to hopefully fix the test failures.

Further test cases to add and apply fixes for:

```php
function foog(): // a comment
array {}

function fooh() // a comment
: array {}
```

should be fixed to become:

```php
function foog(): array {} // a comment
function fooh(): array {} // a comment
```

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+					$this->phpcsFile->fixer->replaceToken( $nullable + 1, '' );
+				}
+			}
+		}
+
+		$first   = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $nullable ?: $colon ) + 1, null, true );
+		$end     = $this->phpcsFile->findNext( array( T_SEMICOLON, T_OPEN_CURLY_BRACKET ), $stackPtr + 1 );
+		$last    = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $end - 1, null, true );
+		$invalid = (int) $this->phpcsFile->findNext( array( T_STRING, T_NS_SEPARATOR, T_RETURN_TYPE ), $first, $last + 1, true );
+		if ( $invalid ) {
```
JDGrimes: Is there a particular reason to check for invalid tokens here? Doesn't that just indicate a parse error/live coding? If so, we shouldn't be giving an error here, just bowing out if we can't continue.

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+			}
+
+			return;
+		}
+
+		$returnType = trim( $this->phpcsFile->getTokensAsString( $first, $last - $first + 1 ) );
+
+		if ( $first === $last
+			&& ! in_array( $returnType, $this->simple_return_types, true )
+			&& in_array( strtolower( $returnType ), $this->simple_return_types, true )
```
JDGrimes: This array could be flipped and `isset()` used instead. Although I guess we're trying to stay close to the upstream version.

WordPress/Sniffs/Functions/ReturnTypeSniff.php
GaryJones: Maybe, but if we can determine that it appears to be _not_ invalid, then we can also then check and fix any of the simple return types not being in lowercase as well.

If it's an invalid token, then it appears to try and remove it (but may leave the colon - I don't see any unit tests covering this).

WordPress/Sniffs/Functions/ReturnTypeSniff.php
GaryJones: There is no upstream as such, so flipped array in 7bcea29eddf726e218c53d6a244e88f86fbf28e1.

WordPress/Sniffs/Functions/ReturnTypeSniff.php
GaryJones: Possible test cases:
```php
function a(): invalidtoken {}
function b(): invalid_token {}
function c() :invalid_token {}
```

But what should they be fixable to (if anything)?

JDGrimes: I don't think that we should attempt to fix them. We don't attempt to fix or even warn about syntax errors elsewhere. See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/793

Also, those examples are valid tokens, aren't they? (`T_STRING` is valid, might be a class name.) It would have to be something like this:

```
function a(): $foo {}
```

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+					$this->phpcsFile->fixer->replaceToken( $token, '' );
+					-- $token;
+				} while ( T_CLOSE_PARENTHESIS !== $this->tokens[ $token ]['code'] );
+				$this->phpcsFile->fixer->endChangeset();
+			}
+		}
+
+		// Only one space after colon.
+		if ( T_WHITESPACE !== $this->tokens[ $colon + 1 ]['code'] ) {
+			$error = 'There must be a space after colon and before return type declaration.';
```
jrfnl: Suggestion: "_There must be one space between the colon and the return type. None found._"

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+			$fix   = $this->phpcsFile->addFixableError( $error, $colon, 'NoSpaceAfterColon' );
+			if ( true === $fix ) {
+				$this->phpcsFile->fixer->addContent( $colon, ' ' );
+			}
+		} elseif ( ' ' !== $this->tokens[ $colon + 1 ]['content'] ) {
+			$error = 'There must be exactly one space after colon and before return type declaration.';
```
jrfnl: Suggestion: "There must be exactly one space between the colon and the return type. Found: %s" - where %s should be replaced with either "newline" or "x spaces", similar to how it's done here: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/PSR2/Sniffs/ControlStructures/ControlStructureSpacingSniff.php#L81-L87

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	/**
+	 * Processes this test, when one of its tokens is encountered.
+	 *
+	 * @param int $stackPtr The position of the current token in the stack.
+	 * @return null Return before end if Return type declaration contains an invalid token.
+	 */
+	public function process_token( $stackPtr ) {
+		$colon = (int) $this->phpcsFile->findPrevious( T_COLON, $stackPtr - 1 );
```
jrfnl: 1. Don't cast the return to `(int)`, token index `0` exists.
2. You probably want to limit the search to the statement: `$this->phpcsFile->findPrevious( T_COLON, $stackPtr - 1, null, false, null, true );`
3. Always check that the return is not `false` before using it. Shouldn't ever happen in this case, but it's better to be safe than sorry. I've come across too many tokenizer bugs not to.
```php
if ( false === $colon ) {
	// Shouldn't happen, but just in case.
	return;
}

File: WordPress/Sniffs/Functions/ReturnTypeSniff.php [hunk 6]
```diff
@@ -0,0 +1,141 @@
...
+		return array(
+			T_RETURN_TYPE,
+		);
+	}
+
+	/**
+	 * Processes this test, when one of its tokens is encountered.
+	 *
+	 * @param int $stackPtr The position of the current token in the stack.
+	 * @return null Return before end if Return type declaration contains an invalid token.
```
jrfnl: `Return type` => `return type`

`@return null` => `@return void`

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+	 */
+	public function process_token( $stackPtr ) {
+		$colon = (int) $this->phpcsFile->findPrevious( T_COLON, $stackPtr - 1 );
+
+		// Space before colon disallowed.
+		if ( T_CLOSE_PARENTHESIS !== $this->tokens[ $colon - 1 ]['code'] ) {
```
jrfnl: `if ( isset( $this->tokens[ $colon - 1 ] ) && ...`

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+			$error = 'There must be no space before colon before a return type.';
+			$fix   = $this->phpcsFile->addFixableError( $error, $colon - 1, 'SpaceBeforeColon' );
+
+			if ( true === $fix ) {
+				$this->phpcsFile->fixer->beginChangeset();
+				$token = $colon - 1;
+				do {
+					$this->phpcsFile->fixer->replaceToken( $token, '' );
+					-- $token;
+				} while ( T_CLOSE_PARENTHESIS !== $this->tokens[ $token ]['code'] );
```
jrfnl: Probably superfluous, but still: `while ( $token > 0 && ...`

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+			}
+		} elseif ( ' ' !== $this->tokens[ $colon + 1 ]['content'] ) {
+			$error = 'There must be exactly one space after colon and before return type declaration.';
+			$fix   = $this->phpcsFile->addFixableError( $error, $colon + 1, 'TooManySpacesAfterColon' );
+			if ( true === $fix ) {
+				$this->phpcsFile->fixer->replaceToken( $colon + 1, ' ' );
+			}
+		}
+
+		$nullable = (int) $this->phpcsFile->findNext( T_NULLABLE, $colon + 1, $stackPtr );
```
jrfnl: Don't cast to `(int)` and check in the `if` against `false`.

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+		if ( $nullable ) {
+			// Check if there is space after nullable operator.
+			if ( T_WHITESPACE === $this->tokens[ $nullable + 1 ]['code'] ) {
+				$error = 'Space is not not allowed after nullable operator.';
```
jrfnl: `not not`

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+				$fix   = $this->phpcsFile->addFixableError( $error, $nullable + 1, 'SpaceAfterNullable' );
+				if ( $fix ) {
```
jrfnl: `if ( true === $fix )`

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+					$this->phpcsFile->fixer->replaceToken( $nullable + 1, '' );
+				}
+			}
+		}
+
+		$first   = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $nullable ?: $colon ) + 1, null, true );
+		$end     = $this->phpcsFile->findNext( array( T_SEMICOLON, T_OPEN_CURLY_BRACKET ), $stackPtr + 1 );
+		$last    = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $end - 1, null, true );
+		$invalid = (int) $this->phpcsFile->findNext( array( T_STRING, T_NS_SEPARATOR, T_RETURN_TYPE ), $first, $last + 1, true );
```
jrfnl: Don't cast to (int) and check in the if against false.

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+				$fix   = $this->phpcsFile->addFixableError( $error, $nullable + 1, 'SpaceAfterNullable' );
+				if ( $fix ) {
+					$this->phpcsFile->fixer->replaceToken( $nullable + 1, '' );
+				}
+			}
+		}
+
+		$first   = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $nullable ?: $colon ) + 1, null, true );
+		$end     = $this->phpcsFile->findNext( array( T_SEMICOLON, T_OPEN_CURLY_BRACKET ), $stackPtr + 1 );
+		$last    = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $end - 1, null, true );
```
jrfnl: Check `$end !== false` before using it. This would throw notices when code is checked in live coding situations in an IDE.

WordPress/Sniffs/Functions/ReturnTypeSniff.php
jrfnl: I'm in agreement with @JDGrimes here. I'd actually say that this whole check for invalid tokens should be removed, as in line 109 -122.

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+			$fix   = $this->phpcsFile->addFixableError( $error, $invalid, 'InvalidToken', $data );
+			if ( true === $fix ) {
+				$this->phpcsFile->fixer->replaceToken( $invalid, '' );
+			}
+
+			return;
+		}
+
+		$return_type = trim( $this->phpcsFile->getTokensAsString( $first, $last - $first + 1 ) );
+
```
jrfnl: This whole check can be simplified as we already received the `T_RETURN_TYPE` as the `$stackPtr` and simple return types are only one token:
```php
$contentLC = strtolower( $this->tokens[ $stackPtr ]['content'] );
if ( isset( $this->simple_return_types[ $contentLC ] && $contentLC !== $this->tokens[ $stackPtr ]['content'] ) {
	// Throw fixable error.
	// ... reusing the $contentLC variable...
}
```

WordPress/Sniffs/Functions/ReturnTypeSniff.php
GaryJones: The reasons for the casts (here, and one or two other places in this PR), is that it is then treated as an int, but the fact it could be a bool returned, makes my IDE sad.

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -0,0 +1,141 @@
...
+		// Space before colon disallowed.
+		if ( T_CLOSE_PARENTHESIS !== $this->tokens[ $colon - 1 ]['code'] ) {
+			$error = 'There must be no space before colon before a return type.';
+			$fix   = $this->phpcsFile->addFixableError( $error, $colon - 1, 'SpaceBeforeColon' );
+
+			if ( true === $fix ) {
+				$this->phpcsFile->fixer->beginChangeset();
+				$token = $colon - 1;
+				do {
+					$this->phpcsFile->fixer->replaceToken( $token, '' );
```
GaryJones: I'd come across the disappearing comments but removed them from the tests until I'd come up with a solution.

WordPress/Tests/Functions/ReturnTypeUnitTest.inc.fixed
```diff
@@ -0,0 +1,80 @@
+<?php
+
+function fooa(): array {} // Good.
+function foob(): array {} // Bad.
+function fooc(): array {} // Bad.
+function food(): array {}
```
GaryJones: No, I took the comment out of the .fixed file to make the tests pass for now :-)

WordPress/Sniffs/Functions/ReturnTypeSniff.php
```diff
@@ -74,67 +79,48 @@ public function process_token( $stackPtr ) {
...
+			$error = 'There must be one space between the colon and the return type. None found.';
 			$fix   = $this->phpcsFile->addFixableError( $error, $colon, 'NoSpaceAfterColon' );
 			if ( true === $fix ) {
 				$this->phpcsFile->fixer->addContent( $colon, ' ' );
 			}
 		} elseif ( ' ' !== $this->tokens[ $colon + 1 ]['content'] ) {
-			$error = 'There must be exactly one space after colon and before return type declaration.';
-			$fix   = $this->phpcsFile->addFixableError( $error, $colon + 1, 'TooManySpacesAfterColon' );
+			$error = 'There must be exactly one space between the colon and the return type. Found: %s';
+			$data = array( 'more than one' ); /* @var @todo Count actual whitespaces */
```
GaryJones: WIP :-)

GaryJones:
> We've had discussions about licensing issues with copying code from other standards before. You've indicated that the sniff is largely based on work from @webimpress for the Zend Framework standard which is copyrighted and licensed under BSD-3. I'm not a legal copyright expert, so is this usage permitted

My understanding is, that as long as we keep copyright / credits, BSD-3 can be used in MIT projects.

In this case (and it may not make a difference), their work is not yet on their master branch, or within a release, so it may not even count as being published.

With the PR feedback given, it's also had some significant changes to it, with sections deleted, and more test cases addressed )i.e. more than just syntax changes).


 > or would it be better/preferred to use Composer to add the ZF standard to a project and just include that one sniff in a custom ruleset ?

Unless @webimpress strongly dictates otherwise, I'd be against this. As mentioned, more test cases have been added and catered for, and even the error messages have more of WPCS flavour about them.

I'm more than happy to add more explicit credits as to the original source, but there is definitely some non-trivial changes made, and I'd hope that in the spirit of open source, these changes could be added back into the Zend Coding Standard, or otherwise make it upstream so everyone could benefit.

> Code style
> 
> I did a compare between this version and the original and - aside from code style -, there are hardly any differences.

As above - there are more than just code style changes now (since you wrote the comment). Theirs is an unreleased Sniff, and ours handles things in a better way.

 
> What about function declarations with quite some variables/defaults making for a very long line ? By the looks of the unit tests, the return type declaration is now fixed to be on the same line as the closing parenthesis. Should new lines be allowed ? Possibly as a configurable property, like in several of the Squiz sniffs ignoreNewlines ?

There's at least one test case that allows for `): array {` to be on its own line. Happy to add more test cases if you feel a case is missing.


> I see no mention of "exactly one space between the end of the return type and the brace". While this is already checked via another sniff (Generic.Functions.OpeningFunctionBraceKernighanRitchie), it should possibly be mentioned in the WP Core coding standards handbook.
> For that matter, the "brace should be on the same line as the class/function declaration" rule is implied in code samples used in the handbook in the "space usage" section, but it isn't actually made explicit either. /cc @pento

As you said, it's not in the handbook, and covered by another sniff anyway, so I'm not sure what you need this PR to clarify here?


> See inline comments. Gone through it with a fine toothcomb. Hope you'll find it useful.

Very useful, thanks - this was my first attempt at amending the internals of a Sniff (albeit starting from a very good base from @webimpress), so it was a case of jumping in at the deep-end and figuring out what the workflow should be. Definitely useful to have detailed feedback!

jrfnl:
> In this case (and it may not make a difference), their work is not yet on their master branch, or within a release, so it may not even count as being published.

Copyright starts at the moment something is written. Not when it is published.

If I understand it correctly, the sniff is not yet merged into the ZFCS ? In that case, it might well be that their license does not apply (yet) and that it falls under the ""normal" copyright, i.e. all rights reserved unless otherwise mentioned and the sniff itself did not contain any such mentions, so that would - for the moment - make it off-limits.

> I'm more than happy to add more explicit credits as to the original source, but there is definitely some non-trivial changes made, and I'd hope that in the spirit of open source, these changes could be added back into the Zend Coding Standard, or otherwise make it upstream so everyone could benefit.

I'd like to hear from @webimpress here, but yes, as - since the original PR was made - the changes have become more extensive, it can be seen as a derivative work.
And yes, I, for one, would very much cheer you on for giving back the changes made here to the original source.

> As you said, it's not in the handbook, and covered by another sniff anyway, so I'm not sure what you need this PR to clarify here?

Nothing for this PR here, but I did wonder if the handbook could do with clarification on that point and this sniff inspired that thought which is why I mentioned it here.

webimpress:
@jrfnl @GaryJones 
Sorry, I've been on holidays.
I had a look on your discussion here, and based on that I've made couple changes to sniff in ZendCodingStandard. The latest version is here:
https://github.com/zendframework/zend-coding-standard/blob/feature/sniffs/src/ZendCodingStandard/Sniffs/Formatting/ReturnTypeSniff.php

The main thing I have changed there is possibility to configure the sniff - you can set how many spaces must be before/after colon and after nullable operator.

Probably it still could be improved, but all tests provided here also pass with my sniff.
As you can see ZendCodingStandard is going to be big library with multiple sniffs.
I think it would be nice if you'd like to use the library thru composer than copy it across. Then you can also use another sniffs which I have implemented there. I know that it is not yet released and probably it will take some time, so it's your decision what to do with it.

GaryJones:
@webimpress Thanks for following up :-)

Would your plan be to get this sniff into PHPCS repo itself?

GaryJones:
Side note for this PR: WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceAfterCloseParenthesis may need adjusting, as it complains of a "space being needed between opening control structure and closing parenthesis" when the space before the colon is missing.

jrfnl:
> Side note for this PR: WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceAfterCloseParenthesis may need adjusting, as it complains of a "space being needed between opening control structure and closing parenthesis" when the space before the colon is missing.

Related to #1071 - The `_T_FUNCTION` and `T_CLOSURE` (and `T_USE` for that matter) tokens have been added to that sniff in WPCS, but are not control structures. They should be handled separately completely as the control structure sniff is not suitable for those.

jrfnl:
Just thinking: should there be some unit tests covering closures in combination with return type declarations ?

PR #1089, opened 2017-08-05
New `WordPress.WP.DiscouragedConstants` sniff

### LINKED ISSUE #97: Flag use of TEMPLATEPATH and STYLESHEETPATH
It is highly desirable to use `get_template_directory()` and `get_stylesheet_directory()` because these can be filtered by plugins, whereas constants cannot.

jrfnl:
This sniff is based on a sniff previously pulled to the TRT fork of WPCS as `Theme.DeprecatedConstants`.

Related issue: WPTRT/WordPress-Coding-Standards/issues/23
Original PR: WPTRT/WordPress-Coding-Standards/pull/30
Covers the list of constants found in: Otto42/theme-check/pull/162

Differences between that sniff and this one:
* This sniff is a **_lot_** more comprehensive in preventing false positives.
    The original sniff was one of the first ones I wrote and it's kind of sweet to look back at that code which I wrote over a year ago and see how much I've learned about writing sniffs in the mean time.
* The original sniff was called `Theme.DeprecatedConstants`, but in reality, most of these constants aren't officially deprecated, though their usage is discouraged and the constants being addressed are not 100% related to themes either. With that in mind, I've renamed the sniff to `WP.DiscouragedConstants` and changed the error level from Error to Warning.
    Also see: https://core.trac.wordpress.org/ticket/18298
* This version of the sniff also addresses the concerns raised in WPTRT/WordPress-Coding-Standards/issues/ 110 about themes `define`-ing any of these constants and leverages the `AbstractFunctionParameterSniff` to do so.

Includes quite extensive unit tests. More are definitely welcome, especially to prevent false positives.

I've added the sniff to the `WordPress-Extra` ruleset.

Based on https://vip.wordpress.com/documentation/code-review-what-we-look-for/#using-theme-constants and https://core.trac.wordpress.org/ticket/18298#comment:2, I've also added the sniff to the `WordPress-VIP` ruleset. /cc @sboisvert @david-binda @ntwb

Fixes #97
Will also fix WPTRT/WordPress-Coding-Standards/issues/110 once the original sniff gets removed from the TRT fork and this sniff is added to the TRT ruleset instead.

----
Update: This PR also adds two new utility methods to the `WordPress\Sniff` class: `is_class_constant()` and `valid_direct_scope()`.

### DIFF
--- a/WordPress/Tests/WP/DiscouragedConstantsUnitTest.inc
+++ b/WordPress/Tests/WP/DiscouragedConstantsUnitTest.inc
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * Make sure that global constants are correctly identified.
+ *
+ * The below are all ok.
+ */
+// Ok, not a constant.
+namespace STYLESHEETPATH {}
+namespace MY\OTHER\STYLESHEETPATH\NAMESPACE {}
+use STYLESHEETPATH;
+use Something, STYLESHEETPATH, SomethingElse;
+class STYLESHEETPATH {
+	const STYLESHEETPATH = 'something';
+	private function STYLESHEETPATH() {}
+}
+class ABC extends STYLESHEETPATH {}
+class DEF implements STYLESHEETPATH {}
+interface STYLESHEETPATH {}
+trait STYLESHEETPATH {}
+const STYLESHEETPATH = 'something';
+$a = new STYLESHEETPATH;
+$a = new STYLESHEETPATH();
+function STYLESHEETPATH() {}
+echo STYLESHEETPATH();
+echo My\UsedAsNamespace\STYLESHEETPATH\something;
+My\UsedAsNamespace\STYLESHEETPATH\something::something_else();
+if ( $abc instanceof STYLESHEETPATH ) {}
+
+goto STYLESHEETPATH;
+// Something.
+STYLESHEETPATH:
+// Something.
+
+// Ok, not a global constant.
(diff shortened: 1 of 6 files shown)


WordPress/Tests/WP/DiscouragedConstantsUnitTest.inc
```diff
@@ -0,0 +1,73 @@
...
+use Something, STYLESHEETPATH, SomethingElse;
+class STYLESHEETPATH {
+	const STYLESHEETPATH = 'something';
+	private function STYLESHEETPATH() {}
+}
+class ABC extends STYLESHEETPATH {}
+class DEF implements STYLESHEETPATH {}
+interface STYLESHEETPATH {}
+trait STYLESHEETPATH {}
+const STYLESHEETPATH = 'something';
```
JDGrimes: Why is this OK, if calling `define()` is not OK? I'd think that if we sniff for `define()`, we ought to also sniff for `const` outside of a class or namespace.

jrfnl: Good question. I added a lot of code to prevent false positives and only addressed the `define()` bit after that. Clearly forgot to also change things for `const`.

I'm going to change this.

jrfnl: Done, see separate commit.

PR #1093, opened 2017-08-07
Add new `CodeAnalysis.AssignmentInCondition` sniff to `Extra`

jrfnl:
Most of the time variable assignments being made in conditions are unintentional and are in actual fact typos which were intended as a comparison.
In those cases when it isn't a typo, this is a typical code smell which is a prime candidate for refactoring.

This sniff will warn when any such a variable assignment is found in a condition.

Note: the sniff does currently **not** detect variable assignments in the conditional part of ternaries.

Inspired by [this twitter thread](https://twitter.com/jrf_nl/status/893891839903838209), or rather: the responses I received to my tweet.

Note: the fact that WP uses Yoda conditions does **not** invalidate the need for this sniff.
The Yoda conditions sniff only checks comparison operators, not assignment operators.

I've run this sniff over WP Core to:
a) test the sniff and
b) see what it would turn up

And found 1004 cases in 847 files and visual verification of a sample of theses results showed no false positives.

AFAICS the cases found in WP Core are intentional assignments within conditions, however, as said above, that is an indicator that the code in question is due for some refactoring as this is definitely a code smell.

This sniff is a duplicate of the same which I've pulled upstream: squizlabs/PHP_CodeSniffer/pull/1594
The upstream sniff has been merged (rather quickly) and is due to be included in PHPCS 3.1.0. Once the minimum PHPCS requirement for WPCS has gone up beyond that version (i.e. after PHPCS 2.x support has been dropped), we can safely remove the WPCS version of the sniff.

### DIFF
--- a/WordPress/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
+++ b/WordPress/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
@@ -169,7 +169,7 @@ class AssignmentInConditionSniff extends Sniff {
 
 			if ( true === $hasVariable ) {
 				$this->phpcsFile->addWarning(
-					'Variable assignment found within a condition. Did you mean to do a comparison ?',
+					'Variable assignment found within a condition. Did you mean to do a comparison?',
 					$hasAssignment,
 					'Found'
 				);
@@ -181,4 +181,4 @@ class AssignmentInConditionSniff extends Sniff {
 
 	}
 
-} // End class.
+}
--- a/WordPress/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
+++ b/WordPress/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
@@ -1,0 +1,184 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPress\Sniffs\CodeAnalysis;
+
+use WordPress\Sniff;
+use PHP_CodeSniffer_Tokens as Tokens;
+
+/**
+ * Detects variable assignments being made within conditions.
+ *
+ * This is a typical code smell and more often than not a comparison was intended.
+ *
+ * Note: this sniff does not detect variable assignments in the conditional part of ternaries!
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   0.14.0
+ *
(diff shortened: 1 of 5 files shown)


WordPress/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
```diff
@@ -0,0 +1,184 @@
...
+
+				// If this is a function call or something, we are OK.
+				if ( T_CLOSE_PARENTHESIS === $this->tokens[ $i ]['code'] ) {
+					break;
+				}
+			}
+
+			if ( true === $hasVariable ) {
+				$this->phpcsFile->addWarning(
+					'Variable assignment found within a condition. Did you mean to do a comparison ?',
```
GaryJones: No space needed before question mark.

WordPress/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
```diff
@@ -0,0 +1,184 @@
...
+				);
+			}
+
+			$startPos = $hasAssignment;
+
+		} while ( $startPos < $closer );
+
+	}
+
+} // End class.
```
GaryJones: No comment needed.

WordPress/Tests/CodeAnalysis/AssignmentInConditionUnitTest.inc
```diff
@@ -0,0 +1,94 @@
...
+for ( $i = 0; $i = 100; $i++ ) {}
+for ( $i = 0; $i = 100 && $b = false; $i++ ) {}
+
+do {
+} while ( $sample = false );
+
+while ( $sample = false ) {}
+
+if ($a = 123) :
+endif;
```
GaryJones: How about:

```php
if( $a = (int) $b ){}
```

? Seems to be covered already, but won't hurt to show it in a test case.

WordPress/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
jrfnl: I'll change this, but be aware that the upstream version has been merged already and does include the space, so the message will change back by the time we switch over to the upstream version (unless someone changes it there too).

To be honest, most of the time it takes some time before something is merged upstream, so I thought I'd still have time to update the PR there after pulling it here, but for some reason this one got merged *very* quickly upstream ;-)

WordPress/Tests/CodeAnalysis/AssignmentInConditionUnitTest.inc
jrfnl: The sniff only examines the left side of the assignment to see if there's a variable. The right side of the condition is never looked at, so this unit test does not cover anything (new).

PR #1094, opened 2017-08-07
Allow for setting the minimum supported WP version from the command line

jrfnl:
This PR addresses the comments made here: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/826#discussion_r127965474

This change allows for setting the `minimum_supported_version` property for all three `WP.Deprecated...` sniffs in one go, either via the command-line or by setting a config variable in a custom ruleset.

As this config variable is no longer directly related to a sniff, I've chosen to make the name slightly more specific/descriptive: `minimum_supported_wp_version` <= note the `_wp_`.

Usage:
```bash
phpcs . --standard=WordPress --runtime-set minimum_supported_wp_version 4.5
```

or in a ruleset:
```xml
<config name="minimum_supported_wp_version" value="4.5"/>
```

The above one liner can replace the custom property which would otherwise need to be set for all three sniffs individually.

Overrule order:
1. Value passed via the command line
2. `<config>` property in the ruleset
3. Individual sniff `<property>`'s set via the ruleset
4. Default value

Simplest way to test this is by running PHPCS over the `WP.DeprecatedParameter` sniff unit tests:
```bash
phpcs -p -s ./WordPress/Tests/DeprecatedParametersUnitTest.inc --standard=WordPress --sniffs=WordPress.WP.DeprecatedParameters

phpcs -p -s ./WordPress/Tests/DeprecatedParametersUnitTest.inc --standard=WordPress --sniffs=WordPress.WP.DeprecatedParameters --runtime-set minimum_supported_wp_version 4.8

phpcs -p -s ./WordPress/Tests/DeprecatedParametersUnitTest.inc --standard=WordPress --sniffs=WordPress.WP.DeprecatedParameters --runtime-set minimum_supported_wp_version 3.8
```

Additional notes:
* I've moved the `$custom_test_class_whitelist` property up in the `WordPress\Sniff` class, so both `public` properties are now at the top of the file.

-----
 To do once this PR is merged:
- [ ] Update the customizable properties wiki page with information about this feature

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -12,6 +12,7 @@ namespace WordPress;
 use PHP_CodeSniffer_Sniff as PHPCS_Sniff;
 use PHP_CodeSniffer_File as File;
 use PHP_CodeSniffer_Tokens as Tokens;
+use WordPress\PHPCSHelper;
 
 /**
  * Represents a PHP_CodeSniffer sniff for sniffing WordPress coding standards.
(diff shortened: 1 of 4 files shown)


grappler:
This breaks backwards compatibility? Just checking so that I know to update my custom rulesets.

jrfnl:
> This breaks backwards compatibility? Just checking so that I know to update my custom rulesets.

No, it does not break backwards compatibility, the custom properties are still supported as they were. You can just overrule them now in one go using the `<config>` setting / `--runtime-set` or replace them with `<config>`, but you don't *need* to.

jrfnl:
FYI: Info about the effect of this PR has now been added to the wiki.

PR #1097, opened 2017-08-07
XSS.EscapeOutput sniff: Fix issue #933 - namespace separators.

### LINKED ISSUE #933: \ before global namespace functions causes an escaping function error.
Related to #764.

This method is in a namespaced class:

```php
public function do_footer_nav() {
	echo \wp_kses_post(
		\genesis_get_nav_menu(
			[
				'menu_class'     => 'menu genesis-nav-menu menu-footer',
				'theme_location' => 'footer',
			]
		)
	);
}
```

The presence of the `\` before the function calls from the global namespace, means that an error of `Expected next thing to be an escaping function, ... not "\"` appears.

I've not looked into the change needed, but if the sniff could allow an optional `\` before those function checks, that would be great :-)

Using fully-qualified function calls is faster. Details and benchmarks are [here](https://github.com/Roave/FunctionFQNReplacer). This is from an inspection from the PHP Inspections extension for PhpStorm.

jrfnl:
This simple change means that namespace separators will be be ignored completely by the check for output escaping which fixes the immediate issue.

For a more thorough fix, the logic of the function would need to be refactored to take namespaced functions into account as well, but that's for another day.

Fixes #933

### DIFF
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -291,6 +291,11 @@ class EscapeOutputSniff extends Sniff {
 				continue;
 			}
 
+			// Ignore namespace separators.
+			if ( T_NS_SEPARATOR === $this->tokens[ $i ]['code'] ) {
+				continue;
+			}
+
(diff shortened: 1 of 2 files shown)


GaryJones:
Works perfectly for my case :-)

PR #1099, opened 2017-08-09
New `WhiteSpace.PrecisionAlignment` sniff for `Core`

### LINKED ISSUE #977: Faulty core fixes: the case of the moving space
#### Summary:
While https://github.com/squizlabs/PHP_CodeSniffer/pull/1404 definitely is a huge improvement in the spaces -> tabs fixing, I'm still seeing an issue with this sniff when there is a combination of tabs and spaces for indentation and the spaces are not there for alignment.

Most likely cause: `Generic.WhiteSpace.DisallowSpaceIndent`

#### Example:

`./src/wp-admin/update.php` line 127 (and in lots of other places too)

Source:
```php
	 		wp_die( $api ); // Tab - Space - Tab - Tab
```

Fixed as:
```php
			 wp_die( $api ); // Tab - Tab - Tab - Space
```

Expected fix:
```php
			wp_die( $api ); // Tab - Tab - Tab
```

### LINKED ISSUE #983: Faulty core fixes: mixed tab/space indentation
There are some instances where spaces are used instead of tabs or a mixture of spaces and tabs.
While the auto-fixer can correct most of this, if the spaces used is not a multitude of 4, it will - correctly - assume that the extra spaces are used for precision alignment.

I don't think we can fix this as there is no way for the sniffer to know whether this "precision alignment" was intentional or accidental.

What we _could_ do however, is to throw a warning when precision alignment is found to alert someone to manual inspect the code.

Opinions ?

#### Example:

`./src/wp-admin/post.php` line 51 (and various other files)

Source: (five spaces)
```php
if ( ! $sendback ||
     strpos( $sendback, 'post.php' ) !== false ||
     strpos( $sendback, 'post-new.php' ) !== false ) {
 	// something
}
```

Fixed as: (tab + space)
```php
if ( ! $sendback ||
	 strpos( $sendback, 'post.php' ) !== false ||
	 strpos( $sendback, 'post-new.php' ) !== false ) {
 	// something
}
```

jrfnl:
This sniff complements the upstream `Generic.WhiteSpace.DisallowSpaceIndent` sniff.

The `Generic.WhiteSpace.DisallowSpaceIndent` sniff makes sure that tabs are used for indentation instead of spaces.
However, the sniff allows for precision alignment of `# spaces < tab_width`.

In WP it is not customary to use precision alignment, though there are a few exceptions where it is used intentionally.

This means that precision alignment can not be auto-fixed.

We can, however, throw a warning when precision alignment is detected which is what this sniff implements.

Notes:
* This sniff will detect precision alignment in PHP, JS as well as CSS files.
* The sniff adds a new `precision alignment` whitelist comment.
* The sniff will handle the "space before star"precision alignment for comments without issue.
* The sniff adds a new custom property which allows users to pass an array of tokens to ignore for the purposes of this sniff. By default nothing is ignored.

Includes unit tests.

I've added the sniff to the `Core` ruleset as this is one of the issues which came out of the initial review of auto-fixed core files.

A run of the sniff over the `/src/` directory of core (using the current `phpcs.xml.5.dist` basis) results in 281 infractions in 68 files. See the attachment for details: 
[20170809-precision-alignment.txt](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/files/1210448/20170809-precision-alignment.txt)

A visual inspection shows no false positives and that nearly all of the infractions found should be corrected. There are only very few, where by the looks of it, the precision alignment was actually intended as such.

/cc @pento 

Fixes #983

-----

#### To do once the sniff is merged:
- [ ] Add the new whitelist flag to the wiki page.
- [ ] Add the new property to the wiki page.

### DIFF
--- a/WordPress/Tests/WhiteSpace/PrecisionAlignmentUnitTest.css
+++ b/WordPress/Tests/WhiteSpace/PrecisionAlignmentUnitTest.css
@@ -1,0 +1,5 @@
+#login-container {
+    margin-left: -225px;
+	width: 450px;
+	 height: 450px; /* Bad. */
+}
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -46,6 +46,7 @@
 
 	<!-- Covers rule: Use real tabs and not spaces. -->
 	<rule ref="Generic.WhiteSpace.DisallowSpaceIndent"/>
+	<rule ref="WordPress.WhiteSpace.PrecisionAlignment"/>
 
 	<!-- Generic array layout check. -->
 	<!-- Covers rule: For associative arrays, values should start on a new line.
--- a/WordPress/Sniffs/WhiteSpace/PrecisionAlignmentSniff.php
+++ b/WordPress/Sniffs/WhiteSpace/PrecisionAlignmentSniff.php
@@ -1,0 +1,181 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPress\Sniffs\WhiteSpace;
+
+use WordPress\Sniff;
+use WordPress\PHPCSHelper;
+
+/**
+ * Warn on line indentation ending with spaces for precision alignment.
+ *
+ * WP demands tabs for indentation. In rare cases, spaces for precision alignment can be
+ * intentional and acceptable, but more often than not, this is a typo.
+ *
+ * The `Generic.WhiteSpace.DisallowSpaceIndent` sniff already checks for space indentation
+ * and auto-fixes to tabs.
+ *
+ * This sniff only checks for precision alignments which can not be corrected by the
+ * `Generic.WhiteSpace.DisallowSpaceIndent` sniff.
+ *
+ * As this may be intentional, this sniff explicitly does *NOT* contain a fixer.
+ *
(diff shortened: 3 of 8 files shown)


grappler:
Out of interest would code like this be affected by this sniff?

```php
add_action( 'script_loader_src', array( $this, 'block_script_load'   ),  100 );
add_action( 'admin_init',        array( $this, 'remove_update_crons' )       );
```

WordPress/Tests/WhiteSpace/PrecisionAlignmentUnitTest.css
```diff
@@ -0,0 +1,5 @@
+#login-container {
+    margin-left: -225px;
```
GaryJones: [space][space][space][space] is allowable by this sniff?

JDGrimes:
@grappler No, I don't think so. It only checks alignment at the start of the line, i.e., after any indentation, but before any other code.

WordPress/Tests/WhiteSpace/PrecisionAlignmentUnitTest.css
jrfnl: Yes it is, as that will correctly be replaced by a tab by the `Generic.WhiteSpace.DisallowSpaceIndent`, so is not the concern of this sniff.

jrfnl:
> Out of interest would code like this be affected by this sniff?

@grappler What @JDGrimes said :wink:

jrfnl:
Just checking: this one is waiting on @pento's opinion before merge ?

pento:
👍🏻 on throwing the warning, I'll trust y'all on the quality of the PR. 🙂

PR #1105, opened 2017-08-16
ArrayDeclarationSpacing: add a property to make an exception for single item associative arrays

### LINKED ISSUE #1104: Forced new lines with associative arrays of one item
`WordPress.Arrays.ArrayDeclarationSpacing.AssociativeKeyFound` insists on multi–line array notation in _all_ cases, including array only having single item.

So to satisfy it `[ 'foo' => 'bar' ]` needs to be rewritten as
```php
[
    'foo' => 'bar',
]
```
which can be hugely space–wasting within function calls and such:
```php
function_call( 'foo', [
    'bar' => 'baz',
], 'stuff' );
```
While [letter of the rules](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation) makes no distinction for number of items, I feel applying this literally to single–item arrays is counter–productive.

Though maybe should be clarified in the rules upstream first? Not sure whom to poke about it.

jrfnl:
Allow for a property to be set in a custom ruleset to allow single item associative arrays to be single line.

If someone can come up with a better (less lengthy) property name: please let me know and I'd happily adjust the PR.

I've chosen to implement this as a `true`/`false` toggle.
The toggle defaults to `false` based on the rule as it is in the handbook.

Alternatively, it could be implemented numerically, i.e. number of items needed before the associative array rule takes effect with a default of `0`, but that felt like it would make the rule very arbitrary where one project would have it at `1`, another at `3` etc., which would counter the use of having the rule at all.

Fixes #1104

----

#### TO DO

- [ ] Add info on the new property to the customizable properties wiki page

### DIFF
--- a/WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
+++ b/WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
@@ -36,8 +36,20 @@ use PHP_CodeSniffer_Tokens as Tokens;
  *                 which were not covered elsewhere. The `ArrayDeclaration` sniff has
  *                 now been deprecated.
  * @since   0.13.0 Class name changed: this class is now namespaced.
+ * @since   0.14.0 Single item associative arrays can now be excepted from the
+ *                 "must be multi-line" rule using the
+ *                 `allow_single_item_single_line_associative_arrays` property.
  */
 class ArrayDeclarationSpacingSniff extends Sniff {
+
+	/**
+	 * Whether or not to allow single item associative arrays to be single line.
+	 *
+	 * @since 0.14.0
+	 *
+	 * @var bool Defaults to false.
+	 */
+	public $allow_single_item_single_line_associative_arrays = false;
 
 	/**
 	 * Token this sniff targets.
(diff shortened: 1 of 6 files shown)


Rarst:
Should it default to `false` ?

I think the reworded rule doesn’t apply to single–item case altogether, so should the old behavior remain as default?

jrfnl:
> I think the reworded rule doesn’t apply to single–item case altogether, so should the old behavior remain as default?

The reworded rule has been reworded again and still applies to single-line items as it stands now.
So yes, for now, the default should remain `false`.

Rarst:
Was it? This is what I am seeing in rules as of right now:

> For associative arrays, each item should start on a new line when the array contains more than one item:
```php
$query = new WP_Query( array( 'ID' => 123 ) );

$query = new WP_Query( array( 
[tab]'post_type'   => 'page',
[tab]'post_author' => 123,
[tab]'post_status' => 'publish',
) );
```

jrfnl:
You're right, I read that wrong. In that case, do we need a property at all ?

That is of course, aside from the fact that a proper discussion about what the rule should be based on merits not prior art, would still be preferable. https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1104#issuecomment-322838091

Rarst:
I think the only argument for keeping this behavior as an available option is if someone _does_ want to keep enforcing it.

jrfnl:
I've updated the PR, changing the default value of the new property to `true`. The unit tests and documentation has been adjusted accordingly.

The added commit also includes a minor change to improve the phrasing of the error message depending on the setting and a changed error code.

jrfnl:
FYI: Property info has been added to the wiki.

Issue #1128, opened 2017-09-08
PHPCBF only partially fixes spaces


mikaelz:
With `phpcbf --standard=phpcs.xml phpcstest.php` some spaces are missing

[pastebin of phpcs.xml](https://pastebin.com/x0c0syBQ). `installed_paths` is at master 1f64b1a0b5b789822d0303436ee4e30e0135e4dc

Before:
```
function processOnHoldOrder($order) {
    if ('gopaybinder' == $order->get_payment_method() && $days_old > 1) {
        $order->update_status('failed', 'Update by '.JOB_NAME);
        fwrite($log, "Changed status to failed: {$order->id}".PHP_EOL);
    }

    if ('bacs' == $order->get_payment_method() && $days_old >= 14) {
        $order->update_status('failed', 'Update by '.JOB_NAME);
        fwrite($log, "Changed status to failed: {$order->id}".PHP_EOL);
    } elseif ('bacs' == $order->get_payment_method() && $days_old >= 3 && $days_old < 14) {
        if (!empty($mails)) {
            foreach ($mails as $mail) {
                if ($mail->id == 'new_order') {
                    $mail->trigger($order);
                    $order->add_order_note('Email notification sent by '.JOB_NAME, false, true);
                    fwrite($log, "Sent email notification: {$order->id}".PHP_EOL);
                }
            }
        }
    }
}
```

After:
```
function processOnHoldOrder( $order) {  // only one space
	if ('gopaybinder' == $order->get_payment_method() && $days_old > 1) {
		$order->update_status( 'failed', 'Update by ' . JOB_NAME );
		fwrite( $log, "Changed status to failed: {$order->id}" . PHP_EOL );
	}

	if ('bacs' == $order->get_payment_method() && $days_old >= 14) { // no spaces
		$order->update_status( 'failed', 'Update by ' . JOB_NAME );
		fwrite( $log, "Changed status to failed: {$order->id}" . PHP_EOL );
	} elseif ('bacs' == $order->get_payment_method() && $days_old >= 3 && $days_old < 14) { // no spaces
		if ( ! empty( $mails )) { // missing space
			foreach ($mails as $mail) { // no spaces
				if ($mail->id == 'new_order') { // no spaces
					$mail->trigger( $order );
					$order->add_order_note( 'Email notification sent by ' . JOB_NAME, false, true );
					fwrite( $log, "Sent email notification: {$order->id}" . PHP_EOL );
				}
[...]

PR #1129, opened 2017-09-08
Update example ruleset based on feedback

jrfnl:
Issue #1128 raised the interesting point that example exclusions maybe taken over literally by people who are not as familiar with PHPCS.

This minor change should prevent that.


jrfnl:
P.S.: build failure is unrelated to this change, but to do with an upstream change in the unit test suite. I'm working on a separate PR for that.

jrfnl:
Rebased after merge of #1130 so we can get a passing build.

PR #1135, opened 2017-09-10
Add new Arrays.MultipleStatementAlignment sniff to `Core`

### LINKED ISSUE #1122: Faulty core fix: Array elements aren't aligned at the =>
Given this example:

```php
<?php

array(
	'aaa' => 'b',
	'c' => 'd',
	'eeeeeee' => 'f',
);
```

The array elements aren't being aligned at the `=>`. This behaviour isn't explicitly noted in the handbook, just [recommended](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation). I can make it more explicit in the Hanbook, thought.

jrfnl:
This new sniff enforces alignment of the double arrow assignment operator to the same column for each item in a multi-item, multi-line array.

For single line arrays it forces the space between the index key and the double arrow to be exactly one space.

The sniff has three public properties which can be set from a ruleset:
- `ignoreNewline`: Set to allow new line(s) between the item key and the double arrow. Defaults to `true`.
    When set to true, the item will be ignored completely when a new line is encountered between the item key and the double arrow.
- `exact`: Whether the alignment has to be exactly 1 space from the end of the widest index. Defaults to `true`.
    When set to `false`, the largest index key + 1 space is taken as a minimum and if a predominant number of items is already aligned at a position which is more than the minimum, this predominant alignment position is taken as the preferred alignment.
    This allows for cleaner diffs as the array does not have to be realigned when the largest key is taken out.
- `maxColumn`: The maximum column at which the double arrow can be placed. Defaults to `1000`.
    This property has been added to aid in managing line-length.

As double arrow alignment is (currently) a recommendation, not a requirement, all checks in the sniff throw `warning`s, even though they are auto-fixable.

Additionally, I've run this sniff against the WP core to make sure it would not cause new fixer conflicts (it doesn't).
I've done a perfunctory visual check on some of the output and so far have not come across any real issues with the new sniff.

The sniff has been added to the `Core` ruleset with a `maxColumn` value of `60`.

The `60` is arbitrary. I've chosen a number which would allow for a balance based on readability, keeping a preferred line-length of 100 and max line length of 120 in mind.
:question: Argued opinions for a different value will gladly be taken into account. /cc @pento 

:question: Also: should the default for `exact` be `true` or `false` ? and/or should the WP ruleset specify this property as `false` ?

Fixes #1122

In a separate commit the issues this sniff throws up for the current WPCS codebase are fixed.

----

#### Note for trac ticket 41057
If there is any - unintentional - precision alignment, this will affect the fixes this sniff makes to the double arrow alignment.

As precision alignment can not be auto-fixed (as it cannot be determined whether this was intentional or not), I would strongly suggest doing a manual verification of the precision alignment warnings and fixing these where relevant **_prior_** to running phpcbf over the core codebase to create the patch for ticket 41057.

I.e.: get the patch to fix unintentional precision alignment in *before* running phpcbf for the final patch.

----

Side-note: this sniff is a candidate to upstream to PHPCS itself at some point as part of the solution for https://github.com/squizlabs/PHP_CodeSniffer/issues/582
With that in mind, the defaults for the sniff properties have been made generic and the unit tests are in two flavours:
* Tab-based long array keyword
* Space based short array

-----
To do when merged:
- [ ] Add information about this sniff and its properties to the wiki


jrfnl:
N.B. regarding the unit tests not passing - looks like there is a difference in behaviour between PHP 5 and 7 which I may have overlooked (or which affects PHPCS upstream). I will be looking into this.

jrfnl:
The unit test failure had to do with a difference in the sorting mechanisms between PHP 5 and 7 when sorting an array with several equal values.

I've added a new commit which takes that situation better into account for the sniff and should solve the unit test failure.

JDGrimes:
> ❓ Also: should the default for exact be true or false ? and/or should the WP ruleset specify this property as false ?

I think the default should be `true`. The core ruleset can specify `false` if that is desired, but I'm indifferent there.

----

Overall, I favor alignment, but I do not think I would be able to use this sniff as it is. I don't find the alignment of arrays composed entirely of multi-line values (like nested arrays, in particular) to be helpful. At best, it does almost nothing to make the code more readable. Actually, I find that it makes it less readable, because my eye is drawn to the `=> array(` hanging there out in space instead of to the key or the nested array's contents. It also makes unnecessary work and frustration trying to align things that may not even be close enough to have any obvious visual relationship. I think that it is telling that we've instinctively *not* used alignment in many (all?) of these cases, judging from the diff. There are also places where we considered newlines to be a break in alignment (again to considerable visual benefit, IMO). 

In other words, I would want to see the sniff changed to:

- Have an options to ignore the alignment for arrays which contain only multi-line values (or better yet, force them not to be aligned).
- Possibly also have the option to respect empty lines as breaks in alignment. (Though really, that is probably not a common case.) As in the assignment alignment sniff, I guess this could potentially be a sufficient fix that the previous suggestion wouldn't be needed.

Of course, if the sniff behavior as it is is what is desired for core, then you don't need to feel compelled to change it just for me. I'd just not use it, or I could modify it to support the options I want. But perhaps it warrants further discussion, as to whether this is actually what core wants.

Just look at [this](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/99386ffda08b2de98863fc2b6151bd59a2ed3e7b/WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php#L49-L235) vs [this](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/VIP/RestrictedFunctionsSniff.php#L49-L235). I find the latter to be more consumable.

jrfnl:
> .... But perhaps it warrants further discussion, as to whether this is actually what core wants.

Yes, please! More discussion and example cases and suggestions on how they should be handled is very welcome :+1: 

> I think that it is telling that we've instinctively not used alignment in many (all?) of these cases, judging from the diff. 

Considering how many arrays are used in the sniffs, the diff (for WPCS) is actually not that big.
It's mainly the more recently introduced large multi-level arrays which cause the large diff and a few times unit tests having been added and the alignment not having been adjusted to keep the diff clean.

> In other words, I would want to see the sniff changed to:
>
> * Have an options to ignore the alignment for arrays which contain only multi-line values

This sounds doable, but what about mixed arrays in that case ? I.e. arrays containing both single line as well as multi-line values ? They should be aligned ?

An alternative could be to only align the lowest level nested array and ignore alignment for higher level arrays, though that would run into the same issue with mixed arrays.

I'm wondering though if we are not making things more complicated for people by creating exceptions (as is being discussed elsewhere for JS vs PHP coding standards).

> (or better yet, force them not to be aligned).

In that case, I would suggest to change it to "force them to have 1 space" as the operator alignment sniff allows more space before assignment operators (to allow for alignment sniffs like this one).

> * Possibly also have the option to respect empty lines as breaks in alignment. (Though really, that is probably not a common case.) As in the assignment alignment sniff, I guess this could potentially be a sufficient fix that the previous suggestion wouldn't be needed.

To be quite fair, that is nowhere near as easy for the array alignment sniff as for the assignment alignment sniff.
For the latter, the start/end lines of the "group" are fluid based on those kind of breaks. For arrays, the start/end is fixed, so this would cause a lot of redundancy with (re-)calculating what the alignment should be each time.

JDGrimes:
> what about mixed arrays in that case ?

I guess when most of the values are single line, alignment should still be required; when most of the values are multi-line, alignment would be optional. Or perhaps it would be simpler if alignment was always optional when any multi-line values were present. (Or at least, if the sniff had the option to make this optional.)

> I'm wondering though if we are not making things more complicated for people by creating exceptions

Potentially, although core could choose to be more strict if desired (not enabling the option to make alignment optional). In this case, I don't really think that the exception would tend to cause confusion, because in many cases where it would apply, it is unlikely that alignment (or lack thereof) would be very obvious at all—the things being aligned would be far enough apart that unless the editor automatically handled alignment, you probably wouldn't go to the trouble of doing it manually.

jrfnl:
@JDGrimes et all

I build in the option requested earlier this week, I just didn't have time to finish it off before now.
I'm currently working on setting up a ridiculous amount of unit tests to make sure this sniff works as expected :heart_eyes_cat: 

I've changed the sniff to have a public `alignMultilineItems` property with the following possible settings:
```php
/*
 * - 'always':   Default. Align all arrays items regardless of single/multi-line.
 * - 'never':    Never align array items which span multiple lines.
 *               This will enforce one space between the array index and the
 *               double arrow operator for multi-line array items, independently
 *               of the alignment of the rest of the array items.
 *               Multi-line items where the arrow is already aligned with the
 *               "expected" alignment, however, will be left alone.
 * - number:     Only align the operator for multi-line arrays items if #number %
 *               (or more) of the items is single-line.
 *               The % is calculated against all array items (with and without
 *               assignment operator).
 *               The expected alignment will be calculated based only on the
 *               items being aligned.
 *               Examples:
 *               * Setting this to `0` means that alignment will be enforced
 *                 unless *all* array items are multi-line.
 *                 This is probably the most commonly desired situation.
 *               * Setting this to `50` means that the majority of array items
 *                 need to be single line before alignment is enforced for
 *                 multi-line items in the array.
 *               * Setting this to `100` is useless as in that case there are
 *                 no multi-line items in the array anyway.
 *               Multi-line items where the arrow is already aligned with the
 *               "expected" alignment, however, will be left alone.
 */
```

Take note of the following:
> Multi-line items where the arrow is already aligned with the "expected" alignment, however, will be left alone.

I added that as it feels counter-intuitive to change ("correct") the alignment to 1 space if the alignment was already based on the widest index key.

How does this sound ? :+1: / :-1:

jrfnl:
> Only align the operator for multi-line arrays items if #number % (or more) of the items is single-line.

I'm a bit in doubt about whether this will be intuitive enough to use as the expected number refers to the % of single line items.

Opinions ?

JDGrimes:
> I'm a bit in doubt about whether this will be intuitive enough to use as the expected number refers to the % of single line items.

Yeah, it does seem unintuitive that 0 = always and 100 = ~~never~~ *meaningless*.

Could it just be changed so that it is the percentage of multi-line items? In that case, it would be "Only align the operator for multi-line arrays items if #number % (or *less*) of the items is *multi*-line." So 0 would = never and 100 would = always. 

The "or less" doesn't seem particularly intuitive either though. Maybe better stated as "the minimum percentage of multi-line items before alignment is optional." Or just, "the percentage above which multi-line item alignment is optional." Or, "the percentage below which multi-line item alignment is required."

If it is just too unintutive, we don't have to have that option, I guess—"never" will still allow for alignment to be used as desired for multi-line items, just never *requiring* it.

jrfnl:
> Could it just be changed so that it is the percentage of multi-line items? In that case, it would be "Only align the operator for multi-line arrays items if #number % (or less) of the items is multi-line." So 0 would = never and 100 would = always.

@JDGrimes If I did that, then your original case could not be addressed: 
> I don't find the alignment of arrays composed entirely of multi-line values (like nested arrays, in particular) to be helpful.

This would translate to "_never, but only if 100% of items is multi-line_".

I'm honestly struggling with what the best way to set up the option is. The implementation is not so much the issue, but for this option to be usable, people should be able to understand it.

jrfnl:
Ok, thinking it over yet again - what about this:

Removing the numeric option and instead have only string based options , but more of them - not as configurable, but hopefully covering the most desired options:
* `always` = always align multi-line items
* `never` = never align multi-line items
* :heavy_plus_sign: `condition:minority-align` = only align multi-line items if less than 50% of all items is multi-line
* :heavy_plus_sign: `condition:minority-noalign` = don't align multi-line items if less than 50% of all items is multi-line, align in all other situations
* :heavy_plus_sign:  `condition:majority-align` = only align multi-line items if 50% or more of all items is multi-line
* :heavy_plus_sign: `condition:majority-noalign` = don't align multi-line items if 50% or more of all items is multi-line, align in all other situations
* :heavy_plus_sign: `condition:all-align` = only align multi-line items if all items in the array are multi-line
* :heavy_plus_sign: `condition:all-noalign` = don't align multi-line items if all items in the array are multi-line, align in all other situations

jrfnl:
@WordPress-Coding-Standards/wpcs-collaborators Anyone got any feedback on my latest proposal above ?

JDGrimes:
> Removing the numeric option and instead have only string based options , but more of them - not as configurable, but hopefully covering the most desired options:

This doesn't seem ideal either, and my impression is that it is more complex than the other option.

What about a number and a comparison operator: `<=90`, `>=100`, etc. Default operator would be `<=`, I guess. But using `>=100` would allow my original case, I think. Or we could allow `==` operator. Operators would maybe be: `>`, `<`, `>=`, `<=`, `==`. 

Does that sound reasonable?

jrfnl:
> What about a number and a comparison operator: <=90, >=100, etc. Default operator would be <=, I guess. But using >=100 would allow my original case, I think. Or we could allow == operator. Operators would maybe be: >, <, >=, <=, ==.

I like that idea. Thanks!

As for the implementation, I think this will be pretty complex to implement without using `eval()`, but I'll figure out a way.

JDGrimes:
> As for the implementation, I think this will be pretty complex to implement without using eval(), but I'll figure out a way.

Wouldn't it be pretty simple to parse with regular expressions? (But you've probably already come up with something better than that... 😄 )

jrfnl:
> Wouldn't it be pretty simple to parse with regular expressions?

The parsing is not the problem, the evaluating however... I guess we'll just end up with a massive switch ;-)

jrfnl:
And... as soon as I hit "Comment", I thought off a better way. Leave it with me ;-)

jrfnl:
Ok, done.

I've rebased the PR, removed the "fix own codebase" commit, added a new commit which adds a `alignMultilineItems` option - see commit message and inline documentation for the details -, added lots of unit tests, and re-ran the fixer over the WPCS codebase excepting 100% multi-line arrays from being aligned and reate a new "fix own codebase" commit based on that.

As far as I'm concerned, this should now be good to go, except for determining what the WP defaults should be for the properties.

@JDGrimes What do you think ?

pento:
Running this on Core looks good, didn't notice any problems a brief visual inspection.

👍🏻 on the `maxColumn` value of `60`.

jrfnl:
Rebased to make sure no new violations against this sniff were introduced in our own codebase in the mean time. As @pento seems to be ok with the addition and the config of it, can this be merged ?

jrfnl:
FYI: The wiki custom properties page has been updated to include information on the public properties provided by this sniff.

PR #1136, opened 2017-09-10
Add sniff to check for alignment of assignment operators to `Core`

jrfnl:
This sniff has already been added and removed a couple of times previously, see the discussion in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/382#r29970107 amongst others.

Adding the `maxPadding` property will hopefully make the sniff more acceptable.

Sister-PR to the array arrow alignment sniff PR as requested in #1122.

The value for `maxPadding` which I've chosen is arbitrary - similar to my notes about `maxColumn` in #1135 - and could probably use some scrutiny before deciding on the real number to use. /cc @pento 

If it would change, both commits in this PR will change.

--- 
[Edit]: Oh and I ran this sniff (combined with #1135) over WP Core as well to check for new fixer conflicts and none were found, so we're good in that respect.

### DIFF
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -403,7 +403,7 @@ class EscapeOutputSniff extends Sniff {
 						// If we're able to resolve the function name, do so.
 						if ( $mapped_function && T_CONSTANT_ENCAPSED_STRING === $this->tokens[ $mapped_function ]['code'] ) {
 							$functionName = $this->strip_quotes( $this->tokens[ $mapped_function ]['content'] );
-							$ptr = $mapped_function;
+							$ptr          = $mapped_function;
 						}
 					}
 
@@ -482,6 +482,7 @@ class EscapeOutputSniff extends Sniff {
 				$customEscapeFunctions,
 				$this->escapingFunctions
 			);
+
 			$this->addedCustomFunctions['escape']   = $this->customEscapingFunctions;
 			$this->addedCustomFunctions['sanitize'] = $this->customSanitizingFunctions;
 		}
@@ -491,6 +492,7 @@ class EscapeOutputSniff extends Sniff {
 				$this->customAutoEscapedFunctions,
 				$this->autoEscapedFunctions
 			);
+
 			$this->addedCustomFunctions['autoescape'] = $this->customAutoEscapedFunctions;
 		}
 
@@ -500,6 +502,7 @@ class EscapeOutputSniff extends Sniff {
 				$this->customPrintingFunctions,
 				$this->printingFunctions
 			);
+
 			$this->addedCustomFunctions['print'] = $this->customPrintingFunctions;
 		}
 	}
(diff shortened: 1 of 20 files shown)


WordPress/Sniffs/XSS/EscapeOutputSniff.php
```diff
@@ -478,7 +478,7 @@ protected function mergeFunctionLists() {
 				);
 			}
 
-			$this->escapingFunctions = $this->merge_custom_array(
+			$this->escapingFunctions                = $this->merge_custom_array(
```
JDGrimes: As noted in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1122#issuecomment-328516737, I'd prefer to see a blank line added after this assignment rather than realigning it with the assignments below it.

jrfnl: I've done this now for the large alignments (in the mergeFunctionLists/Whitelist functions) and in one or two other places where it made sense.

Added the change as a separate commit. If ok, I'll squash it into the previous one.
If you'd like more changes, let me know, though from reviewing the others, these groupings make sense (to me).

JDGrimes: Thanks, looks good to me.

jrfnl: Ok, squashed the wo code base fix related commits into one. Ready to merge again.

jrfnl:
Anyone still have an opinion on this ?

> The value for maxPadding which I've chosen is arbitrary - similar to my notes about maxColumn in #1135 - and could probably use some scrutiny before decising on the real number to use.

GaryJones:
> Anyone still have an opinion on this ?

Fine with me.

jrfnl:
Rebased for merge conflicts.

@pento Had a chance to look at this yet ?

jrfnl:
Rebased for merge conflicts. @pento Could you please have a look ?

pento:
Looks good! No problem with the default value for `maxPadding`.

pento:
Oh, here's a thing I've come across. This pattern is fairly common:

```php
<?php

$foo['some_key'] = __( 'Cat' );
/* translators: foo */
$foo['some_other_key'] = __( 'Foo' );
```

The assignments are only separated for the translator note, otherwise they'd be on adjacent lines. I would expect them to be aligned, the current PR does not.

jrfnl:
> The assignments are only separated for the translator note, otherwise they'd be on adjacent lines. I would expect them to be aligned, the current PR does not.

The sniff in this PR sees a blank line and comments as "delimiters", i.e. they end the current group for the purpose of alignment and start a new group when followed by an assignment afterwards.

The sniff does not allow for a way around that and there are no alternative sniffs for this available. The only alternative would be creating a WP native sniff instead. :confused:

pento:
Okay, don't worry about it, then. We can fix them manually.

jrfnl:
> We can fix them manually.

Fixing them manually will not stop the sniff complaining about the alignment being off though, so I'm not sure how that will help in the long run if/when PHPCS will be enabled on each commit to core.

pento:
Oh, right. Well, I guess we can live with it looking a little weird.

jrfnl:
Rebased to make sure no new violations against this sniff were introduced in our own codebase in the mean time. As @pento seems to be ok with the addition and the config of it, can this be merged ?

jrfnl:
Rebase to fix the merge conflict.

PR #1141, opened 2017-09-13
[WordPress.WP.I18n] Allow omitting $text_domain arg when 'default'

westonruter:
This will facilitate the WordPress core to ship with `WordPress.WP.I18n` enabled with a ruleset that has:

```xml
		<properties>
			<property name="text_domain" value="default" />
		</properties>
	</rule>
```

There seems to be an issue with unit tests whereby the `codingStandardsChangeSetting` directives aren't getting recognized. Help appreciated.

### DIFF
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -345,7 +345,7 @@ class I18nSniff extends Sniff {
 
 		if ( empty( $tokens ) || 0 === count( $tokens ) ) {
 			$code = $this->string_to_errorcode( 'MissingArg' . ucfirst( $arg_name ) );
-			if ( 'domain' !== $arg_name || ! empty( $this->text_domain ) ) {
+			if ( 'domain' !== $arg_name || ( ! empty( $this->text_domain ) && ! in_array( 'default', $this->text_domain, true ) ) ) {
 				$this->addMessage( 'Missing $%s arg.', $stack_ptr, $is_error, $code, array( $arg_name ) );
 			}
 			return false;
--- a/WordPress/Tests/WP/I18nUnitTest.1.inc
+++ b/WordPress/Tests/WP/I18nUnitTest.1.inc
@@ -5,7 +5,7 @@
 __( 'string' ); // OK - no text domain known, so not checked.
 __( 'string', 'something' ); // OK - no text domain known, so not checked.
 
-// @codingStandardsChangeSetting WordPress.WP.I18n text_domain my-slug,default
+// @codingStandardsChangeSetting WordPress.WP.I18n text_domain my-slug
 
 __( 'string' ); // Bad - no text domain passed.
 __( 'string', 'something' ); // Bad - text domain mismatch.
@@ -112,7 +112,7 @@ __( 'String with a literal %%', 'my-slug'); // Ok, replacement would be a single
 __( 'String with a literal %% and a %s placeholder', 'my-slug'); // Ok.
 __( 'A replacement variable can not start with "%%cf_" or "%%ct_" as these are reserved for the WPSEO standard variable variables for custom fields and custom taxonomies. Try making your variable name unique.', 'my-slug' ); // Ok.
 
-// The domain 'default' was also added to the text domains.
+// @codingStandardsChangeSetting WordPress.WP.I18n text_domain my-slug,default
 __( 'String default text domain.', 'default' ); // Ok.
 __( "String default text domain.", "default" ); // Ok.
 
@@ -150,6 +150,11 @@ Not
 really.
 EOD
 , 'my-slug' ); // OK.
+
+// @codingStandardsChangeSetting WordPress.WP.I18n text_domain default
+__( 'String default text domain.', 'my-slug' ); // Bad because text_domain is only 'default'.
+__( 'String default text domain.', 'default' ); // Ok because domain is being explicit.
+__( 'String default text domain.' ); // Ok because default domain is 'default' and it matches one of the supplied configured text domains.
 
 // @codingStandardsChangeSetting WordPress.WP.I18n text_domain null
 // @codingStandardsChangeSetting WordPress.WP.I18n check_translator_comments true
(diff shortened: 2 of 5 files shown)


jrfnl:
@westonruter The reason the unit tests are not working is that passing a `text_domain` from the command line always trumps the property in the ruleset or inline.
This comes from the original set up of the unit tests: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/560/files#diff-3891ffe5e426f12e595144ee2213fbf6R13

I never changed that as I presumed there must have been a reason for using that method to set the property...

If we do change that to use the "normal" unit test `@codingStandardsChangeSetting`, we should add a third unit test file and leave the `setUp()` and `tearDown()`, but change them to only run for the third unit test file. That way we can still test that passing a `text_domain` from the command line works as well.

Let me know if you need pointers for making that change.

---- 
Other than that, I haven't looked at the code in detail yet (but will try to do so later tonight).
One question came to mind though: if the `text_domain` default is passed, should the sniff maybe "force" the `text_domain` in I18n function call to **not** be set ? That way at least the code would be consistent instead of having most calls without text domain and some with the `default` textdomain.

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -345,7 +345,7 @@ protected function check_argument_tokens( $context ) {
 
 		if ( empty( $tokens ) || 0 === count( $tokens ) ) {
 			$code = $this->string_to_errorcode( 'MissingArg' . ucfirst( $arg_name ) );
-			if ( 'domain' !== $arg_name || ! empty( $this->text_domain ) ) {
+			if ( ( 'domain' !== $arg_name || ! empty( $this->text_domain ) ) && ! in_array( 'default', $this->text_domain, true ) ) {
```
jrfnl: Why use `in_array()` instead of a straight up string compare ?

westonruter: Because the `text_domain` is an array, like `array( 'customize-posts', 'default' )`.

westonruter:
@jrfnl thanks for the in-depth response.

> Let me know if you need pointers for making that change.

It looks like the change will require more time then I have to spare. So I'll have to leave it for you or another contributor to implement that.

> One question came to mind though: if the text_domain default is passed, should the sniff maybe "force" the text_domain in I18n function call to not be set ? That way at least the code would be consistent instead of having most calls without text domain and some with the default text domain.

Good question. If the `text_domain` in the ruleset is `default` and the `$text_domain` has explicitly been set to `default` as well, then this would seem like a good redundancy _warning_ to trigger, though it shouldn't be an outright error.

WordPress/Sniffs/WP/I18nSniff.php
jrfnl: And so it is... :woman_facepalming: :tired_face:

jrfnl:
> It looks like the change will require more time then I have to spare. So I'll have to leave it for you or another contributor to implement that.

See #1142 ;-)

Note: this PR will need to be rebased/partially redone after the above PR has been merged. Then again, once that's done, it should be easy to get the unit tests for this PR to pass.

>  If the text_domain in the ruleset is default and the $text_domain has explicitly been set to default as well, then this would seem like a good redundancy warning to trigger, though it shouldn't be an outright error.

I would like to suggest making it a `fixableWarning` in that case.

westonruter:
Rebased… but I didn't get the expected results.

WordPress/Tests/WP/I18nUnitTest.1.inc
```diff
@@ -151,5 +151,9 @@ really.
 EOD
 , 'my-slug' ); // OK.
 
-// @codingStandardsChangeSetting WordPress.WP.I18n text_domain null
```
jrfnl: This line should not be removed. It **_has_** to be at the end of the file to reset the properties for the other test files, so it should come back below the newly added test cases.

WordPress/Tests/WP/I18nUnitTest.1.inc.fixed
```diff
@@ -112,7 +112,7 @@ __( 'String with a literal %%', 'my-slug'); // Ok, replacement would be a single
 __( 'String with a literal %% and a %s placeholder', 'my-slug'); // Ok.
 __( 'A replacement variable can not start with "%%cf_" or "%%ct_" as these are reserved for the WPSEO standard variable variables for custom fields and custom taxonomies. Try making your variable name unique.', 'my-slug' ); // Ok.
 
-// The domain 'default' was also added to the text domains.
+// @codingStandardsChangeSetting WordPress.WP.I18n text_domain my-slug,default
```
jrfnl: See above

WordPress/Tests/WP/I18nUnitTest.1.inc.fixed
```diff
@@ -151,5 +151,9 @@ really.
 EOD
 , 'my-slug' ); // OK.
 
-// @codingStandardsChangeSetting WordPress.WP.I18n text_domain null
```
jrfnl: See above

WordPress/Sniffs/WP/I18nSniff.php
jrfnl: The braces are incorrectly placed on this line which is one of the reasons why the unit tests are failing.
```php
if ( 'domain' !== $arg_name || ( ! empty( $this->text_domain ) && ! in_array( 'default', $this->text_domain, true ) ) ) {}
```

westonruter:
Looks like that fixed it. Thanks! I amended my commit.

jrfnl:
Only question remaining now is:

> > If the text_domain in the ruleset is default and the $text_domain has explicitly been set to default as well, then this would seem like a good redundancy warning to trigger, though it shouldn't be an outright error.
> 
> I would like to suggest making it a fixableWarning in that case.

jrfnl:
Oh and I noticed I'd made a minor mistake in my previous PR - the reset line at the end of the first two test files should be:
```php
// @codingStandardsChangeSetting WordPress.WP.I18n text_domain null
```
Want me to send in a separate PR for that or will you fix that in this one ?

westonruter:
> I would like to suggest making it a fixableWarning in that case.

I think this can be a separate PR.

jrfnl:
@westonruter Thanks for fixing that. Only still needs to be fixed in the `1.inc.fixed` file for the build to pass.

westonruter:
@jrfnl thank you for supporting me in the writing of this.

GaryJones:
There's currently no such plugin at https://wordpress.org/plugins/default/, but if there were, they may really struggle with i18n!

PR #1145, opened 2017-09-15
[Extra] Add PEAR.Files.IncludingFile sniff

### LINKED ISSUE #1007: Check for parentheses around require(_once)/include_(once)
There doesn't appear to be any checks if files are included as:

```php
include '/path/to/file.php';
```
or
```php
include ( '/path/to/file.php' );
```

(Repeat for `include_once`, `require`, `require_once`.)

WordPress Core apparently does a mixture of both.

Generally, there is no different between them, except when a comparison is done against it - see Example 4 on http://us1.php.net/manual/en/function.include.php. Otherwise, the `(` `)` are redundant and can be safely removed.

There is a [PEAR sniff](https://github.com/squizlabs/PHP_CodeSniffer/blob/b95ff2c3b122a3ee4b57d149a57d2afce65522c3/src/Standards/PEAR/Sniffs/Files/IncludingFileSniff.php#L54) that could be used / adapted from.

This was [mentioned](https://wordpress.slack.com/archives/C5VCTJGH3/p1498611873198551) in the #core-coding-standards channel, so while any addition would go into `WordPress-Extras` right now, it may be that the Handbook has been updated, so this could go into `WordPress-Core` standard instead.

### LINKED ISSUE #1143: Add sniff to check best practices around include/require statements to Extra ?
There is a sniff available upstream to check that non-conditional file includes use `require(_once)` and that conditional file includes use `include(_once)`. It also checks that no parenthesis are used for these statements (not needed as these are language constructs, not function calls).

I realize this may seem a little opinionated, but this is a valid best practice.

If a `require` fails, it will produce a fatal `E_COMPILE_ERROR` which for an unconditional file include is warranted.
If a `include` fails, it will produce a warning for the file include and possibly a plenitude of other warnings/errors if the include was not used conditionally.

I propose adding this sniff to the `Extra` ruleset.

@WordPress-Coding-Standards/wpcs-collaborators  Opinions ?

Refs:
* http://php.net/manual/en/function.require.php
* http://php.net/manual/en/function.include.php

jrfnl:
As discussed in #1143 

Closes #1143 
Closes #1007

### DIFF
--- a/Test/phpcs2-bootstrap.php
+++ b/Test/phpcs2-bootstrap.php
@@ -44,9 +44,9 @@ https://is.gd/contributing2WPCS
 }
 
 // Load the PHPCS test classes and the WPCS versions where necessary.
-include_once PHPCS_DIR . $ds . 'tests' . $ds . 'AllTests.php';
-include_once __DIR__ . $ds . 'Standards' . $ds . 'AllSniffs.php';
-include_once __DIR__ . $ds . 'Standards' . $ds . 'AbstractSniffUnitTest.php';
+require_once PHPCS_DIR . $ds . 'tests' . $ds . 'AllTests.php';
+require_once __DIR__ . $ds . 'Standards' . $ds . 'AllSniffs.php';
+require_once __DIR__ . $ds . 'Standards' . $ds . 'AbstractSniffUnitTest.php';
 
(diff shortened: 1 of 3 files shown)


GaryJones:
Had to restart a stalled [job](https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/275911964), but all good now.

PR #1152, opened 2017-09-19
Add customisable property to set the posts_per_page variable

dingo-d:
WordPress VIP rule restricts the number of `posts_per_page` variable to 100, with this change you will be able to set the limit of the `posts_per_page` variable in your `ruleset.xml` like:

~~~xml
<rule ref="WordPress.VIP.PostsPerPage">
	<properties>
		<property name="posts_per_page" value="200"/>
	</properties>
</rule>
~~~

### DIFF
--- a/WordPress/Sniffs/VIP/PostsPerPageSniff.php
+++ b/WordPress/Sniffs/VIP/PostsPerPageSniff.php
@@ -29,8 +29,9 @@ class PostsPerPageSniff extends AbstractArrayAssignmentRestrictionsSniff {
 	 *
 	 * Posts per page limit to check against.
 	 *
-	 * @var int
 	 * @since 0.14.0
+	 *
+	 * @var int
 	 */
 	public $posts_per_page = 100;
 
--- a/WordPress/Sniffs/VIP/PostsPerPageSniff.php
+++ b/WordPress/Sniffs/VIP/PostsPerPageSniff.php
@@ -20,6 +20,7 @@ use WordPress\AbstractArrayAssignmentRestrictionsSniff;
  *
  * @since   0.3.0
  * @since   0.13.0 Class name changed: this class is now namespaced.
+ * @since   0.14.0 Added the posts_per_page property.
  */
 class PostsPerPageSniff extends AbstractArrayAssignmentRestrictionsSniff {
 
@@ -29,6 +30,7 @@ class PostsPerPageSniff extends AbstractArrayAssignmentRestrictionsSniff {
 	 * Posts per page limit to check against.
 	 *
 	 * @var int
+	 * @since 0.14.0
 	 */
 	public $posts_per_page = 100;
 
@@ -62,7 +64,7 @@ class PostsPerPageSniff extends AbstractArrayAssignmentRestrictionsSniff {
 	 *                       with custom error message passed to ->process().
 	 */
 	public function callback( $key, $val, $line, $group ) {
-		$key = strtolower( $key );
+		$key                  = strtolower( $key );
 		$this->posts_per_page = (int) $this->posts_per_page;
 
 		if (
--- a/WordPress/Sniffs/VIP/PostsPerPageSniff.php
+++ b/WordPress/Sniffs/VIP/PostsPerPageSniff.php
@@ -24,6 +24,15 @@ use WordPress\AbstractArrayAssignmentRestrictionsSniff;
 class PostsPerPageSniff extends AbstractArrayAssignmentRestrictionsSniff {
 
 	/**
+	 * Posts per page property
+	 *
+	 * Posts per page limit to check against.
+	 *
+	 * @var int
+	 */
+	public $posts_per_page = 100;
+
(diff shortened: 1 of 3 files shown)


WordPress/Tests/VIP/PostsPerPageUnitTest.inc
```diff
@@ -1,8 +1,11 @@
 <?php
-
+// @codingStandardsChangeSetting WordPress.VIP.PostsPerPage posts_per_page 200
```
jrfnl: For this sniff to be tested properly:
1. The sniff needs to be tested with the default values, i.e. the way the unit test case file was originally.
2. New unit tests have to be added for testing that changing the property works.

To do this:
* revert the unit test case file to the way it was before.
* add the above line at the bottom
* add one or more test cases to test that changing the property has the desired effect.
* add a second line like the above below the property change test case to reset the property value to the default value.
* adjust the line numbers in the unit test file

This will look something like (to be added at the bottom of the original file):
```php
// @codingStandardsChangeSetting WordPress.VIP.PostsPerPage posts_per_page 50
 $query_args['posts_per_page'] = 50; // OK.
 $query_args['posts_per_page'] = 100; // Bad.
 $query_args['posts_per_page'] = 200; // Bad.
 $query_args['posts_per_page'] = 300; // Bad.
// @codingStandardsChangeSetting WordPress.VIP.PostsPerPage posts_per_page 200
 $query_args['posts_per_page'] = 50; // OK.
 $query_args['posts_per_page'] = 100; // OK.
 $query_args['posts_per_page'] = 200; // OK.
 $query_args['posts_per_page'] = 300; // Bad.
// @codingStandardsChangeSetting WordPress.VIP.PostsPerPage posts_per_page 100
```

WordPress/Sniffs/VIP/PostsPerPageSniff.php
```diff
@@ -24,6 +24,15 @@
 class PostsPerPageSniff extends AbstractArrayAssignmentRestrictionsSniff {
 
 	/**
+	 * Posts per page property
+	 *
+	 * Posts per page limit to check against.
+	 *
```
jrfnl: Missing `@since` tag.
It would also be a good idea to add a note about this change to the sniff class changelog at the top of the file.

WordPress/Sniffs/VIP/PostsPerPageSniff.php
```diff
@@ -54,17 +63,19 @@ public function getGroups() {
 	 */
 	public function callback( $key, $val, $line, $group ) {
 		$key = strtolower( $key );
+		$this->posts_per_page = (int) $this->posts_per_page;
```
jrfnl: Either align the `=` signs or add a blank line before this line.

WordPress/Sniffs/VIP/PostsPerPageSniff.php
dingo-d: The correct since tag is `0.3.0` or `0.13.0`? There are two since tags above the class definition, not sure which one to use.

WordPress/Sniffs/VIP/PostsPerPageSniff.php
dingo-d: Will align the equal signs, other sniffs are aligning that way so it's better for the consistency.

WordPress/Sniffs/VIP/PostsPerPageSniff.php
jrfnl: The correct since tag would be `0.14.0` as that's the release in which this property is added ;-)

WordPress/Tests/VIP/PostsPerPageUnitTest.inc
dingo-d: I corrected this, should be ok now. Changed both the .inc and .php unit test files.

WordPress/Sniffs/VIP/PostsPerPageSniff.php
dingo-d: Sure thing :D Fixed

WordPress/Sniffs/VIP/PostsPerPageSniff.php
```diff
@@ -20,10 +20,21 @@
...
  */
 class PostsPerPageSniff extends AbstractArrayAssignmentRestrictionsSniff {
 
 	/**
+	 * Posts per page property
+	 *
+	 * Posts per page limit to check against.
+	 *
+	 * @var int
+	 * @since 0.14.0
```
jrfnl: Minor nitpick: the `@var` tag should be the last in the doc block and should have one blank line above it, so:
```php
 * @since 0.14.0
 * 
 * @var int
 */
```

dingo-d: No problem will fix it now :)

JDGrimes:
Thanks for the PR @dingo-d !

Issue #1157, opened 2017-09-22
Proposal for restructuring the rulesets & some of the sniffs


jrfnl:
As a follow up on issues #676, #740, #1067, #1156, I would like to propose a restructuring.

I'm opening this as a separate issue as I would like to get as many eyes as possible on this proposal and would like a (preferably unanimous) :+1: / :-1: / :eyes: (= seen it, no opinion either way) vote from all the @WordPress-Coding-Standards/wpcs-admins on whether to go ahead with this. /cc @grappler @david-binda 

Please vote separately for proposal 1 and proposal 2.

# TL;DR
**Proposal to split the rulesets up into additional separate rulesets for codestyle and best practices and move some of the sniffs around to make adoption of the best practice sniffs from WPCS easier for the wider WP community.**

---
## Justification
While WPCS is based on the Core handbook and originally was probably primarily intended for usage by Core, the WordPress infrastructure is a lot wider than just Core and the way the current rulesets are set up are hindering adoption of WPCS in the wider WordPress community.

## Example usecases
* Plugin/Theme developer preferring to use PSR2 for codestyle, but still wanting the benefits of the WPCS best practice sniffs - see #1067
* Plugin/Theme developer choosing not to support PHP 5.2, but setting the minimum at for instance PHP 5.6 and wanting to check the code style of modern PHP constructs in a way that closely matches the WP code style.
* VIP/Plugin/Theme reviewers wanting to just and only check whether plugins/themes comply with best practices without being inundated by messages about code style issues - also see [Slidedeck: Using WPCS to review plugins and themes](https://speakerdeck.com/jrf/for-non-developers)

## Proposal 1: changing the rulesets
Adding 7 new (subset) rulesets:

New Ruleset | Notes
----- | ----
`WordPress-Core-Style` | See #740 for an initial proposal (albeit slightly outdated)
`WordPress-Extra-Style` | Would contain sniffs like `Squiz.Scope.MethodScope`, `Generic.PHP.Syntax` and `WordPress.Classes.ClassInstantiation`
`WordPress-Modern-Style` | Would contain sniffs like `PSR2.Namespaces.NamespaceDeclaration`, but also the ReturnTypeSniff from PR #1084 and the `use` declaration sniffs I've been working on to fix #1071
`WordPress-Core-Best-Practices` | See #740 for an initial proposal
`WordPress-Extra-Best-Practices` | Would contain nearly everything else now in `Extra`
[...]

PR #1159, opened 2017-09-22
CronInterval: make the minimum cron interval configurable

jrfnl:
This change allows for passing a minimum cron interval through a property in a custom ruleset.

Includes unit tests.

Includes one additional minor documentation fix for a previously added property.

As per a comment I made in proposal 2 in #1157

- [ ] Add info about the new property to the wiki


jrfnl:
@westonruter Good point, but that would only be the case if the cron event was scheduled with the same arguments.
However, this sniff does not check the *scheduling* of cron events, just the addition of cron *schedules*, so that is not something we could check conclusively in this sniff and adding a warning without that additional check will quickly cause irritation, so in that case we would also need to add a new whitelist flag ... etc

jrfnl:
@westonruter Any feedback in response to my previous comment ?

jrfnl:
FYI: Wiki updated with info about the new custom property.

PR #1160, opened 2017-09-23
Extra: Add new `PreparedSQLPlaceholders` sniff

### LINKED ISSUE #19: Add sniff to require a $wpdb->prepare() to have a 2nd argument

### LINKED ISSUE #641: Validate the placeholders in SQL queries.
> %s is used for string placeholders and %d is used for integer placeholders. Note that they are not ‘quoted’! 

Ref: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#formatting-sql-statements

We should make sure that no other placeholders are passed and that placeholders are not numbered.

---

Additional related issues from issue https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/19:

> Could we count the %s, %d etc. and make sure there's enough arguments to cover the placeholders?

Ref: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/19#issuecomment-81494168

> Related: [WP#22262](https://core.trac.wordpress.org/ticket/22262).

Ref: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/19#issuecomment-81266615

---

Could possibly be added to the `WordPress.WP.PreparedSQL` sniff.

jrfnl:
WordPress 4.8.2 introduced a security fix which removed an undocumented, but well used feature in the `$wpdb->prepare()` method.

Officially only the `%s`, `%d` and `%f` placeholders were supported.
In reality, most forms of `sprintf()` type placeholders formats were supported before WP 4.8.2.
See: https://core.trac.wordpress.org/changeset/41496

This already has caused havoc, breaking big plugins like Yoast SEO and WordFence.

This new sniff will check that `$wpdb->prepare()` is used correctly by checking for the following issues:
- Are only the following placeholders used ? %d, %f (%F) and %s.
- Are literal % signs properly escaped ?
- Are placeholders left unquoted in the query string ? (as they should be)
- Is the correct number of replacement variables passed ?

Includes unit tests.

The sniff has been added to the `Extra` ruleset, though it could be argued that it should be added to `Core` as most of the above points are already [in the handbook](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#formatting-sql-statements).

Refs:
* https://wordpress.org/news/2017/09/wordpress-4-8-2-security-and-maintenance-release/
* https://developer.wordpress.org/reference/classes/wpdb/prepare/
* https://github.com/Yoast/wordpress-seo/issues/7886
* https://wordpress.org/plugins/wordfence/#developers (see changelog for 6.3.19)
* https://core.trac.wordpress.org/ticket/41925

Fixes #641
Fixes #19

----
**Additional notes:**
The `is_wpdb_method_call()` method has been moved from the `PreparedSQL` sniff to the WP base `Sniff` to enable this sniff to use it as well.

Additionally, the re-usability and stability of the `is_wpdb_method_call()` method has been improved:
* Add recognition of static function calls using `::` operator.
* Pass the methods to check for via a property to allow for child classes to be more flexible.
* Only set the properties if such a property exists in the child class.
* Don't sniff the method during live coding, i.e. when no close parenthesis is found yet.

Includes unit tests.


jrfnl:
> Related: #641 :smile:

You know what ? Looking at that issue and at the handbook, let's just go ahead and add this sniff to `Core` instead of to `Extra`.

jrfnl:
Just for the fun of it, ran this sniff over core: result:
* 2 false positives for wrong replacement number (rare cases, but may need a whitelist flag)
* 6 quoted placeholders: https://core.trac.wordpress.org/ticket/41983

jrfnl:
Oh and found a bug... let me fix that...

jrfnl:
I have reported some (non-security) issues this sniff identified to bbPress and Yoast SEO as well.

Would @Ipstenu, @Otto42or @grappler be interested in running it over the whole plugin/theme repo ?

grappler:
I think it would be interesting to run on all of the themes but I don't expect many issues as there is no need to for most themes to run direkt MYSQL.

jrfnl:
In relation to the issues I opened in Trac - it turns out the documentation of the `wpdb::prepare()` and the `wpdb_esc_like()` methods may need to be updated to improve the information on using a literal `%` in combination with `LIKE` in queries.
If that yields the need for changes in this sniff, I will address that ASAP, preferably before the 0.14.0 release.

Ref: https://core.trac.wordpress.org/ticket/41983#comment:3

PR #1162, opened 2017-09-23
Contributing: update info about unit testing

jrfnl:
Related to #1130

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -66,7 +66,7 @@ When you introduce a new whitelist comment, please don't forget to update the [w
 ## Pre-requisites
 * WordPress-Coding-Standards
 * PHP CodeSniffer 2.9.x or 3.x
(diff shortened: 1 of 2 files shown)


.github/CONTRIBUTING.md
```diff
@@ -89,6 +89,7 @@ The easiest way to do this is to add a `phpunit.xml` file to the root of your WP
 ```xml
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.1/phpunit.xsd"
```
GaryJones: Can we bump the version in the schema URL, to also highlight that we support PHPUnit 6.x?

jrfnl: Good point. Done + also done for the `phpunit.xml.dist` file.

PR #1163, opened 2017-09-23
Improve fixes for PEAR.Files.IncludingFile.BracketsNotRequired

### LINKED ISSUE #1153: Fixer for PEAR.Files.IncludingFile.BracketsNotRequired assumes no spacing in parens
The fixer for `PEAR.Files.IncludingFile.BracketsNotRequired` (added in #1145) turns this:

```php
require_once( '/path/to/file.php' );
```

Into this:

```php
require_once  '/path/to/file.php' ;
```

- The space before the closing parenthesis is left.
- A new space is added after `require_once`, in addition to the space that already followed the opening parenthesis.

In other words, the fixer assumes that no extra whitespace will be here. Ideally, it should be whitespace agnostic. But because the sniff comes from upstream, it is possible that we'd need to create a custom sniff for our purposes instead.

JDGrimes:
Adds two sniffs that correct spacing on either side of the `require*`
after the parentheses are removed:

- `Squiz.WhiteSpace.SemicolonSpacing` (added to `Core`) — Requires no
  whitespace before the semicolon.
- `Squiz.WhiteSpace.LanguageConstructSpacing` (added to `Extra`) —
  Requires exactly one space after a control structure.

I ran both of these over some code to see if there were any issues, and the only problem that turned up was with the semicolon sniff on this code:

```php
		for ( ; $next <= $end; ++$next ) {
```

```
---------------------------------------------------------------------------------------------------------------------------------------
---------------------------------------------------------------------------------------------------------------------------------------
 62 | ERROR | [x] Space found before semicolon; expected "(;" but found "( ;" (Squiz.WhiteSpace.SemicolonSpacing.Incorrect)
---------------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
---------------------------------------------------------------------------------------------------------------------------------------
```

Of course, removing the space will trigger:

```
  62 | ERROR | [x] No space after opening parenthesis is prohibited
     |       |     (WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceAfterOpenParenthesis)
```

I could not find a workaround, adding a `/* comment */` did not help. I could refactor the code, of course, but in this case `$next` is assigned a value outside for readability. Perhaps the `SemicolongSpacing` sniff should ignore `for` loops anyway, since there is already a `Squiz.ControlStructures.ForLoopDeclaration` sniff specifically to check their formatting. We could see about fixing that upstream, ~~but IMO it doesn't need to be a blocker for merging this~~.

Looks like there is actually [the same issue in WPCS](https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/278955953#L571) though:

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/e96329ae758b95a5402d674935b442e073fef756/WordPress/Sniffs/Arrays/ArrayIndentationSniff.php#L423

So we may have to address that before merging.

Fixes #1153

### DIFF
--- a/WordPress/Tests/WhiteSpace/SemicolonSpacingUnitTest.inc
+++ b/WordPress/Tests/WhiteSpace/SemicolonSpacingUnitTest.inc
@@ -1,0 +1,12 @@
+<?php
+
+/*
+ * Test that the sniff does *not* throw incorrect errors for whitespace before semi-colons in
+ * "empty" parts of a `for` control structure.
+ */
+for ($i = 1; ; $i++) {}
+for ( ; $ptr >= 0; $ptr-- ) {}
+for ( ; ; ) {}
+
+// But it should when the whitespace is between a condition and a semi-colon.
+for ( $i = 1 ; ; $i++ ) {}
--- a/WordPress/Tests/WhiteSpace/SemicolonSpacingUnitTest.inc.fixed
+++ b/WordPress/Tests/WhiteSpace/SemicolonSpacingUnitTest.inc.fixed
@@ -1,0 +1,12 @@
+<?php
+
+/*
+ * Test that the sniff does *not* throw incorrect errors for whitespace before semi-colons in
+ * "empty" parts of a `for` control structure.
+ */
+for ($i = 1; ; $i++) {}
+for ( ; $ptr >= 0; $ptr-- ) {}
+for ( ; ; ) {}
+
+// But it should when the whitespace is between a condition and a semi-colon.
+for ( $i = 1; ; $i++ ) {}
--- a/WordPress/Tests/WhiteSpace/SemicolonSpacingUnitTest.php
+++ b/WordPress/Tests/WhiteSpace/SemicolonSpacingUnitTest.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * Unit test class for WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPress\Tests\WhiteSpace;
+
+use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
+
+/**
+ * Unit test class for the SemicolonSpacing sniff.
+ *
+ * Only the WPCS specific difference with the upstream sniff is being tested.
+ * The upstream sniff is tested upstream.
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   0.14.0
+ */
+class SemicolonSpacingUnitTest extends AbstractSniffUnitTest {
+
+	/**
+	 * Returns the lines where errors should occur.
+	 *
+	 * @return array <int line number> => <int number of errors>
+	 */
+	public function getErrorList() {
+		return array(
+			12 => 1,
+		);
+
+	}
+
+	/**
+	 * Returns the lines where warnings should occur.
+	 *
+	 * @return array <int line number> => <int number of warnings>
+	 */
+	public function getWarningList() {
+		return array();
+
+	}
+
+} // End class.
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -405,7 +405,10 @@
 	<rule ref="Generic.Files.EndFileNewline"/>
 
 	<!-- No whitespace should come before semicolons. -->
-	<rule ref="Squiz.WhiteSpace.SemicolonSpacing"/>
+	<!-- Extends upstream sniff. Should defer to upstream version once minimum PHPCS requirement
+		 has gone up to version containing bugfix.
+		 https://github.com/squizlabs/PHP_CodeSniffer/pull/1691 -->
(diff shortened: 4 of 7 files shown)


jrfnl:
Oh dear. Pff... for the WPCS bit, I suppose we could hackily solve it by using `for ( $ptr = $ptr; ....)`, but I do think the original code is a valid case of setting the `for()` up this way.
I suppose more of the same errors would happen with `for ( ; ; )` which is completely valid PHP, though not always intuitive for everyone.

jrfnl:
Just thinking out loud here: as the Squiz based standard is sparse on spaces, I very much doubt, we'd be able to get a fix in upstream for this.
I think the easiest way to get round this would be to extend the upstream sniff and overload the `process()` method with something like this:
```php
public function process() {
	// Check if the $stackPtr has 'nested_parenthesis'
	// If so, get the last (deepest) one, take the key to get the ptr to the parenthesis opener
	// Check if the parenthesis opener has a parenthesis owner and if so, if that's a `for`.
	// If so, return.
	// If not, parent::process()
}
```

JDGrimes:
> think the easiest way to get round this would be to extend the upstream sniff and overload the process() method with something like this:

That's essentially what I had in mind for the upstream fix, thinking that there is no need for it to overlap with the sniff specifically for `for` loops.

But perhaps best to just extend the sniff as you say.

jrfnl:
@JDGrimes You may want to open an issue about it upstream anyway to feel out if a PR would be welcomed or not.
As an upstream PR would only go in 3.2 or higher, having an extended sniff in WPCS until the time the WPCS minimum requirement has gone up would be needed anyway for now.

jrfnl:
@JDGrimes Would you happen to have the chance to update the PR per the above comments over the next few days ?

JDGrimes:
> @JDGrimes Would you happen to have the chance to update the PR per the above comments over the next few days ?

Unfortunately, I probably won't. ☹️ Been meaning to leave a comment here to that effect, but kept hoping I would have time.

jrfnl:
@JDGrimes Thanks for letting me know. As the related sniff `PEAR.Files.IncludingFile` went into 0.14.0, these should go into 0.14.0 as well, so I've rebased the PR and added an additional commit making the changes as discussed above.

I've opened an issue upstream with a proposal on how to deal with the issue regarding `for()` in the long run: https://github.com/squizlabs/PHP_CodeSniffer/issues/1690, though I'm tempted to just send in a PR with the same fix as applied here.

This PR should now pass the build, so should be good to go.

jrfnl:
Rebased for merge conflict.

WordPress/Tests/WhiteSpace/SemicolonSpacingUnitTest.inc
```diff
@@ -0,0 +1,12 @@
+<?php
+
+/*
+ * Test that the sniff does *not* throw incorrect errors for semi-colons in
```
GaryJones: Clarification:
"for semi-colons" => "for spaces before semi-colons"

WordPress/Tests/WhiteSpace/SemicolonSpacingUnitTest.inc
```diff
@@ -0,0 +1,12 @@
...
+ * "empty" parts of a `for` control structure.
+ */
+for ($i = 1; ; $i++) {}
+for ( ; $ptr >= 0; $ptr-- ) {}
+for ( ; ; ) {}
+
+// But it should when the semi-colon follows a condition.
```
GaryJones: Clarification:
"the semi-colon" => "the space before a semi-colon"

WordPress/Tests/WhiteSpace/SemicolonSpacingUnitTest.inc.fixed
```diff
@@ -0,0 +1,12 @@
+<?php
+
+/*
+ * Test that the sniff does *not* throw incorrect errors for semi-colons in
```
GaryJones: Clarification, as above.

WordPress/Tests/WhiteSpace/SemicolonSpacingUnitTest.php
```diff
@@ -0,0 +1,48 @@
...
+	 * Returns the lines where warnings should occur.
+	 *
+	 * @return array <int line number> => <int number of warnings>
+	 */
+	public function getWarningList() {
+		return array();
+
+	}
+
+} // End class.
```
GaryJones: Didn't these comments get dropped?

jrfnl: Only for long control structures, not necessarily for functions/classes.

WordPress/Tests/WhiteSpace/SemicolonSpacingUnitTest.inc
jrfnl: Fixed, just used slightly different phrasing (whitespace instead of spaces as whitespace can also be tabs or new lines)

jrfnl:
@GaryJones Thanks for the feedback. I've fixed-up the existing commit as it doesn't warrant its own commit.

jrfnl:
FYI: the upstream fix has been merged, so the WPCS version of the sniff can be removed once the WPCS minimum requirement has gone up to PHPCS `3.2.0`.

PR #1166, opened 2017-09-23
Fix suggestions for unsafe _ex() function

tfrommen:
WordPress does not have any `esc_attr_ex()` or `esc_html_ex()` functions, we have to `echo` the non-echoing variants, which do exist.

### DIFF
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -77,7 +77,7 @@ class EscapeOutputSniff extends Sniff {
 	 */
 	protected $unsafePrintingFunctions = array(
 		'_e'  => 'esc_html_e() or esc_attr_e()',
-		'_ex' => 'esc_html_ex() or esc_attr_ex()',
+		'_ex' => 'echo esc_html_x() or echo esc_attr_x()',
 	);
 
 	/**
--- a/WordPress/Tests/XSS/EscapeOutputUnitTest.inc
+++ b/WordPress/Tests/XSS/EscapeOutputUnitTest.inc
@@ -24,7 +24,7 @@ function custom_column_display( $column, $post_id )
 {
     global $post;
     switch ( $column ) {
-        case 'some_number' : 
+        case 'some_number' :
             echo (int) $test;
             echo (int) get_post_meta( $post_id, SOME_NUMBER, true );
         break;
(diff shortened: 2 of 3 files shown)


jrfnl:
Hi Thorsten, congrats on your first WPCS PR ;-)

Nice catch on those functions not existing, Not sure how they made it in, though there is an open ticket for adding these: https://core.trac.wordpress.org/ticket/32840

The phrase is being used in the following error message text: `Expected next thing to be an escaping function (like %s), not '%s'`, so would result in:
`Expected next thing to be an escaping function (like echo esc_html_x() or echo esc_attr_x()), not '_ex'`

Looking at that, adding the word `echo` to that phrase makes the sentence incorrect and confusing.

What about changing it to `esc_html_x() or esc_attr_x()` instead ?

tfrommen:
Ha, yes, thanks.

Hm, people unfamiliar with the individual gettext functions might really just replace `_ex` with, say, `esc_html_x`, and end up without any output...?

Maybe the echoing functions should get another message, for example, the following:

>Expected output to be run through an escaping function (like %s), not '%s'

?

JDGrimes:
>> Expected output to be run through an escaping function (like %s), not '%s'

I think that message makes sense. I don't see a problem using it for all unexpected functions. But maybe change "not" to "found":

> Expected output to be run through an escaping function (like %s), found '%s'

So we'd have:

> Expected output to be run through an escaping function (like echo esc_html_x() or echo esc_attr_x()), found '_ex'

Opinions?

tfrommen:
> Opinions?

Well, I'm 👍, but that's no news, I guess... 😉

jrfnl:
With the change in the message to `found` it does make more sense and would be more in line with other messages in WPCS as well.

tfrommen:
I updated my change, and created a new PR, #1209, with the discussed change to the message.

Merge in whatever order you like best. 🙂

jrfnl:
@tfrommen Thanks for doing that. I've left some feedback in the other PR.

Once the message change has been merged, I would be ok with this PR going in in it's original form, i.e. https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1166/commits/8b55d9abe495ef02bca0415682cb26c0400385e1

jrfnl:
> Once the message change has been merged, I would be ok with this PR going in in it's original form, i.e. 8b55d9a

PR #1209 has been merged :+1:

As a side-note, I just noticed there is no unit test covering the `_ex()` function with which to test this PR ...

tfrommen:
@jrfnl I adapted the message, again, and also made sure the functionality is covered in tests.

tfrommen:
OK, that sucks with the test file...

Am I the first to mention that this kind of _test_ is really far from ideal? 😅
The only way to _easily_ update tests is to append stuff...

jrfnl:
@tfrommen You're a star ;-)

Still seeing a merge conflict though. Rebasing the PR should solve it.

> Am I the first to mention that this kind of test is really far from ideal?

Actually, it makes it darn easy to make sure everything is covered with integration tests, so I'm not complaining.

tfrommen:
Done, @jrfnl.

>> Am I the first to mention that this kind of test is really far from ideal?
> Actually, it makes it darn easy to make sure everything is covered with integration tests, so I'm not complaining.

Well, I was talking about **how** the _tests_ are set up, meaning you have that ugly and expressionless array of line numbers and error counts. But I'll just be done with it. 🙂

jrfnl:
Thanks @tfrommen !

PR #1167, opened 2017-09-23
Bump version constraint of suggested package to support PHP 5.3

tfrommen:
Set version constraint to allow for use with PHP 5.3 (thanks, @GaryJones 🙂).


tfrommen:
Ah, didn't see/expect that.

PR #1169, opened 2017-09-25
PreparedSQLPlaceholders: Fix unclear error message

jrfnl:
The error message was suffering from exactly the same as it was reporting... :facepalm:

### DIFF
--- a/WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
+++ b/WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
@@ -119,7 +119,7 @@ class PreparedSQLPlaceholdersSniff extends Sniff {
 					foreach ( $matches[0] as $match ) {
 						if ( '%' === $match ) {
 							$this->phpcsFile->addError(
-								'Found unescaped literal "%" character.',
+								'Found unescaped literal "%%" character.',
 								$i,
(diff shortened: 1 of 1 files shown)


WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
```diff
@@ -119,7 +119,7 @@ public function process_token( $stackPtr ) {
 					foreach ( $matches[0] as $match ) {
 						if ( '%' === $match ) {
 							$this->phpcsFile->addError(
-								'Found unescaped literal "%" character.',
+								'Found unescaped literal "%%" character.',
```
GaryJones: This doesn't make sense now - two characters (both `%`), yet the final word of the message suggests a single character.

jrfnl: That's the whole point. Both the `addError()/addWarning()` methods as well as `$wpdb->prepare()` use a `sprintf()` like function to parse in variables into the error message/SQL query.
The `%` is used as as placeholder indicator. To use a literal `%`, you need to escape it by adding a second `%`.

As in -without this fix, the error message displayed became:
`Found unescaped literal " character.`
With the fix:
`Found unescaped literal "%" character.`

PR #1170, opened 2017-09-25
Fix markdown link at README.md

simison:
Fixes broken markdown syntax for an internal link at README.md


JDGrimes:
Thanks @simison ! 😃

PR #1175, opened 2017-10-05
PrecisionAlignment: fix false positive in combination with incorrectly aligned comments

jrfnl:
If the stars for a docblock/multi-line inline comment  were not properly aligned, the sniff would throw an error even though there was no precision alignment.
The incorrect alignment of the _stars_ is not this sniffs concern, but for another sniff.

Includes unit tests.

As the sniff is new to `0.14.0`, this bugfix should be included in the upcoming release.

### DIFF
--- a/WordPress/Tests/WhiteSpace/PrecisionAlignmentUnitTest.1.inc
+++ b/WordPress/Tests/WhiteSpace/PrecisionAlignmentUnitTest.1.inc
@@ -45,3 +45,12 @@
 /*
 
 */
+
+// Testing that incorrectly aligned docblocks and multi-line inline comments do not trigger errors.
+	/**
+	* Provision some options
+	*/
+
+	/*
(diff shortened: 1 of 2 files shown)


WordPress/Tests/WhiteSpace/PrecisionAlignmentUnitTest.1.inc
```diff
@@ -45,3 +45,12 @@
 /*
 
 */
+
+// Testing that incorrectly alignment docblocks and multi-line inline comments do not trigger errors.
```
JDGrimes: >incorrectly alignment

should be

>incorrectly aligned

😄

GaryJones: Fixed.

PR #1178, opened 2017-10-06
New RestrictedPHPFunctions sniff to initially forbid `create_function`

jrfnl:
The following has been added to the handbook:

> create_function() function, which internally performs an eval(), is deprecated in PHP 7.2. Both of these must not be used.

https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#clever-code

`create_function` was already discouraged in `Extra` via the `WordPress.PHP.DiscouragedPHPFunctions` sniff.
As it is now forbidden for Core, this check has been moved a new `WordPress.PHP.RestrictedPHPFunctions` sniff and added to the `Core` ruleset.

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -348,6 +348,10 @@
 		<message>eval() is a security risk so not allowed.</message>
 	</rule>
 
+	<!-- Rule: create_function() function, which internally performs an eval(),
+	     is deprecated in PHP 7.2. Both of these must not be used. -->
+	<rule ref="WordPress.PHP.RestrictedPHPFunctions"/>
+
 
 	<!--
 	#############################################################################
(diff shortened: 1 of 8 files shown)


JDGrimes:
GitHub offered for me to resolve the conflict with the web editor, so I thought I would give it a try, but apparently I messed something up in the process: https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/284188513

I think I must have accidentally left part of the  git markers in. Anyway, this may need to be fixed manually @jrfnl . 😞

jrfnl:
@JDGrimes No worries. I've rebased it. Should be all good to go now.

jrfnl:
Hang on.. Just realized, this needs another change.

jrfnl:
Ok, done. Added a new commit with the additional change.

The `Extra` and `WordPress` rulesets excluded `create_function` previously as it was a valid alternative to anonymous functions which were not available in PHP 5.2. This exclusion has now been removed.

jrfnl:
FYI: updated the wiki page section on "existing exclusions" to be in line with the change made here

Issue #1120, opened 2017-09-04
Faulty core fixes: ScopeIndentSniff doesn't align a multiline comment correctly


pento:
Given the following example:

```php
<?php

function a() {
/*
 * A comment.
 */
}
```

The comment will be indented, but the 2nd and 3rd line are missing the space to align their `*` correctly.

pento:
There's a similar issue with docblocks, given this example:

```php
<?php

function a() {
/**
 * A docblock
 */
}
```

The `/**` line of the docblock is indented, but subsequent lines are not.

jrfnl:
I'd need to check, but I seem to remember that is covered by one of the comment related sniffs in the `Docs` standard.

In case that would be a stand-alone sniff, should it be moved to the `Core` standard ?

pento:
Yah, I think it would need to be moved - the current `Core` behaviour is incorrect. The other option would be to include selected rules from `Docs` in the `phpcs.xml.dist` file, but that seems like it'll cause problems for anyone who wants to use `Core` rules for their own project, but doesn't copy the core `phpcs.xml.dist`.

jrfnl:
@pento I've been looking into this. There is no isolated enough upstream sniff which can be used for the multi-line comments, however, for the docblock comments - as long as these are file/function/class/interfaces/property docblocks (i.e. not hook documentation docblocks) - the `Squiz.Commenting.DocCommentAlignment` sniff could be added to the core `phpcs.xml.dist` file to fix this.

Want to have a try with that ?

I'll have a think about writing a WPCS native sniff for multi-line comment alignment. My current line of thinking is: don't touch the first line (leave that to the `ScopeIndent` sniff to sort out) and align the subsequent lines based on the alignment of the comment opener. Independently of the comment style (`/* */` vs `/** */` - not `//`).

How does that sound ?

pento:
Unfortunately, `Squiz.Commenting.DocCommentAlignment` isn't a huge help - there are a few places in `tests` it fixes, but all of the problems in `src` are either hook docblocks, or multiline comments.

I like the sound of the WPCS sniff behaviour.

jrfnl:
@pento Ok, thanks for the response. Leave it with me.

jrfnl:
@pento @DrewAPicture I have a question regarding multi-line non-docblock comments:
[...]

PR #1193, opened 2017-10-11
Core: Add `Generic.Functions.FunctionCallArgumentSpacing` sniff

### LINKED ISSUE #1189: Faulty core fixes: space isn't added after a comma in a function call
Given the following test case:

```php
<?php

a( 1,2 );
```

The lack of space after the comma in the function call isn't detected.

Core ref: src/wp-admin/install-helper.php

Tested with PHPCS `master`, WPCS `develop`.

jrfnl:
Adds sniff to check for whitespace around comma's within a function call. Allows for new lines. All errors thrown are fixable.

For closures it will also check whitespace around the equals sign for default values.

Example cases:
```php
a( 1,2 ); // NoSpaceAfterComma

a( 1   ,2 ); // SpaceBeforeComma + NoSpaceAfterComma

a( 1,   2 ); // TooMuchSpaceAfterComma

a(
   1,
   2
); // OK
```

Fixes #1189

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -213,6 +213,8 @@
 	<!-- Related to issue #970 / https://github.com/squizlabs/PHP_CodeSniffer/issues/1512 -->
 	<rule ref="WordPress.Functions.FunctionCallSignatureNoParams"/>
 
+	<rule ref="Generic.Functions.FunctionCallArgumentSpacing"/>
+
 	<!-- Rule: Perform logical comparisons, like so: if ( ! $foo ) { -->
 
 	<!-- Covers rule: When type casting, do it like so: $foo = (boolean) $bar; -->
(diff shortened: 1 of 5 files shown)


pento:
This doesn't seem to be fixing the problems for me in `src/wp-admin/install-helper.php` (the `$wpdb->get_col()` call in `maybe_create_table()`, for example).

jrfnl:
> This doesn't seem to be fixing the problems for me in src/wp-admin/install-helper.php (the $wpdb->get_col() call in maybe_create_table(), for example).

Silly question, but are you sure you are running with *this* branch ? The sniff has not been merged to WPCS yet.

I've just ran a phpcbf test over `src/wp-admin/install-helper.php` with just this sniff and with the whole `WordPress-Core` ruleset and the fixes look fine to me.

The only thing I can see which isn't correct is the docblock indents, but that's another issue - #1120.

pento:
> Silly question, but are you sure you are running with this branch ?

That's not a silly question, it was exactly the problem. It works correctly when I use the correct version! 😁

PR #1197, opened 2017-10-12
Improve whitelist comments + issue with PreparedSQL sniff

jrfnl:
**This PR contains a bugfix and an enhancement. Pulled in one go as they would otherwise conflict.**

While working on some improvements for the `PreparedSQLPlaceholders` sniff, which includes a whitelist comment, I came across the following issue:

```php
// This works fine.
$sql = $wpdb->prepare( "SELECT * FROM $wpdb->users WHERE id = %d AND user_login = %s", $replacements ); // WPCS: PreparedSQLPlaceholders replacement count OK.

/*
 * This would trigger a "Use placeholders and $wpdb->prepare();
 * found // WPCS: PreparedSQLPlaceholders replacement count OK." error.
 */
$sql = $wpdb->prepare( // WPCS: PreparedSQLPlaceholders replacement count OK.
	"SELECT *
		FROM $wpdb->users
		WHERE id = %d
			AND user_login = %s",
	$replacements
);
```

**Bugfix:** That the `WP.PreparedSQL` sniff triggers an error for comments is clearly wrong, so that is fixed in the first commit.

Next I had a look at the `has_whitelist_comment()` method as - based on the number of questions I've seen about the whitelist comment placement and the number of times I've struggled with it myself in the past -, it doesn't always seem intuitive where to place the whitelist comment.

Whitelist comments only worked if they were:
* the last token in an embedded PHP snippet
* òr at the end of the line for the token which triggers the whitelist check.

However, for a "normal" user, it is often unclear what token triggers the whitelist check, so people find it hard to correctly place the whitelist comments.

**Enhancement:** The second commit changes the behaviour of the whitelist comment method and now allows for both the above as well as a whitelist comment directly after the semi-colon ending the statement.

I do wonder why this wasn't implemented when the method was originally and if there were pertinent reasons not to do so. Maybe @JDGrimes or @westonruter knows ?

So taking the above example, this will now work:
```php
$sql = $wpdb->prepare(
	"SELECT *
		FROM $wpdb->users
		WHERE id = %d
			AND user_login = %s",
	$replacements
); // WPCS: PreparedSQLPlaceholders replacement count OK.
```

Unit tests for both are, of course, included.

#### TODO when merged
- [ ] Update wiki whitelist comment page

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -1105,18 +1105,20 @@ abstract class Sniff implements PHPCS_Sniff {
 	/**
 	 * Find whitelisting comment.
 	 *
-	 * Comment must be at the end of the line, and use // format.
+	 * Comment must be at the end of the line or at the end of the statement
+	 * and must use // format.
 	 * It can be prefixed or suffixed with anything e.g. "foobar" will match:
 	 * ... // foobar okay
 	 * ... // WPCS: foobar whitelist.
 	 *
 	 * There is an exception, and that is when PHP is being interspersed with HTML.
-	 * In that case, the comment should come at the end of the statement (right
+	 * In that case, the comment should always come at the end of the statement (right
 	 * before the closing tag, ?>). For example:
 	 *
 	 * <input type="text" id="<?php echo $id; // XSS OK ?>" />
 	 *
 	 * @since 0.4.0
+	 * @since 0.14.0 Whitelist comments at the end of the statement are now also accepted.
 	 *
 	 * @param string  $comment  Comment to find.
 	 * @param integer $stackPtr The position of the current token in the stack passed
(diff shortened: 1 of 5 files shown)


JDGrimes:
> I do wonder why this wasn't implemented when the method was originally and if there were pertinent reasons not to do so. Maybe @JDGrimes or @westonruter knows ?

I think that when I added this method I was trying to make it as efficient as possible, because it gets called by quite a few sniffs on every single token. So we just checked at the end of the line. (There was inconsistency in how different sniffs handled this up to that point, so I just decided to standardize one way or the other.) I think it was only later that the bug with interspersed HTML was fixed. So now that we get the end of the statement anyway, there is no reason not to check both places that I can see.

jrfnl:
@JDGrimes Thanks for your response.

PR #1201, opened 2017-10-17
AutoEscapedFunctions: add `get_submit_button()`

jrfnl:
Looks like this function correctly escapes everything for output and does not have a filter which could undo things.

Ref: https://developer.wordpress.org/reference/functions/get_submit_button/

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -216,6 +216,7 @@ abstract class Sniff implements PHPCS_Sniff {
 		'get_search_form'           => true,
 		'get_search_query'          => true,
 		'get_sidebar'               => true,
+		'get_submit_button'         => true,
 		'get_template_part'         => true,
 		'get_the_author_link'       => true,
 		'get_the_author'            => true,


westonruter:
Normally, yes. However, I could do something evil and pass bad `$other_attributes` like either of:

```php
$other_attributes = '><script>alert("EVIL")</script><';
```
or
```php
$other_attributes = array(
    '><script>alert("EVIL")</script><' => '',
);
```
This would be unlikely, but we wouldn't be able to guarantee where `$other_attributes` is coming from statically.

jrfnl:
Cricky. You're right of course, though only when `$other_attributes` is passed as a string as for an array it's correctly escaped.

The above would result in broken HTML though, but I suppose it could be rewritten as `onLoad="runEvilScript();"`, so you're point is valid (though chances are very small).

westonruter:
@jrfnl actually even when `$other_attributes` is passed as an array it is vulnerable because only the array values get escaped. The attribute names do not:

```php
        foreach ( $other_attributes as $attribute => $value ) {
            $attributes .= $attribute . '="' . esc_attr( $value ) . '" '; // Trailing space is important
        }
```

So as I see it, a key of `'><script>alert("EVIL")</script><'` would get injected.

jrfnl:
Closing this PR for now. I do think it would be a good idea for a Trac ticket to be opened for this function to be improved in Core, though I suspect it won't be accepted what with that it would break some plugins doing really funky (and very stupid) stuff.....

PR #1207, opened 2017-10-20
AssignmentInCondition: add separate errorcode for assignment found in while

jrfnl:
A `while()` condition is the only control structure in which it is sometimes legitimate to use an assignment in condition.

Having a separate error code will enable people to selectively exclude warnings for this.

N.B.: same change has been pulled upstream.

N.B.2: I've not applied this to the second error message as when it's a non-variable assignment, it doesn't matter whether it's in a loop or not, it's always wrong.

### DIFF
--- a/WordPress/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
+++ b/WordPress/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
@@ -193,10 +193,15 @@ class AssignmentInConditionSniff extends Sniff {
 			}
 
 			if ( true === $hasVariable ) {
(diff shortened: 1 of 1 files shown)


jrfnl:
> Fine with me, though it may mean someone who had previously excluded all Found messages, now gets FoundInWhileCondition popping up, which they will then need to exclude as well.

True, but chances are that they only excluded `Found` because of a `while` structure, so this will enable them to *include* `Found` and *exclude* messages for `while` constructs, which should be better ;-)

jrfnl:
On that note: the sniff has not been in a release yet, so only people using bleeding edge could have done that.

PR #1208, opened 2017-10-21
PreparedSQLPlaceholders: massive improvements

jrfnl:
I initially wanted to fine-tune the sniff to throw more helpful messages for `%` SQL wildcards found in the value part of a `LIKE` query based on https://core.trac.wordpress.org/changeset/41660 .
It took some time to figure out the right way to isolate those out correctly.

While working on that, I made a number of significant other improvements to the sniff making it far more accurate and helpful.

Please see the individual commits for more details.
The PR will also be easiest to review using the individual commits.

Extensive unit tests have been added for all additional checks and improvements.

----

### Todo once this has been merged:
- [ ] Update the wiki page for the new whitelist flag


WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
```diff
@@ -23,6 +23,14 @@
  *   placeholders found or individual parameters for each placeholder should
  *   be passed.
  *
+ * The sniff allows for a specific pattern with a variable number of placeholders
+ * created using code along the lines of:
+ * `sprintf( 'query .... IN (%s) ...', implode( ',', array_fill( 0, count( $something), '%s' ) ) )`.
```
JDGrimes: There is a space missing after `$something`.

WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
```diff
@@ -217,4 +439,128 @@ public function process_token( $stackPtr ) {
...
+
+		return $regex_quote;
+	}
+
+	/**
+	 * Analyse a sprintf() query wrapper to see if it contains a specific code pattern
+	 * to deal correctly with `IN` queries.
+	 *
+	 * The pattern we are searching for is:
+	 * `sprintf( 'query ....', implode( ',', array_fill( 0, count( $something), '%s' ) ) )`
```
JDGrimes: Ditto.

WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
```diff
@@ -217,4 +439,128 @@ public function process_token( $stackPtr ) {
...
+
+		return $found;
+	}
+
+	/**
+	 * Analyse an implode() function call to see if it contains a specific code pattern
+	 * to dynamically create placeholders.
+	 *
+	 * The pattern we are searching for is:
+	 * `implode( ',', array_fill( 0, count( $something), '%s' ) )`
```
JDGrimes: And here too.

WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
jrfnl: Good point. Fixed.

For the example I agree that the space should be there. As for the sniff: that should work independently of code-style.

jrfnl:
@JDGrimes While I fixed up those docs, I realized I should make one more tiny message precision improvement and add something about `LIKE` to the class docblock, so I've added both these changes in a separate new commit to make them reviewable rather than fixing up the original commits.

WordPress/Sniff.php
```diff
@@ -36,6 +36,15 @@
 abstract class Sniff implements PHPCS_Sniff {
 
 	/**
+	 * Regex to get complex variables from T_DOUBLE_QUOTED_STRING or T_HEREDOC.
+	 *
+	 * @since 0.14.0
+	 *
+	 * @var string
+	 */
+	const REGEX_COMPLEX_VARS = '`(?:(\{)?(?<!\\\\)\$)?(\{)?(?<!\\\\)\$(\{)?(?P<varname>[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)(?:->\$?(?P>varname)|\[[^\]]+\]|::\$?(?P>varname)|\([^\)]*\))*(?(3)\}|)(?(2)\}|)(?(1)\}|)`';
```
GaryJones: Where has this regex come from? 

Can it be broken down and commented (with the flag that ignores whitespace)?

jrfnl: I wrote the regex (what did you expect ?).

I did so nearly a year ago when I needed the same `strip_interpolated_variables()` function for the [PHPCompatibility standard](https://github.com/wimg/PHPCompatibility/blob/9a9d15aa50b6ffec7a0e7683377bef338c64911f/PHPCompatibility/Sniff.php#L26) and pulled it in https://github.com/wimg/PHPCompatibility/pull/314 which includes [extensive unit tests of the function](https://github.com/wimg/PHPCompatibility/blob/9a9d15aa50b6ffec7a0e7683377bef338c64911f/PHPCompatibility/Tests/BaseClass/FunctionsTest.php#L441-L504).

The varname part of the regex comes from http://php.net/manual/en/language.variables.basics.php

I don't mind documenting the regex, but that will easily take me two hours or more as I'll have to decipher it myself again first... 

If you insist on it being documented, I'd suggest opening an issue so it will be done in a separate PR at some point.

PR #1209, opened 2017-10-22
Adapt message for unsafe printing functions.

tfrommen:
As discussed in #1166, this PR adapts the message when using an unsafe printing function.

### DIFF
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -243,7 +243,7 @@ class EscapeOutputSniff extends Sniff {
 
 		if ( isset( $end_of_statement, $this->unsafePrintingFunctions[ $function ] ) ) {
 			$error = $this->phpcsFile->addError(
-				"Expected next thing to be an escaping function (like %s), not '%s'",
+				"Expected output to be run through an escaping function (like %s), found '%s'.",
 				$stackPtr,
 				'UnsafePrintingFunction',
 				array( $this->unsafePrintingFunctions[ $function ], $function )
--- a/WordPress/Sniffs/XSS/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/XSS/EscapeOutputSniff.php
@@ -225,9 +225,9 @@ class EscapeOutputSniff extends Sniff {
 			// Report on what is very likely a PHP short open echo tag outputting a variable.
 			if ( preg_match( '`\<\?\=[\s]*(\$[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*(?:(?:->\S+|\[[^\]]+\]))*)[\s]*;?[\s]*\?\>`', $this->tokens[ $stackPtr ]['content'], $matches ) > 0 ) {
 				$this->phpcsFile->addError(
-					'Expected next thing to be an escaping function, not %s.',
+					"All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks), found '%s'.",
 					$stackPtr,
-					'OutputNotEscaped',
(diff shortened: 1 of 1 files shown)


tfrommen:
Oh my. Originally, I just wanted to fix a reference to non-existing functions... 😅 

Anyway, sorry, but I'm not following. Would you like all three messages to be _instructive_? This will look odd with two parts in parentheses, and also with all of that in one pairs of parentheses.

We could include the reference (which is just _namedropping_, as we don't include real links) in a separate, second sentence, but since there actually are two distinct Handbooks for plugin and theme development, this is not straight forward. Personally, I would not include any reference.

`echo`-ing:

> All output should be run through an escaping function (like %s).

Printing functions:

> All output should be run through an escaping function (like %s), found '%s'.

But if you ask me, I would just leave off the currently used printing function, and have a single message for all three situations.

This would mean we can make the `$unsafePrintingFunctions` array just keep the function names (and adapt the check to be `in_array()`, of course) and close my other, original PR...

Waiting for feedback on this.

jrfnl:
> Oh my. Originally, I just wanted to fix a reference to non-existing functions... :sweat_smile:

That happens sometimes :wink: (See #1208, I only wanted to change the message for LIKE queries... )

> Anyway, sorry, but I'm not following.

Sorry for being unclear. Unfortunately as those lines haven't been touched (yet), I couldn't leave comments in-line.

Let me try and clarify.

1. For the error being thrown for non-escaped output using short echo:
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/d7d8e96a820c52e5eec60098463b2d7f74eb451d/WordPress/Sniffs/XSS/EscapeOutputSniff.php#L227-L232

    I'm suggesting to change the message to something along the lines of:
    `All output should be run through an escaping function (see the WP Developer Handbooks for 'Data Validation'), found '%s'`
(same as lower down) and the error code to `OutputNotEscapedShortEcho`.

2. For the one you changed:
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/d7d8e96a820c52e5eec60098463b2d7f74eb451d/WordPress/Sniffs/XSS/EscapeOutputSniff.php#L245-L250
     I'm suggesting to change the message to:
    `All output should be run through an escaping function (like %s), found '%s'.` 

    The reference to the handbooks is not needed here as the message already suggests the most appropriate alternatives.

3. For the last one:
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/d7d8e96a820c52e5eec60098463b2d7f74eb451d/WordPress/Sniffs/XSS/EscapeOutputSniff.php#L442-L447
    I'm suggesting to change the message to:
    `All output should be run through an escaping function (see the WP Developer Handbooks for 'Data Validation'), found '%s'`


> We could include the reference (which is just namedropping, as we don't include real links) in a separate, second sentence, but since there actually are two distinct Handbooks for plugin and theme development, this is not straight forward. Personally, I would not include any reference.

I don't agree. Having the page name there is helpful for people to know what to search for on the WP site.

It's not always devs who've made a conscious choice to use WPCS who will be presented with these messages. They are also used in the (new) Theme Check plugin and already by a number of theme reviewers, so some devs will be confronted with the message(s) out of context.

> But if you ask me, I would just leave off the currently used printing function, and have a single message for all three situations.

Again, I don't agree. This sniff may throw several messages for one output call, which would become very confusing if it didn't reference the variable which is not being escaped.

Also, some output calls might be very long and concatenate a lot of different parts over several lines with most of the variables escaped or using auto-escaped functions, but one. Having the "found %s" there, makes it easier to fix the right one.

All in all, I do still believe your original PR has merit. Especially combined with the other changes as I've outlined above.

tfrommen:
Sooo, I hope this makes sense, @jrfnl ! 🙂

PR #1213, opened 2017-10-23
I18n: improve handling of "default" text-domain

### LINKED ISSUE #1212: I18n: only allow omitting text-domain if the only text-domain passed is `default`
Further on #1141, if `default` is passed as part of a group of text-domains `my-textdomain,default`, the sniff will no longer throw errors when the text domain is omitted.

This also means that it will no longer throw errors when `my-textdomain` was the intended text domain and it was accidentally omitted.

That makes this check a lot less useful as just by adding `default` to the list of allowed domains, you are essentially disabling it.

What about changing the logic to this ?
* if `default` is the *only* text domain passed, allow omitting it.
* if `default` is passed as part of a group of text-domains, throw a warning about a missing text-domain and mention to use the `default` text domain if a core translation was intended.

/cc @westonruter

jrfnl:
PR #1141 changed the behaviour of the I18n sniff to ignore a missing text domain argument if `default` was one of the allowed text-domains.

This was overshooting a little as this effectively disabled the complete "missing domain argument" check if `default` was one of the passed text-domains.

This PR changes that behaviour as well as addressing the remaining comment made in the original PR:
> If the text_domain in the ruleset is `default` and the `$text_domain` has explicitly been set to default as well, then this would seem like a good redundancy warning to trigger, though it shouldn't be an outright error.
>
> I would like to suggest making it a fixableWarning in that case.

Ref: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1141#issuecomment-329317824

If `default` is _**the only text-domain**_ set in the ruleset:
* The sniff will now *not* trigger an error for a missing text-domain (= same as in PR #1141).
* Trigger an auto-fixable warning about the text-domain argument being redundant when the text-domain in a gettext function call is explicitly set to `default`.

If `default` is _**one of an array of text-domains**_ set in the ruleset:
* The sniff will now trigger a warning instead of an error for a missing text-domain. The warning has a slightly different error message and a different error code. This fixes the "overshooting".

Additionally, this PR also:
* Fixes a potential "undefined index" notice for line 375.
* Reduces some code duplication in the examination of text string tokens.

Includes adjusted unit tests.

Fixes #1212


jrfnl:
:blush:

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -341,15 +372,32 @@ protected function check_argument_tokens( $context ) {
...
 		if ( empty( $tokens ) || 0 === count( $tokens ) ) {
 			$code = $this->string_to_errorcode( 'MissingArg' . ucfirst( $arg_name ) );
-			if ( 'domain' !== $arg_name || ( ! empty( $this->text_domain ) && ! in_array( 'default', $this->text_domain, true ) ) ) {
+			if ( 'domain' !== $arg_name ) {
+				$this->addMessage( 'Missing $%s arg.', $stack_ptr, $is_error, $code, array( $arg_name ) );
+				return false;
+			}
+
+			// Ok, we're examining a text domain, now deal correctly with the 'default' text domain.
+			if ( true === $this->text_domain_is_default ) {
```
GaryJones: This could be a separate `if` (and the next `elseif` becomes an `if`) since we return here.

WordPress/Tests/WP/I18nUnitTest.1.inc
```diff
@@ -151,9 +147,16 @@ really.
...
+__( "String default text domain.", "my-slug" ); // Ok.
+__( "String default text domain.", "default" ); // Ok.
+__( "String default text domain.", 'something' ); // Bad, text-domain mismatch.
+__( 'String default text domain.' ); // Warning, use default.
+
 // @codingStandardsChangeSetting WordPress.WP.I18n text_domain default
 __( 'String default text domain.', 'my-slug' ); // Bad because text_domain is only 'default'.
-__( 'String default text domain.', 'default' ); // Ok because domain is being explicit.
+__( 'String default text domain.', 'default' ); // Warning, text domain can be omitted.
+__( 'String default text domain.', /*explicitely set for reason. */ 'default' ); // Warning, text domain can be omitted (not auto-fixable.
```
GaryJones: Typo: `explicitly` (and needs a space before it)

WordPress/Sniffs/WP/I18nSniff.php
jrfnl: Done.

WordPress/Tests/WP/I18nUnitTest.1.inc
jrfnl: Fixed

PR #1216, opened 2017-10-25
Add WP_Ajax_UnitTestCase to \WordPress\Sniff::$test_class_whitelist

westonruter:
I was getting a `WordPress.Files.FileName.InvalidClassFileName` error for test when I extended `WP_Ajax_UnitTestCase`.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -839,10 +839,15 @@ abstract class Sniff implements PHPCS_Sniff {
 	 * @var string[]
 	 */
 	protected $test_class_whitelist = array(
-		'WP_UnitTestCase'            => true,
-		'WP_Ajax_UnitTestCase'       => true,
-		'PHPUnit_Framework_TestCase' => true,
-		'PHPUnit\Framework\TestCase' => true,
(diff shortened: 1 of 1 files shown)


GaryJones:
The additions all seem very specific to WP Core (the software, not the Standard). i.e. it's not applicable when I'm running PHPCS against my plugin test files. 

The existing three (red in this PR) are all very potentially applicable to someone creating unit tests for a theme or plugin.

Is there a way for WP Core to add these exclusions via it's own `phpcs.xml.dist` instead?

jrfnl:
> Is there a way for WP Core to add these extra exclusions via it's own phpcs.xml.dist instead, and have them be merged with the existing three?

There is. :wink:  https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#custom-unit-test-classes

Having said that, I'm in favour of this PR as it's quite customary for plugins to extend the WP core test classes within their Unit test suite.

GaryJones:
Not something I've ever done but I won't stand in the way of this PR going in.

PR #1217, opened 2017-10-30
Naming Conventions: make exceptions for core hooks intended to be reused

JDGrimes:
Some core actions and filters are intended to be called by plugins and themes, so we don't need to throw an error when they are.

### DIFF
--- a/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
+++ b/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
@@ -92,6 +92,18 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	);
 
 	/**
+	 * A list of core hooks that are allowed to be called by plugins and themes.
+	 *
+	 * @since 0.14.0
+	 *
+	 * @var array
+	 */
+	protected $whitelisted_core_hooks = array(
+		'widget_title'   => true,
+		'add_meta_boxes' => true,
+	);
+
+	/**
 	 * Returns an array of tokens this test wants to listen for.
 	 *
 	 * @since 0.12.0
@@ -572,6 +584,13 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 
 		$is_error    = true;
 		$raw_content = $this->strip_quotes( $parameters[1]['raw'] );
+
+		if (
+			'define' !== $matched_content
+			&& isset( $this->whitelisted_core_hooks[ $raw_content ] )
+		) {
+			return;
+		}
 
 		if ( $this->is_prefixed( $raw_content ) === true ) {
 			return;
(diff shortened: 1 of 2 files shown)


JDGrimes:
Any thoughts on also adding the `'upgrader_process_complete'` action to this list? My plugin adds some custom upgraders that call this action, but it doesn't seem as obvious that it is intended to be reused this way. But if nobody has objections to adding it, I'll do so.

jrfnl:
> Any thoughts on also adding the 'upgrader_process_complete' action to this list? My plugin adds some custom upgraders that call this action, but it doesn't seem as obvious that it is intended to be reused this way. But if nobody has objections to adding it, I'll do so.

I'm not sure that is such a typical exception that it qualifies to make it into this list (though TGMPA obviously uses it as well).
The sniff already offers a whitelist comment `// WPCS: prefix ok.` which can be used in those cases.

JDGrimes:
> I'm not sure that is such a typical exception that it qualifies to make it into this list (though TGMPA obviously uses it as well).

Yeah, I was thinking along similar lines.



> The sniff already offers a whitelist comment // WPCS: prefix ok. which can be used in those cases.

Right. I think I'll just use that for now.

GaryJones:
@JDGrimes Although your commit passed, the subsequent merge commit failed:

https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/builds/295128533

jrfnl:
@GaryJones As this involves all of (and only) the runs against PHPCS `master`, I suspect this has to do with a bad upstream commit.

jrfnl:
I've added a comment to the relevant commit upstream: https://github.com/squizlabs/PHP_CodeSniffer/commit/1f07a69c00ef3b933f295f6c047e297b679677e7#commitcomment-25303320

jrfnl:
Never mind: looks like this was already fixed upstream in a commit made after our Travis run had run its course: https://github.com/squizlabs/PHP_CodeSniffer/commit/a1db83c90d67bfa8373029d03bfacc2c84f66304

jrfnl:
I've restarted the builds for `develop` and the `0.14.0` release PR to confirm.

GaryJones:
Would `the_content` and `the_excerpt` be suitable for inclusion here as well?

jrfnl:
I was also wondering about `the_title` ?

GaryJones:
My simplified use case right now, is:

```php
foreach ( $featured_offers as $featured_offer ) {
    // ...do other stuff with $featured_offer->ID and $featured_offer->post_title...
    echo wp_kses_post( apply_filters( 'the_excerpt', $featured_offer->post_excerpt ) ); // WPCS: prefix ok.
}
```

Are you suggesting that setting up `the_post()` (and later `wp_reset_postdata()`) and calling `the_excerpt()` would be the way to go?

jrfnl:
Not suggesting anything at the moment. I'd need to look into this more before having a proper opinion ;-)

PR #1220, opened 2017-11-01
PreparedSQLPlaceholders: update for WP 4.8.3 changes

jrfnl:
WP 4.8.3 brings back numbered and formatted placeholders. The sniff was originally written to detect *against* those, so to prevent confusing devs terribly, something had to be done quickly before the `0.14.0` release.

So here goes.

* Allow for normal `vsprintf()` placeholders, but only for digits, floats and strings - `dfFs`.
* Continue to detect unsupported placeholders and unescaped literal `%`'s, though this has now become a far more complex set to detect.
* Continue to error on quoted _simple_ placeholders.
* **New**: Throw a `warning` on _unquoted_ **complex** placeholders. These will not be quoted by `$wpdb->prepare()` as these are often used to pass in partial query strings.
* Downgrade the mismatched placeholders vs replacements message to a `warning`. As the passing in of _partial query strings_ has been brought back, the chances of a number mismatch have increased, so a warning seems more appropriate.

The regexes are "variations on the same theme" as the regex which is used in the `WP.I18n` sniff, but with some significant differences.

The first regex is not that different, except that it better protects against issues with a literal `%` (`%%`) being followed by a placeholder. It is also limited to the allowed placeholder types.
The second regex has more significant differences, both in the greediness as well as the matching of "whatever comes after" the `%`.
The documentation of the regexes illustrate this in more detail.

Includes updated & adjusted unit tests to test for the new reality.

This should be the last PR for WPCS `0.14.0`. I'll update the changelog in a moment to account for these changes, so it's good to go as soon as this is merged.

/cc @pento


WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
```diff
@@ -43,6 +43,77 @@
 class PreparedSQLPlaceholdersSniff extends Sniff {
 
 	/**
+	 * These regexes copied from http://php.net/manual/en/function.sprintf.php#93552
+	 * and adjusted for limitations in `$wpdb->prepare()`.
+	 *
+	 * Near duplicate of the one used in the WP.I18n sniff, but with less types allowed.
```
GaryJones: Typo: less -> fewer

WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
```diff
@@ -43,6 +43,77 @@
...
+		(?:
+			%                          # Start of placeholder.
+			(?:[0-9]+\\\\?\$)?         # Optional ordering of the placeholders.
+			[+-]?                      # Optional sign specifier.
+			(?:
+				(?:0|\'.)?                 # Optional padding specifier - excluding the space.
+				-?                         # Optional alignment specifier.
+				[0-9]*                     # Optional width specifier.
+				(?:\.(?:[ 0]|\'.)?[0-9]+)? # Optional precision specifier with optional padding character.
+				|                      # Only recognize the space as padding in combination with a width specifier.
```
GaryJones: Alignment of comment seems to off by 4.

WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
```diff
@@ -43,6 +43,77 @@
...
+		(?<!%)                     # Don\'t match a literal % (%%).
+		(
+			%                          # Start of placeholder.
+			(?!                        # Negative look ahead.
+				%[^%]                       # Not a correct literal % (%%).
+				|
+				%%[dfFs]                    # Nor a correct literal % (%%), followed by a simple placeholder.
+			)
+			(?:[0-9]+\\\\??\$)?+        # Optional ordering of the placeholders.
+			[+-]?+                     # Optional sign specifier.
```
GaryJones: Comment alignment off.

WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
```diff
@@ -43,6 +43,77 @@
...
+			(?:
+				(?:0|\'.)?+                 # Optional padding specifier - excluding the space.
+				-?+                         # Optional alignment specifier.
+				[0-9]*+                     # Optional width specifier.
+				(?:\.(?:[ 0]|\'.)?[0-9]+)?+ # Optional precision specifier with optional padding character.
+				|                      # Only recognize the space as padding in combination with a width specifier.
```
GaryJones: Comment alignment off.

GaryJones:
L19: Does the documentatio need updating?

L21: Says that placeholders should not be quoted, but then includes a check for unquoted complex placeholders - does L21 need clarifying?

Can't find any explanation in the docs about what counts as a "simple" placeholder vs what counts as a "complex" placeholder?

WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
jrfnl: Fixed.

WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
jrfnl: Fixed

WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
jrfnl: That's on purpose to make it clear that it divides the regex part this is in ( ... | ... ) and the docs explain what the second part is about.

WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
jrfnl: See above.

jrfnl:
Regarding the class docblock documentation: updated, including indication of the difference between simple and complex placeholders.

PR #1233, opened 2017-11-20
Added Enqueued functions check for version and in footer

snightingale:
I have created a custom Sniff based from:

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1146

This checks that IF source ($src) if specified then it will check to make sure there is a Version ($ver), and it will error if not.

Also added another check for In Footer ($in_footer) to warn the user that no value is set.

PR #1237, opened 2017-11-22
Update default minimum WP version

jrfnl:
The minimum version should be three versions behind the latest WP release, so what with 4.9 having come out last week, it should now be 4.6.

@JDGrimes Want to run your deprecated function magic to see if we need to add functions/classes/parameters to the lists ?

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -78,7 +78,7 @@ abstract class Sniff implements PHPCS_Sniff {
 	 *
 	 * @var string WordPress version.
 	 */
-	public $minimum_supported_version = '4.5';
+	public $minimum_supported_version = '4.6';
 
 	/**
 	 * Custom list of classes which test classes can extend.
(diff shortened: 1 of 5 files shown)


jrfnl:
The build is running into the same fixer conflict issue as #1233. I'll look into that when I find some time and open a separate PR for that if needs be. (This one would nee d to be rebased afterwards)

jrfnl:
Rebased. I expect the build will now pass.

JDGrimes:
@jrfnl I've run the tool, and it looks like there are some newly deprecated functions in 4.9. Here is a diff: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/compare/feature/update-minimum-wp-version...feature/update-minimum-wp-version-diff

The other differences I think are mostly because of the scanner's limitations (it doesn't detect class methods), but it might be good to double-check on some.

jrfnl:
@JDGrimes Thanks for that! I've re-verified all the differences.

* The methods are covered by ticket #731 and should be removed once that tickets gets fixed.
* The function differences are fine. The sniff is correct in this case and WP Core is lagging behind. These have all been previously reported in https://core.trac.wordpress.org/ticket/41121 (still open).
    I have added comments to the sniff to annotate this and save me from having to do the same check again next time ;-)

Other than that, the four newly deprecated functions have of course been added, so this PR should be good to go AFAICS.

JDGrimes:
Should this be merged as part of 0.15.0, or should it wait?

jrfnl:
@JDGrimes As 4.9 is out and it's a small change, I'd be ok with it to be merged in for 0.15.0.

PR #1238, opened 2017-11-23
Fix build failure

jrfnl:
The build failure has to do with an upstream fixer conflict between two sniffs fighting over tabs vs spaces.
This is related to a recent upstream change and should be reported there for further attention.

As the space is not important for this particular unit test, let's just prevent the conflict from happening for now.


jrfnl:
Issue reported upstream in https://github.com/squizlabs/PHP_CodeSniffer/issues/1759

Issue #1138, opened 2017-09-11
Issues with closures and `use` as arguments


tfrommen:
I'm having issues when I feed closures that `use` local variables into function calls, for example, `add_action()`.

**Example `test.php` file:**

```php
<?php
$foo = 'foo';
$bar = 'bar';
add_action( function () use (
	$foo,
	$bar
) {

	return "{$foo}{$bar}";
} );
```

**Run:**

```sh
phpcs test.php --standard=WordPress --sniffs=PEAR.Functions.FunctionCallSignature
```

**Output:**

```text
-----------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-----------------------------------------------------------------------------------------------------
 7 | ERROR | [x] Multi-line function call not indented correctly; expected 4 spaces but found 0
-----------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-----------------------------------------------------------------------------------------------------
```

This is also true when passing the closure as argument to a mehod, of course:

```php
$foo->inject( function () use (
	$foo,
	$bar
) { ... } );
```

----

Am I doing something _wrong_? What would be the WPCS-conform way of formatting this? I'm pretty sure it is **not** the following, which happens to be what PHPCS is fine with:

```php
<?php
$foo = 'foo';
$bar = 'bar';
add_action( function () use (
	$foo,
	$bar
	) {

	return "{$foo}{$bar}";
} );
```

jrfnl:
Hi @tfrommen, I'm going to have to take a look at the PEAR sniff to see if it can actually deal with closures with use statements. If not, we may need to patch the upstream sniff.

Your issue is loosely related to #1071 in which we have been discussing creating dedicated sniffs for checking the formatting of `use` statements, though by the looks of your issue, the current proposal would definitely conflict with the PEAR sniff as it is now.

PR #1242, opened 2017-11-26
Issue 1157/Proposal 2: rename sniffs

jrfnl:
While there currently isn't consensus yet about proposal 1 of issue #1157, proposal 2 seems widely supported.

To that end, I have prepared the necessary changes.

This is PR 1 in a series of three PRs to address this and by far the largest one.

This PR basically takes care of all the sniff renames, sniff deprecations and `error` to `warning` downgrades as proposed, with the exception of the splitting of the `PostsPerPage` sniff which I will address in a separate PR.

I've elected to do all the renames in one go as they all require changes to the `ruleset.xml` file for the top-level `WordPress` ruleset and would otherwise all conflict or have to be pulled one by one, each waiting on the previous one.

All changes have been made in separate commits though for easier review and better traceability in the git history.


GaryJones:
Reminder: Check wiki for old references too.

JDGrimes:
This looks good to me, but I wonder if we should do some additional renames while we are at it.

For example, I'm unclear on the difference between `PHP`, `CodeAnalysis`, and `Functions`.

Going down through the list of our different sniff groupings:

- `Arrays` looks OK.
- `CSRF` is going to be deprecated by this PR.
- `Classes` contains only one sniff that seems OK where it is, I guess. Although as with `Functions` (see below), what should go here as opposed to other places (`PHP` or `CodeAnalysis`) may be kind of fuzzy.
- `CodeAnalysis` seems a reasonable category, but how is it different from `PHP`? Should some sniffs be moved here?
- `DB` is fine.
- `Files` I see no problem with. Though on the other hand, the sniff that is there could probably just as easily be moved to `NamingConventions`.
- `Functions` contains three sniffs: `DontExtract`, `FunctionCallSignatureNoParams`, and `FunctionRestrictions` (which is deprecated). Why should `DontExtract` be here, rather than in `PHP` (with other function related sniffs)? And why should `FunctionCallSignatureNoParams` be here, rather than in `Whitespace`? And is this category supposed to be about function declarations, or about function usages? Currently it is about both, but we are inconsistent as to what we place here and what we place elsewhere. Should we just deprecate this category entirely?
- `NamingConventions` looks good.
- `PHP` seems to hold most of our generic function-related sniffs. It also contains some sniffs that could just as easily go under `CodeAnalysis`, like `YodaConditions` and `StrictComparisons`. Those sniffs could probably be used for JS as well as PHP. I'm not sure that we really even include JS sniffing in our vision for WPCS, which makes the `PHP` category kind of redundant.
- `VIP` I guess is dealt with pretty well by this PR already.
- `Variables` contains only `VariableRestrictions` (which is deprecated), and `GlobalVariables`.  Since the latter is specifically related to WordPress globals, it might be reasonable to move it to `WP` and get rid of this category entirely. Other variable-related sniffs seem to fall into other categories already (like `NamingConventions`).
- `WP` looks good to me, for sniffs specifically related to WordPress-provided functions/classes/etc., that don't fall into other discrete categories like `Security` or `DB` (as fixed by this PR).
- `WhiteSpace` looks good.
- `XSS` is dealt with by this PR, and will now be deprecated.

So basically, `CodeAnalysis`, `Functions`, `PHP`, and `Variables` could probably use some work. `Classes` and `Files` might also be ripe for change. Or would it just be unnecessary churn? Thoughts?

I realize that this initial proposal was mainly intended to address ease of including/excluding DB and Security sniffs in particular, and preparing VIP for deprecation. So maybe any other changes should be dealt with separately?

jrfnl:
@JDGrimes As a general rule for both the proposal as well as most sniffs where I've added new categories, I've looked towards PHPCS upstream to see where similar sniffs upstream were placed and tried to use categories which were consistent with upstream.

> Classes contains only one sniff that seems OK where it is, I guess. Although as with `Functions` (see below), what should go here as opposed to other places (`PHP` or `CodeAnalysis`) may be kind of fuzzy.

Upstream `Classes` is used for any class specific (formatting) sniffs. Example sniffs from upstream in this category (various standards): `DuplicateClassName`, `OpeningBraceSameLine`, `ClassDeclaration`, `PropertyDeclaration`, `SelfMemberReference`, `ValidClassName`.

We mainly use upstream sniffs for these things at the moment, except for the `ClassInstantiation` for which there is no upstream sniff.

> `CodeAnalysis` seems a reasonable category, but how is it different from `PHP`? Should some sniffs be moved here?

To me, the difference between the `CodeAnalysis` category and the `PHP` category is in the kind of things being analysed.
* `CodeAnalysis` - detect sloppy or messy code. Often thing which the PHP Mess Detector would detect as well, i.e. `EmptyStatement`. Both the sniffs we have in that category will be removed in due time in favour of the same sniffs upstream.
    - The `AssignmentInCondition` sniff has already been merged upstream, but our minimum required version will need to go up before we can use the upstream version (also upstream currently does not detect assignments in condition in ternaries yet)
    - The `EmptyStatement` sniff was added to deal with some empty statement detection missing in the upstream sister-sniff, but needed for the core project. I'm not sure by heart whether I've already pulled our additions upstream. If not, they should be.
* `PHP` - detect certain PHP constructs or features on which we have an opinion about whether to use these or not.

With that in mind, `PHP.StrictComparisons` _could_ be moved to `CodeAnalysis`. However, it is also still well placed in the `PHP` category, especially as the `StrictInArray` sniff is there too, which is very much linked to specific PHP functions and therefore seems correctly placed in `PHP`.

_Side note: both of these sniffs would be good candidates to pull upstream in the foreseeable future._

The only "odd one out" I see in the PHP category would be `YodaConditions`, but I'm not sure where else that should be placed.
Personally, I'd rather have a discussion whether that rule is still needed at all, now the `AssignmentInCondition` sniff is available.

> Files I see no problem with. Though on the other hand, the sniff that is there could probably just as easily be moved to `NamingConventions`.

AFAICS, the difference between `NamingConventions` and `Files` is:
* `NamingConventions` - naming of code constructs. Looks for a construct declaration and then examines the name.
* `File` - anything file specific. Upstream this includes sniffs like `ByteOrderMark`, `EndFileNewline`, `LineLength`, `(No)ClosingTag` and `OneObjectStructurePerFile`.
    Generally speaking these sniffs register the `T_OPEN_TAG`, scan a whole file in one go and prevent PHPCS from triggering the sniff a second time.

With that in mind, the WP sniffs follow the exact same logic.

> Functions contains three sniffs: `DontExtract`, `FunctionCallSignatureNoParams`, and `FunctionRestrictions` (which is deprecated). Why should `DontExtract` be here, rather than in PHP (with other function related sniffs)?

`DontExtract` is being moved to PHP as part of this PR :sunglasses: 

> And why should `FunctionCallSignatureNoParams` be here, rather than in `Whitespace`? And is this category supposed to be about function declarations, or about function usages? Currently it is about both, but we are inconsistent as to what we place here and what we place elsewhere. Should we just deprecate this category entirely?

Upstream, this category is used similar to `Classes` and contains both sniffs related to function declarations as well as function calls. Just like classes is being used for both class declaration sniffs as well as class instantiation sniffs.

There are a couple of open issues - #1138 comes to mind - related to function call signatures, for which I'm considering writing a WPCS specific sniff which would then be placed in this category.

Basically, `WhiteSpace` is used upstream (and in WPCS) only for generic whitespace sniffs which are not linked to a language structure which creates scope - with the emphasis on that last bit: _creates scope_ -.


> PHP seems to hold most of our generic function-related sniffs. It also contains some sniffs that could just as easily go under CodeAnalysis, like YodaConditions and StrictComparisons.

 See my comments about this above.

> Those sniffs could probably be used for JS as well as PHP. I'm not sure that we really even include JS sniffing in our vision for WPCS, which makes the PHP category kind of redundant.

If I remember correctly, we do in some of the whitespace and file related sniffs.
Personally I'd be happy for WPCS to also cover JS things, as long as they aren't already covered by JSLint or JSHint.

> `Variables` contains only `VariableRestrictions` (which is deprecated), and `GlobalVariables`. Since the latter is specifically related to WordPress globals, it might be reasonable to move it to WP and get rid of this category entirely. Other variable-related sniffs seem to fall into other categories already (like NamingConventions).

Upstream there is no `Variables` category. I'd be in favour of moving the `GlobalVariables` sniff to the `WP` category and maybe even to rename it to `GlobalVariablesOverwrite` to make it more explicit what the sniff detects.
The category could then be removed in the drop 2.x support/remove all deprecations release in summer 2018.

JDGrimes:
@jrfnl All of that makes a lot of sense.

> Upstream there is no Variables category. I'd be in favour of moving the GlobalVariables sniff to the WP category and maybe even to rename it to GlobalVariablesOverwrite to make it more explicit what the sniff detects.
> The category could then be removed in the drop 2.x support/remove all deprecations release in summer 2018.

I'd be in favor of this as well.

jrfnl:
I've added an additional commit to handle the rename of the `Variables.GlobalVariables` sniff to `WP.GlobalVariablesOverride`.

Using `Override` instead of `Overwrite` for consistency as that's what's been used elsewhere in the sniff.

GaryJones:
@jrfnl Good to merge from me.

jrfnl:
As this has now been merged, I'd like to change the wiki as well. I will annotate that the change in sniff names happened in version `1.0.0` as discussed in #1247 and I will change the milestone name from `0.15.0` to `1.0.0`

GaryJones:
I'm getting the following on `dev-develop` branch:

>  1 | WARNING | [ ] The "WordPress.XSS.EscapeOutput" sniff has been renamed to "WordPress.XSS.EscapeOutput". Please update your custom ruleset. (WordPress.XSS.EscapeOutput.FoundPropertyForDeprecatedSniff)
>    1 | WARNING | [ ] The "WordPress.CSRF.NonceVerification" sniff has been renamed to "WordPress.Security.NonceVerification". Please update your custom ruleset. (WordPress.CSRF.NonceVerification.FoundPropertyForDeprecatedSniff)

The second one seems sensible (though I'm not sure why it's appearing since my `phpcs.xml.dist` simply includes `WordPress` and excludes `WordPress-VIP` i.e. nothing referenced by name), but the first one seems like a wrong message.

jrfnl:
@GaryJones Good catch! The message text should be fixed via #1263.

As for why the messages are showing... not sure... Could you point me to the ruleset so I can run some tests ?

GaryJones:
@jrfnl My PR has just been merged, so see https://github.com/NicktheGeek/genesis-featured-widget-amplified/blob/hotfix/0.9.1/.phpcs.xml.dist 

With the corrected messages:

screenshot 2017-12-20 19 26 20

GaryJones:
@jrfnl So, it seems to be this:

1. XSS.EscapeOutputSniff is loaded.
2. `XSS.EscapeOutput.DeprecatedSniff` has severity `0` [set](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/cda66cba6fab213b7a549499d125b6b278f60315/WordPress/ruleset.xml#L95-L97) in `WordPress/ruleset.xml`, so the [first warning](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/cda66cba6fab213b7a549499d125b6b278f60315/WordPress/Sniffs/XSS/EscapeOutputSniff.php#L59-L63) doesn't get reported. All good.
3. There is then a bunch of checks for to see whether any custom properties have been set - it checks the public custom properties against the protected properties that are used internally after the custom ones have been merged in. 
4. The merging in is done in [`mergeFunctionLists()`](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/cda66cba6fab213b7a549499d125b6b278f60315/WordPress/Sniffs/Security/EscapeOutputSniff.php#L462-L509), and this is [called](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/cda66cba6fab213b7a549499d125b6b278f60315/WordPress/Sniffs/Security/EscapeOutputSniff.php#L198) at the top of the `Security.EscapeOutputSniff`'s (parent) `process_token()` method.
5. However, this parent `process_token()` is not [called](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/cda66cba6fab213b7a549499d125b6b278f60315/WordPress/Sniffs/XSS/EscapeOutputSniff.php#L83) until _after_ those "public property vs protected property" checks are done i.e. on the first round, the checks are never going to pass.

A quick solution would be to add `$this->mergeFunctionLists();` to the top of the `process_token()` calls in deprecated Sniffs - that way, the custom properties have been merged with the internal protected property. Certainly, it seems to remove the warnings in my case, though I'm not sure if there is any side-effect of running `$this->mergeFunctionLists();` twice. If not, then that should be good to go.

screenshot 2017-12-20 20 47 30

jrfnl:
@GaryJones I'm been looking into this as well. The defaults for the `addedCustomproperties` sub-values are `null` and not `array()` which is why the compare fails.
I'm currently trying to figure out why I set those to `null` initially and whether they can safely be changed to `array()`. If so, that will be the solution.
I'm also trying to figure out why you only get this for these two sniffs and not for, for instance, the `ValidatedSanitizedInput` sniff which should also be affected.

> A quick solution would be to add $this->mergeFunctionLists(); to the top of the process_token() calls in deprecated Sniffs - that way, the custom properties have been merged with the internal protected property. 

That would also prevent the deprecation warnings from being thrown when the custom properties *have* been set using the old sniff names, which is exactly the situation for which we *want* to throw the messages, so that would not work.

jrfnl:
FYI: I've updated the wiki pages to use the new sniff names.

PR #1249, opened 2017-12-06
AssignmentInCondition: fix bug where nested parentheses were presumed to be related to a ternary

### LINKED ISSUE #1227: False-positive detection of assignment/condition
**Error Code:** `WordPress.CodeAnalysis.AssignmentInCondition.Found`

PHP 7, and WPCS as well as PHPCS latest stable.

----

Hey there, me again! 👋 🙂 

I just stumbled over some _interesting_ issue.

Consider the following code, to be found in `foo.php`:

```php
<?php
/**
 * Foo.
 *
 * @package foo
 */

( function () {
	$foo = 42;

	return 1 === 2 ? 'a' : 'b';
} );
```

The code is the best I could come up in trying to simplify my real code example, which is a lot more complex). Actually, the callback is being fed to some constructor, but just wrapping it with parentheses already does the trick (or bug).

When running `$ phpcs foo.php --standard=WordPress`, we get the following output:

```text
FILE: /path/to/foo.php
----------------------------------------------------------------------------------------------------
----------------------------------------------------------------------------------------------------
 9 | WARNING | Variable assignment found within a condition. Did you mean to do a comparison?
----------------------------------------------------------------------------------------------------

Time: 526ms; Memory: 8Mb
```

L9 is `$foo = 42;`, so where is the condition?

I played around with this, and all of the following _variations_ do not trigger the sniff:

```php
// No ternary.
( function () {
	$foo = 42;

	return 1 === 2;
} );

// No variable definition.
( function () {

	return 1 === 2 ? 'a' : 'b';
} );

// Store closure in a variable.
$foo = function () {
	$foo = 42;

	return 1 === 2 ? 'a' : 'b';
};

/** Don't put function in parentheses (which means I have to give it a name, of course). */
function foo() {
	$foo = 42;

	return 1 === 2 ? 'a' : 'b';
}

// Use the function body (as a block).
{
	$foo = 42;

	return 1 === 2 ? 'a' : 'b';
}
```

I am going to ignore the Sniff/message for now, but it would be great to have this fixed. :)

Thanks,
Thorsten

jrfnl:
For "assignment in condition" check with ternaries, only ternaries in parentheses are checked as it is neigh impossible to determine where the condition starts otherwise.
Obviously, for that to work, the sniff should only take parentheses into account which are actually related to the ternary.

Fixed now.

Fixes #1227

~~P.S.: there may be more "delimiters" which should be taken into consideration, not just the semicolon, but I couldn't come up with valid example code at this moment. (head in other things). Please let me know if you can think of some, including example code, and I'll adjust the sniff.~~

[Edit]:
The renewed commit now takes a number of additional delimiters into account.
It also adds a separate errorcode for assignments found in ternaries.

### DIFF
--- a/WordPress/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
+++ b/WordPress/Sniffs/CodeAnalysis/AssignmentInConditionSniff.php
@@ -137,9 +137,24 @@ class AssignmentInConditionSniff extends Sniff {
 				$opener = $this->tokens[ $prev ]['parenthesis_opener'];
 				$closer = $prev;
 			} elseif ( isset( $token['nested_parenthesis'] ) ) {
-				end( $token['nested_parenthesis'] );
+				$closer = end( $token['nested_parenthesis'] );
 				$opener = key( $token['nested_parenthesis'] );
-				$closer = $stackPtr;
+
+				$next_statement_closer = $this->phpcsFile->findEndOfStatement( $stackPtr, array( T_COLON, T_CLOSE_PARENTHESIS, T_CLOSE_SQUARE_BRACKET ) );
+				if ( false !== $next_statement_closer && $next_statement_closer < $closer ) {
+					// Parentheses are unrelated to the ternary.
+					return;
+				}
+
+				$prev_statement_closer = $this->phpcsFile->findStartOfStatement( $stackPtr, array( T_COLON, T_OPEN_PARENTHESIS, T_OPEN_SQUARE_BRACKET ) );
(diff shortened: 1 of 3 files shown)


jrfnl:
Hang on.... thought off some more examples. Will rework when I have a chance.

jrfnl:
Ok, so it took a little while before I could get back to this one, but all done now. The ternary check has been hardened against the false positives as previously reported as well as several other similar cases I came up with.
Additionally I've added a separate errorcode for assignments found in ternary conditions.

PR is ready for review now.

PR #1251, opened 2017-12-06
Fixable precision alignment warnings

shadyvb:
Makes precision alignment warnings fixable by `phpcbf`.

### DIFF
--- a/WordPress/Sniffs/WhiteSpace/PrecisionAlignmentSniff.php
+++ b/WordPress/Sniffs/WhiteSpace/PrecisionAlignmentSniff.php
@@ -115,10 +115,13 @@ class PrecisionAlignmentSniff extends Sniff {
 				continue;
 			}
 
-			$spaces = 0;
+			$spaces  = 0;
+			$fixable = false;
 			switch ( $this->tokens[ $i ]['type'] ) {
 				case 'T_WHITESPACE':
-					$spaces = ( $this->tokens[ $i ]['length'] % $this->tab_width );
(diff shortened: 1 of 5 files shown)


jrfnl:
@shadyvb There is a very good reason why there is no fixer for this sniff: the sniff can not distinguish between accidental misalignment and deliberate precision alignment.

In my opinion, this sniff should **not** contain an auto-fixer.

Please see the history of the sniff and the related tickets for more information.

Refs:
* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/commits/31f18a7168ecff69a738d7fa7687bb56902241ed/WordPress/Sniffs/WhiteSpace/PrecisionAlignmentSniff.php
* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1099

shadyvb:
@jrfnl I agree this will have a good percentage of false positives, but I was under the impression I can run the fixer interactively, which doesn't seem to be possible, unless I missed something obvious.

This is almost useless, except if you fix and then check everything manually once more, which might not be convenient for some ( mixing false positives with proper fixes ).

Closing down 👍

jrfnl:
@shadyvb Just checking: can the branch be deleted ?

Regarding running the fixer interactively: yes and no: - you can run phpcs (not phpcbf) interactively using the `-a` option, but that means it will stop after each file, report on what errors are in the file and ask you to re-run on the file (after you've manually fixed errors), continue to the next file or quit altogether.

AFAIK there is no option to individually approve each fix, other than by cherrypicking lines when you make the commit in git, but that has nothing to do with PHPCS 😒

PR #1256, opened 2017-12-07
Update README.md

ramonamorea:
Added the guide to set up WPCS when using Eclipse with XAMPP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@
     + [Sublime Text](#sublime-text)
     + [Atom](#atom)
     + [Visual Studio](#visual-studio)
+    + [Eclipse with XAMPP](#eclipse-with-xampp)
 * [Running your code through WPCS automatically using CI tools](#running-your-code-through-wpcs-automatically-using-ci-tools)
     + [Travis CI](#travis-ci)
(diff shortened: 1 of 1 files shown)


JDGrimes:
The [Travis failure](https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/313142388#L707-L759) appears to me to be a bug in PHPCS, and not any issue with WPCS or this PR. In that case, I suppose it should be reported upstream.

jrfnl:
Correct, the Travis error is unrelated. I've identified which upstream commit is causing it & have reported it there: https://github.com/squizlabs/PHP_CodeSniffer/commit/21315b25d9ecfd9adc07d7c2c552527f5578e586#r26132171

jrfnl:
The above mentioned issue should be fixed now. I've just restarted the job in Travis. Fingers crossed that the build now passes :crossed_fingers:

jrfnl:
Thank you @ramonamorea !

As a side-note for anyone who has some time to spare: the list with IDE specific instructions in the readme is pretty long by now, so maybe it's a good idea to move some more of these to separate wiki pages with just a reference to the wiki in the Readme.... ?

dingo-d:
I think a link to wiki page is a good solution.

PR #1260, opened 2017-12-14
PostsPerPage: split sniff

jrfnl:
As noted in PR #1242, there was one sniff rename (split) not covered by that PR: `PostsPerPage`. This PR takes care of that one.

Split this sniff into two: one in VIP checking for disabling of pagination, one in WP checking for high pagination limit.

The WP version now throws a `warning`. This is reset to an error for use by VIP via the VIP ruleset.

### DIFF
--- a/WordPress/Tests/WP/PostsPerPageUnitTest.inc
+++ b/WordPress/Tests/WP/PostsPerPageUnitTest.inc
@@ -1,0 +1,30 @@
+<?php
+
+$args = array(
+	'posts_per_page' => 999, // Bad.
+	'posts_per_page' => -1, // OK.
+	'posts_per_page' => 1, // OK.
+	'posts_per_page' => '1', // OK.
+);
+
+_query_posts( 'nopaging=true&posts_per_page=999' ); // Bad.
+_query_posts( 'numberposts=999' ); // Bad.
+
+$query_args['posts_per_page'] = 999; // Bad.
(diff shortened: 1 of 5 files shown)


WordPress/Tests/WP/PostsPerPageUnitTest.inc
```diff
@@ -0,0 +1,30 @@
+<?php
+
+$args = array(
+	'posts_per_page' => 999, // Bad.
+	'posts_per_page' => -1, // OK.
```
LC43: why is this OK? on VIP -1 was set as Bad

LC43: found a ticket for this: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1379

jrfnl: That ticket is not the right one. See ticket #1157.

The original sniff has been split into two different sniffs:
* A WP one which checks for high pagination limits
* A VIP one which checks for disabling of pagination,

The VIP one will be removed in WPCS 2.0.

LC43: thanks @jrfnl for the quick reply!

the WP doesn't check for -1, which is unlimited, in my opinion, worse than a high limit.
should it check also for it?

PR #1261, opened 2017-12-14
Add a couple of extra sniffs to `Extra`

jrfnl:
A couple of sniffs which were previously in VIP are also useful for `Extra`. The messages in these sniffs have - where appropriate - been downgraded to warnings.

This commit adds these sniffs to the `Extra` ruleset.

Note: There are three more candidates for this:
* `DB.DirectDatabaseQuery`
* `DB.SlowDBQuery`
* `Security.ValidatedSanitizedInput`

I've not included the first two (yet) as those may need some more discussion first.
The third one is not included as it throws a lot of noise messages and has quite some open issues surrounding it, which should be addressed first IMO.

### DIFF
--- a/WordPress-Extra/ruleset.xml
+++ b/WordPress-Extra/ruleset.xml
@@ -156,6 +156,11 @@
 		 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/919 -->
 	<rule ref="WordPress.Classes.ClassInstantiation"/>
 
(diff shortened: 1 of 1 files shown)


GaryJones:
It's funny, as when I was reviewing the renaming PR, I started to ask the question as to whether that meant that Extra would benefit now some of them are less VIP-specific, but I checked the code again and saw that they didn't.

jrfnl:
@GaryJones *grin* Some things are better pulled step-by-step, also to avoid conflicts. I prepared these PRs all in one go and am pulling them one after another. One more after this one.

PR #1264, opened 2017-12-21
New Security.SafeRedirect sniff to encourage safe redirects

JDGrimes:
Introduces a sniff that warns about the use of the `wp_redirect()` function, suggesting
`wp_safe_redirect()` instead, to avoid malicious redirects.

This is based on the warning currently in `VIP.RestrictedFunctions`, but
in a more generic form.

I've added the sniff to the `WordPress-Extra` ruleset.

### DIFF
--- a/WordPress-Extra/ruleset.xml
+++ b/WordPress-Extra/ruleset.xml
@@ -103,6 +103,9 @@
 
 	<rule ref="WordPress.Security.EscapeOutput"/>
 
+	<!-- Encourage use of wp_safe_redirect() to avoid open redirect vulnerabilities. -->
+	<rule ref="WordPress.Security.SafeRedirect"/>
+
 	<!-- Verify that a nonce check is done before using values in superglobals.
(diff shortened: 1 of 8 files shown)


WordPress-Extra/ruleset.xml
```diff
@@ -103,6 +103,9 @@
 
 	<rule ref="WordPress.Security.EscapeOutput"/>
 
+	<!-- Encourage use of wp_safe_redirect() to avoid open redirect vulnerabilities. -->
```
jrfnl: Please add the link to the issue being addressed by this sniff in the comment (as previously discussed and done elsewhere in the ruleset).

JDGrimes:
> The advice to use exit() after a redirect is not VIP specific, so would not be a bad thing to keep (until there is a dedicated sniff which checks for exit()/die() after a range of functions).

Yeah, I removed it thinking of the need for a separate sniff, but I guess it does make sense to keep it for now.



> Code duplication can be removed from the VIP.RestrictedFunctions sniff in favour of including this sniff in the VIP ruleset as well.

OK, wasn't sure about that. Will do.

PR #1267, opened 2017-12-26
WP.I18n: Fix false positive on method calls.

### LINKED ISSUE #1266: False positive for Missing $domain arg
I activated the text domain verification with:
```
	<properties>
		<property name="text_domain" type="array" value="polylang" />
	</properties>
</rule>
```
and excluded the `WordPress.WP.I18n.LowLevelTranslationFunction` sniff.

However the code
`$mo->translate( $string )`
generates an error: `Missing $domain arg. (WordPress.WP.I18n.MissingArgDomain)`.

I would expect this error for a missing domain in a call to `__()` or any high level I18n functions, but the `translate` method of the `Translations` WordPress class doesn't have any domain argument. See: https://github.com/WordPress/WordPress/blob/4.9.1/wp-includes/pomo/translations.php#L94-L98, so I don't expect that using a `translate` method with only one argument would generate such error.

I am using PHPCS 2.9.1 and WPCS 0.14.

jrfnl:
The `WP.I18n` sniff did not properly verify whether the `T_STRING` found was actually a function call, in contrast to a method call or a namespaced function call.

I've fixed this now by extending the `AbstractFunctionRestriction` sniff and using the logic contained therein to make sure that only plain function calls are being addressed.

Includes unit tests.

A future further improvement could be extending the `AbstractFunctionParameter` sniff instead to benefit from using the logic from the `get_function_call_parameters()` utility function.

For now, I've just made the function parenthesis check slightly more stable, by making it code style independent and adding an extra safeguard for live coding situations.

Fixes #1266

### DIFF
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -9,7 +9,7 @@
 
 namespace WordPress\Sniffs\WP;
 
-use WordPress\Sniff;
+use WordPress\AbstractFunctionRestrictionsSniff;
 use WordPress\PHPCSHelper;
 use PHP_CodeSniffer_Tokens as Tokens;
 
@@ -27,8 +27,11 @@ use PHP_CodeSniffer_Tokens as Tokens;
  *                   as a comma-delimited list.
  *                   `phpcs --runtime-set text_domain my-slug,default`
  * @since   0.13.0 Class name changed: this class is now namespaced.
+ * @since   1.0.0  This class now extends the AbstractFunctionRestrictionSniff.
+ *                 The parent `exclude` property is, however, disabled as it
+ *                 would disable the whole sniff.
  */
-class I18nSniff extends Sniff {
+class I18nSniff extends AbstractFunctionRestrictionsSniff {
 
 	/**
 	 * These Regexes copied from http://php.net/manual/en/function.sprintf.php#93552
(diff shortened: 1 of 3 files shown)


WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -185,21 +202,40 @@ public function process_token( $stack_ptr ) {
 			}
 		}
 
-		if ( '_' === $token['content'] ) {
+		// Prevent exclusion of the i18n group.
+		$this->exclude = '';
+
+		// Note: this will give false positives if a method called `_` is used.
+		if ( '_' === $this->tokens[ $stack_ptr ]['content'] ) {
```
JDGrimes: Is there a particular reason not to move this check as well, other than the fact that `_` isn't a part of the `i18n` group? Couldn't it be added as part of a second group and checked in the `process_matched_token()` method? That way it wouldn't catch false-positives from `_()` methods either. (Which admittedly seems unlikely, but I see no reason not to exclude that possibility.)

jrfnl: > Is there a particular reason not to move this check as well, other than the fact that _ isn't a part of the i18n group?

I didn't move for exactly that reason :grin: 

> Couldn't it be added as part of a second group and checked in the process_matched_token() method? That way it wouldn't catch false-positives from _() methods either.

I think there are more elegant ways to do this, I'll have a fiddle with it.

jrfnl: Reading your comment again, I now understand what you mean by "second group". Yes, that would actually work perfectly. Changing things now.

jrfnl: Done. Added second commit. Feel free to squash on merge.

I've also moved the other check which checks for a particular function down to below the parenthesis check as an extra safeguard against false positives.

PR #1268, opened 2017-12-26
Readme: move IDE setup instructions to the wiki

jrfnl:
Following up on https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1256#issuecomment-350631622

I've created wiki pages containing the information which is (was) in the readme about configuring [Sublime Text](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Setting-up-WPCS-to-work-in-Sublime-Text) and [Atom](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Setting-up-WPCS-to-work-in-Atom) for use with PHPCS & WPCS.

This PR removes that information from the Readme and adds links to the wiki pages in it's place.

It also simplifies the IDE list to a bullet list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,11 +16,7 @@
     + [Recommended additional rulesets](#recommended-additional-rulesets)
 * [How to use](#how-to-use)
     + [Command line](#command-line)
-    + [PhpStorm](#phpstorm)
-    + [Sublime Text](#sublime-text)
-    + [Atom](#atom)
-    + [Visual Studio](#visual-studio)
-    + [Eclipse with XAMPP](#eclipse-with-xampp)
+    + [Using PHPCS and WPCS from within your IDE](#using-phpcs-and-wpcs-from-within-your-ide)
 * [Running your code through WPCS automatically using CI tools](#running-your-code-through-wpcs-automatically-using-ci-tools)
     + [Travis CI](#travis-ci)
 * [Fixing errors or whitelisting them](#fixing-errors-or-whitelisting-them)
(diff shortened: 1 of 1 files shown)


jrfnl:
FYI: I've also updated the wiki homepage for the new pages + added a custom sidebar to the wiki

GaryJones:
What advantage does putting info in the wiki have compared to having it in the readme?

jrfnl:
@GaryJones These were my reasons for proposing the change:
* The Readme was getting pretty long with the various IDE configuration explanations which are only relevant to a smaller percentage of our users.
* Also, two of the IDE config instructions link to external resources, three were in the Readme with screenshots and all, which could give the wrong impression of a preference for certain IDEs. Having them all be links takes that impression away.
* This way information which is relevant to nearly all users is in the readme and the other information is in the wiki, but with links to it from the readme making it easily discoverable anyhow.

ntwb:
I think this makes sense, I'd thought about reformatting this section of the readme with updated screenshots that were all of a similar size to make readability a little prettier.

That said, the downside of GitHub Wiki's is discoverability and keeping up with any changes to the wiki, I find vandals aka spammers often target these on GitHub and "following" wiki changes is, for the most part, a manual process of occasionally checking the edit history manually /shrug

GaryJones:
> I find vandals aka spammers

We can limit the wiki to those with push access:

screenshot 2017-12-28 10 33 44

jrfnl:
> We can limit the wiki to those with push access:

@GaryJones Let's leave that for now. We've not had any trouble (yet) and this allows community contributions to the wiki, like we had this month.
If we do run into trouble, it's easy enough to revert a change to a wiki page and limit the editing rights then.

PR #1271, opened 2017-12-30
AlternativeFunctions: detect `strip_tags()` and advise using `wp_strip_all_tags()` instead

jrfnl:
The advantage of `wp_strip_all_tags()` over `strip_tags()` is that it does not leave the content of removed `<script>` and `<style>` tags behind.

> This differs from `strip_tags()` because it removes the contents of the `<script>` and `<style>` tags. E.g. `strip_tags( '<script>something</script>' )` will return ‘something’. `wp_strip_all_tags()` will return ”

Refs:
* https://developer.wordpress.org/reference/functions/wp_strip_all_tags/
* http://php.net/manual/en/function.strip-tags.php

### DIFF
--- a/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
+++ b/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
@@ -84,6 +84,14 @@ class AlternativeFunctionsSniff extends AbstractFunctionRestrictionsSniff {
 				),
 			),
 
+			'strip_tags' => array(
+				'type'      => 'warning',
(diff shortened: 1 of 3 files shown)


JDGrimes:
Another benefit of `wp_strip_all_tags()` is that it doesn't have the dangerous second parameter of `strip_tags()`, which can be used to whitelist certain tags to keep. The problem with that parameter is that when the tags are kept, so are any attributes they have, and thus scripts can be executed via `onclick` etc. In other words, `strip_tags()` by itself is not safe if the second parameter is used. But `wp_strip_all_tags()` is fine, since it doesn't support that feature.

PR #1274, opened 2018-01-03
Improve handling short echo open tag

jrfnl:
Apart from the `PrecisionAlignment` sniff (PR #1273), there were two more sniffs which did not handle the PHP short echo open tag, while they should.

This PR fixes that for both.

Includes unit tests.

**Notes:**
* Similar to #1273, no work-around has been created for PHP 5.3 with short open tags turned off.
    For the `EmptyStatement` sniff, the unit tests for short open echo tags have been put in a separate file as the `.fixed` file would not match on PHP 5.3 with short open tags turned off.
    Once PHP 5.3 support is dropped, the unit tests files can be merged together.
* I've reviewed all sniff which referred to the `T_OPEN_TAG` token. For the other sniffs which use that token, no adjustments were needed.


WordPress/Tests/Files/FileNameUnitTests/some-view.inc
```diff
@@ -0,0 +1 @@
+<div><?= 'text' ?></div>
```
GaryJones: This file appears to be the same as SomeView.inc, and I don't see any reference to it?

jrfnl: * The contents of the file *is* the same - both files are intended to trigger the `FileName` sniff while using only short open tags.
* The filename however is **not** the same. The first file `SomeView.inc` is supposed to trigger an error. This file is there to confirm that no error is triggered when the file name complies with the basic "_lowercase with dashes_" rule. It is not mentioned in the test case file as it's not supposed to trigger an error.
* As the testcase file gathers the files to be tested using a `glob()` call, it will, however, be tested.

jrfnl:
FYI: the travis failure has to do with the `Internal.NoCode` warning being thrown on PHP 5.3 in combination with `short_open_tag` being off. I've amended the commits to account for that now.

Issue #1239, opened 2017-11-24
PrefixAllGlobals: make independent of case naming conventions


jrfnl:
I think the `PrefixAllGlobals` sniff should be more code style independent, i.e. remove the need for an underscore between the prefix and the class/function/variable name.

Should the sniff really care about whether people use Camel_Caps, camelCase, snake_case or any other variant of naming conventions, as long as whatever name is used, is prefixed ?

Just wondering what the opinions are about this.

Related to #763

Issue #1236, opened 2017-11-21
PrefixAllGlobals/OverrideGlobals: detect non-prefixed variables when created in control structures


jrfnl:
The declaration of the new `$something` variable in code like the below is currently not detected by the `PrefixAllGlobals`, nor the `GlobalVariables` sniff while it does create a new non-prefixed variable and could easily dangerously override a WP native global.

```php
$prefix_something = array(...);
foreach ( $prefix_something as $something ) {
    // Do something.
}

while ( ( $something = function_call() ) === true ) {
	// Do something.
}
```

Along the same lines, but generally less problematic, the `$i` variable declaration in the code below is also not detected.
```php
for ( $i = 0; $i < 10; $i++ ) {
	// Do something.
}
```

Of course, this only applies when these constructs are placed directly in the global namespace.

JDGrimes:
Yes, I noticed this a while back and didn't end up getting an issue created. I believe that it is because of the check for whether or not the variable is within a condition of any kind. I guess it is intended to be directed at functions, but ends up picking up conditions as well.

jrfnl:
Just been running some tests. Looks like that from the examples above, it's only the `foreach` which is not being picked up.
The `while` and `for` both report an error already (providing they are in the global namespace).

PR #1276, opened 2018-01-04
PrefixAllGlobals: bug fix - don't trigger on closure parameter definition

### LINKED ISSUE #1275: PHPCS incorrectly thinks a function argument with a default value is a global variable
Take this example:

```php
<?php

add_shortcode( 'hello', function( $attrs, $content = null ) {

} );
```

PHPCS will give this error:

```
 3 | ERROR | Variables defined by a theme/plugin should start with the theme/plugin prefix. Found: "$content". (WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedVariableFound)
```

This error is incorrect. `$content` is not a global, therefore, it does not need to be prefixed with the theme/plugin prefix. I think that the fact that this is an anonymous function is throwing it off.

I'm using PHPCS 3.2.2, Wordpress-Coding-Standards 0.14.0, and PHP version 7.1.12.

jrfnl:
Fixes #1275

This PR also contains a minor fix to the unit test case files (in a separate commit):
* Custom properties should be reset at the end of the unit test file in which they are used.


jrfnl:
FYI: while I'm looking at this sniff now anyway, I'm preparing follow-on PRs to fix #1239 and #1236. Will pull each one in turn as the changes to the unit test case files would cause conflicts otherwise.

PR #1277, opened 2018-01-04
AlternativeFunctions: add warning about using PHP native rand generation

### LINKED ISSUE #1248: WP.AlternativeFunctions: suggest wp_rand()
Inspired by ticket https://core.trac.wordpress.org/ticket/42782: should we add `wp_rand()` to the `WP.AlternativeFunctions` sniff ?

Ref: https://developer.wordpress.org/reference/functions/wp_rand/

jrfnl:
Fixes #1248

I've included the discouraged random seed generation functions in this PR as:
* they are not needed when using `wp_rand()`
* they are discouraged for use in PHP anyway

Refs:
* https://developer.wordpress.org/reference/functions/wp_rand/
* http://php.net/manual/en/function.rand.php
* http://php.net/manual/en/function.mt-rand.php
* http://php.net/manual/en/function.srand.php
* http://php.net/manual/en/function.mt-srand.php

### DIFF
--- a/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
+++ b/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
@@ -92,6 +92,24 @@ class AlternativeFunctionsSniff extends AbstractFunctionRestrictionsSniff {
 				),
 			),
 
+			'rand_seeding' => array(
+				'type'      => 'warning',
+				'message'   => '%s() is discouraged. Rand seeding is not necessary when using the wp_rand() function (as you should).',
+				'functions' => array(
+					'srand',
+					'mt_srand',
+				),
+			),
+
+			'rand' => array(
+				'type'      => 'warning',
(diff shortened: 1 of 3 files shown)


WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -92,6 +92,24 @@ public function getGroups() {
...
+					'mt_srand',
+				),
+			),
+
+			'rand' => array(
+				'type'      => 'warning',
+				'message'   => '%s() is discouraged. Use the far less predictable wp_rand() instead.',
+				'functions' => array(
+					'rand',
+					'mt_rand'
```
JDGrimes: Missing comma here is causing the Travis build failure.

https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/325103528#L785-L794

jrfnl: Thanks, Fixed now.

PR #1279, opened 2018-01-04
PrefixAllGlobals: handle variables created by foreach

### LINKED ISSUE #1236: PrefixAllGlobals/OverrideGlobals: detect non-prefixed variables when created in control structures

jrfnl:
Variables created by control structures, such as `$b` in `foreach ( $a as $b )` should also be prefixed when in the global namespace.

Verified that variable assignments in other control structures, `if`, `while`, `for` are correctly detected already and added unit tests to make sure they will continue to be detected correctly.

~~Goto labels are a special kind of control structure and while these are restricted to the same file and context, I have not been able to find any confirmation that `goto` labels have a concept of "local" versus "global".~~

~~Therefore `goto` labels when used will always need to be prefixed too.~~

This PR takes care of this.

Fixes #1236

### DIFF
--- a/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
+++ b/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
@@ -435,10 +435,15 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	 *                  normal file processing.
 	 */
 	protected function process_variable_assignment( $stackPtr ) {
-
-		// We're only concerned with variables which are being defined.
-		// `is_assigment()` will not recognize property assignments, which is good in this case.
-		if ( false === $this->is_assignment( $stackPtr ) ) {
+		/*
+		 * We're only concerned with variables which are being defined.
+		 * `is_assigment()` will not recognize property assignments, which is good in this case.
+		 * However it will also not recognize $b in `foreach( $a as $b )` as an assignment, so
+		 * we need a separate check for that.
+		 */
+		if ( false === $this->is_assignment( $stackPtr )
+			&& false === $this->is_foreach_as( $stackPtr )
+		) {
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


JDGrimes:
> Goto labels are a special kind of control structure and while these are restricted to the same file and context, I have not been able to find any confirmation that goto labels have a concept of "local" versus "global".
> 
> Therefore goto labels when used will always need to be prefixed too.

I'm not convinced this is necessary. As far as I can tell, a `goto` label is to its context about the same thing that a method is to its class. Since we don't require class methods to be prefixed, I don't see the argument for prefixing `goto` labels. `goto` labels must be unique within their context, but I [they don't need to be unique across contexts](https://3v4l.org/Su6ud).

jrfnl:
@JDGrimes In that case, only goto label in the global scope would need to be checked, shouldn't be much of a problem to change the code for that.

It would be nice to find some official documentation about this though.

JDGrimes:
Even in "global scope," I don't think that it is actually in global scope, but file scope. Let me test and see.

jrfnl:
@JDGrimes I know you can't `goto` across files, but does the name not get registered in the global name table ?

JDGrimes:
File `a.php`:

```php
<?php

include 'b.php';

goto a;

a:
	echo "a() a\n";
```

File `b.php`:

```php
<?php

goto a;

a:
	echo "b() a\n";
```

```
$ php a.php
b() a
a() a
```

In other words, no, the name does not get registered in the global name table.

jrfnl:
@JDGrimes Ok, thanks for testing that. I will remove the commit dealing with `goto` and remove the unit tests from the test file.

jrfnl:
Fixed ;-)

PR #1284, opened 2018-01-09
WordPress-VIP: Ensure we use WordPress.com not WordPress.

philipjohn:
We should always refer to "WordPress.com VIP" and never "WordPress VIP".
This change updates a few places where wording was slightly wrong.

The name of the standard remains "WordPress-VIP" as that's probably too
ingrained in folks processes to change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ You can use the following as standard names when invoking `phpcs` to select snif
   - `WordPress-Docs` - additional ruleset for [WordPress inline documentation standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/)
   - `WordPress-Extra` - extended ruleset for recommended best practices, not sufficiently covered in the WordPress core coding standards
     - includes `WordPress-Core`
-  - `WordPress-VIP` - extended ruleset for [WordPress VIP.com coding requirements](http://vip.wordpress.com/documentation/code-review-what-we-look-for/)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -116,7 +116,7 @@ You can use the following as standard names when invoking `phpcs` to select snif
   - `WordPress-Docs` - additional ruleset for [WordPress inline documentation standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/)
   - `WordPress-Extra` - extended ruleset for recommended best practices, not sufficiently covered in the WordPress core coding standards
     - includes `WordPress-Core`
-  - `WordPress-VIP` - extended ruleset for [WordPress VIP coding requirements](http://vip.wordpress.com/documentation/code-review-what-we-look-for/)
+  - `WordPress-VIP` - extended ruleset for [WordPress VIP.com coding requirements](http://vip.wordpress.com/documentation/code-review-what-we-look-for/)
```
jrfnl: This correction looks incorrect ;-)

GaryJones:
Good catch. I'm +1 for adding the clarifications.

README.md
philipjohn: Damn it, I corrected that and must have not committed!

philipjohn:
> There are also numerous error messages which refer to VIP context or VIP platform. Error messages can be safely changed if deemed appropriate, so feel free to expand this PR if you think that clarification is needed in those.

When it's just "VIP" on it's own, that's probably fine. It's probably excessive to be having "WordPress.com" every single time :)

jrfnl:
Thanks @philipjohn !

PR #1295, opened 2018-02-03
Improve recognition of test classes

jrfnl:
Based on a non-GH report by @GaryJones .

This commit enhances the `Sniff::is_test_class()` method to handle namespaced test classes better.

* Will now correctly recognize whether a class is in a namespace and adjust the response based on that.
    Previously, namespaces classes where the classname was one of the whitelisted global classes would pass, even though they were not the same class.
    The same happened with extended classes.
* The other side of this was that a namespaced custom test class where the name was split between the namespace declaration and the extended class declaration was not correctly recognized as the whitelisted test class. Now it will be.
* Improves handling of FQN used for the extended class name, i.e. prefixed with a `\` to indicate global namespace.
* Adds tolerance for custom unit test classes being passed through a custom ruleset with a `\` prefix.

Included unit tests for all the above mentioned cases.
While the unit tests are all of the `WordPress.Files.FileName` sniff variety, the changes will benefit all sniffs which check for test classes, i.e. as the `PrefixAllGlobals` and the `GlobalVariablesOverride` sniffs.

**N.B.:** `use` statements are explicitly **not** examined for importing of the whitelisted test classes as with the variety of syntaxes, this would get very complicated very quickly.
After all, users can add an `<exclude-pattern>` for a particular sniff to their custom ruleset to selectively exclude the test directory.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -1233,6 +1233,7 @@ abstract class Sniff implements PHPCS_Sniff {
 	 *   or a custom whitelisted unit test class.
 	 *
 	 * @since 0.12.0 Split off from the `is_token_in_test_method()` method.
+	 * @since 1.0.0  Improved recognition of namespaced class names.
 	 *
 	 * @param int $stackPtr The position of the token to be examined.
 	 *                      This should be a class, anonymous class or trait token.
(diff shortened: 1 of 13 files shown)


jrfnl:
> Some files are showing on GH that they have no trailing linespace at the end of the files.

Good catch, I thought I'd fixed them all before sending in the PR, but clearly missed some. Fixed up (amended) the commit now.

>  These should presumably be fixed to pass internal PHPCS checks?

There is no internal PHPCS check for this. As the unit test case files are meant to have code which fails against sniffs, they are not checked for code style.

PR #1297, opened 2018-02-07
Fix `posts_per_page` sniff

justnorris:
This fixes https://github.com/Automattic/VIP-Coding-Standards/issues/29

phpcs is looking for `T_COMMA, T_SEMICOLON` for `$valEnd` - things like comments, spaces and closing parentheses are also included when there is no `T_COMMA`. 

For example, this is not reported:
`
$args = array( 'posts_per_page' => -1 ); // Bad
`

At first I tried to just append closing parentheses and closing square brackets to the tokens to look for, but that didn't work out because in some situations, with a missing comma - it would still include comments and spaces.

This approach is going to find a `$valEnd` (either semicolon or a comma) then try to step back until it finds a valid value (but not further back than `$valStart`). If a string/number/bool/constant is found it'll overwrite existing `$valEnd` with the backtracked position.


jrfnl:
@justnorris Thanks for the PR.

In my opinion, your PR as it is, unfortunately should not be accepted as it is making incorrect assumptions about the type of content which can be in an array value which is not appropriate for an abstract class which is intended to be used by more than just the sniff being addressed here.

I have to admit that I think a complete rewrite of that abstract class for more accurate token walking and code style independence is probably in order.

jrfnl:
@justnorris @tomjn Just checking - as the VIP sniffs are going to be deprecated in WPCS, can this PR be closed ?

tomjn:
Yes it can be closed, @justnorris we can build this into a sniff in our own repo

PR #1308, opened 2018-02-26
PHP.StrictInArray: don't throw an error for no parameters found

jrfnl:
Adds the provided code sample as a unit test.

Also adjusts a unit test in the deprecated function test case file, to make sure when a similar syntax is used with a deprecated function, that an error is correctly thrown.

Fixes 1306

### DIFF
--- a/WordPress/Sniffs/PHP/StrictInArraySniff.php
+++ b/WordPress/Sniffs/PHP/StrictInArraySniff.php
@@ -101,19 +101,4 @@ class StrictInArraySniff extends AbstractFunctionParameterSniff {
 		}
 	}
 
-	/**
-	 * Process the function if no parameters were found.
-	 *
-	 * @since 0.11.0
-	 *
-	 * @param int    $stackPtr        The position of the current token in the stack.
-	 * @param array  $group_name      The name of the group which was matched.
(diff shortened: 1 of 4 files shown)


WordPress/Sniffs/PHP/StrictInArraySniff.php
```diff
@@ -101,19 +101,4 @@ public function process_parameters( $stackPtr, $group_name, $matched_content, $p
...
-	 *
-	 * @since 0.11.0
-	 *
-	 * @param int    $stackPtr        The position of the current token in the stack.
-	 * @param array  $group_name      The name of the group which was matched.
-	 * @param string $matched_content The token content (function name) which was matched.
-	 *
-	 * @return void
-	 */
-	public function process_no_parameters( $stackPtr, $group_name, $matched_content ) {
```
GaryJones: What value was this providing before, that it no longer needed?

jrfnl: None really ;-) I think I added it because we could, not so much because it added real value.

We're not checking if other PHP/WP functions which expect parameters receive them and PHP will throw a warning when it receives too few parameters anyhow.

Aside from that, this check widens the scope of the sniff to more than its mandate.

PR #1349, opened 2018-05-07
Remove all functions from the auto escape list that don't return a value

### LINKED ISSUE #1348: Remove all functions that don't have a return value from list of auto escaped functions
I've been auditing the functions that WPCS lists as auto escaped to ensure that they are all actually safe.

There is a large number of functions present in the list which don't return anything (rather they echo their data) and therefore I don't believe they should be there. Most of the functions have been there from the start of WPCS.

PR coming up which demonstrates the list.

cc @westonruter as he introduced most of the original list at the beginning of WPCS and might have some insight.

johnbillion:
Fixes #1348


GaryJones:
Has a merge conflict.

NielsdeBlaauw:
I've reviewed the functions removed in the current pull request. I agree with all functions removed in 1ffc42776eb120da5783fdef8195afa7cc53b759
 
I've also gone through the list of remaining items in the 'autoEscapedFunctions' property.
I've found the follow functions that should be reconsidered for this pull-request:

- `bloginfo` - only echoes, no return
- `body_class` - only echoes, no return
- `get_search_query` - with improper usage this function might not escape
- `post_type_archive_title` - prefix does not actually seem to be escaped
- `single_cat_title` - prefix does not actually seem to be escaped
- `single_month_title` - prefix does not actually seem to be escaped
- `single_post_title` - prefix does not actually seem to be escaped
- `single_tag_title` - prefix does not actually seem to be escaped
- `the_date` - first parameter is not escaped, which might be vulnerable.
- `vip_powered_wpcom` - Seems to be exclusively part of VIP. Does this touch the https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1309 issue?

`wp_list_*`, `wp_nav_menu`, `wp_register` functions don't escape some parts, but those are on purpose. However, they don't filter out dangerous html properties.

jrfnl:
@johnbillion Thank you for the PR. Would you mind rebasing it on the current `develop` to make it mergable again ?

@NielsdeBlaauw Thank you for your detailed review & I look forward to your pull request for your other finds!

jrfnl:
@johnbillion Just checking - did you see my previous message asking to rebase ?

GaryJones:
Closing in favour of https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1547 since @johnbillion hasn't fixed the merge conflict.

GaryJones:
@NielsdeBlaauw Sorry - I missed your comment. If you want to open a PR with your recommended additional changes here, then we can take a look :-) Thanks!

PR #1357, opened 2018-05-16
Composer: add note about why the DealerDirect plugin is suggested

jrfnl:
Just suggesting a package does not explain to a user why it would be interesting for them to install it and I've already had question about the suggestion.

The `suggest` entry in the `composer.json` file has a value which is not actually intended for version constraints, but to add a hint to the user about why they should install the package.

Up to now, the value in the WPCS `composer.json` was used for a version constraint. While this is still a good idea to mention in this case as the plugin needs a certain version to be compatible with the same PHP versions WPCS supports, adding the hint on why the package is suggested should help users even more.

Ref: https://getcomposer.org/doc/04-schema.md#suggest

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
 		"squizlabs/php_codesniffer": "^2.9.0 || ^3.0.2"
 	},
 	"suggest" : {
-		"dealerdirect/phpcodesniffer-composer-installer": "^0.4.3"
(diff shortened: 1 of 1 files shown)


jrfnl:
Just checking: any particular reason why this wasn't merged with two approvals ? Was the build not finished yet or something ?

GaryJones:
I approved 4 minutes after JD it seems, but I hadn't spotted JDs approval (or it hadn't shown up), and so closed the tab.

jrfnl:
That explains it. Thanks!

PR #1359, opened 2018-05-19
Fix deprecated sniff warnings

### LINKED ISSUE #1325: Renamed sniff reported on line 1
Originally posted to https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1242#issuecomment-353124042

----

I'm getting the following when using the develop branch on a plugin's codebase:

1 | WARNING | [ ] The "WordPress.XSS.EscapeOutput" sniff has been renamed to "WordPress.XSS.EscapeOutput". Please update your custom ruleset. (WordPress.XSS.EscapeOutput.FoundPropertyForDeprecatedSniff)
1 | WARNING | [ ] The "WordPress.CSRF.NonceVerification" sniff has been renamed to "WordPress.Security.NonceVerification". Please update your custom ruleset. (WordPress.CSRF.NonceVerification.FoundPropertyForDeprecatedSniff)

The second one seems sensible (though I'm not sure why it's appearing since my phpcs.xml.dist simply includes WordPress and excludes WordPress-VIP i.e. nothing referenced by name), but the first one seems like a wrong message.

----

The message text was fixed in #1263, but not sure why the violations are appearing in the first place. I'm regularly using `^0.14` rather than `dev-develop` which I had been up until now.

jrfnl:
This PR actually fixes a number of things which I missed in the deprecation PR the first time around.

#### Deprecation messages: Simplify & improve code

The `addWarning()` and `addError()` methods return a boolean value signifying whether or not the notice has been thrown. This will in most cases be `true`, but will return `false` when the notice for whatever reason is being ignored.

Instead of hard-setting the `$this->thrown` value to true, let's use the return value instead so if the message is ignored for one file, it will still be thrown for the next.

#### Deprecation messages: fix messages being thrown when they shouldn't be

* Change the default values for the `$addedCustomFunctions` properties.
* Update the compares for the custom properties within the deprecated sniffs to prevent "deprecated property" messages being thrown unnecessarily.

Includes minor efficiency fix: No need to do the property compares if the message has already been thrown, so compare whether the message has already been thrown first.

To test this:
1. Run with the `WordPresss` ruleset over a project and don't see the superfluous notices anymore.
2. Use a ruleset along the lines of the below and make sure that for the deprecated sniffs, the deprecation messages are actually being thrown.

```xml
<?xml version="1.0"?>
<ruleset name="Test">

	<autoload>path/to/WordPress/PHPCSAliases.php</autoload>

	<rule ref="WordPress-Core"/>

	<rule ref="WordPress.WP.PreparedSQL">
		<type>warning</type>
	</rule>

	<rule ref="WordPress.Functions.DontExtract">
		<type>warning</type>
	</rule>

	<rule ref="WordPress.XSS.EscapeOutput">
		<type>warning</type>
	</rule>
</ruleset>
```

#### VIP ruleset: prevent deprecation messages being thrown when sniffs are not explicitly included through a custom ruleset

This was already done for the `WordPress` ruleset, but not for the `WordPress-VIP` ruleset.

#### Fix minor whitespace issue in test file 

Fixes #1325

### DIFF
--- a/WordPress-VIP/ruleset.xml
+++ b/WordPress-VIP/ruleset.xml
@@ -124,4 +124,101 @@
 	<!-- https://vip.wordpress.com/documentation/vip/code-review-what-we-look-for/#use-wp_safe_redirect-instead-of-wp_redirect -->
 	<rule ref="WordPress.Security.SafeRedirect"/>
 
+	<!--
+	#############################################################################
+	Account for deprecated sniffs.
+	These directives can be removed when the deprecated sniffs are removed.
+	#############################################################################
+	-->
+
+	<!-- Prevent deprecation notice when the sniff is not explicitely included. -->
+	<rule ref="WordPress.VIP.DirectDatabaseQuery.DeprecatedSniff">
+		<severity>0</severity>
+	</rule>
+
+	<!-- Prevent duplicate messages from deprecated sniff. -->
+	<rule ref="WordPress.VIP.DirectDatabaseQuery.SchemaChange">
+		<severity>0</severity>
+	</rule>
+	<rule ref="WordPress.VIP.DirectDatabaseQuery.DirectQuery">
+		<severity>0</severity>
+	</rule>
+	<rule ref="WordPress.VIP.DirectDatabaseQuery.NoCaching">
+		<severity>0</severity>
+	</rule>
+
+	<!-- Prevent deprecation notice when the sniff is not explicitely included. -->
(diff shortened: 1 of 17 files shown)


jrfnl:
@GaryJones @JDGrimes Just wondering if either of you saw I pulled this or if the GH mail/other notifications system was broken when I did.

GaryJones:
Just not had time to look through it, sorry.

Client deadlines :-)

jrfnl:
No worries, I was just wondering. Take your time.

PR #1364, opened 2018-05-31
User capabilities should be used not roles nor deprecated capabilities

grappler:
This comes from https://github.com/WPTRT/WordPress-Coding-Standards/pull/36 This was initially worked on for the theme review ruleset but we realised it would be better for WPCS.

I kept the first commit to give props where props are due.


WordPress/Tests/WP/UseCapabilitiesNotRolesUnitTest.inc
```diff
@@ -0,0 +1,41 @@
+<?php
+
```
GaryJones: Why the extra lines here?

grappler: They were there from the original commit and I was too lazy to remove them and update the warning and error list. 🙃

WordPress/Sniffs/WP/UseCapabilitiesNotRolesSniff.php
```diff
@@ -0,0 +1,234 @@
...
+	 * The group name for this group of functions.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @var string
+	 */
+	protected $group_name = 'caps_not_roles';
+
+	/**
+	 * List of functions that accept roles and capabilities as an agrument.
```
jrfnl: `agrument` => `argument`

WordPress/Sniffs/WP/UseCapabilitiesNotRolesSniff.php
```diff
@@ -0,0 +1,234 @@
...
+	 *
+	 * The number represents the position in the function call
+	 * passed variables, here the capability is to be listed.
```
jrfnl: `The number represents the position in the function call passed variables, here the capability is to be listed.`

:point_up:  This sentence seems a bit wonky.

Did you mean: `The number represents the parameter position in the function call, where the capability is supposed to be listed.` ?

WordPress/Sniffs/WP/UseCapabilitiesNotRolesSniff.php
```diff
@@ -0,0 +1,234 @@
...
+	 * The list is sorted alphabetically.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @var array Function name with parameter position.
+	 */
+	protected $target_functions = array(
```
jrfnl: Even though it's disabled for new installs, there is also the `add_links_page()` function.

WordPress/Sniffs/WP/UseCapabilitiesNotRolesSniff.php
```diff
@@ -0,0 +1,234 @@
...
+		}
+
+		$matched_parameter = $this->strip_quotes( $parameters[ $position ]['raw'] );
+		if ( isset( $this->core_capabilities[ $matched_parameter ] ) ) {
+			return;
+		}
+
+		if ( isset( $this->core_roles[ $matched_parameter ] ) ) {
+			$this->phpcsFile->addError(
+				'Capabilities should be used instead of roles. Found "%s" in function "%s"',
```
jrfnl: I'd either add parentheses to the function name, as in `Found "%s" in call to function "%s()"` or would just leave the function name part away `Found "%s"`. The line number and the parameter content should be clear enough.

WordPress/Sniffs/WP/UseCapabilitiesNotRolesSniff.php
```diff
@@ -0,0 +1,234 @@
...
+				$stackPtr,
```
jrfnl: As these function are often formatted as multi-line function call, I'd suggest changing this to `$parameter['start']` so it will point to the line where the capability was found on.

```php
    add_menu_page(
        __( 'Custom Menu Title', 'textdomain' ),
        'custom menu',
        'manage_options',
        'myplugin/myplugin-admin.php',
        '',
        plugins_url( 'myplugin/images/icon.png' ),
        6
    );
```

WordPress/Sniffs/WP/UseCapabilitiesNotRolesSniff.php
```diff
@@ -0,0 +1,234 @@
...
+		'edit_private_pages'          => true,
+		'read_private_pages'          => true,
+		'edit_published_posts'        => true,
+		'upload_files'                => true,
+		'publish_posts'               => true,
+		'delete_published_posts'      => true,
+		'edit_posts'                  => true,
+		'delete_posts'                => true,
+		'read'                        => true,
+		'level_10'                    => true,
```
jrfnl: If I remember correctly, the use of "level_xx" capabilities was deprecated ages ago. Should these be put in a separate list and warned against when found ?

WordPress/Sniffs/WP/UseCapabilitiesNotRolesSniff.php
```diff
@@ -0,0 +1,234 @@
...
+				$stackPtr,
+				'RoleFound',
+				array(
+					$matched_parameter,
+					$matched_content,
+				)
+			);
+		} else {
+			$this->phpcsFile->addWarning(
+				'"%s" is an unknown role or capability. Check the "%s()" function call to ensure it is a capability and not a role.',
```
jrfnl: Be aware that this will give a lot of false positives/unnecessary warnings as the `$capability` a plugin needs is often defined in a class property/(class) constant and passed to the function that way.

As I can imagine that this check is useful for the Theme Review Team, maybe have it disabled by default with the ability to enable it via a custom property ?

```php
class My_Plugin {
    const CAPABILITY = 'manage_options';

    public function do_something() {
        if ( current_user_can( self::CAPABILITY ) ) {}
    }
}
```

WordPress/Tests/WP/UseCapabilitiesNotRolesUnitTest.inc
```diff
@@ -0,0 +1,41 @@
...
+add_options_page( $pagetitle , $menu_title , $xxx , 'menu_slug' , 'function' ); // Warning.
+add_menu_page( $pagetitle , 'menu_title' , 'xxx' , 'handle' , 'function' , 'icon_url' ); // Warning.
+add_utility_page(
+ 'page_title' , 'menu_title' , 'super_admin' , 'menu_slug' , 'function', 'icon_url'  // Error.
+);
+add_submenu_page(
+  'parent_slug' ,
+  'page_title' ,
+  'menu_title' ,
+  $varible , // Warning.
```
jrfnl: `$varible` => `$variable`

WordPress/Sniffs/WP/UseCapabilitiesNotRolesSniff.php
```diff
@@ -0,0 +1,234 @@
...
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+
+		$position = $this->target_functions[ $matched_content ];
+		if ( ! isset( $parameters[ $position ] ) ) {
+			return;
+		}
+
+		$matched_parameter = $this->strip_quotes( $parameters[ $position ]['raw'] );
```
jrfnl: This will only work when the `$capability` is passed as a single text string.

Note: If the warning is to be removed/disabled by default (see my other remark), this should be fine to leave as is.

WordPress/Sniffs/WP/UseCapabilitiesNotRolesSniff.php
```diff
@@ -0,0 +1,234 @@
...
+	 */
+	protected $group_name = 'caps_not_roles';
+
+	/**
+	 * List of functions that accept roles and capabilities as an agrument.
+	 *
+	 * The number represents the position in the function call
+	 * passed variables, here the capability is to be listed.
+	 * The list is sorted alphabetically.
+	 *
```
jrfnl: Maybe mention that these functions are defined in `wp-admin/includes/plugin.php` to make live easier for updating the list at a later point time.

jrfnl:
Oh and another thing: the sniff should probably be added to a ruleset ;-) I'd suggest adding it to the `Extra` ruleset initially.

WordPress/Sniffs/WP/UseCapabilitiesNotRolesSniff.php
```diff
@@ -14,14 +14,30 @@
...
 class UseCapabilitiesNotRolesSniff extends AbstractFunctionParameterSniff {
 
+	/**
+	 * Only check for known capabilites.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @var boolean
+	 */
+	public $checkOnlyKnownCaps = true;
```
jrfnl: Please use `$snake_case` property names here and below.

FYI: the WPCS native CS check does not check for this as often upstream properties are used/overruled, but (new) WPCS native properties should be snake_case.

WordPress/Tests/WP/UseCapabilitiesNotRolesUnitTest.inc
```diff
@@ -0,0 +1,42 @@
...
+add_posts_page( 'page_title', 'menu_title', 'administrator', 'menu_slug', 'function' ); // Error.
+add_media_page( 'page_title', 'menu_title', 'editor', 'menu_slug', 'function' ); // Error.
+add_pages_page( 'page_title', 'menu_title', 'author', 'menu_slug', 'function' ); // Error.
+add_comments_page( 'page_title', 'menu_title', 'contributor', 'menu_slug', 'function' ); // Error.
+add_theme_page( 'page_title', $menu_title , 'subscriber', 'menu_slug', 'function' );  // Error.
+add_plugins_page( 'page_title', 'menu_title', 'super_admin', 'menu_slug', 'function' );  // Error.
+add_users_page( 'page_title', 'menu_title', 'administrator', 'menu_slug', 'function' );  // Error.
+add_management_page( 'page_title', 'menu_title', 'editor', 'menu_slug', 'function' );  // Error.
+add_options_page( 'page_title', 'menu_title', 'contributor', 'menu_slug', 'function' );  // Error.
+// @codingStandardsChangeSetting WordPress.WP.UseCapabilitiesNotRoles checkOnlyKnownCaps false
```
jrfnl: Please add a blank line above this to make it easier to see that a property is being changed and something else is now being tested.

jrfnl: Also: please _reset_ the property at the end of the block of tests so additional unit tests which may be added later can expect the default properties to be in effect.

WordPress/Tests/WP/UseCapabilitiesNotRolesUnitTest.inc
jrfnl: You may as well remove them as this is a new sniff.

WordPress/Sniffs/WP/UseCapabilitiesNotRolesSniff.php
grappler: I thought the notice was being shown in the correct line as I had a similar example in the unit test. As I was double checking to reply here I realised it was not the case. It has been fixed now.

jrfnl:
@grappler Hmm.. weird, I can't seem to respond to "outdated" comments anymore... GH playing up.

>  I have moved them to a separate list but I feel that the deprecation notice should be in a separate sniff.

Could you explain why you think this should be a separate sniff ?

> I kept them in this sniff as they are still valid regardless if they are deprecated.

If there is not going to be a warning for the deprecated levels, than having the separate list & doing the `array_merge()` is just overhead.

> I thought the notice was being shown in the correct line as I had a similar example in the unit test. As I was double checking to reply here I realised it was not the case. It has been fixed now.

I understand why you added the extra logic to determine the line as the `start` may be on the previous line, however, in that case, why not just use `$parameter['end']` ?

If you do think the extra logic has value, it will need to be changed to take comments and annotations into account, i.e. use `Tokens::$emptyTokens` not `T_WHITESPACE`.

Other than that, my previous remark about the `get_all_capabilities()` method preventing unit tests with the `$custom_capabilities` from working still stands, just as the fact that that functionality *does* need unit tests.

And the PHP 5.3 issue also still needs attention.

WordPress/Tests/WP/UseCapabilitiesNotRolesUnitTest.inc
```diff
@@ -0,0 +1,47 @@
...
+);
+if ( author_can( $post, $capability ) ) { } // Warning.
+if ( author_can( $post, 'FFFFF' ) ) { }  // Warning.
+if ( current_user_can( 'SSSSS' ) ) { } // Warning.
+if( current_user_can_for_blog( '3', 'EEEEEE' ) ) { }  // Warning.
+// @codingStandardsChangeSetting WordPress.WP.UseCapabilitiesNotRoles check_only_known_caps true
+
+// @codingStandardsChangeSetting WordPress.WP.UseCapabilitiesNotRoles custom_capabilities custom_cap
+if ( author_can( $post, 'custom_cap' ) ) { } // OK.
+if ( author_can( $post, 'read' ) ) { } // OK.
```
jrfnl: Maybe also add:
```php
// @codingStandardsChangeSetting WordPress.WP.UseCapabilitiesNotRoles check_only_known_caps false
if ( author_can( $post, 'unknown_custom_cap' ) ) { } // Warning.
// @codingStandardsChangeSetting WordPress.WP.UseCapabilitiesNotRoles check_only_known_caps true
```
... to test that the warning is still thrown correctly even when custom capabilities have been registered.

A unit test with several custom caps would also not be a bad thing. Doesn't have to be a separate test, you could just change the settings line here (use a comma-delimited list without spaces) and test both custom caps.

grappler:
OK, I will rename the sniff to `WordPress.WP.Capabilities` and add a deprecated notice.

With two of my Unit Test additions `$matched_parameter` as `'super_admin' // Error.` and `'super_admin' /* Error */`. I see two ways of getting around this. One to use regex somehow or the other to strip PHP comments out of the strings. What do you think?

jrfnl:
> With two of my Unit Test additions $matched_parameter as 'super_admin' // Error. and 'super_admin' /* Error */. I see two ways of getting around this. One to use regex somehow or the other to strip PHP comments out of the strings. What do you think?

That's why walking the parameter tokens instead of using `raw` to test for a match is generally the way to go.
If you do a for loop from `start` to `end` and disregard any empty tokens, concatenate all text string tokens and have a toggle for "unexpected tokens" (variables, constants), both this as well as silly things like: `'super' . '_admin'` can be detected reliably.

grappler:
> That's why walking the parameter tokens instead of using raw to test for a match is generally the way to go.

It ended up being such an easy fix 😄 

I think I have fixed all of the comments.

jrfnl:
I've added a commit to this PR to allow the builds to pass. This PR identified a fixer conflict which is why the build was failing. The fixer conflict involved two upstream sniffs.

I've opened PR https://github.com/squizlabs/PHP_CodeSniffer/pull/2056 to fix the fixer conflict.

For now, the unit test case file from this PR will be ignored for the fixer conflict check. Once the upstream PR has been merged, my commit should be reverted.

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,301 @@
...
+		}
+
+		$next_not_empty = $this->phpcsFile->findNext(
+			Tokens::$emptyTokens,
+			$parameters[ $position ]['start'],
+			$parameters[ $position ]['end'] + 1,
+			true
+		);
+
+		$matched_parameter = $this->strip_quotes( $this->tokens[ $next_not_empty ]['content'] );
```
jrfnl: Token walking means doing a `for()` loop between the param start and end. You are now just checking the first token and not even checking that this is a text string token....

This will break on all of the following examples (silly code, but you get my point):
```php
if ( author_can( $post, editor() ) ) { } // Should not error for using role.
if ( author_can( $post, 'super' . '_admin' ) ) { } // Should give error for using role.

// @codingStandardsChangeSetting WordPress.WP.Capabilities check_only_known_caps false
if ( author_can( $post, editor() ) ) { } // Should give warning about unknown capability, not error for using role.
if ( author_can( $post, install_plugins() ) ) { } // Should give warning about unknown capability.
if ( author_can( $post, 'install_' . 'plugins' ) ) { } // OK.
if ( author_can( $post, 'install_plugins' . $something ) ) { } // Should give warning about unknown capability.
if ( author_can( $post, 'install_plugins' . '_for_theme' ) ) { } // Should give warning about unknown capability.
// @codingStandardsChangeSetting WordPress.WP.Capabilities check_only_known_caps true
```

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,301 @@
...
+			return;
+		}
+
+		$custom_capabilities = $this->merge_custom_array( $this->custom_capabilities, array(), true );
+		if ( isset( $custom_capabilities[ $matched_parameter ] ) ) {
+			return;
+		}
+
+		if ( isset( $this->deprecated_capabilities[ $matched_parameter ] ) ) {
+			$this->phpcsFile->addError(
```
jrfnl: **_[No action needed, just noting this down for the future]_**

As the capability is only deprecated, should this be an `error` or a `warning` ?

Based on the logic used in other sniffs which check for deprecations, it should be an `error` as the version in which the capability was deprecated is more than three WP releases previous to the current release.

However, that would mean that a check should be done against the `$minimum_supported_version` of WP and that the deprecation version should be added to the `$deprecated_capabilities` array as the value for each item.

As all currently deprecated capabilities were deprecated in WP 3.0.0, I'm happy to leave this as is for now. As soon as other capabilities would become deprecated, this would need to be adjusted though.

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,301 @@
...
+				'The capatibility "%s" found in function call "%s()" has been deprecated since WordPress version 3.0.',
```
jrfnl: `capatibility` -> `capability`

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,301 @@
...
+				$next_not_empty,
+				'DeprecatedCapability',
```
jrfnl: Not a blocker, but the error code could just be `Deprecated` as `Capability` is already in the sniff name, so the full error code would be `WordPress.WP.Capability.Deprecated`.

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,301 @@
...
+				)
+			);
+			return;
+		}
+
+		if ( false === $this->check_only_known_caps ) {
+			$this->phpcsFile->addWarning(
+				'"%s" is an unknown role or capability. Check the "%s()" function call to ensure it is a capability and not a role.',
+				$next_not_empty,
+				'UnknownCapabilityFound',
```
jrfnl: Not a blocker, but the error code could just be `Unknown` as `Capability` is already in the sniff name, so the full error code would be `WordPress.WP.Capability.Unknown`.

WordPress/Sniffs/WP/CapabilitiesSniff.php
grappler: Would it make sense to have this code as a utility function? I think there are other classes that would profit from it. 

```php
// More extensive check to prevent people circumventing the sniff.
$text                 = '';
$found_variable_token = false;

for ( $i = $parameters[1]['start']; $i <= $parameters[1]['end']; $i++ ) {
    if ( isset( Tokens::$stringTokens[ $this->tokens[ $i ]['code'] ] ) === false ) {
        if ( T_VARIABLE === $this->tokens[ $i ]['code'] ) {
            $found_variable_token = true;
        }

        return;
    }

    $text .= $this->strip_quotes( $this->tokens[ $i ]['content'] );
}
```
Copied the code from https://github.com/WPTRT/WordPress-Coding-Standards/pull/26/

grappler:
Thank you for the review.

- _A test that nothing happens when the expected parameter is not set_. ✅
- _A test that no warning is being thrown for an unknown capability when $check_only_known_caps is true (default)_. ✅
- _Testing that the $custom_capabilities property works as expected when $check_only_known_caps is true (default). I.e. the current tests are within the check_only_known_caps false block, so only the combination is tested._ [I am not sure this is testable. There never will be an error as the custom capability is not a role nor an unknown cap.]
- _Also the unit tests on line 45 and 54 are testing the same thing_. [I wanted to make sure the error was the same after changing the setting.]

WordPress/Sniffs/WP/CapabilitiesSniff.php
jrfnl: I was thinking about that, but I don't think there has been much need for it before now in WPCS.

All the same, I'd be happy for this to become a utility function, however, the "is there a variable ?" check in this code snippet would probably be a different question whenever used. Also in the above sample, it is presumed only single/double quoted strings are used, not heredoc/nowdoc, so some changes would be needed and I'd guess that the utility should in that case just retrieve a string of all text string tokens found between `$start` and `$end`, not do the "is there are variable" check ?

P.S.: looking at the code now, I'm still going to make one more change to the WPTRT sniff this came from ;-)

WordPress/Sniffs/WP/CapabilitiesSniff.php
jrfnl: @grappler I saw you went ahead and made this change already anyway :two_hearts:

Two minor remarks about that:
* The use of the `Sniff::minimum_supported_version` property still needs to be annotated in the class docblock with a `@uses` tag.
* As, at this moment, there are no capabilities which are deprecated in the last three versions, this needs a dedicated set of unit tests like so:
```php
add_options_page( 'page_title', 'menu_title', 'level_10', 'menu_slug', 'function' ); // Error.
// @codingStandardsChangeSetting WordPress.WP.Capabilities minimum_supported_version 3.5
add_options_page( 'page_title', 'menu_title', 'level_10', 'menu_slug', 'function' ); // Error.
// @codingStandardsChangeSetting WordPress.WP.Capabilities minimum_supported_version 2.8
add_options_page( 'page_title', 'menu_title', 'level_10', 'menu_slug', 'function' ); // Warning.
// @codingStandardsChangeSetting WordPress.WP.Capabilities minimum_supported_version 4.6
```

WordPress/Sniffs/WP/CapabilitiesSniff.php
grappler: OH, I forgot that the code was checking for two things. When I mentioned the idea I was thinking about returning a string of all of the text string token.

jrfnl:
Further on https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1364#issuecomment-396014775 :

> I've added a commit to this PR to allow the builds to pass. This PR identified a fixer conflict which is why the build was failing. The fixer conflict involved two upstream sniffs.
> 
> I've opened PR squizlabs/PHP_CodeSniffer#2056 to fix the fixer conflict.
> 
> For now, the unit test case file from this PR will be ignored for the fixer conflict check. Once the upstream PR has been merged, my commit should be reverted.

The upstream PR has been merged, so my earlier commit to ignore this sniff for the fixer check can now be reverted.

jrfnl:
@grappler Any chance that this PR will get some final love anytime soon ? Would be great to be able to add it to WPCS!

jrfnl:
@grappler Just wondering if you'll have a chance to finish this off in the near future.

If you haven't got time or lost interest, please let us know and we'll see if we can find someone to take over.

PR #1365, opened 2018-06-03
New WhitespaceBeforePHPOpenTag sniff

jrfnl:
New sniff to detect whitespace before the first PHP open tag to prevent "headers already send" errors.

This sniff has been added to the `Core` ruleset.

Includes auto-fixer.
Includes unit tests.

N.B.: I've also tried to add a unit test with a unicode zero-width space, but haven't managed to get that working, so for now, I've dropped that test.

N.B.2: I've opened an issue upstream to see if there would be interest for this sniff in PHPCS itself - https://github.com/squizlabs/PHP_CodeSniffer/issues/2038

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -425,6 +425,7 @@
 	-->
 	<!-- Important to prevent issues with content being sent before headers. -->
 	<rule ref="Generic.Files.ByteOrderMark"/>
+	<rule ref="WordPress.PHP.WhitespaceBeforePHPOpenTag"/>
 
 	<!-- All line endings should be \n. -->
 	<rule ref="Generic.Files.LineEndings">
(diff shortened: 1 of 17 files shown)


jrfnl:
The upstream issue yielded that the `Squiz.WhiteSpace.SuperfluousWhiteSpace` sniff checks for this and as that sniff is included already in the `Core` ruleset, I'm closing this PR.

Note: the upstream sniff does not handle unicode whitespace found at the beginning of the file, but IMO, this should be fixed upstream.

GaryJones:
Can this PR branch be deleted?

jrfnl:
It can, was just waiting for a response upstream about the unicode. I'll be pulling the line I used here for that upsteam.

jrfnl:
Upstream PR has been submitted.

PR #1366, opened 2018-06-04
Check for deprecated parameter values

grappler:
Follow on from #826

Solves part of #576

Added to the `Extra` ruleset with the other deprecated sniffs. I have added checks for `do_action_deprecated()` and `apply_filters_deprecated()` as they work well in this sniff but I am not sure if they fit in correctly.

### DIFF
--- a/WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
+++ b/WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
@@ -1,0 +1,216 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPress\Sniffs\WP;
+
+use WordPress\AbstractFunctionParameterSniff;
+use PHP_CodeSniffer_Tokens as Tokens;
+
+/**
+ * Check for usage of deprecated parameter values in WP functions and provide alternative based on the parameter passed.
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   1.0.0
+ *
+ * @uses    \WordPress\Sniff::$minimum_supported_version
+ */
+class DeprecatedParameterValuesSniff extends AbstractFunctionParameterSniff {
+
+	/**
+	 * The group name for this group of functions.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @var string
+	 */
+	protected $group_name = 'wp_deprecated_parameter_values';
+
+	/**
+	 * Array of function, argument, and replacement function for deprecated argument.
+	 *
+	 * The list of deprecated parameter values can be found by
+	 * looking for `_deprecated_argument()`.
+	 * The list is sorted alphabetically by function name.
+	 * Last updated for WordPress 4.9.6.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @var array Multidimensional array with parameter details.
+	 *     $target_functions = array(
+	 *         (string) Function name. => array(
+	 *             (int) Target parameter position, 1-based. => array(
+	 *                 (string) Parameter value. => array(
+	 *                     'alt'     => (string) Suggested alternative.
+	 *                     'version' => (int) The WordPress version when deprecated.
+	 *                 )
+	 *             )
+	 *         )
+	 *     );
+	 */
+	protected $target_functions = array(
+		'add_settings_field' => array(
+			4 => array(
+				'misc' => array(
+					'alt'     => 'another settings group',
+					'version' => '3.0.0',
+				),
+				'privacy' => array(
+					'alt'     => 'another settings group',
+					'version' => '3.5.0',
+				),
+			),
+		),
+		'add_settings_section' => array(
+			4 => array(
+				'misc' => array(
+					'alt'     => 'another settings group',
+					'version' => '3.0.0',
+				),
+				'privacy' => array(
+					'alt'     => 'another settings group',
+					'version' => '3.5.0',
+				),
+			),
+		),
+		'bloginfo' => array(
+			1 => array(
+				'home' => array(
+					'alt'     => 'the "url" argument',
+					'version' => '2.2.0',
+				),
+				'siteurl' => array(
+					'alt'     => 'the "url" argument',
+					'version' => '2.2.0',
+				),
+				'text_direction' => array(
+					'alt'     => 'is_rtl()',
+					'version' => '2.2.0',
+				),
+			),
+		),
+		'get_bloginfo' => array(
+			1 => array(
+				'home' => array(
+					'alt'     => 'the "url" argument',
+					'version' => '2.2.0',
+				),
+				'siteurl' => array(
+					'alt'     => 'the "url" argument',
+					'version' => '2.2.0',
+				),
+				'text_direction' => array(
+					'alt'     => 'is_rtl()',
+					'version' => '2.2.0',
+				),
+			),
+		),
+		'register_setting' => array(
+			1 => array(
+				'misc' => array(
+					'alt'     => 'another settings group',
+					'version' => '3.0.0',
+				),
+				'privacy' => array(
+					'alt'     => 'another settings group',
+					'version' => '3.5.0',
+				),
+			),
+		),
+		'unregister_setting' => array(
+			1 => array(
+				'misc' => array(
+					'alt'     => 'another settings group',
+					'version' => '3.0.0',
+				),
+				'privacy' => array(
+					'alt'     => 'another settings group',
+					'version' => '3.5.0',
+				),
+			),
+		),
+	);
+
+	/**
+	 * Process the parameters of a matched function.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		$this->get_wp_version_from_cl();
+		$param_count = \count( $parameters );
+		foreach ( $this->target_functions[ $matched_content ] as $position => $parameter_args ) {
+
+			// Stop if the position is higher then the total number of parameters.
+			if ( $position > $param_count ) {
+				break;
+			}
+
+			$this->process_parameter( $matched_content, $parameters[ $position ], $parameter_args );
+		}
+	}
+
+	/**
+	 * Process the parameter of a matched function.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameter       Array with start and end token positon of the parameter.
+	 * @param array  $parameter_args  Array with alternative and WordPress deprecation version of the parameter.
+	 *
+	 * @return void
+	 */
+	protected function process_parameter( $matched_content, $parameter, $parameter_args ) {
+
+		$parameter_position = $this->phpcsFile->findNext(
+			Tokens::$emptyTokens,
+			$parameter['start'],
+			$parameter['end'] + 1,
+			true
+		);
+
+		if ( false === $parameter_position ) {
+			return;
+		}
+
+		$matched_parameter = $this->strip_quotes( $this->tokens[ $parameter_position ]['content'] );
+		if ( ! isset( $parameter_args[ $matched_parameter ] ) ) {
+			return;
+		}
+
+		$message = 'The parameter value "%s" has been deprecated since WordPress version %s.';
+		$data    = array(
+			$matched_parameter,
+			$parameter_args[ $matched_parameter ]['version'],
+		);
+
+		if ( ! empty( $parameter_args[ $matched_parameter ]['alt'] ) ) {
+			$message .= ' Use %s instead.';
+			$data[]   = $parameter_args[ $matched_parameter ]['alt'];
+		}
+
+		$is_error = version_compare( $parameter_args[ $matched_parameter ]['version'], $this->minimum_supported_version, '<' );
+		$this->addMessage(
+			$message,
+			$parameter_position,
+			$is_error,
+			$this->string_to_errorcode( 'Found' ),
+			$data
+		);
+	}
+
+}
--- a/WordPress/Tests/WP/DeprecatedParameterValuesUnitTest.inc
+++ b/WordPress/Tests/WP/DeprecatedParameterValuesUnitTest.inc
@@ -1,0 +1,29 @@
+<?php
+
+// All will give an ERROR.
+
+get_bloginfo( 'home' );
+get_bloginfo( 'siteurl' );
+get_bloginfo( "text_direction" );
+echo bloginfo( 'home' );
+echo bloginfo( "siteurl" );
+echo bloginfo( 'text_direction' );
+add_settings_field( '', '', '', 'misc' );
+add_settings_field( '', '', '', 'privacy' );
+add_settings_section( '', '', '', 'misc' );
+add_settings_section( '', '', 'privacy', 'privacy' );
+register_setting( 'misc' );
+register_setting( 'privacy' );
+unregister_setting( 'misc' );
+unregister_setting( 'privacy' );
+
+add_settings_field( '', '', '' ); // OK.
+
+// Dynamic values will not give a notice.
+$value = 'text_direction';
+get_bloginfo( $value );
+get_bloginfo( CONSTANT );
+echo bloginfo( function_name() );
+
+get_bloginfo( 'wpurl' ); // OK.
+add_settings_field( '', '', '', /* deliberately not passed */ ); // Valid for PHP 7.3+.
--- a/WordPress-Extra/ruleset.xml
+++ b/WordPress-Extra/ruleset.xml
@@ -119,6 +119,7 @@
 	<rule ref="WordPress.WP.DeprecatedFunctions"/>
 	<rule ref="WordPress.WP.DeprecatedClasses"/>
 	<rule ref="WordPress.WP.DeprecatedParameters"/>
+	<rule ref="WordPress.WP.DeprecatedParameterValues"/>
 	<rule ref="WordPress.WP.AlternativeFunctions"/>
 	<rule ref="WordPress.WP.DiscouragedConstants"/>
 	<rule ref="WordPress.WP.DiscouragedFunctions"/>
--- a/WordPress/Tests/WP/DeprecatedParameterValuesUnitTest.php
+++ b/WordPress/Tests/WP/DeprecatedParameterValuesUnitTest.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * Unit test class for WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPress\Tests\WP;
+
+use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
+
+/**
+ * Unit test class for the DeprecatedParameterValues sniff.
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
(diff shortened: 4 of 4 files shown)


WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,231 @@
...
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		$param_count = count( $parameters );
+		foreach ( $this->target_functions[ $matched_content ] as $position => $paramerter_args ) {
+
+			if ( $position > $param_count ) {
+				break;
+			}
+
+			if ( ! isset( $parameters[ $position ] ) ) {
```
jrfnl: Can you give me a code example explaining what this is testing ?

AFAICS, this could only ever happen if a) the `$position > $param_count` as is already tested above or b) when no parameter is passed for one of the preceding parameters which would always cause a parse error.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,231 @@
...
+	 * and then alphabetically.
+	 * Last updated for WordPress 4.9.6.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @var array Multidimensional array with parameter details.
+	 *    $target_functions = array(
+	 *        (string) Function name. => array(
+	 *            (int) Target parameter position, 1-based. => array(
+	 *                (string) Parameter value. => array(
```
jrfnl: This sub-array setup will only work if the expected type for the value passed is a text string token or an integer.

How is this sniff going to handle deprecated parameter values of another type ? Like booleans, floats etc.

N.B.: the format of the complete array is "free" with the only requirement being the function names being the top-level keys. You don't have to follow the examples set by other sniffs for the array format if it's not the best format for this sniff.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,231 @@
...
+			if ( $position > $param_count ) {
+				break;
+			}
+
+			if ( ! isset( $parameters[ $position ] ) ) {
+				continue;
+			}
+
+			$matched_parameter    = $this->strip_quotes( $parameters[ $position ]['raw'] );
+			$is_dynamic_parameter = $this->phpcsFile->findNext(
```
jrfnl: What are you trying to do here ?

The above sub-array only allows for parameter values which are `T_CONSTANT_ENCAPSED_STRING` or `T_LNUMBER`.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,231 @@
...
+
+			$message        = '%s( %s ) was found.';
+			$stack_position = $stackPtr;
+			$data           = array(
+				$matched_content,
+				$parameters[ $position ]['raw'],
+			);
+
+			if ( $is_dynamic_parameter ) {
+				$message       .= ' The "%s" parameter is dynamic. Please check the value manually.';
```
jrfnl: I don't think this error belongs in WPCS. This will throw up *way* too many false positives and only cause annoyance.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,231 @@
...
+				$stack_position = $is_dynamic_parameter;
+				$data[]         = $matched_parameter;
+			} else {
+				$message       .= ' The "%s" parameter has been deprecated since WordPress version %s.';
+				$stack_position = $this->phpcsFile->findNext(
```
jrfnl: What are you trying to do here ? Find the start of the param ? If so, use a negative check against `Tokens::$emptyTokens`.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,231 @@
...
+				$data[]         = $matched_parameter;
+				$data[]         = $paramerter_args[ $matched_parameter ]['version'];
+			}
+
+			if ( ! empty( $paramerter_args[ $matched_parameter ]['alt'] ) ) {
+				$message .= ' Use %s instead.';
+				$data[]   = $paramerter_args[ $matched_parameter ]['alt'];
+			}
+
+			if ( $is_dynamic_parameter || version_compare( $paramerter_args[ $matched_parameter ]['version'], $this->minimum_supported_version, '>=' ) ) {
```
jrfnl: I would much prefer the `$error_type` to be set when the message is set instead of having a condition like this at the end of the function.
You can then use the `$this->addMessage()` function to throw the error.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,231 @@
...
+				$this->phpcsFile->addWarning( $message, $stack_position, $matched_content . 'Found', $data );
```
jrfnl: Couple of remarks about the error code:
1. If the message about dynamic parameters would stay, it should have a different error code.
2. The `$matched_content` for the error code should be wrapped in a call `$this->string_to_errorcode()`, though having it as is doesn't actually make the error code properly modular as, based on the above array, most of the time, it would still cover two or three different deprecated values.
    I'm not saying the code _should_ be modular, but if you want it to be, then it should be properly modular.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,231 @@
...
+				$stack_position = $this->phpcsFile->findNext(
+					array( T_CONSTANT_ENCAPSED_STRING, T_ARRAY, T_FALSE, T_TRUE, T_NULL, T_LNUMBER ),
+					$parameters[ $position ]['start'],
+					$parameters[ $position ]['end'] + 1
+				);
+				$data[]         = $matched_parameter;
+				$data[]         = $paramerter_args[ $matched_parameter ]['version'];
+			}
+
+			if ( ! empty( $paramerter_args[ $matched_parameter ]['alt'] ) ) {
```
jrfnl: Do we know anything about when the `alt`ernative was introduced ? If so, should that info be added to the array and the alternative only displayed when the `minimum_supported_version` actually allows for it ?

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
jrfnl: If you want to use `$this->minimum_supported_version`, you need to call `$this->get_wp_version_from_cl();` somewhere before using it as otherwise passing the minimum supported WP version via the command-line will not work.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,231 @@
...
+		),
+		'do_action_deprecated' => array(
+			1 => array(
+				'refresh_blog_details' => array(
+					'alt'     => 'the action "clean_site_cache"',
+					'version' => '4.9.0',
+				),
+			),
+		),
+		'do_action_deprecated' => array(
```
jrfnl: This entry needs to be merged with the previous one. Also see my remark about the sort order.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,231 @@
...
+				$parameters[ $position ]['start'],
+				$parameters[ $position ]['end'] + 1,
+				true
+			);
+
+			if ( ! $is_dynamic_parameter && ! isset( $paramerter_args[ $matched_parameter ] ) ) {
+				continue;
+			}
+
+			$message        = '%s( %s ) was found.';
```
jrfnl: This message template is misleading if the parameter examined is not the first parameter in the function call.

WordPress/Tests/WP/DeprecatedParameterValuesUnitTest.inc
```diff
@@ -0,0 +1,29 @@
+<?php
+
+// all will give an ERROR.
```
jrfnl: `all` => `All` (one more in another comment below)

WordPress/Tests/WP/DeprecatedParameterValuesUnitTest.inc
```diff
@@ -0,0 +1,29 @@
...
+unregister_setting( 'misc' );
+unregister_setting( 'privacy' );
+
+// all will give a WARNING.
+$value = 'text_direction';
+get_bloginfo( $value );
+get_bloginfo( CONSTANT );
+echo bloginfo( function_name() );
+
+// Method names within a class should be fine.
```
jrfnl: These can be removed. The abstract sniff is already extensively tested elsewhere.

WordPress/Tests/WP/DeprecatedParameterValuesUnitTest.inc
```diff
@@ -0,0 +1,29 @@
+<?php
+
```
jrfnl: I'm missing unit tests testing the conditions on line 176 and 180.

Similarly, maybe add one unit test where one of the deprecated values is used in another position in the function call ?

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,231 @@
...
+
+	/**
+	 * Array of function, argument, and replacement function for deprecated argument.
+	 *
+	 * The list of deprecated parameter values can be found by
+	 * looking for `_deprecated_argument()`, `apply_filters_deprecated()`
+	 * and `do_action_deprecated()`.
+	 * The list is sorted first by the WordPress version when deprecated
+	 * and then alphabetically.
+	 * Last updated for WordPress 4.9.6.
```
jrfnl: Just wondering: how complete is this list at this moment ?

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
grappler: > How is this sniff going to handle deprecated parameter values of another type ? Like booleans, floats etc.

I am trying to think of code example where this would be needed. I came up with the following test but I can't think of a situation where this would be needed.

```php
function bar_foo( $value ) {
	if ( false === $value ) {
		// Deprecated. Use String instead.
	}
}
```

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
grappler: I re-checked it yesterday.

grappler:
Thank you for the review. I have removed `$s_dynamic_parameter` but I would like to add a possibility to extend this class for WPTRT so that `$s_dynamic_parameter` can be added there.

grappler:
> I have removed `$s_dynamic_parameter` but I would like to add a possibility to extend this class for WPTRT so that `$s_dynamic_parameter` can be added there.

I have decided against making any changes here. We can just override `process_parameters()` in WPTRT if need be.

@jrfnl This PR can be merged if OK.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,196 @@
...
+				$message .= ' Use %s instead.';
+				$data[]   = $parameter_args[ $matched_parameter ]['alt'];
+			}
+
+			$is_error = version_compare( $parameter_args[ $matched_parameter ]['version'], $this->minimum_supported_version, '<' );
+			$this->addMessage(
+				$message,
+				$parameter_position,
+				$is_error,
+				$this->string_to_errorcode( $matched_content . 'Found' ),
```
jrfnl: My previous remark about the error code not being properly modular still applies: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1366#discussion_r192923365

WordPress/Tests/WP/DeprecatedParameterValuesUnitTest.inc
jrfnl: Still missing a unit test with a non-deprecated value (https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1366#discussion_r192925928):
`get_bloginfo( 'wpurl' ); // OK.`

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,196 @@
...
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		$this->get_wp_version_from_cl();
+		$param_count = count( $parameters );
```
jrfnl: I've just implemented using FQN for functions which can benefit from it in #1414. If/when you are updating the PR anyway, it would be appreciated if you could change this function call to `\count()` (just add the backslash).

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,196 @@
...
+				'privacy' => array(
+					'alt'     => 'another settings group',
+					'version' => '3.5.0',
+				),
+			),
+		),
+		'bloginfo' => array(
+			1 => array(
+				'home' => array(
+					'alt'     => 'echo esc_url( home_url() )',
```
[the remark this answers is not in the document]
grappler: This initially came from the Theme Review Team theme check plugin. As `(get_)bloginfo()` is not a very good function it makes sense to call the real functions directly. The code is clearer as to when it needs to be escaped.

I asked this question in 2016 in #core Slack channel but got no response https://wordpress.slack.com/archives/C02RQBWTW/p1479063392002489

Related trac tickets [#35014](https://core.trac.wordpress.org/ticket/35014) [#44223](https://core.trac.wordpress.org/ticket/44223).

The reason is that I think `home_url()` is the better function. I happy to change it as it is not an official alternative.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,196 @@
...
+			}
+
+			$parameter_position = $this->phpcsFile->findNext(
+				Tokens::$emptyTokens,
+				$parameters[ $position ]['start'],
+				$parameters[ $position ]['end'] + 1,
+				true
+			);
+
+			$matched_parameter = $this->strip_quotes( $this->tokens[ $parameter_position ]['content'] );
```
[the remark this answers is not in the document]
grappler: This test case is not valid PHP. I tried a few different things but I don't think `$parameter_position` will ever return false as checks earlier make sure of that.

I can add check before if you wish as defensive code.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
grappler: Sorry, I did not fully understand what is needed to make it modular. At the moment the error code would `WordPress.WP.DeprecatedParameterValues.register_settingFound`

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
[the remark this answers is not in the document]
jrfnl: The above example will be valid code as of PHP 7.3 (trailing comma's will be allowed in function calls). Even so, while I agree that the code sample was not valid before that, the sniff should function without throwing notices, even when parse/fatal errors are encountered and it would not at this moment.
So yes, this simple defensive coding check is warranted.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
jrfnl: What I mean by this is that the error code as it it, would silence the message for _two_ different errors: `register_setting()` in combination with having either `misc` or `privacy` as the value for the first parameter. This non-modularness gets worse if there would be more deprecated parameters at different positions in the same function.

So making this truly modular, the error code should encompass:
1. the function detected
2. the position/name of the parameter which is problematic
3. the value detected

The current errorcode only takes the first bit into account.

I'm not saying that's the way to go as that may be overdoing it, but either make the errorcode modular or not - i.e. you could just have `Found` as the errorcode -.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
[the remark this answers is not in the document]
grappler: I will add the check but before I do I wanted to mention that `add_settings_field( '', '', '', /* deliberately not passed */ );` does not cause any errors with the current code.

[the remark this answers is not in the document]
jrfnl: > does not cause any errors with the current code.

... but it did look at token 0 for the content - which in this case is the PHP open tag. (false is being cast to 0)

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,216 @@
...
+				'privacy' => array(
+					'alt'     => 'another settings group',
+					'version' => '3.5.0',
+				),
+			),
+		),
+		'bloginfo' => array(
+			1 => array(
+				'home' => array(
+					'alt'     => 'the "url" agrument',
```
jrfnl: Spelling: `agrument` => `argument` (here and the other four places)

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,216 @@
...
+	protected function process_parameter( $matched_content, $parameter, $parameter_args ) {
+
+		$parameter_position = $this->phpcsFile->findNext(
+			Tokens::$emptyTokens,
+			$parameter['start'],
+			$parameter['end'] + 1,
+			true
+		);
+
+		if ( ! $parameter_position ) {
```
jrfnl: Please check against `false`. While not problematic in this particular case, the `findPrevious()` sister-method can also return `0`, so strict checking is a good habit to get into for these things.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
```diff
@@ -0,0 +1,216 @@
...
+	 *
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameter       Array with start and end token positon of the parameter.
+	 * @param array  $parameter_args  Array with alternative and WordPress deprecation version of the parameter.
+	 *
+	 * @return void
+	 */
+	protected function process_parameter( $matched_content, $parameter, $parameter_args ) {
+
+		$parameter_position = $this->phpcsFile->findNext(
```
jrfnl: I'm wondering whether this part (line 179 up to 188) shouldn't be part of the `process_parameters()` method above.
Feels like this is a check you would always need to do.
Was there any particular reason for moving it to the `process_parameter()` method ?

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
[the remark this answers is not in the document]
jrfnl: I'm ok with having `home_url()` here as long as it is documented inline that this is a deliberate choice to suggest a different alternative than what WP itself suggest.

Either that of - as you've done now - the official alternative, both options are fine by me.

WordPress/Sniffs/WP/DeprecatedParameterValuesSniff.php
grappler: I did think about it. I decided not to as I would need to pass `$parameter_position` as another argument to `process_parameter()`.

jrfnl: Fair enough :+1:

grappler:
Thanks @jrfnl ! 🎉 I have merged the commit to one.

jrfnl:
Thank you @grappler for all your work on this sniff!

jrfnl:
FYI: I've updated the section about setting the `minimum_supported_version` in the [Customizable sniff properties](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#minimum-wp-version-to-check-for-usage-of-deprecated-functions-classes-and-function-parameters) wiki page to reflect this sniff having been added.

PR #1374, opened 2018-06-09
AlternativeFunctions: improve handling of file_get_contents()

### LINKED ISSUE #943: file_get_contents incorrectly flagged as bad behaviour
`file_get_contents()` is being flagged as bad with the suggested replacement being `wp_remote_get()`. This is wrong for local files, because `wp_remote_get()` does not load files from the filesystem (unless I'm mistaken?)

jrfnl:
The `AlternativeFunctions` sniff was throwing two different warnings whenever `file_get_contents()` was encountered, one advising to use `wp_remote_get()`, one advising to use the `WP_FileSystem`.

Aside from that, the warning is only really relevant when retrieving a remote URL. Using `file_get_contents()` for local files should not trigger a warning as discussed in #943.

This PR eliminates the notice about using the `WP_FileSystem` in favour of `wp_remote_get()`.
It also adds a parameter check to attempt to determine whether a local or a remote file is being retrieved and eliminates the `warning` when it is clear that `file_get_contents()` is used only to retrieve a local file.

Includes unit tests.

Fixes #943

### DIFF
--- a/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
+++ b/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
@@ -70,7 +70,7 @@ class AlternativeFunctionsSniff extends AbstractFunctionRestrictionsSniff {
 
 			'file_get_contents' => array(
 				'type'      => 'warning',
-				'message'   => '%s() is discouraged. Use wp_remote_get() instead.',
+				'message'   => '%s() is discouraged. Use wp_remote_get() for remote URLs instead.',
 				'since'     => '2.7.0',
 				'functions' => array(
 					'file_get_contents',
@@ -90,7 +90,6 @@ class AlternativeFunctionsSniff extends AbstractFunctionRestrictionsSniff {
 					'fread',
 					'fwrite',
 					'file_put_contents',
-					'file_get_contents',
 				),
 			),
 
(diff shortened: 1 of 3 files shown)


WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -167,6 +166,44 @@ public function process_matched_token( $stackPtr, $group_name, $matched_content
...
+				}
+
+				if ( strpos( $params[1]['raw'], 'http:' ) !== false
+					|| strpos( $params[1]['raw'], 'https:' ) !== false
+				) {
+					// Definitely a URL, throw notice.
+					break;
+				}
+
+				if ( preg_match( '`\b(?:ABSPATH|WP_(?:CONTENT|PLUGIN)_DIR|WPMU_PLUGIN_DIR|TEMPLATEPATH|STYLESHEETPATH|(?:MU)?PLUGINDIR)\b`', $params[1]['raw'] ) === 1 ) {
```
GaryJones: Feels like these regexes could be centralised for potential re-use, or the whole `preg_match()` could be inside a named local / parent sniff method to become a little more self-documenting.

jrfnl: @GaryJones Good question.

IIRC, so far, we haven't needed anything like it elsewhere. Generally speaking, I'd move them to be more re-usable once a secondary use-case has been identified.

All the same, I'm not fussed about this, so what about adding them as class constants to the `Sniff` class and using the `x` modifier to spread them over several lines with documentation ?

Moving the whole function call to a boolean return function feels wrong as that would prevent sniffs which use the regexes from getting access to the `$matches` if so desired.

GaryJones: Could the new method return the matches, and then the use case here cast to a bool if needed?

jrfnl: > Could the new method return the matches, and then the use case here cast to a bool if needed?

The thing is `preg_match()` has two results:
* The function return value: `1` for a match, `0` for no match or `false` for an error/fault regex etc
* The `$matches` array which may or may not be filled with matches.

By returning just the `$matches`, we would loose the nuances the function return value provides.

I'm wondering if we're not trying to create a solution for a non-issue. Have you got a (future) sniff in mind which would need these regex(es) or an existing one which could benefit from using them ?

If not, let's leave the abstraction until the time when we do have a second use-case as that might also clarify what the best way to abstract this would be.

PR #1378, opened 2018-06-14
Added Enqueued functions check for version and in footer

### LINKED ISSUE #1146: Enhancement: add new sniff to verify optimized use of `wp_enqueue_...` functions
I'd like to propose a new sniff to be added to the `Extra` ruleset to verify the following two things:
* That all calls to these methods which pass a `$src`, also pass something for `$ver`.
    
   Reason: Not passing the `$ver` argument can cause issues with browser caching/the wrong version of a script/stylesheet being used.
* That all calls to the script enqueue methods pass the `$in_footer` argument and warn if the `$in_footer` argument is set to `false`.
    
    Reason: The default value for `$in_footer` is `false`. Not passing the argument means the script will load in the HTML header, i.e. in a layout rendering blocking manner. This is rarely needed.
    By demanding the argument to be present at least the dev will have to think about this and (hopefully) will set the value to `true` in most cases.

Loosely related to #867

Refs;
* https://developer.wordpress.org/reference/functions/wp_register_script/
* https://developer.wordpress.org/reference/functions/wp_enqueue_script/
* https://developer.wordpress.org/reference/functions/wp_register_style/
* https://developer.wordpress.org/reference/functions/wp_enqueue_style/
* https://codex.wordpress.org/Function_Reference/wp_register_style

moorscode:
Addition to [this PR ](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1233)

The checks have been changed to use the tokens to check against unwanted tokens.
This allows for more accurate checking.

Decided to not refrain from throwing a warning about the 5th argument when the 4th is missing, so the user can fix this in one go, instead of having to run the sniffs again and see that new things appear.

Added additional tests for better coverage.

Closes #1233 
Fixes #1146


CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ This projects adheres to [Semantic Versioning](http://semver.org/) and [Keep a C
 
 _Nothing yet._
 
+
```
GaryJones: 0.14.1 is already released, so not sure why this is part of this PR?

WordPress/Sniff.php
```diff
@@ -2645,4 +2645,38 @@ public function is_use_of_global_constant( $stackPtr ) {
...
+		if ( false === $hasNonFalse ) {
+			return true;
+		}
+
+		// Find out it has a number value.
+		$numberToken = $this->phpcsFile->findNext( T_LNUMBER, $parameter['start'], ( $parameter['end'] + 1 ) );
+		if ( $numberToken ) {
+			// If the number is 0, it will evaluate to false and needs to be caught.
+			$token = $this->tokens[ $numberToken ];
+			if ( $token['content'] === '0' ) {
```
GaryJones: Does this need to be reversed?

WordPress/Sniff.php
```diff
@@ -2645,4 +2645,38 @@ public function is_use_of_global_constant( $stackPtr ) {
...
+		$hasNonFalse = $this->phpcsFile->findNext( $falseTokens, $parameter['start'], ( $parameter['end'] + 1 ), true );
+		// If no false tokens are found, we are good.
+		if ( false === $hasNonFalse ) {
+			return true;
+		}
+
+		// Find out it has a number value.
+		$numberToken = $this->phpcsFile->findNext( T_LNUMBER, $parameter['start'], ( $parameter['end'] + 1 ) );
+		if ( $numberToken ) {
+			// If the number is 0, it will evaluate to false and needs to be caught.
```
GaryJones: The `T_LNUMBER`, according to https://secure.php.net/manual/en/tokens.php, can be values in multiple bases - i.e. `0x13`. Presumably, this could also be something equating to 0, like `0x000`? If so, then the identity check `=== '0'` below will incorrectly not match.

GaryJones: Put another way, could it be a string like `'000'`?

CHANGELOG.md
jrfnl: The original sniff was based on `master` and pulled against `develop`. @moorscode has rebased the PR now against `develop`, so this should be sorted.

WordPress/Sniffs/PHP/EnqueuedCheckSniff.php
```diff
@@ -0,0 +1,128 @@
...
+
+		/*
+		 * Version Check
+		 *
+		 * Check to make sure the version is set explicitly.
+		 */
+
+		if ( ! isset( $parameters[4] ) || $parameters[4]['raw'] === 'null' ) {
+			$this->phpcsFile->addError(
+				'Version not set for %s.',
```
jrfnl: What about something along the lines of: `Script/Style version not set in call to %s(). This means new versions of the script/style will not always be loaded due to browser caching.`

WordPress/Sniffs/PHP/EnqueuedCheckSniff.php
```diff
@@ -0,0 +1,128 @@
...
+					'Version ($ver) is not explicitly set for %s; Please supply an explicity version ($ver) argument',
+					$stackPtr,
+					'NoExplicitVersion',
+					array( $matched_content )
+				);
+			}
+		}
+
+		/*
+		 * In footer Check
```
jrfnl: `Check` => `check` or `In footer Check` => `In Footer Check`

WordPress/Sniffs/PHP/EnqueuedCheckSniff.php
```diff
@@ -0,0 +1,128 @@
...
+		 *
+		 * Check to make sure that $in_footer is set to true.
+		 * It will warn the user to make sure it is intended.
+		 *
+		 * Only wp_register_script and wp_enqueue_script need this check,
+		 * as it is not available to wp_register_style and wp_enqueue_style.
+		 */
+		switch ( $matched_content ) {
```
jrfnl: This can be simplified:

```php
if ( strpos( $matched_content, '_style' ) !== false ) {
    return;
}

// ... rest of the code
```

WordPress/Sniffs/PHP/EnqueuedCheckSniff.php
```diff
@@ -0,0 +1,128 @@
...
+			case 'wp_register_script':
+			case 'wp_enqueue_script':
+				if ( ! isset( $parameters[5] ) || $this->is_falsy_parameter( $parameters[5] ) ) {
```
jrfnl: I'm not sure if we should throw a warning if the param has been explicitly set to `false`. In that case, I'd consider it a conscious choice of the developer.

WordPress/Tests/PHP/EnqueuedCheckUnitTest.inc
```diff
@@ -0,0 +1,28 @@
+<?php
+
```
jrfnl: Missing a unit test where `$src` is not set (`wp_enqueue_script( $handle )`).

WordPress/Tests/PHP/EnqueuedCheckUnitTest.php
```diff
@@ -0,0 +1,56 @@
+<?php
+/**
+ * Unit test class for WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPress\Tests\PHP;
```
jrfnl: `PHP` => `WP`

WordPress/Tests/PHP/EnqueuedCheckUnitTest.php
```diff
@@ -0,0 +1,56 @@
...
+class EnqueuedCheckUnitTest extends AbstractSniffUnitTest {
+
+	/**
+	 * Returns the lines where errors should occur.
+	 *
+	 * @return array <int line number> => <int number of errors>
+	 */
+	public function getErrorList() {
+		return array(
+			3 => 1,
```
jrfnl: Please align the arrows (doesn't PHPCS throw an error about this in the Travis check ?)

WordPress/Tests/PHP/EnqueuedCheckUnitTest.php
```diff
@@ -0,0 +1,56 @@
...
+			3  => 1,
+			16 => 1,
+			17 => 1,
+			18 => 1,
+			19 => 1,
+			22 => 1,
+		);
+	}
+
+} // End class.
```
jrfnl: We've recently removed all `// End` comments in the WPCS code. Please remove them from this PR as well.

WordPress/Sniffs/PHP/EnqueuedCheckSniff.php
```diff
@@ -0,0 +1,128 @@
...
+						'In footer ($in_footer) is set to `false` for %s; ' .
+						'It is recommended to load scripts in the footer, please check that this is intended.',
+						$stackPtr,
+						'NotInFooter',
+						array( $matched_content )
+					);
+				}
+		}
+	}
+} // End class.
```
jrfnl: We've recently removed all `// End` comments in the WPCS code. Please remove them from this PR as well.

moorscode:
Hi @dingo-d, yes any specifically set value will be accepted as valid and will not throw any error or warning!

moorscode:
Progress note: Currently the `eval` is failing because the regex needs to be updated.

jrfnl:
> Basically, it will return a time of the last modification of the file, which is useful to use if you are building a theme using webpack, gulp or grunt for instance, and you want to make sure that nothing is being cached when you are developing a theme.

@dingo-d To be fair, that is actually an anti-pattern and would be a good addition to this sniff at a later stage.

Using `filemtime()` during development is a good choice, but it should never make it into production.
Using a function call - especially one to the file system - where a simple text string would suffice, is bad practice.

dingo-d:
> Using filemtime() during development is a good choice, but it should never make it into production.
Using a function call - especially one to the file system - where a simple text string would suffice, is bad practice.

Will take that into account when refactoring the boilerplate. Thanks!

jrfnl:
@moorscode Just checking in: will you have time over the next few days to update the PR so it's ready for a final review ?

moorscode:
This is now ready for final review.

WordPress/Tests/WP/EnqueuedResourceParametersUnitTest.inc
```diff
@@ -8,16 +8,22 @@ wp_register_script( 'someScript-js', 'https://domain.com/someScript.js' , array(
 wp_register_script( 'someScript-js', 'https://domain.com/someScript.js' , array( 'jquery' ), 0, true ); // Error - 0, false or NULL are not allowed
 wp_register_script( 'someScript-js', 'https://domain.com/someScript.js' , array( 'jquery' ), false, true ); // Error - 0, false or NULL are not allowed.
 wp_register_script( 'someScript-js', 'https://domain.com/someScript.js' , array( 'jquery' ), null, true ); // Error - 0, false or NULL are not allowed.
+wp_register_script( 'someScript-js', 'https://domain.com/someScript.js' , array( 'jquery' ), 0.0, true ); // Error - 0, false or NULL are not allowed
```
GaryJones: `domain.com` is a real address. These tests should use `example.com`.

WordPress/Sniff.php
```diff
@@ -2643,5 +2643,4 @@ public function is_use_of_global_constant( $stackPtr ) {
 
 		return true;
 	}
-
```
JDGrimes: This line doesn't need to be removed.

WordPress/Tests/WP/EnqueuedResourceParametersUnitTest.inc
```diff
@@ -0,0 +1,43 @@
+<?php
+
+wp_register_script( 'someScript-js', 'https://domain.com/someScript.js' , array( 'jquery' ) ); // Error - missing $ver, Warning - In Footer is set to a falsy (default) value.
```
JDGrimes: I second @GaryJones here, it would be better to use `example.com`.

moorscode:
Thanks, I have applied the two remarks.

WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
```diff
@@ -0,0 +1,155 @@
...
+				'Resource version not set in call to %s(). This means new versions of the resource will not always be loaded due to browser caching.',
+				$stackPtr,
+				'MissingVersion',
+				array( $matched_content )
+			);
+		} else {
+			// The version argument should have a non-false value.
+			if ( $this->is_falsy( $parameters[4]['start'], $parameters[4]['end'] ) ) {
+				$this->phpcsFile->addError(
+					'Version ($ver) is not explicitly set for %s; This means that the WordPress core version will be used, this is not recommend for Plugin or Theme development.',
```
GaryJones: Re-word / fix typo:

```
'Version parameter is not explicitly set for %s; This means that the WordPress core version will be used which is not recommended for plugin or theme development.',
```

WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
```diff
@@ -0,0 +1,155 @@
...
+		 * as it is not available to wp_register_style and wp_enqueue_style.
+		 */
+		if ( 'wp_register_script' !== $matched_content && 'wp_enqueue_script' !== $matched_content ) {
+			return;
+		}
+
+		if ( ! isset( $parameters[5] ) ) {
+			// If in footer is not set, throw a warning about the default.
+			$this->phpcsFile->addWarning(
+				'In footer ($in_footer) is set set explicitly %s; ' .
```
GaryJones: Typo: "set set" -> "not set"

WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
```diff
@@ -0,0 +1,155 @@
...
+				'It is recommended to load scripts in the footer, please set this value to `true`, or `false` if it should be loaded in the header.',
```
GaryJones:
```
'It is recommended to load scripts in the footer. Please set this value to `true` to load it in the footer, or explicitly `false` if it should be loaded in the header.',
```

WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
```diff
@@ -0,0 +1,155 @@
...
+		$falseTokens[] = T_FALSE;
+
+		// Find anything excluding the false tokens.
+		$hasNonFalse = $this->phpcsFile->findNext( $falseTokens, $start, $end + 1, true );
+		// If no non-false tokens are found, we are good.
+		if ( false === $hasNonFalse ) {
+			return true;
+		}
+
+		// Find out it has a number value.
```
GaryJones: Missing "if" in this comment.

WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
```diff
@@ -0,0 +1,155 @@
...
+		$numberToken = $this->phpcsFile->findNext( array( T_DNUMBER, T_LNUMBER ), $start, $end + 1 );
+		if ( $numberToken ) {
+			$value = trim( $this->phpcsFile->getTokensAsString( $start, ( $end - $start ) + 1 ) );
+
+			// If there are characters that we don't want to eval, return not falsy.
+			if ( preg_match( '#^[\s(\d+(\.\d+)?)*/-]+$#', $value ) > 0 ) {
+				// Evaluate the argument to figure out the outcome is false or not.
+				return false === eval( "return (bool) $value;" ); // @codingStandardsIgnoreLine - No harm here.
```
GaryJones: Can `@codingStandardsIgnoreLine` be replaced with `// phpcs:ignore ...` here?

WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
```diff
@@ -0,0 +1,155 @@
...
+			return true;
+		}
+
+		// Find out it has a number value.
+		$numberToken = $this->phpcsFile->findNext( array( T_DNUMBER, T_LNUMBER ), $start, $end + 1 );
+		if ( $numberToken ) {
+			$value = trim( $this->phpcsFile->getTokensAsString( $start, ( $end - $start ) + 1 ) );
+
+			// If there are characters that we don't want to eval, return not falsy.
+			if ( preg_match( '#^[\s(\d+(\.\d+)?)*/-]+$#', $value ) > 0 ) {
```
GaryJones: Can any of these groups be marked as non-capturing (adding `?:` inside the `(`)?

WordPress/Tests/WP/EnqueuedResourceParametersUnitTest.inc
```diff
@@ -0,0 +1,42 @@
...
+
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ) ); // Error - missing $ver, Warning - In Footer is set to a falsy (default) value.
+
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), /* comment */ '1.1.1', true ); // Ok.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), '1.1.1', $in_footer ); // Ok, the value is set.
+
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), 0, true ); // Error - 0, false or NULL are not allowed
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), false, true ); // Error - 0, false or NULL are not allowed.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), null, true ); // Error - 0, false or NULL are not allowed.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), 0.0, true ); // Error - 0, false or NULL are not allowed
```
GaryJones: How does it handle a hexadecimal zero like `0x` / `0x0`?

WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
moorscode: I do not know why not, but this has been copied from https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/81a265a2bee11d2b9a040496792c647aa5c97dca/WordPress/Sniffs/WP/CronIntervalSniff.php#L181

So perhaps @jrfnl can shed some light on the specific choice here.

WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
```diff
@@ -0,0 +1,155 @@
...
+		$falseTokens[] = T_FALSE;
+
+		// Find anything excluding the false tokens.
+		$hasNonFalse = $this->phpcsFile->findNext( $falseTokens, $start, $end + 1, true );
+		// If no non-false tokens are found, we are good.
+		if ( false === $hasNonFalse ) {
+			return true;
+		}
+
+		// If there is a number, it can be a falsy-value.
```
jrfnl: This whole block still needs a little work. Also, the docblock needs to reflect the impreciseness of the method - i.e. `true` means "falsy value found", while `false` can means "no falsy value found or undertermined"

Aside from the regex being invalid - you can't have groups within a character class -, you would also not catch string `'0'` when passed as the parameter value this way. Or `'0' . '.0'` for that matter.

May I suggest an approach which may be slightly more stable ?
* Walk the tokens between `start` and `end`.
* If you encounter an `$emptyToken`, continue.
* If you encounter a token which is not any of in the below list, return `false` (undetermined).
    - `T_NULL`
    - `T_FALSE`
    - `T_TRUE`
    - `T_LNUMBER`
    - `T_DNUMBER`
    - `T_CONSTANT_ENCAPSED_STRING`
    - `T_START_NOWDOC`
    - `T_NOWDOC`
    - `T_END_NOWDOC`
    - `T_OPEN_PARENTHESIS`
    - `T_CLOSE_PARENTHESIS`
    - `T_STRING_CONCAT`
    - `Tokens::$assignmentTokens`
    - `Tokens::$comparisonTokens`
    - `Tokens::$operators`
    - `Tokens::$booleanOperators`
    - `Tokens::$castTokens`
   
    As this array will not change between runs, you can just set the basic token list as a property using the `TOKEN_CODE => true` format and add the token arrays to it from within the `register()` method.
    You can then just do an `isset` in this function to check against the list and exit if something is not on the list.
* If the token is on the list, add it to a temporary variable.
* Once the token walking is done, just do the `false === eval( $tempVar )` as all the tokens I mentioned above will be safe in that context.

For an example of a sniff where I've used a similar approach: https://github.com/wimg/PHPCompatibility/blob/945a8ece200f61e0ea75af5c8242e1eea5ba9b17/PHPCompatibility/Sniffs/PHP/DiscouragedSwitchContinueSniff.php#L174-L186

I believe that using the above approach, all calculations, integers using a different base and the most common results of type juggling etc will be evaluated correctly.

Regarding the `// @codingStandardsIgnoreLine` - the sniff this is copied from is an old sniff from before PHPCS 3.2.0 and as the WPCS minimum PHPCS requirement hasn't gone up yet to 3.2+, existing annotations have not been changed yet.
I agree with @GaryJones though that ignore annotations which are being newly introduced should probably use the new style as the WPCS native CS check in Travis is run on `master` anyway.

WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
```diff
@@ -0,0 +1,155 @@
...
+				'Resource version not set in call to %s(). This means new versions of the resource will not always be loaded due to browser caching.',
+				$stackPtr,
+				'MissingVersion',
+				array( $matched_content )
+			);
+		} else {
+			// The version argument should have a non-false value.
+			if ( $this->is_falsy( $parameters[4]['start'], $parameters[4]['end'] ) ) {
+				$this->phpcsFile->addError(
+					'Version parameter is not explicitly set for %s; This means that the WordPress core version will be used which is not recommended for plugin or theme development.',
```
jrfnl: Suggestion: `Version parameter is not explicitly set or has been set to an equivalent of "false" for %s`

WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
jrfnl: @moorscode and me discussed this IRL today and I've made the suggested changes in a separate commit (with @moorscode's blessing).

moorscode:
I have rebased the commit to develop and cleaned up the commit history (with the awesome assistance of @jrfnl)

jrfnl:
Thanks @moorscode for all your hard work on this! I'd be happy to merge this in. @JDGrimes do you still want to have another look or can I dismiss your earlier review ?

JDGrimes:
🎉 Thanks again @moorscode !

Issue #1371, opened 2018-06-09
Sniff to detect proper usage of preg_quote() ?


jrfnl:
I'd like to suggest adding two new sniffs around the usage of [`preg_quote()`](http://php.net/manual/en/function.preg-quote.php).

* Verify the `$regex` parameter of any PCRE function calls to make sure that if the regex is being build up by concatenating parts together, that any `$variable` parts are wrapped in a call to `preg_quote()`.
    ```php
    preg_match( '`http[s]?://[^\s<>\'"()]*' . preg_quote( $match_data[0], '`' ) . '`', $content ); // OK.
    preg_match( '`http[s]?://[^\s<>\'"()]*' . $match_data[0] . '`', $content ); // Warning.
    preg_match( $regex, $content ); // OK, ignore as impossible to check reliably.
    ```
* Verify that the optional second parameter `$delimiter` of `preg_quote()` is always passed.
    Too often I come across code where it is missing and unless it is passed, the default `/` delimiters are presumed, which is often wrong.

Initially these sniffs would go into `Extra`, but I'd encourage the WP Core team to consider accepting them into the `Core` ruleset.

Opinions ?

jrfnl:
Re-opened as the first bullet point has not been addressed yet.

jrfnl:
An issue has been opened in PHPCSExtra for this: https://github.com/PHPCSStandards/PHPCSExtra/issues/190

Once PHPCSExtra has the sniff(s) for this, those should be added to WPCS `Extra` (and possibly `Core`).

Issue #764, opened 2017-01-01
Improve sniff handling of modern PHP code


jrfnl:
While WP still has a minimum requirement of PHP 5.2 and a lot of related projects code accordingly, there are also numerous projects out there which have set their plugin/theme minimum requirement at PHP 5.3 or higher.

Most sniffs currently don't take things like namespaces into account so the below code would trigger an error (even though the `fwrite` function in this case is namespaced).

While this may not be the best example (or good practice in the first place), I would like to suggest we review code to see if we can make it more forward compatible with modern PHP code.

```php
namespace MyNamespace;

function fwrite() {
	// A compat layer or something.
}

class My_Class {
	function do_something() {
		fwrite( $param );
	}
}
```

jrfnl:
Another typical one which is currently not or hardly taken into account: `T_HEREDOC`

jrfnl:
Let's try & keep track of the reviews done:

Globally checked (I might well have missed something) that sniffs allow correctly for:
- [x] short array tokens
- [x] closures/anonymous functions
- [x] anonymous classes
- [x] interfaces
- [x] traits
- [ ] `try/catch/finally` constructs
- [ ] `__NAMESPACE__`, `__TRAIT__`, `__DIR__` magic constants
- [x] `T_NOWDOC` and related tokens
- [x] `T_HEREDOC` and related tokens
- [ ] namespaces
- [ ] use statements
- [ ] type hints
- [ ] yield (for generators)
- [ ] ternary without middle `$x ? : $y`
- [ ] new operators: `T_POW`, `T_POW_EQUAL`, `T_ELLIPSIS`, `T_SPACESHIP`, `T_COALESCE`, `T_COALESCE_EQUAL`

N.B.: I don't "own" this ticket. Everyone is warmly invited to contribute, both in ideas what to check for as well as doing the reviews!

maurobringolf:
I am not sure if this is the right place to ask, but phpcs gives me wierd error messages around a closure used in a multiline function call:

```php
array_map( function( $item ) {
    echo 'hello';
    },
    $arr
);
```

I get this: *"Line indented incorrectly; expected 0 tabs, found 1"*. If I do not indent the closing brace it says *"Multi-line function call not indented correctly; expected 4 spaces but found 0"*. I see that you checked closures, so what would be the way to format this?

Thanks for any help

jrfnl:
[...]

PR #1387, opened 2018-06-25
Sniff::get_function_call_parameters(): correctly handle closures when passed as param

jrfnl:
Closures can be declared within function calls when the function expects a call-back.
This situation was only partially handled within the `Sniff::get_function_call_parameters()` method.

Most of the time when a comma is encountered within a closure, it will be at a different nesting level than the original function call/array, which meant it would be disregarded and the method would return the correct results.
However, there are some, albeit rare, situations in which the comma would be at the target nesting level, causing the method to split the closure into two or more "parameters".

As far as I know, this bug has not caused any issues for any of the WPCS sniffs so far, but the situation should be handled correctly even so.

I came across this while working on #1371, the PR for which will contain a unit test covering this fix.

Additionally, the same fix has been pulled to the PHPCompatibility standard with dedicated unit tests.
See: https://github.com/wimg/PHPCompatibility/pull/682

Related to #764

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -2096,7 +2096,7 @@ abstract class Sniff implements PHPCS_Sniff {
 		$next_comma  = $opener;
 		$param_start = ( $opener + 1 );
 		$cnt         = 1;
-		while ( $next_comma = $this->phpcsFile->findNext( array( T_COMMA, $this->tokens[ $closer ]['code'], T_OPEN_SHORT_ARRAY ), ( $next_comma + 1 ), ( $closer + 1 ) ) ) {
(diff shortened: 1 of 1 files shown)


GaryJones:
Any unit tests needed for this?

jrfnl:
@GaryJones 

> I came across this while working on #1371, **the PR for which will contain a unit test covering this fix**.
> 
> Additionally, the same fix has been pulled to the PHPCompatibility standard with **dedicated unit tests**.
See: wimg/PHPCompatibility#682

While the unit tests in PHPCompatibility are of course in a different repo, the function used is the same in both (as I wrote & pulled it) and PHPCompatibility is better set up to unit test utility methods, while WPCS is not.

The unit test in the `PregQuote` sniff will cover the change sufficiently for WPCS once it is pulled.

PR #1401, opened 2018-06-28
README: add a few more badges

### LINKED ISSUE #1393: Packagist indicates the wrong branch as "unstable"
The `dev-master` branch is seen on Packagist as the `unstable` branch. This should be the `dev-develop` branch instead.

Who has admin access to the listing on Packagist to change this ? @westonruter ?

jrfnl:
Badges offer a quick overview of a project.

* Show `stable` and `unstable` version with their respective build status and last update date.
* Make the minimum supported PHP version and the version the code is tested on, more easily discoverable.
* Show the License and Contributor count more prominently

Notes:
* The `Tested on` badge will need manual updating each time a new PHP version is added to the Travis build matrix, but as this is approximately once a year, I figured that shouldn't be a problem.
* For the badges created via `img.shields.io`, I've noticed that if there are more than a couple, they may not always display properly.
    To counter this, I've only used them when there was no good alternative and added a `maxAge` value to the `svg` URLs to try and get the images to cache which will hopefully make the display more stable.
    - For custom badges which will rarely change, I've set the `maxAge` to four weeks.
    - For the stable release info, I've set the `maxAge` to half an hour.
    - For the PHP minimum version info and the contributors count, I've set the `maxAge` to an hour.

### Example of what it looks like:

screenshot

Closes #1393

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,21 @@
+<div aria-hidden="true">
+
 [![Latest Stable Version](https://poser.pugx.org/wp-coding-standards/wpcs/v/stable)](https://packagist.org/packages/wp-coding-standards/wpcs)
-[![Build Status](https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards.png?branch=master)](https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards)
-[![GitHub Release Date](https://img.shields.io/github/release-date/WordPress-Coding-Standards/WordPress-Coding-Standards.svg?maxAge=1800)](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/releases)
+[![Travis Build Status](https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards.png?branch=master)](https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards)
+[![Release Date of the Latest Version](https://img.shields.io/github/release-date/WordPress-Coding-Standards/WordPress-Coding-Standards.svg?maxAge=1800)](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/releases)
 :construction:
 [![Latest Unstable Version](https://img.shields.io/badge/unstable-dev--develop-e68718.svg?maxAge=2419200)](https://packagist.org/packages/wp-coding-standards/wpcs#dev-develop)
-[![Build Status](https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards.png?branch=develop)](https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards)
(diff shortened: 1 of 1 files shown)


jrfnl:
> As things stand, the values from these buttons are completely inaccessible to screen reader users. 

Thanks for bringing attention to that :+1: I agree we should consider this.

> Either the values need to be made accessible, using plain text, 

The problem with that is that most values are dynamically generated, so unless we want to manually maintain the button (answer: no...), there is no way we can add that data either in markdown or HTML.

I can try and improve the `alt` values a little as they seem to be taken from the `![...]` part of the image markdown.

> or the whole badge area needs to be wrapped in an aria-hidden (though Markdown may not render inside a HTML block).

That seems to me the most logical choice. And I don't mean that as a cop-out. (Nearly) all information contained in the badges is also mentioned in one way or another in the readme itself, so the badges are just some visual sugar for a quick overview.

@GaryJones What do you think ? Any other options you can see ?

GaryJones:
In the Preview of the changes at least, GitHub appears to be stripping the extra `div` you added, so there's no change for screen readers.

It may need be HTML anchors and images (with anchors being `aria-hidden="true"`), rather than markdown.

jrfnl:
@GaryJones I've done some investigating and it looks like `aria-hidden` will be stripped no matter how I add it. See: https://github.com/jch/html-pipeline/issues/215

This is the file which whitelists what is allowed: https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/sanitization_filter.rb

I'd happily send in a PR to that repo of course, but as the original issue has been open for three years, I'm not sure how much traction that would get, let alone when it will take effect for this file.

In the mean time, how do we solve this for this PR ? Would adding the badges in plain HTML *without* `alt` tags be any better ? That way screen readers wouldn't have anything to read out or do they default to saying `image` or something ?
I'll happily adjust the PR, but I need some help here on what's the best approach/solution.

jrfnl:
@GaryJones Woha! It looks like my comment in the repo mentioned above is having effect straight away. A PR has been opened to allow the attributes: https://github.com/jch/html-pipeline/pull/298. :tada: 

Still not sure of course when it will have effect on GH-wide, but I'm inclined to say, let's merge this in the knowledge that the accessibility issue will be solved automatically in the near future once the PR has been merged, a new version of the gem has been released and GH has updated their gems to the latest versions.

GaryJones:
These don't appear to be loading correctly, at least for me - console says they are 504 timing out:

screenshot 2018-07-02 10 39 07

jrfnl:
> These don't appear to be loading correctly, at least for me - console says they are 504 timing out:

@GaryJones I noticed a similar issue earlier, they should all display once the caching has kicked in.

> For the badges created via img.shields.io, I've noticed that if there are more than a couple, they may not always display properly.
> To counter this, I've only used them when there was no good alternative and added a maxAge value to the svg URLs to try and get the images to cache

The badges are currently created by various different services. `shields.io` is the problematic service, so if you know of (another) alternative for those badges... ?

Issue #1309, opened 2018-02-27
VIP ruleset removal


tomjn:
At the moment there's a VIP ruleset in the WPCS, but this is confusing, especially when VIPs and VIP itself make use of a ruleset located here named `WordPressVIPMinimum`. It's confusing for our clients and misleading for WPCS users.

Instead, VIP would like to move the ruleset to our own repo, or integrate it so that it's no longer necessary. At that point talking about the two should become a lot easier, and the possibility of migrating rules back to WPCS should be more feasible. We also have plans to change severity and messages specifically for the VIP ruleset so the extra flexibility of having all VIP stuff in a VIP repo would be good

Are there any particular notes implementation wise or suggestions to keep in mind?

jrfnl:
@tomjn Could you give us a status update please ?

tomjn:
When it got to the point were I would have to deprecate rather than remove things, the scope of what needed doing far outstripped the available time, so nothing further was done on that front, and I focused elsewhere. I haven't had the time to circle back to this, and I don't know when I will.

Having said that, feel free to burn the VIP ruleset with the fire of a thousand suns, with extreme prejudice.

aristath:
While I agree with this one, the VIP ruleset does have some useful stuff in it...
Particularly the validation & sanitization rulesets

tomjn:
I agree and they should be migrated into the other rules, but right now:

 - Nobody at VIP uses the ruleset
 - Our clients are told not to use it
 - Anybody who does use these rules gets a lot of existing deprecation warnings about sniffs
 - It's a source of confusion for all involved

We have the funeral party ready at hand, we just need someone to do the deed and put it out of its misery

GaryJones:
@tomjn Are folks at VIP using https://github.com/Automattic/VIP-Coding-Standards/ and/or https://github.com/Automattic/VIP-Go-PHPCS-Blockers consistently then?

tomjn:
We use https://github.com/Automattic/VIP-Coding-Standards/, which is ran on all code before it gets reviewed, all commits that arrive in SVN, and via CI for client git repos. If the clients don't run it which we encourage them to do, then it gets ran at our end

https://github.com/Automattic/VIP-Go-PHPCS-Blockers can be ignored and needs archiving, it's creation is more of an internal miscommunication, which became the `WordPress-VIP-Go` ruleset in the other repo

jrfnl:
[...]

PR #1403, opened 2018-06-30
Add note regarding WordPress-VIP ruleset

GaryJones:
Quick and temporary workaround to [reduce the chance of new folks using it](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1309#issuecomment-401496440):

> I honestly believe this ruleset is damaging, and the longer it remains the higher the chances of new people finding it and using it.

See #1309.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,8 +116,9 @@ You can use the following as standard names when invoking `phpcs` to select snif
   - `WordPress-Docs` - additional ruleset for [WordPress inline documentation standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/)
   - `WordPress-Extra` - extended ruleset for recommended best practices, not sufficiently covered in the WordPress core coding standards
     - includes `WordPress-Core`
-  - `WordPress-VIP` - extended ruleset for [WordPress.com VIP coding requirements](https://vip.wordpress.com/documentation/code-review-what-we-look-for/)
-    - includes `WordPress-Core`
+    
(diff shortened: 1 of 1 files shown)


GaryJones:
> It was originally for [WordPress.com VIP coding requirements] - is this sentence missing the word intended or is it just me ?

Added, and clarified further.

> I'm a bit wary about using the word soon in this context all things considering.

Changed to "in due course".

GaryJones:
This was meant as a quick workaround whilst the finer points of how such a removal would be done. It was building on Tom's information that the VIP team no longer use it, or recommend it to their clients.

I think making an _actual_ change to the characteristics of WPCS, like removing the ruleset, should be its own PR, even if *that* includes a further change to this readme about its inclusion.

JDGrimes:
Just to make sure that we are all on the same page, @jrfnl was suggesting that the `WordPress.VIP` sniffs be excluded from the `WordPress` ruleset, so that the VIP sniffs would no longer be used by WPCS by default—only if they were explicitly included via a custom ruleset. 

I think that makes sense, as otherwise, I'd think a lot of folks would still end up using the VIP sniffs via the `WordPress` ruleset. It doesn't make sense to tell people not to use it, and then offer them a ruleset that includes it, IMO.

tomjn:
I can say VIP doesn't recommend the ruleset fullstop, client or not client

jrfnl:
Closing as @GaryJones' textual improvements were implemented in #1410.

PR #1411, opened 2018-07-04
Performance: use FQN for global constants

jrfnl:
Constants in PHP are namespaced as well, however, when PHP cannot find the constant in the namespace, it will fall through to the global namespace.

The step to first check within the namespace can be skipped by either using `use const ...` (PHP 5.6+) or by prefixing the constant with a `\`.

As WPCS uses a lot of global constants, some performance gain can be archived by using the FQN for constants.

This PR implements this throughout the codebase.

@JDGrimes It would be great if you'd have the chance to run a performance test to see what the effect is of this PR ;-)

### DIFF
--- a/Test/Standards/AbstractSniffUnitTest.php
+++ b/Test/Standards/AbstractSniffUnitTest.php
@@ -100,7 +100,7 @@ abstract class AbstractSniffUnitTest extends PHPUnit_Framework_TestCase {
     {
         $testFiles = array();
 
-        $dir = substr($testFileBase, 0, strrpos($testFileBase, DIRECTORY_SEPARATOR));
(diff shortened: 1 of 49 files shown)


jrfnl:
> One of the inspections in my IDE also highlights global function calls that could have an explicit FQN when it gives (supposed) performance benefits.

You are reading my mind ? Already done as well, just figured that should be a separate PR and as it would conflict with this one, it will be pulled after this has been merged.

JDGrimes:
> @JDGrimes It would be great if you'd have the chance to run a performance test to see what the effect is of this PR ;-)

I haven't gotten around to setting up Blackfire since last updating PHP, so unfortunately I won't likely have a chance to benchmark this. :(

jrfnl:
@JDGrimes Never mind. Would have been nice to see what the effect was, but it definitely shouldn't have a negative effect, so we should be good.

PR #1413, opened 2018-07-05
AbstractVariableRestrictions: remove superfluous token constant

jrfnl:
The `T_VAR` constant represents the token for the `var` keyword which was used in PHP 4 to declare class properties.

Searching for that in this context is irrelevant.


GaryJones:
Approved, but `var` is still valid - so its possible that some code might still be using it?

jrfnl:
> but var is still valid - so its possible that some code might still be using it?

Yes, `var` is absolutely still valid, however the code which this PR addresses is code which is examining the use of class properties/constants, i.e. as stated a few lines above the change:
`// Object var, ex: $foo->bar / $foo::bar / Foo::bar / Foo::$bar .`

The touched line is trying to find the `bar` bits of the above examples. In that context searching for the `var` keyword is nonsense.

PR #1414, opened 2018-07-05
Performance: use FQN for functions which can use PHP7 compile time opcodes

jrfnl:
Functions in PHP are namespaced as well, however, when PHP cannot find the function in the namespace, it will fall through to the global namespace.

The step to first check within the namespace can be skipped by either using `use func ...` (PHP 5.6+) or by prefixing the function with a `\`.

While this offers a small performance gain as a general practice, using this for the special compiled functions should offer a noticeable performance gain when running WPCS on PHP 7+.
The special compiled functions are replaced by opcodes at compile time, but only if it's clear at compile time that the global PHP native functions will be used.
For the list of functions, see: https://github.com/php/php-src/blob/f2db305fa4e9bd7d04d567822687ec714aedcdb5/Zend/zend_compile.c#L3872

This PR implements this throughout the codebase.

@GaryJones You meant this, didn't you ?

### DIFF
--- a/Test/Standards/AbstractSniffUnitTest.php
+++ b/Test/Standards/AbstractSniffUnitTest.php
@@ -80,7 +80,7 @@ abstract class AbstractSniffUnitTest extends PHPUnit_Framework_TestCase {
             self::$phpcs = new PHP_CodeSniffer();
         }
 
-        $class = get_class($this);
+        $class = \get_class($this);
         $this->standardsDir = $GLOBALS['PHP_CODESNIFFER_STANDARD_DIRS'][$class];
 
     }//end setUp()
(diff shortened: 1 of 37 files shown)


GaryJones:
Yes, though I haven't yet understood why certain functions are "special compiled" and why some aren't. Why strlen, and not substr, for instance? I guess it's down to PHP internals that I don't appreciate.

jrfnl:
@GaryJones I can't speak for the details of the PHP internals, but it has to do with the implementation of the [Abstract Syntax Tree](https://wiki.php.net/rfc/abstract_syntax_tree), so if you want to know more, that's where to start digging ;-)

PR #1421, opened 2018-07-12
Fix false positive violation for return type colon

### LINKED ISSUE #1420: Spacing of : for return declaration when including an argument
Related to #764.

When a function has a return declaration AND at least one parameter, then a warning is thrown unless there is a space before the `:`. My understanding is that WPCS doesn't define whether a space should be included here or not. The rest of the PHP world doesn't include a space, and WPCS ONLY flags the lack of space when a parameter is present.

Example:

```php
<?php
/**
 * Foo
 *
 * @package   Test
 */

/**
 * Foo1
 *
 * @return string
 */
function prefix_foo1(): string {
	return '';
}

/**
 * Foo2
 *
 * @param string $input String.
 * @return string String
 */
function prefix_foo2( $input ): string {
	return '';
}

/**
 * Foo3
 *
 * @param string $input String.
 * @return string String
 */
function prefix_foo3( string $input ): string {
	return '';
}


/**
 * Foo4
 *
 * @param string $input String.
 * @return string String
 */
function prefix_foo4( string $input ) : string {
	return '';
}

/**
 * Foo5
 *
 * @param string $input String.
 * @return string String
 */
function prefix_foo5( $input ) : string {
	return '';
}
```

- Foo1: No argument, no warning.
- Foo2: Argument, warning.
- Foo3: Argument with type declaration, warning.
- Foo4: Argument, space before `:`, no warning.
- Foo5: Argument with type declaration, space before `:`, no warning.

Here are the warnings:

```
 23 | ERROR | [x] Space between opening control structure and closing parenthesis is required (WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceAfterCloseParenthesis)
 33 | ERROR | [x] Space between opening control structure and closing parenthesis is required (WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceAfterCloseParenthesis)
```

The general nature of the warning means it can't easily be excluded - either `phpcs:disable` and `phpcs:enable` lines need to be added, or the whole check needs to be severity `0`, which then means other instances of this whitespace issue would not be reported.

GaryJones:
Fixes #1420.

### DIFF
--- a/WordPress/Sniffs/WhiteSpace/ControlStructureSpacingSniff.php
+++ b/WordPress/Sniffs/WhiteSpace/ControlStructureSpacingSniff.php
@@ -344,6 +344,10 @@ class ControlStructureSpacingSniff extends Sniff {
 				}
 
 				if ( \T_WHITESPACE !== $this->tokens[ ( $parenthesisCloser + 1 ) ]['code']
+					&& ! ( // Do NOT flag : immediately following ) for return types declarations.
+						\T_COLON === $this->tokens[ ( $parenthesisCloser + 1 ) ]['code']
+						&& in_array( $this->tokens[ $this->tokens[ $parenthesisCloser ]['parenthesis_owner'] ]['code'], array( \T_FUNCTION, \T_CLOSURE ), true )
+					)
 					&& ( isset( $scopeOpener ) && \T_COLON !== $this->tokens[ $scopeOpener ]['code'] )
 				) {
 					$error = 'Space between opening control structure and closing parenthesis is required';
@@ -355,7 +359,7 @@ class ControlStructureSpacingSniff extends Sniff {
 				}
 			}
 
-			// Ignore this for function declarations. Handled by the OpeningFunctionBraceKrnighanRitchie sniff.
+			// Ignore this for function declarations. Handled by the OpeningFunctionBraceKernighanRitchie sniff.
 			if ( \T_FUNCTION !== $this->tokens[ $stackPtr ]['code']
 				&& \T_CLOSURE !== $this->tokens[ $stackPtr ]['code']
 				&& isset( $this->tokens[ $parenthesisOpener ]['parenthesis_owner'] )
--- a/WordPress/Tests/WhiteSpace/ControlStructureSpacingUnitTest.1.inc
+++ b/WordPress/Tests/WhiteSpace/ControlStructureSpacingUnitTest.1.inc
@@ -274,3 +274,19 @@ function hi(): array
 {
     return [];
 }
+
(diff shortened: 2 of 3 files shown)


WordPress/Sniffs/WhiteSpace/ControlStructureSpacingSniff.php
```diff
@@ -344,6 +344,7 @@ public function process_token( $stackPtr ) {
 				}
 
 				if ( \T_WHITESPACE !== $this->tokens[ ( $parenthesisCloser + 1 ) ]['code']
+					&& T_COLON !== $this->tokens[ ( $parenthesisCloser + 1 ) ]['code']
```
jrfnl: I'm trying to think of code samples on which this will break, but it's late and my brain is shutting down.

Maybe - to be on the safe side -, this condition should be combined with a check that the `$stackPtr` which triggered the sniff is either a `T_FUNCTION` or a `T_CLOSURE` token ?

GaryJones: I would agree, but I'm not sure what `$stackPtr` would point to at this moment of the process.

I considered trying to find the next `{`, and then see if the scope_condition was a `T_FUNCTION`, or other similar things to find out of the `:` is indeed using somewhere inside `function ... {`.

I'll see if I can review what `$stackPtr` should contain.

GaryJones: Finally got my head around the approach I was trying to take. See 07bc4ba.

jrfnl: Sorry, I was in a meeting. I think the code can actually be simpler:
You already have the `$parenthesisCloser` and - if I remember correctly - parenthesis which belong to a function declaration will have a `parenthesis_owner` set in the token array, so you could look at that and check if that is a `T_FUNCTION` or `T_CLOSURE`.

GaryJones: Awesome, thanks - https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1421/commits/798be40c739796b15f648dc28c9f2ce7367b1f59.

GaryJones: And https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1421/commits/d2685e9453b77b21f3d7e970a0507cd6ecff2707 updates again to support closures.

PR #1433, opened 2018-07-25
Composer: Make script DRY

GaryJones:
Also allows the functionality to be run directly (`composer install-codestandards`) instead of having to do a full `composer install` or `composer update`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
 	},
 	"type"       : "phpcodesniffer-standard",
 	"scripts"    : {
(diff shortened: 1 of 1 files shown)


GaryJones:
No worries if this gets deferred until `1.0.1` - it's only for WPCS development anyway.

jrfnl:
I'd like to merge this as it is a sensible change, but I personally don't use Composer for sniff development. 

@WordPress-Coding-Standards/wpcs-admins Anyone around who does use Composer for sniff development and can test this PR to confirm it can be merged ?

PR #1450, opened 2018-08-16
New WPCS native `NoSilencedErrors` sniff

### LINKED ISSUE #1449: NoSilencedErrors: allow for valid use-cases
Currently, all error silencing is being reported with a `warning`.

There are, however, valid use-cases for the error reporting operator `@`, even though rare, where no amount of error checking can help out.

In particular, this applies to certain PHP functions which will throw an `E_WARNING` when the function failed.

An example of such a function is [`unlink()`](http://php.net/manual/en/function.unlink.php). You can check whether the `file_exists()`, whether it is readable/writable/executable, but there is no way to easily and reliably check if the file is "deletable" - which may not be the case due to file permissions. So, when using `unlink()`, it should, IMHO, be acceptable to use the error silencing operator.

I'd like to suggest creating a WPCS native version of the upstream `NoSilencedErrors` sniff (which is so simple that extending it would be silly).
In the WPCS native version, we could add a whitelist of functions which may be preceded by the error silencing operator and omit throwing a `warning` in that case.

Some other candidates:
* [`chdir`](http://php.net/manual/en/function.chdir.php)
* [`file_exists()`](http://php.net/manual/en/function.file-exists.php) - throws an `E_WARNING` in case of failure.
* [`filesize()`](http://php.net/manual/en/function.filesize.php) - throws an `E_WARNING` in case of error.
* [`fopen()`](http://php.net/manual/en/function.fopen.php) - throws an `E_WARNING` if the file failed to open.
* [`is_dir`](http://php.net/manual/en/function.is-dir.php) - throws an `E_WARNING` on failure.
* [`mkdir()`](http://php.net/manual/en/function.mkdir.php) - throws an `E_WARNING` if the relevant permissions prevent creating the directory.
* [`opendir()`](http://php.net/manual/en/function.opendir.php) - throws an `E_WARNING` if the directory can not be opened due to permission restrictions or filesystem errors.

... or maybe we should just whitelist all relevant [filesystem](http://php.net/manual/en/ref.filesystem.php) and [directory](http://php.net/manual/en/ref.dir.php) functions and possibly most [stream](http://php.net/manual/en/ref.stream.php), [ftp](http://php.net/manual/en/ref.ftp.php) and [gz](http://php.net/manual/en/ref.zlib.php) functions.

**Note:** This sniff is included in the `Core` ruleset, so while a number of these functions are discouraged for use by plugins and themes, there are valid use-cases for these functions in Core.
For plugins/themes, another sniff takes care of throwing a warning for using these functions.

Additionally, we could also consider blanket allowing error silencing in unit test classes, where this could be perfectly valid.

Opinions ? Suggestions for additional functions which should be whitelisted ?

jrfnl:
Until now, WPCS used the PHPCS native `Generic.PHP.NoSilencedErrors` sniff. This sniff would throw warnings for all uses of the `@` error silencing operator. However, there are a number of PHP functions for which the use of the error silencing operator is warranted as no amount of error checking can prevent warnings from being thrown.

This PR adds a new WPCS native `PHP.NoSilencedErrors` sniff which will:
* not throw warnings when the error silencing operator is used in combination with one of the whitelisted functions.
* throw warnings in all other cases.
* allows for a user-defined list of additional whitelisted functions to be passed via a custom ruleset.
* displays a brief snippet of code to show the context in which the error silencing operator is being used.

To get to the current list of whitelisted functions, I've checked:
* The PHP manual for every single function in the [filesystem](http://php.net/manual/en/ref.filesystem.php), [directory](http://php.net/manual/en/ref.dir.php),  [stream](http://php.net/manual/en/ref.stream.php), [ftp](http://php.net/manual/en/ref.ftp.php), [gz](http://php.net/manual/en/ref.zlib.php) and [ssh2](http://php.net/manual/en/ref.ssh2.php) extensions.
* Ran the sniff over WP Core and used the reported metrics to verify a number of additional functions for which WP Core currently uses error silencing.

Functions for which there is a way to avoid the warning being thrown have **not** been added to the list. So, for example, the `header()` function does not qualify, as the warning could be prevented by wrapping the function call in a `headers_send()` condition.

For WP Core, the original sniff yielded 286 warnings.
This sniff gets rid of 119 of those, which should be considered unfixable, leaving 167 violations still to be addressed in WP Core. /cc @pento 

Fixes #1449

#### To do once this PR is merged:
- [ ] Add the sniff properties for this sniff to the Customizable Sniff Properties wiki page.

### DIFF
--- a/WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
+++ b/WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
@@ -166,13 +166,13 @@ class NoSilencedErrorsSniff extends Sniff {
 		$this->custom_whitelist = array_map( 'strtolower', $this->custom_whitelist );
 
 		/*
-		 * Check if the error silencing if done for one of the whitelisted functions.
+		 * Check if the error silencing is done for one of the whitelisted functions.
 		 */
-		$nextNonEmpty = $this->phpcsFile->findNext( $this->empty_tokens, ( $stackPtr + 1 ), null, true, null, true );
-		if ( false !== $nextNonEmpty && \T_STRING === $this->tokens[ $nextNonEmpty ]['code'] ) {
-			$hasParenthesis = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $nextNonEmpty + 1 ), null, true, null, true );
-			if ( false !== $hasParenthesis && \T_OPEN_PARENTHESIS === $this->tokens[ $hasParenthesis ]['code'] ) {
-				$function_name = strtolower( $this->tokens[ $nextNonEmpty ]['content'] );
+		$next_non_empty = $this->phpcsFile->findNext( $this->empty_tokens, ( $stackPtr + 1 ), null, true, null, true );
+		if ( false !== $next_non_empty && \T_STRING === $this->tokens[ $next_non_empty ]['code'] ) {
+			$has_parenthesis = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $next_non_empty + 1 ), null, true, null, true );
+			if ( false !== $has_parenthesis && \T_OPEN_PARENTHESIS === $this->tokens[ $has_parenthesis ]['code'] ) {
+				$function_name = strtolower( $this->tokens[ $next_non_empty ]['content'] );
 				if ( isset( $this->function_whitelist[ $function_name ] ) === true
 					|| in_array( $function_name, $this->custom_whitelist, true ) === true
 				) {
@@ -183,10 +183,10 @@ class NoSilencedErrorsSniff extends Sniff {
 		}
 
 		// Prepare the "Found" string to display.
-		$context_length = (int) $this->context_length;
-		$endOfStatement = $this->phpcsFile->findEndOfStatement( $stackPtr, T_COMMA );
-		if ( ( $endOfStatement - $stackPtr ) < $context_length ) {
-			$context_length = ( $endOfStatement - $stackPtr );
+		$context_length   = (int) $this->context_length;
+		$end_of_statement = $this->phpcsFile->findEndOfStatement( $stackPtr, \T_COMMA );
+		if ( ( $end_of_statement - $stackPtr ) < $context_length ) {
+			$context_length = ( $end_of_statement - $stackPtr );
 		}
 		$found = $this->phpcsFile->getTokensAsString( $stackPtr, $context_length );
 		$found = str_replace( array( "\t", "\n", "\r" ), ' ', $found ) . '...';
--- a/WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
+++ b/WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
@@ -1,0 +1,208 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPress\Sniffs\PHP;
+
+use WordPress\Sniff;
+use PHP_CodeSniffer_Tokens as Tokens;
+
+/**
+ * Discourage the use of the PHP error silencing operator.
+ *
+ * This sniff allows the error operator to be used with a select list
+ * of whitelisted functions, as no amount of error checking can prevent
+ * PHP from throwing errors when those functions are used.
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   1.1.0
+ */
+class NoSilencedErrorsSniff extends Sniff {
+
+	/**
+	 * Number of tokens to display in the error message to show
+	 * the error silencing context.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @var int
+	 */
+	public $context_length = 6;
+
+	/**
+	 * User defined whitelist.
+	 *
+	 * Allows users to pass a list of additional functions to whitelist
+	 * from their custom ruleset.
+	 *
+	 * @since 1.1.0
+	 *
+	 * @var array
+	 */
+	public $custom_whitelist = array();
+
+	/**
+	 * PHP native function whitelist.
+	 *
+	 * Errors caused by calls to any of these native PHP functions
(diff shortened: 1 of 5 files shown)


WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
```diff
@@ -0,0 +1,208 @@
...
+	 *
+	 * @return array
+	 */
+	public function register() {
+		$this->empty_tokens                    = Tokens::$emptyTokens;
+		$this->empty_tokens[ \T_NS_SEPARATOR ] = \T_NS_SEPARATOR;
+		$this->empty_tokens[ \T_BITWISE_AND ]  = \T_BITWISE_AND;
+
+		return array(
+			\T_ASPERAND,
```
GaryJones: TIL :-)

WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
```diff
@@ -0,0 +1,208 @@
...
+	 *
+	 * @param int $stackPtr The position of the current token in the stack.
+	 */
+	public function process_token( $stackPtr ) {
+		// Handle the user-defined custom function whitelist.
+		$this->custom_whitelist = $this->merge_custom_array( $this->custom_whitelist, array(), false );
+		$this->custom_whitelist = array_map( 'strtolower', $this->custom_whitelist );
+
+		/*
+		 * Check if the error silencing if done for one of the whitelisted functions.
```
GaryJones: Typo: if -> is

WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
```diff
@@ -0,0 +1,208 @@
...
+		 */
+		$nextNonEmpty = $this->phpcsFile->findNext( $this->empty_tokens, ( $stackPtr + 1 ), null, true, null, true );
```
GaryJones: Shouldn't this variable name be underscored to comply with WP CS?

GaryJones:
Any interest in having this upstream? Even if its just the logic of allowing for a whitelist, and WPCS chooses what functions to whitelist.

WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
jrfnl: Fixed, here and in a few other places.

It's funny, I tend to use snake_case for everything, until I name a variable holding the return value of a pascalCase PHPCS native function call,.. :confused:

jrfnl:
> Any interest in having this upstream? Even if its just the logic of allowing for a whitelist, and WPCS chooses what functions to whitelist.

I have an [open PR upstream](https://github.com/squizlabs/PHP_CodeSniffer/pull/2122) for adding a similar change to the error message (adding the "Found %s" context part).

You can ask upstream if there is interest in adding a whitelist property, and if so, I'd be happy to submit a PR.
The reason I didn't is that:
* I implemented this with a WPCS whitelist + the ability for a user-defined custom whitelist. I very much doubt PHPCS would accept a PR which would go that far. If at all, it would probably be *one* whitelist only.
* I've limited the whitelist to global function calls. In PHPCS, chances are, this will call for discussion on methods or even other type of constructs to whitelist. Having a more detailed/flexible whitelist will quickly become very complicated and probably be rejected for PHPCS (based on my experience so far).

Either way, asking doesn't hurt.

JDGrimes:
I was putting together a comment on the original issue, but hadn't gotten around to posting it yet.

Personally, I don't think I would use this sniff, because I would still tend to avoid using the `@` operator. Even though an error may not be completely avoidable in some cases (because the conditions that would cause the error can't be detected ahead of time), that is generally because an actual error of some kind has occurred. Ideally, such a failure would be handled more gracefully, but I expect that in some cases the error from PHP might be useful for debugging purposes. If `unlink()` fails with an error, for example, that indicates a potential filesystem/permissions issue, and I don't want to silence errors about that.

That is my 2¢ anyway. But perhaps there are circumstances where I would consider using the `@` operator valid. I just don't see the need for using it in all of the proposed cases.

----

I would also like to point out that the `@` operator is indiscriminate, in that it silences error reporting for the entire expression. So in an example like this errors could be silenced that are unrelated to the file operation:

```php
unlink( $this->get_path_to_link() );
```

Because any errors from the `get_path_to_link()` method and any functions that it calls will also be silenced.

jrfnl:
> Personally, I don't think I would use this sniff, because I would still tend to avoid using the @ operator. 
> ... Ideally, such a failure would be handled more gracefully, but I expect that in some cases the error from PHP might be useful for debugging purposes.
>
> I would also like to point out that the @ operator is indiscriminate, in that it silences error reporting for the entire expression.

@JDGrimes Good points and I agree with most of it.

The thing is - sniff warnings/errors should be solvable.
Of course, inline ignore annotations could be used, but to use them for situations which just aren't solvable feels wrong.

And indiscriminately removing the `@` and trusting that error displaying is `off` (possibly error logging `on`) while a lot of WP users wouldn't know how to turn it off on misconfigured servers, goes a bit far too.

Using error silencing is still discouraged for all situations, that hasn't changed.
What I've tried to do is find a balance. Discouraging it still, while not throwing an error about it in the least amount of situations possible. This allows people to get a "clean bill of health" for their code from WPCS when error silencing is used only in "sane" situations.

Experienced PHPCS users could just exclude this sniff in their ruleset and use the upstream one instead if they want to.
But... we could make this even easier by adding a public `$use_whitelist` boolean property to the sniff - default value `true` - which would disable the use of the whitelist if set to `false`.

What do you think about that ?

jrfnl:
> But... we could make this even easier by adding a public $use_whitelist boolean property to the sniff - default value true - which would disable the use of the whitelist if set to false.

Just thinking - maybe the default should be `false`. The `WordPress-Core` ruleset setting it to `true` and the `WordPress-Extra` ruleset (re-)setting it to `false` ?

JDGrimes:
@jrfnl That idea sounds good to me.

jrfnl:
@JDGrimes Okidoki. Added a new commit to add the additional public property. I ended up setting the default to `true` as otherwise, I'd have to overrule it in two different rulesets, now only once.

Tested against the `Core`, `Extra` and `WordPress` rulesets and found working.

jrfnl:
Add one more commit which improves the handling of the error message and metric in the rare case that an end-user would set `context_length` to `0`.

Issue #1330, opened 2018-03-28
Buggy multi-line parameters in function and filter calls


azaozz:
This is a follow-up from #1323. 

It seems that the parser has introduced unwanted changes in single function and filter/action parameters: associative arrays were formatted to be multi-line. This was never intended in the original "Associative arrays should be multi-line" rule. 

Please note: **this is** about a single parameter being on multiple lines. Example 1:
```
do_action(
	'upgrader_process_complete', $this, array(
		'action' => 'update',
		'type'   => 'core',
	)
);
```
**This is not** about having each parameter on a new line. When the parameters are long, having each on a new line is better. Example 2:
```
do_action(
	'upgrader_process_complete', 
	'some long parameter that is easier to read when on a new line',
	'another long parameter that is more readable when on a new line'
);
```
#1323 already deals with partially fixing this. I'd go one step further. The best way to fix this bug would be to discourage or even ban single parameters inside function and filter/action calls being on multiple lines. 

In that terms:
- This should be "illegal":
```
do_action(
	'upgrader_process_complete', $this, array(
		'action' => 'update',
		'type'   => 'core',
	)
);
```
- This should be discouraged, possibly illegal too as it makes the source harder to read:
```
do_action(
	'upgrader_process_complete',
	$this,
	array(
		'action' => 'update',
		'type'   => 'core',
	)
);
```
- This should be the preferred/proper way as it is easiest to read:
```
$param = array(
	'action' => 'update',
	'type'   => 'core',
);

do_action( 'upgrader_process_complete', $this, $param );
```
To fix this we will need:
1. A rule that enforces proper multi-line parameters (see example 2 above). Ideally this will be only when at least one parameter is long, lets say over 50 characters.
2. A rule that flags or prevents single parameters that are on multiple lines.
3. Fix for the associative array rule to make it not apply inside function and filter/action calls. Ideally this will flag cases where an array is used "inline" and suggest setting it to a variable above the function/filter/action call.

GaryJones:
At first glance at least, I could get behind this proposal.

I generally like keeping data away from what you do with that data (as is the example here), though it would often mean introducing single-use variables. 

Extracting the array from how it's used also potentially leads to a refactoring opportunity:

```php
protected function get_params() {
    return array(
    	'action' => 'update',
	    'type'   => 'core',
    );
}

...
do_action( 'upgrader_process_complete', $this, $this->get_params() );
```

[...]

PR #1453, opened 2018-08-17
Core ruleset: change behaviour for the `PEAR/FunctionCallSignature` sniff

jrfnl:
The `allowMultipleArguments` property of the `PEAR.Functions.FunctionCallSignature` sniff defaults to `true`.

By setting it to `false`, the sniff behaviour changes and demands that there is just one parameter on each line in a multi-line function call.
The behaviour for single-line function calls is not affected by this change.

This property has been added to the WP Core native custom ruleset in [changeset 43571](https://core.trac.wordpress.org/changeset/43571) as part of [trac ticket 44600](https://core.trac.wordpress.org/ticket/44600).

This PR now formalizes this by adding the property to the `WordPress-Core` ruleset.

Note: The WP Core handbook still needs to be adjusted to mention this change!

While this is a significant change, IMHO it is not a *breaking* change, but an enhancement, i.e. something on which WPCS previously had no opinion, will now start throwing errors.
With that in mind, I'm targeting this PR for WPCS 1.1.0.

Regarding the removal of the severity changes:
Previously these would already kick in when using the fixer, but would *not* throw an error when just checking the code with `phpcs`.

The impact of these changes is as follows:
Original code:
```php
$a = myfunction( $param1,
    $param2, $param3 );
```

... would previously have been auto-fixed as:
```php
$a = myfunction(
	$param1,
	$param2, $param3
);
```

Code layout which would still be allowed when just the property change would be applied:
```php
$a = myfunction( $param1,
    $param2,
    $param3 );
```

Code layout expected and auto-fixed as, because of the additional removal of the severity changes:
```php
$a = myfunction(
	$param1,
	$param2,
	$param3
);
```

/cc @pento

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -217,13 +217,10 @@
 		<properties>
 			<property name="requiredSpacesAfterOpen" value="1"/>
 			<property name="requiredSpacesBeforeClose" value="1"/>
+
+			<!-- ... and for multi-line function calls, there should only be one parameter per line. -->
+			<property name="allowMultipleArguments" value="false"/>
 		</properties>
-	</rule>
-	<rule ref="PEAR.Functions.FunctionCallSignature.ContentAfterOpenBracket">
-		<severity phpcs-only="true">0</severity>
-	</rule>
-	<rule ref="PEAR.Functions.FunctionCallSignature.CloseBracketLine">
-		<severity phpcs-only="true">0</severity>
 	</rule>
 	<!-- Related to issue #970 / https://github.com/squizlabs/PHP_CodeSniffer/issues/1512 -->
 	<rule ref="WordPress.Functions.FunctionCallSignatureNoParams"/>
--- a/WordPress/PHPCSAliases.php
+++ b/WordPress/PHPCSAliases.php
@@ -61,21 +61,23 @@ if ( ! \defined( 'WPCS_PHPCS_ALIASES_SET' ) ) {
 	 * @link https://github.com/squizlabs/PHP_CodeSniffer/issues/1564
 	 * @link https://github.com/squizlabs/PHP_CodeSniffer/issues/1591
 	 */
-	spl_autoload_register( function ( $class ) {
-		// Only try & load our own classes.
-		if ( stripos( $class, 'WordPress' ) !== 0 ) {
-			return;
-		}
+	spl_autoload_register(
+		function ( $class ) {
+			// Only try & load our own classes.
+			if ( stripos( $class, 'WordPress' ) !== 0 ) {
+				return;
+			}
 
-		// PHPCS handles the Test and Sniff classes without problem.
-		if ( stripos( $class, '\Tests\\' ) !== false || stripos( $class, '\Sniffs\\' ) !== false ) {
-			return;
-		}
+			// PHPCS handles the Test and Sniff classes without problem.
+			if ( stripos( $class, '\Tests\\' ) !== false || stripos( $class, '\Sniffs\\' ) !== false ) {
+				return;
+			}
 
-		$file = dirname( __DIR__ ) . DIRECTORY_SEPARATOR . strtr( $class, '\\', DIRECTORY_SEPARATOR ) . '.php';
+			$file = dirname( __DIR__ ) . DIRECTORY_SEPARATOR . strtr( $class, '\\', DIRECTORY_SEPARATOR ) . '.php';
 
-		if ( file_exists( $file ) ) {
-			include_once $file;
+			if ( file_exists( $file ) ) {
+				include_once $file;
+			}
 		}
-	} );
+	);
 }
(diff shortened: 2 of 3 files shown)


pento:
In the Handbook, I believe the following changes are required:

Under **Indentation**, correct the multiline `new WP_Query()` formatting.

Add a new section, which addresses all of #1330, not just this PR:

> **Multiline Function Calls**
> 
> When splitting a function call over multiple lines, each parameter must be on a seperate line. Blank lines and inline comments can be included, but are not required.
> 
> Parameters must only take up a single line. Multi-line values must be assigned to a variable, then variable can be passed to a function.
> 
> ```php
> $bar = array(
> 	'this',
> 	'is',
> 	'WordPress',
> );
> $baz = sprintf(
> 	'Hello, %s!',
> 	'Friend'
> );
> 
> $a = foo(
> 	$bar,
> 	$baz,
> 
> 	/* translators: %s: cat */
> 	sprintf( __( 'The best pet is a %s.' ), 'cat' )
> );
> ```

Does this new section seem reasonable?

WordPress-Core/ruleset.xml
```diff
@@ -217,14 +217,11 @@
 		<properties>
 			<property name="requiredSpacesAfterOpen" value="1"/>
 			<property name="requiredSpacesBeforeClose" value="1"/>
+
+			<!-- ... and for multi-line function calls, there should only be one parameter per line. -->
+			<property name="allowMultipleArguments" value="false" />
```
pento: Unit tests are [complaining](https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/417092571) about the space at end of this tag. 🙃

jrfnl: Thanks. Fixed up now.

jrfnl:
@pento Regarding the proposed handbook text:

> Blank lines and inline comments can be included

Currently blank lines in a multi-line function call trigger an error and are removed by the fixer. Are we sure we want to change that behaviour ?

> inline comments can be included

Based on your remark in #1330, this should be `_single-line_ inline comments`.

> Parameters must only take up a single line. Multi-line values must be assigned to a variable, then variable can be passed to a function.

What about:
_Each parameters must only take up no more than a single line. Multi-line parameter values must be assigned to a variable and then that variable should be passed to the function call._

> ```php
> $bar = array(
> 	'this',
> 	'is',
> 	'WordPress',
> );
> ```

As the example stands, there is no reason for this array to be multi-line, so the example feel contrived.

What about replacing that part with something like:
```php
$bar = array(
 	'use_this' => true,
 	'meta_key' => 'field_name',
);
```

GaryJones:
> Each parameters must only take up no more than a single line. Multi-line parameter values must be assigned to a variable and then that variable should be passed to the function call.

My amendment would be:

> Each parameter must only be on a single line. Multi-line parameter values must be assigned to a variable first, and then that variable should be used in the function call.

----


```
$baz = sprintf(
	'Hello, %s!',
	'Friend'
);
```

Can this be changed so that the string is internationalised (and optionally escaped) with `__()` / `esc_html__()`, and a translator comment, so that it acts as a direct comparison to the `$a` example?

```
$bar = array(
 	'use_this' => true,
 	'meta_key' => 'field_name',
);
$baz = sprintf(
	/* translators: %s: Friend's name */
	esc_html__( 'Hello, %s!', 'yourtextdomain' ),
	$friend_name
);

$a = foo(
	$bar,
	$baz,

	/* translators: %s: cat */
	sprintf( __( 'The best pet is a %s.' ), 'cat' )
);
```

WordPress/PHPCSAliases.php
```diff
@@ -61,21 +61,23 @@ class_alias( 'PHP_CodeSniffer\Standards\Squiz\Sniffs\WhiteSpace\SemicolonSpacing
 	 * @link https://github.com/squizlabs/PHP_CodeSniffer/issues/1564
 	 * @link https://github.com/squizlabs/PHP_CodeSniffer/issues/1591
 	 */
-	spl_autoload_register( function ( $class ) {
-		// Only try & load our own classes.
-		if ( stripos( $class, 'WordPress' ) !== 0 ) {
-			return;
-		}
+	spl_autoload_register(
+		function ( $class ) {
```
GaryJones: Anonymous functions don't need to be treated as a closure and assigned to a variable first? Becuase this would be an exception to the "parameter must only be on a single line" part of the rule.

jrfnl: That will be dealt with in a separate sniff via a warning - see #1330.

This PR just activates the parts of the `PEAR.Functions.FunctionCallSignature` sniff which were disabled and are step 1 to fixing #1330.

jrfnl:
@pento I'm starting to prepare the WPCS 1.1.0 release. Correct me if I'm wrong, but I think the handbook text has not been updated for the change in this PR yet.
Could we make that happen somewhere in the next few days ?

PR #1457, opened 2018-08-19
GlobalVariablesOverride: detect variable overrides in global namespace

### LINKED ISSUE #1395: GlobalVariableOverride doesn't actually detect override in the global namespace
While looking at the `WordPress.WP.GlobalVariableOverride` sniff and fixing some other things, I suddenly realized the sniff does not actually check for overrides in the global namespace, except for when a `global` statement has been encountered.

Looking at the history, it looks like it never did and I cannot determine whether this was a conscious choice or an oversight.

It would be easy enough to add this. As variables are not affected by namespacing, all that is needed really is a check that the `$variable` is not scoped.
(Ok, some more changes are needed, but that's for the PR)

The only problem I can see is false positives for variables which are in a file (view) which is included from within a function in another file, making the variables local to that function.
This will mainly be a problem in themes.

There are two ways to deal with those false positives:
1. [File based] Add a `// phpcs:disable WordPress.WP.GlobalVariableOverride -- file included from within function abc()` in the file header.
2. [Project ruleset based] Add a public property to the sniff which would enable/disable the check for variable overrides in the global namespace, so people can enable/disable this check project-wide from within their custom ruleset.

So, the question is:
* Should the check for overrides of WP global variables in the global namespace be added ?
* And if so, with or without the public property ? and if with, what should be the default ?
    I'd advocate for enabled, but that's just me (better safe than sorry principle).

Opinions ?

Related #26, #213

jrfnl:
This PR changes the `GlobalVariablesOverride` sniff to always check for variable overrides in the global namespace, independently of there being a `global` statement or not.

While this may throw false positives for theme template files/ plugin view files which are `include`-ed from within a function and therefore scoped to that function, there is no _guarantee_ that the file will _always_ be included from within an function in another file, so using the same variable name as used by WP itself, would still be bad practice.

All the same, I've added a public `treat_files_as_scoped` property which when set to `true` from within a custom ruleset, basically reverts the sniff back to the old behaviour.
The default is `false` and setting this property to `true` should be discouraged.

If - in the future - PHPCS would allow limiting the effect of `// phpcs:set` annotations to the file in which the annotation is used, that annotation could be used in the file header for view/template files to revert to the old behaviour _just and only_ for those particular files. See upstream feature request https://github.com/squizlabs/PHP_CodeSniffer/issues/2126 for more information.

Additional notes/fixes:
* Move the `is_foreach_as()` method from the `PrefixAllGlobals` class to the `Sniff` base class so it can also be used by this sniff.
* Made the sniff more efficient by checking for test classes early and skipping over them completely.
* Made the sniff more efficient by checking early for whitelist comments when examining `T_VARIABLE` tokens.
* Bug fix: when examining a `global` statement, nested funcions/closures/classes should always be skipped. Previously, this was only done when the `global` statement was found in the global namespace.
* Don't throw errors for default values for function parameters in function declarations when the function parameter name would be shadowing a WP global variable or the PHP `$GLOBALS` variable.
* Don't throw errors for class properties.
* Improved the error message when an assignment to $GLOBALS is found
    Previously, the error message would just show the rather non-descript `$GLOBALS`, now it will show `$GLOBALS['array_key']`.

See the individual commits for more detailed information.

Fixes #1395

#### To do once the PR is merged:
- [ ] Add information to the wiki about the new customizable property.


jrfnl:
Anyone have an opinion on this ?

jrfnl:
@WordPress-Coding-Standards/wpcs-admins Anything I can do to move this PR forward ? It is a significant change, so a second review (and if deemed acceptable: approval), would be appreciated.

Issue #1454, opened 2018-08-17
Reorganize issue/PR labels


jrfnl:
This repo currently has the following [labels](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/labels) available: 
* `Component: Docs`
* `Component: Extras`
* `Component: Templates`
* `Component: Upstream`
* `Component: VIP Blocker`
* `Component: VIP Caution`
* `Project: auto-fix WP Core / trac-41057`
* `Requirement: phpcs 3.x`
* `Status: Awaiting Feedback`
* `Status: Duplicate`
* `Status: Invalid`
* `Status: Investigation`
* `Status: Maybe Later`
* `Status: PR Exists`
* `Status: PR Required`
* `Status: Review Ready`
* `Status: Wontfix`
* `Type: Bug`
* `Type: Chores`
* `Type: Cleanup`
* `Type: Documentation`
* `Type: Enhancement`
* `Type: Modern PHP`
* `Type: Question`

I'd like to propose to make the following changes:

### Add labels
* `Component: Core` - While the other rulesets each have a Component label, the `Core` ruleset does not, so this label should be added.
* `Meta` or `Repo: meta` - questions about the repo strategy, repo structure, release schedule, dependency minimum version changes, and for instance, this issue itself, are more "meta" questions and there's currently not a good label available to tag these with.
* `Status: help wanted`, `Status: good first issue` - both of these (without the `Status:` prefix) are standard labels which are created by default nowadays for new repositories. They are widely recognized and help new contributors find issues which could be a good fit for them.

### Rename labels
* `Component: Upstream` - PHPCS upstream is not a component of this repo, so I'd like to propose renaming this label to `PHPCS: upstream`.
* `Requirement: phpcs 3.x` should probably be changed to `Requirement: higher phpcs version` or `Requirement: dependency change` - as that way the same label can continue to be used, even when it would become `phpcs 3.3.x` or `phpcs 4.x` in the future or if more dependencies are added, it can be (re-)used for those too.
    The label basically involves anything for which the minimum supported version of a dependency would need to be raised.

### Remove labels
* `Component: Templates` - No idea why this label was ever added.
* `Component: VIP Blocker` - as the VIP sniffs are slated for removal and the VIP CS checks are now contained in their own dedicated repo, this label is superfluous.
* `Component: VIP Caution` - same as above.
[...]

PR #1458, opened 2018-08-21
Create issue templates

GaryJones:
See #1454

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,6 +2,8 @@ Hi, thank you for your interest in contributing to the WordPress Coding Standard
 
 # Reporting Bugs
 
+Please search the repo to see if your issue has been reported already and if so, comment in that issue instead of opening a new one.
+
 Before reporting a bug, you should check what sniff an error is coming from.
 Running `phpcs` with the `-s` flag will show the name of the sniff with each error.
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,29 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+## Bug Summary
+<!-- Please provide a clear and concise description of what the bug is. -->
+
+## Minimal Code Snippet
+<!-- Please provide example code that demonstrates the issue. -->
+
+## Error Code
+<!-- The error code for the sniff that is being triggered (you can see the sniff error codes by running `phpcs` with the `-s` flag). e.g. `WordPress.PHP.NoSilencedErrors.Discouraged` -->
+
+## Versions
+
+| Item            | Version
+| ----------------| -------
+| WPCS            | x.y.z
+| PHP_CodeSniffer | x.y.z
+| PHP             | x.y.z 
+
+
+## Additional Context (optional)
+<!-- Add any other context about the problem here. -->
+
+## Possible Workaround (optional)
+<!-- If you have discovered a workaround, please include it below. -->
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -9,8 +9,6 @@ about: Create a report to help us improve
 Please provide a clear and concise description of what the bug is.
 
 What did you expect to happen? What actually happened?
-
-For bugs with fixers: How the code was fixed? How did you expect the code to be fixed?
 -->
 
 ## Minimal Code Snippet
@@ -20,11 +18,15 @@ For bugs with fixers: How the code was fixed? How did you expect the code to be 
 // Place your code sample here.
 ```
 
+For bugs with fixers: How was the code fixed? How did you expect the code to be fixed?
+
 ## Error Code
 <!--
 The error code for the sniff that is (or should be) being triggered (you
 can see the sniff error codes by running `phpcs` with the `-s` flag).
 e.g. `WordPress.PHP.NoSilencedErrors.Discouraged`
+
+You can leave this section empty if you are reporting a false negative.
 -->
 
 ## Environment
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -16,7 +16,7 @@ For bugs with fixers: How the code was fixed? How did you expect the code to be 
 ## Minimal Code Snippet
 <!-- Please provide example code that demonstrates the issue. Do NOT paste screenshots of code! -->
 
-```
+```php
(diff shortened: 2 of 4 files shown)


GaryJones:
> If you have to choose a template, maybe we should also have a "Generic" template which is mostly empty ?

I'd rather not, simply because it could be used lots by non-committers, which means we then don't get a chance to prompt for the info we really need, which then misses the point of having issue templates in the first place.

The proposed templates can still be amended, but if we're wanting non-committers to use them, then committers should be accepting of using them as well.

> "Please search the repo to see if your issue has been reported already and if so, comment in that issue instead of opening a new one."

That *should* be standard behaviour (but we know it's not). The corollary of adding it though is that folks may then start commenting on closed Issues, or on Issues that are only vaguely (or not at all) related to their issue. One way or another, there's going to be a level of Issue administration, and I don't think adding it to the top of each Issue template will do much. 

Saying that, I've added it into the `CONTRIBUTING.md` document.



> False negatives could relate to a sniff, but the errorcode may not be obvious to the user reporting them.

I've amended the wording slightly in relation to the provision of an error code.

> Should this template maybe also allow for suggestions to up the dependency with arguments of why the minimum version of the dependency should change ?

That's going to be rare enough, that it can count as a feature request issue. That already asks for what their problem is ("I want to use feature X in PHPCS), and what their proposed solution ("Bump the min PHPCS version to ...").



> This template feels very generic and not aimed at the WPCS project.

The wording is pretty much the default wording when using the Issue Template Maker UI.

> Any (sniff) feature request/suggestion should normally be accompanied by code samples of what should be detected.
> And preferably also code samples of code which shouldn't be flagged.

I've updated the wording to add this in.

> Sometimes, a feature has been added in a new WPCS/PHPCS version and people are just using an outdated version.

Maybe, but I think we should handle that on a case by case basis, rather than asking every new feature request to state what their current setup is.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,45 @@
...
+
+What did you expect to happen? What actually happened?
+
+For bugs with fixers: How the code was fixed? How did you expect the code to be fixed?
+-->
+
+## Minimal Code Snippet
+<!-- Please provide example code that demonstrates the issue. Do NOT paste screenshots of code! -->
+
+```
```
jrfnl: The backticks + php ( or js or css ) will create the syntax highlighting. Without the language indication, it will just be a plain grey code block.

jrfnl:
I've just used the "Dependency Change" template to create issue #1477

Feedback based on first usage of it:
* Making the name of the dependency and the version number **bold** feels natural, so might be good to add to the template.
* Aside from the dependency name and version, the first section is quite non-descript and as that info is also in the title (or should be according to the instructions in the template), it doesn't actually add any value.
* I very much missed a block to give a rationale of why the change should be made in WPCS + a block to add reference links.
    Even if no rationale is given, reference links - like a link to the changelog of the version which contains feature or the upstream issue which requested it - should IMHO always accompany a dependency change request.
    I've now added these two blocks myself anyway, but having them below the action list feel counter-intuitive.

GaryJones:
> Feedback based on first usage of it

I've updated the PR for this feedback.

.github/CONTRIBUTING.md
```diff
@@ -2,6 +2,8 @@ Hi, thank you for your interest in contributing to the WordPress Coding Standard
 
 # Reporting Bugs
 
+Please search the repo to see if your issue has been reported already and if so, comment in that issue instead of opening a new one.
```
jrfnl: :+1:

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,45 @@
...
+
+---
+
+## Bug Description
+<!--
+Please provide a clear and concise description of what the bug is.
+
+What did you expect to happen? What actually happened?
+
+For bugs with fixers: How the code was fixed? How did you expect the code to be fixed?
```
jrfnl: I think this might be better placed in the `code snippet` section.
Also: "_How **was** the code ~~was~~ fixed_"

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,45 @@
...
+```php
+// Place your code sample here.
+```
+
+## Error Code
+<!--
+The error code for the sniff that is (or should be) being triggered (you
+can see the sniff error codes by running `phpcs` with the `-s` flag).
+e.g. `WordPress.PHP.NoSilencedErrors.Discouraged`
+-->
```
jrfnl: Maybe add: "_You can leave this section empty if you are reporting as false negative_" ?

.github/ISSUE_TEMPLATE/dependency-change.md
```diff
@@ -0,0 +1,21 @@
...
+
+<!-- PLEASE prefix the title the Issue with the dependency name and version when action should be taken e.g. PHPCS 3.3: ... -->
+
+## Rationale
+
+<!-- Why should this change be made WPCS? -->
+
+## References
+
+<!-- References, like a link to the changelog of the version which contains feature or the upstream issue which requested it -->
```
jrfnl:
```suggestion
<!-- References, like a link to the changelog of the version which contains a feature or the upstream issue which requested it -->
```

.github/ISSUE_TEMPLATE/dependency-change.md
```diff
@@ -0,0 +1,21 @@
...
+name: Dependency Change
+about: A reminder to take action when a WPCS dependency changes
+
+---
+
+<!-- PLEASE prefix the title the Issue with the dependency name and version when action should be taken e.g. PHPCS 3.3: ... -->
+
+## Rationale
+
+<!-- Why should this change be made WPCS? -->
```
jrfnl:
```suggestion
<!-- Why should this change be made in WPCS? -->
```

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Describe the solution you'd like
+<!--
+A clear and concise description of what you want to happen.
+
+Any (sniff) feature request/suggestion should be accompanied by code samples of what should be detected.
+And preferably also code samples of code which shouldn't be flagged.
+-->
+
+## Additional context
```
jrfnl:
```suggestion
## Additional context (optional)
```

GaryJones:
Updated per today's feedback.

jrfnl:
Looks like you'll need to rebase the PR so the build can pass.

GaryJones:
Rebased.

PR #1459, opened 2018-08-22
ValidFunctionName: Add additional unit test for interfaces extending multiple interfaces

jrfnl:
Interfaces can extend multiple other interfaces. Looking at the sniff code, no changes are needed in the sniff to account for this, however, adding an additional unit test to warrant against sniff changes in the future which may not take this into account.

N.B.: I've verified all other usages of the `T_INTERFACE` token in the whole of WPCS and this was the only one which looked at `extends/implements`, so the only one for which interfaces extending multiple other interfaces could be problematic.

See:
* http://php.net/manual/en/language.oop5.interfaces.php#example-208

Loosely related to:
* https://github.com/squizlabs/PHP_CodeSniffer/pull/2128

### DIFF
--- a/WordPress/Tests/NamingConventions/ValidFunctionNameUnitTest.inc
+++ b/WordPress/Tests/NamingConventions/ValidFunctionNameUnitTest.inc
@@ -104,3 +104,8 @@ class My_Soap extends somethingElse {
 }
 
 function __MyTemplateTags() {} // Bad x 2.
+
+// Verify that interfaces which extend multiple other interfaces do not throw unnecessary errors.
+interface MyInterface extends InterfaceA, InterfaceB {
+	function __MyTemplateTags(); // OK.
+}


jrfnl:
Oh cricky... turns out support for interfaces was only added to the PHPCS `File::findExtendedClassName()` method in PHPCS 3.1.0.

Considering that we'll be dropping PHPCS 2.x support in the near future and there have been no bug reports, I don't think it's worth adding a fix to the sniff to make this work with PHPCS 2.x.

So, I'm going to go ahead and change the milestone to `2.0.0`. Once work on that release has started and the PHPCS minimum supported version has been raised, the PR can be rebased and the build should then pass without problem.

jrfnl:
I've rebased this PR now that the "drop PHPCS 2.x" PR has been merged. I expect the build to pass now, so this PR is ready for review.

jrfnl:
Oops.. I seem to have pulled this from my own fork, so updating the PR got a bit screwed up. Never mind. This PR is now replaced by #1562

PR #1464, opened 2018-08-25
PrecisionAlignment: prevent false positive for inline javascript comments

jrfnl:
Inline HTML can contain javascript `<script>` blocks, which in turn can contain multi-line `/* */` style comments.

This sniff should not throw errors for those.

The implemented solution is quite simplistic, but determining whether or not we are within a `<script>` block can be quite complicated as - what with going into and out of PHP - the opener or closer may be in a PHP `echo` statement.

The simple solution implemented now may cause some false negatives, but that is preferable over false positives in this case.

Includes unit test.


jrfnl:
@WordPress-Coding-Standards/wpcs-admins Anything I can do to move this PR forward ?

PR #1466, opened 2018-08-27
AlternativeFunctions: suggest using wp_(un)slash()

jrfnl:
The WP `wp_slash()` and `wp_unslash()` functions are capable of (un)slashing both strings, as well as (nested) arrays, while the PHP native functions can only handle strings.

For input variables, you typically encounter arrays in forms. For instance when a checkbox or dropdown field can accept multiple values. however, the usage of arrays in input variables is definitely not limited to those examples.

With that in mind, I'd like to suggest adding a recommendation to use these functions instead of the PHP native ones to the `AlternativeFunctions` sniff.

### DIFF
--- a/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
+++ b/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
@@ -121,6 +121,25 @@ class AlternativeFunctionsSniff extends AbstractFunctionRestrictionsSniff {
 					'mt_rand',
 				),
 			),
+
+			'addslashes' => array(
+				'type'      => 'warning',
+				'message'   => '%s() is discouraged. Use wp_slash() for improved handling of complex variables instead; or in the context of database queries: use prepared statements.',
+				'since'     => '3.6.0',
+				'functions' => array(
+					'addslashes',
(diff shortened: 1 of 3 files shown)


jrfnl:
@WordPress-Coding-Standards/wpcs-admins Anyone has an opinion on this proposal ?

westonruter:
I agree. I think the original purpose being `wp_unslash()` and `wp_slash()` was that in the eventual case that magic quotes could be turned off, that these functions could then no-op. I doubt that will ever happen, but if it does, using these functions will ensure that things don't break.

JDGrimes:
The other side of what @westonruter pointed out is that these functions should really only be used in the context of dealing with magic quotes. Unrelated slashing and unslashing operations shouldn't use them.

jrfnl:
> Unrelated slashing and unslashing operations shouldn't use them.

Well, for slashing/unslashing related to database operations, the functions shouldn't be used either as prepared SQL should be used instead.

Let's gather some examples of situations in which it *would* be preferred to use the PHP native functions over the WP versions. Depending on the examples, either the sniff could try to account for the context in which the functions are used or this PR will be closed.

Issue #763, opened 2017-01-01
Make best practice sniffs code-style independent


jrfnl:
There are a number of sniffs which currently walk the token tree presuming that code complies with the WP code style rules.
The net end-result of this is that the messages from these sniffs will not be thrown for code which does not comply with the WP code style rules.
As VIP plugins as well as themes generally don't have to comply with the WP code style, but *do* have to comply with a lot of best practices, this means that important messages might be missed.

As a secondary result, it also means that an initial run against the WP rulesets might not show all messages and new messages might start to get thrown *only* once code style issues have been fixed.

I would suggest we should review all (non-code style specific) sniffs for these kind of issues and improve them where necessary.

Particular attention should be paid to:
* Object/Function/Namespace name compares using `strpos()`, `preg_match()` or a strict compare. These are all case-insensitive in PHP so should be compared as such.
* If constant name compares are being done and the `define()` function is checked, it should be checked that the third parameter `$case_insensitive` is either not passed or passed a `false`, otherwise a case-insensitive name check is needed.
* Sniffs doing token "calculations" ( `$token + 1` ) instead of using `findNext/Previous()` in combination with `Tokens::$emptyTokens`

jrfnl:
I believe this has been addressed by now in most sniffs in the codebase, safe for one or two. The `AbstractArrayAssignmentRestrictionsSniff` comes to mind as one of the exceptions.

I'll keep an eye out for these kind of things when going through the sniffs anyway to upgrade them for Utils/PHP 8.x, but suggest we close this issue as "fixed".

PR #1468, opened 2018-08-28
Sniff::is_in_isset_or_empty(): improve code-style independence

jrfnl:
Includes unit tests in the ValidatedSanitizedInput test case file.

Relates to #763

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -1505,7 +1505,8 @@ abstract class Sniff implements PHPCS_Sniff {
 		end( $nested_parenthesis );
 		$open_parenthesis = key( $nested_parenthesis );
 
-		return \in_array( $this->tokens[ ( $open_parenthesis - 1 ) ]['code'], array( \T_ISSET, \T_EMPTY ), true );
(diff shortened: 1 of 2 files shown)


WordPress/Sniff.php
```diff
@@ -1505,7 +1505,8 @@ protected function is_in_isset_or_empty( $stackPtr ) {
 		end( $nested_parenthesis );
 		$open_parenthesis = key( $nested_parenthesis );
 
-		return \in_array( $this->tokens[ ( $open_parenthesis - 1 ) ]['code'], array( \T_ISSET, \T_EMPTY ), true );
```
GaryJones: So when looking for code-style assumptions, a good starting point is `$this->tokens[ $var +/- X ]`?

jrfnl: Yes and no. doing that calculation is very common and correct when it's to set the starting point for a `findNext()`/`findPrevious()` function call.
Code style assumptions are only problematic for best practice sniffs, not code style sniffs and to bypass those assumption you need to take every crappy way of writing PHP into account, so looking for checks using `T_WHITESPACE` is another typical search entry point. This should - for code style independence - be `Tokens::$emptyTokens`.

jrfnl:
@WordPress-Coding-Standards/wpcs-admins Anything I can do to move this PR forward ?

PR #1470, opened 2018-08-28
Release version 1.1.0

jrfnl:
:warning: **DO NOT MERGE (YET)** :warning: 

PR for tracking changes for the 1.1.0 release. Suggested target release date: ~~Monday September 3~~ Wednesday September 5.
- [x] Add changelog for 1.1.0
- [ ] Merge this PR
- [ ] Add release tag & copy & paste the changelog to it
- [ ] Close the 1.1.0 milestone
- [x] Open a new milestone for the next release
- [x] If any open PRs which were milestoned for 1.1.0 do not make it into the release, update their milestone.


.travis.yml
```diff
@@ -4,6 +4,11 @@ dist: trusty
 
 cache:
   apt: true
+  directories:
+    # Cache directory for older Composer versions.
```
GaryJones: Travis always does a `composer self-update` - so why is the old directory needed?

jrfnl: That question should probably have been asked when the PR adding this was pulled....

GaryJones: I hadn't studied how Travis works much at that point. Now I have :-)

jrfnl: Ok, well, my take on this is quite simple:
* Having the extra directory here makes no difference for Travis - if there is nothing in it, nothing will be cached anyway.
* Whether or not the `composer self-update` command should be part of the script is debatable. I'm on the "let's remove it" side as:
    1. Travis generally updates their images regularly enough that the version of Composer shipped with it will not be that much out of date anyway.
    2. If Travis - now or in the future - uses different versions of Composer on images targeting different PHP versions, all the better, as that means our set-up and composer validation would be tested on different versions. Self-update removes that testing.
* If the self-update command would be removed from the script, having both directories in the cache list is the better setup.

GaryJones: > Whether or not the composer self-update command should be part of the script is debatable. I'm on the "let's remove it" side as:

The `composer self-update` I was referring to is [part of the environment setup](https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/421374075#L439), well before our `before_install` scripts start kicking in. We can't remove their `composer self-update`, which is why the old extra path, I believe, is completely redundant.

jrfnl: Ah, ok, that might actually be a semi-recent thing as I distinctly remember that Travis "in the olden days" (which may only be a couple of months ago) didn't do that and you had to add the command to your own script if you wanted to be sure to use the current version of Composer.

jrfnl:
@WordPress-Coding-Standards/wpcs-admins All, I'd like to release version 1.1.0 tomorrow, as in Monday September 10th (GMT).

To be able to do so, I need the following:
* One more approval for the changelog PR #1476 
* Two approvals for this PR
* The WP Core handbook to have been updated for the change made in PR #1453

Can we make this happen within the next +/- 24 hours ?

pento:
The Core handbook has been updated for #1453.

jrfnl:
WPCS 1.1.0 has been released:
Release: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/releases/tag/1.1.0
Release tweet: https://twitter.com/jrf_nl/status/1039199312448761857

Thank you all for working with me to get this done and especially @JDGrimes for the reviewing and merges this weekend.

Issue #1269, opened 2017-12-27
Purpose and future of WordPress-Extra and WordPress-Core


JDGrimes:
> This ticket is a split from #1157 as suggested in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1157#issuecomment-351753423.

## History

In the beginning, [there was just one `WordPress` ruleset](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/tree/04fd547c691ca2baae3fa8e195a46b0c9dd671c5). In #58 it was proposed that the sniffs be split into several subsets (something similar was also proposed in #42):

>A lot of the sniffs are opinionated and may not be reflective of an actual problem. We should have a ruleset XML file that just contains the core sniffs for checking code formatting, and then another ruleset XML that has the extra sniffs (e.g. those labeled with `extra` in GitHub)

The `WordPress-Core` ruleset was created in #64 to be a set of "Non-controversial generally-agreed upon WordPress Coding Standards", and in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/commit/8409c2748e4b95eb6aa9d027e5cf634518886f13#diff-4486463b25768f719431bc763baab473 `WordPress-Extra` was introduced, as "Best practices beyond core WordPress Coding Standards".

Ever since, the `Core` ruleset has essentially reflected the [standards outlined in the handbook](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/), while `Extra` contains additional "best practices".

## Present Discussion

In #1157 it was proposed to split formatting and non-formatting rules into their own subsets. But in the process of the discussion, there was soon talk of renaming `Extra` and `Core` while we were at it.

The main relevant issues raised are:

1. The name `Extra` may not be ideal.
2. The `Extra` ruleset is just a catch-all for stuff not in `Core`.
3. The term `Core` implicitly ties the standard too closely to a single project, instead of the community standards in the handbook.

### Confusion around `Extra`

One reason this came up is because the `Extra` ruleset is sometimes confusing.

>To be honest, I never understood the Extra ruleset. To me, this always sounded like stuff that doesn't really belong somewhere else, or rather: you are not sure about where it does belong. https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1157#issuecomment-331635949

[...]

Issue #1067, opened 2017-07-26
New PSR2 + WP best practices ruleset


jrfnl:
A recent article by @tommcfarlin : https://tommcfarlin.com/using-psrs-in-wordpress-development/ got me thinking about PSR2 vs the WordPress Coding Standards.
I could easily create a ruleset which would allow devs to use PSR2 while still getting the benefits of the WP best practice sniffs and share it somewhere (separate repo/gist).

But what about potentially adding this as an extra ruleset here ? Not to replace any of the current rulesets, just as an alternative ruleset for non-core WordPress-related development.

Opinions ?

Related issues: #763 and #740

dingo-d:
I don't think that it should be added as an extra ruleset since PSR standards are not allowed in core (https://core.trac.wordpress.org/ticket/23357).

For any non-standard development, users can just create their own custom rulesets, no?

jrfnl:
@dingo-d Just checking: you did read that I said _an alternative ruleset for non-core WordPress-related development_ ?

grappler:
There are two ways to go ahead.

1. Create a new ruleset `WordPress-PSR2` with WordPress best practices and PSR2 coding style
1. Separate WordPress best practices from coding style #740. Then users could use `WordPress-Core-Best-Practices`  and `PSR2` without needing a new ruleset.

I think option 2 may be better. As the WordPress Community we should encourage that WordPress coding standards are used but still making it easy for people to use other coding styles.

Though I do same my files different in plugins to match a more modern PHP naming convention then what WordPress has now.

dingo-d:
@jrfnl Yes I did read it, and I still don't see the merit for it, since users can create their own ruleset in any way they want :)

Aside from helping people, who want to include some PSR coding standards with WordPress one, out, I don't know why would this need to be added as a separate ruleset. Better add it to a wiki page, showing users how to achieve that. 

That's just my opinion, didn't mean anything bad, just think that it's not that important :)

JDGrimes:
[...]

Issue #740, opened 2016-12-24
Splitting the Core ruleset into Code style & Core


jrfnl:
Following up on #676:

I've made a start at splitting the ruleset, but would like a second opinion on what should go where.

N.B. the below list includes rules which are currently not covered yet!

### Current proposal:

Proposal | Section | Sub-section | Reasoning
-------- | ------- | ----------- | ---------
Code Style | [Single and Double Quotes](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#single-and-double-quotes) | quote style |
Core | [Single and Double Quotes](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#single-and-double-quotes) | use of esc_attr |
Code Style | [Indentation](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation) | |
Code Style | [Brace Style](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style) | |
Code Style | [Use elseif, not else if](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#use-elseif-not-else-if) | |
Core | [Regular Expressions](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#regular-expressions) | PCRE not POSIX | 
Core | [Regular Expressions](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#regular-expressions) | /e switch | 
Code style | [Regular Expressions](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#regular-expressions) | single quotes | 
Core + Code style | [No Shorthand PHP Tags](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#no-shorthand-php-tags) | | Can lead to bugs cross-PHP version.
Code style | [Remove Trailing Spaces](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#remove-trailing-spaces) | |
Code style | [Space Usage](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#space-usage) | |
Code style | [Formatting SQL statements](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#formatting-sql-statements) | capitalize SQL parts |
Core | [Formatting SQL statements](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#formatting-sql-statements) | update vs slashes |
[...]

PR #1475, opened 2018-08-31
Start to separate coding standards and best practices to their own rulesets

javorszky:
Base ruleset is `WordPress-Core`.

Thinking:
- is it something that is covered by PSR1 or PSR2? -> coding style
- is it something that isn't WP specific, isn't covered by PSR1 or PSR2, but relates to how the code looks and other code styles may or may not have a rule to it? -> coding style
- everything else is a best practice

Have a look at the files. I'll go through WordPress-Extra and separate the extras / overrides from that onto either of these. Not sure how to handle the layered approach to be honest.

Would a WordPress-Extra-CodingStyle and WordPress-Extra-BestPractices make sense that just imported the previous ones?

Won't that fragment the rulesets?

### DIFF
--- a/WordPress-BestPractices/ruleset.xml
+++ b/WordPress-BestPractices/ruleset.xml
@@ -1,0 +1,221 @@
+<?xml version="1.0"?>
+<ruleset name="WordPress Core">
+	<description>Non-controversial generally-agreed upon WordPress Coding Standards</description>
+
+	<autoload>./../WordPress/PHPCSAliases.php</autoload>
+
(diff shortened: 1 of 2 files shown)


jrfnl:
Hi @javorszky Thank you for your willingness to contribute and for this PR.

Unfortunately I will close this at this time as this has been discussed several times before - see issues #740, #1067, #1157, #1269 - without consensus being reached.
I would kindly like to ask you to contribute and continue the discussions in those issue(s). If and when consensus would be reached and this PR would be in line with it, we can always re-open the issue.

PR #1492, opened 2018-09-23
PrefixAllGlobals: add "WordPress" to the prefix blacklist

jrfnl:
While the word "WordPress" can be used as _part of a prefix_, the word on its own should not be allowed as a prefix.

Includes unit test.

### DIFF
--- a/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
+++ b/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
@@ -49,8 +49,9 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	 * @var string[]
 	 */
 	protected $prefix_blacklist = array(
-		'wp' => true,
-		'_'  => true,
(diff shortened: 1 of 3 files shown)


jrfnl:
> How would this affect plugins whose slug included wordpress at the start? e.g. http://wordpress.org/plugins/wordpress-seo ?

@GaryJones Plugins with a slug starting with `WordPress` (still) shouldn't use `WordPress` as their prefix. For the example you give, the prefix could, for instance, be `WordPress_SEO` and that would be fine.

The blacklist - currently (including this PR): `_`, `wp` and `wordpress` - only checks that none of the prefixes passed is on the blacklist (case-insensitively), it doesn't check whether a prefix *starts* with or *contains* the strings in the blacklist.
That is how the blacklist behaviour has been from the start and this PR doesn't change that.

jrfnl:
And then I was suddenly wondering why the sniff would not throw an error for the function declaration with the blacklisted prefix.... found the cause & fixed it in a separate commit.
Commits can (should) be squashed on merge with the first commit message remaining.

PR #1503, opened 2018-10-14
Build/Travis: test builds against PHP 7.3

jrfnl:
Once PHP 7.3-beta came out, the `nightly` build on Travis became PHP 7.4-dev and builds haven't been tested against PHP 7.3 for months now.

Luckily, Travis has *finally* deemed it appropriate to set up a PHP 7.3 alias now RC3 is out, so I've added  PHP 7.3 to the matrix.

PHP 7.3 in combination with PHPCS < 3.3.1 should be allowed to fail as PHPCS wasn't fully compatible with PHP 7.3 until version 3.3.1.

Note: The builds will only fail on a few select (test) cases, but making exceptions for those specific cases, while we're going to be dropping support for earlier PHPCS versions in the near future anyway, seems superfluous. Allowing the builds to fail for those PHP/PHPCS combinations is the simpler solution for now.

See squizlabs/PHP_CodeSniffer#2086 for the details of the fixes made to PHPCS itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ php:
     - 7.0
     - 7.1
     - 7.2
+    - 7.3
     - nightly
 
 env:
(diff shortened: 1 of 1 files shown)


jrfnl:
Note: the 7.3/master build is passing. The 7.2/master build is failing, but the fix for that is in PR #1498

PR #1516, opened 2018-11-01
Sniff::is_class_property: bug fix

jrfnl:
When a class would be nested in parenthesis, the method did not recognize class properties properly.

To check for this without code duplication, the return of the `Sniff::valid_direct_scope()` method has changed. It used to always return a boolean. Now it will return `false` if not in a valid direct scope and the `$stackPtr` to the scope if it is.

Just like the rest of these methods, unit tested in the PHPCompatibility standard.

See also these issues for the same fix elsewhere:
* PHPCompatibility/PHPCompatibility#758
* squizlabs/PHP_CodeSniffer#2214

This fixes the false positive for line 124 in the `GlobalVariablesOverrideUnitTest.1.inc` test file.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -1960,7 +1960,7 @@ abstract class Sniff implements PHPCS_Sniff {
 			'T_ANON_CLASS' => true,
 			'T_TRAIT'      => true,
 		);
-		if ( true === $this->valid_direct_scope( $stackPtr, $valid_scopes ) ) {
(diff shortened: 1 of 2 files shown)


GaryJones:
> When a class would be nested in parenthesis, the method did not recognize class properties properly.

Is there no unit test for this?

jrfnl:
> Is there no unit test for this?

As I stated in the issue description:

> Just like the rest of these methods, unit tested in the PHPCompatibility standard.

Additionally, the existing unit test covers this:
> This fixes the false positive for line 124 in the GlobalVariablesOverrideUnitTest.1.inc test file.

PR #1525, opened 2018-11-06
New `I18nTextDomainFixer` utility sniff

### LINKED ISSUE #1174: Add fixer for TextDomainMismatch
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/d62b34082bf1921c0896ac7a0b6f32a85fe9a576/WordPress/Sniffs/WP/I18nSniff.php#L369

(and a repeat a few lines down) identifies when a text domain does not match what it should.

Since we have the text domain string in the function call, and the allowed text domain(s), could a fixer be added when there is only a single intended domain that does the replacement?

jrfnl:
This sniff can replace the text-domain used in a code-base.
The sniff will fix the text-domains in both I18n function calls as well as in a plugin/theme header.
The sniff is disabled by default.

Passing the following properties will activate the sniff:
- `old_text_domain`: an array with one or more (old) text-domain names which need to be replaced;
- `new_text_domain`: the correct (new) text-domain as a string.

Notes:
* The sniff will do a comprehensive find & replace for the text-domains in *all* internationalization related functions, including deprecated/internal use functions and the `load_text_domain()` functions.
* The sniff will add the (new) text-domain if it is missing and all the preceding function call arguments are in place.
     If any of the preceding function call arguments is (also) missing, the sniff will throw a warning, but will not run the fixer.
* For a scaffold plugin/theme which doesn't use placeholder text-domains to use this sniff to add the text-domain, the `old_text_domain` property *does* still need to be passed.
     In that case, just putting anything in it, will do, i.e. `<property name="old_text_domain" type="array" value="ignore"/>`.
* The `default` text-domain is not allowed as the "new" text-domain as it is reserved for WP Core.
* The sniff will search for the plugin/theme header and fix the `Text Domain: ...` header if found, or add it, if it is missing.
    For a comment to be recognized as the plugin/theme header:
    - At least one required header has to be present.
    - At least three of the possible headers have to be present.
* The sniff will examine all comments to find the plugin/theme header, with these caveats:
    - For CSS, It will only examine the file if it is called `style.css`.
    - It will keep track of whether or not the plugin/theme header has been found already and bow out from examining further comments as soon as it has found the header.
        This is an efficiency mechanism as examining all comments in a code base while we're only looking for one particular one, would make the sniff unnecessarily slow.
        This also means that running this sniff over two or more plugins/themes in one go will not work, at least not for the fixing of the plugin/theme header.
     - The sniff does not know in advance whether a plugin or a theme is being examined, so it will look for the header in every file (save non-`style.css` CSS files).
    - For the unit test files, an exception is build in to prevent us having to have 20+ unit test files just to test the recognition of the headers.

Fixes #1174

### DIFF
--- a/WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
+++ b/WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
@@ -1,0 +1,725 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPress\Sniffs\Utils;
+
+use WordPress\AbstractFunctionParameterSniff;
+use PHP_CodeSniffer_Tokens as Tokens;
+
+/**
+ * Comprehensive I18n text domain fixer tool.
+ *
+ * This sniff can:
+ * - Add missing `text-domain`s.
+ * - Replace `text-domain`s based on an array of `old` values to a `new` value.
+ *
+ * Note: Without a user-defined configuration in a custom ruleset, this sniff will be ignored.
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   1.2.0
+ */
+class I18nTextDomainFixerSniff extends AbstractFunctionParameterSniff {
+
+	/**
+	 * A list of tokenizers this sniff supports.
+	 *
+	 * @since 1.2.0
+	 *
+	 * @var array
+	 */
+	public $supportedTokenizers = array(
+		'PHP',
+		'CSS',
+	);
+
+	/**
+	 * Old text domain(s) to replace.
+	 *
+	 * @since 1.2.0
+	 *
+	 * @var string[]|string
+	 */
+	public $old_text_domain;
+
+	/**
+	 * New text domain.
+	 *
+	 * @since 1.2.0
+	 *
+	 * @var string
+	 */
+	public $new_text_domain = '';
+
+	/**
+	 * The group name for this group of functions.
+	 *
+	 * @since 1.2.0
+	 *
+	 * @var string
+	 */
+	protected $group_name = 'i18nfixer';
+
+	/**
+	 * The WP Internationalization related functions to target for the replacements.
+	 *
+	 * @since 1.2.0
+	 *
+	 * @var array <string function name> => <int parameter position>
+	 */
+	protected $target_functions = array(
+		'load_textdomain'                        => 1,
+		'load_plugin_textdomain'                 => 1,
+		'load_muplugin_textdomain'               => 1,
+		'load_theme_textdomain'                  => 1,
+		'load_child_theme_textdomain'            => 1,
+		'unload_textdomain'                      => 1,
+
+		'__'                                     => 2,
+		'_e'                                     => 2,
+		'_x'                                     => 3,
+		'_ex'                                    => 3,
+		'_n'                                     => 4,
+		'_nx'                                    => 5,
+		'_n_noop'                                => 3,
+		'_nx_noop'                               => 4,
+		'translate_nooped_plural'                => 3,
+		'_c'                                     => 2, // Deprecated.
+
+		'esc_html__'                             => 2,
+		'esc_html_e'                             => 2,
+		'esc_html_x'                             => 3,
+		'esc_attr__'                             => 2,
+		'esc_attr_e'                             => 2,
+		'esc_attr_x'                             => 3,
+
+		'is_textdomain_loaded'                   => 1,
+		'get_translations_for_domain'            => 1,
+
+		// Shouldn't be used by plugins/themes.
+		'translate'                              => 2,
+		'translate_with_gettext_context'         => 3,
+
+		// WP private functions. Shouldn't be used by plugins/themes.
+		'_load_textdomain_just_in_time'          => 1,
+		'_get_path_to_translation_from_lang_dir' => 1,
+		'_get_path_to_translation'               => 1,
+	);
+
+	/**
+	 * Whether a valid new text-domain was found.
+	 *
+	 * @since 1.2.0
+	 *
+	 * @var string
+	 */
+	private $is_valid = false;
+
+	/**
+	 * The new text-domain as validated.
+	 *
+	 * @since 1.2.0
+	 *
+	 * @var string
+	 */
+	private $validated_textdomain = '';
+
+	/**
+	 * Whether the plugin/theme header has been seen and fixed yet.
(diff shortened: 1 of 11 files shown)


WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
```diff
@@ -0,0 +1,725 @@
...
+		'__'                                     => 2,
+		'_e'                                     => 2,
+		'_x'                                     => 3,
+		'_ex'                                    => 3,
+		'_n'                                     => 4,
+		'_nx'                                    => 5,
+		'_n_noop'                                => 3,
+		'_nx_noop'                               => 4,
+		'translate_nooped_plural'                => 3,
+		'_c'                                     => 2, // Deprecated.
```
GaryJones: [`ngettext`](https://developer.wordpress.org/reference/hooks/ngettext/), [`_nc`](https://developer.wordpress.org/reference/functions/_nc/) (deprecated), [`__ngettext`](https://developer.wordpress.org/reference/functions/__ngettext/) (deprecated), and maybe others, should be included?

WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
```diff
@@ -0,0 +1,725 @@
...
+		'translate_with_gettext_context'         => 3,
+
+		// WP private functions. Shouldn't be used by plugins/themes.
+		'_load_textdomain_just_in_time'          => 1,
+		'_get_path_to_translation_from_lang_dir' => 1,
+		'_get_path_to_translation'               => 1,
+	);
+
+	/**
+	 * Whether a valid new text-domain was found.
```
GaryJones: Let's keep all instance (at least in comments) as two words: "text domain". Not a single word, not hyphenated.

WordPress/Tests/Utils/I18nTextDomainFixerUnitTest.3.inc
```diff
@@ -0,0 +1,119 @@
...
+
+// @codingStandardsChangeSetting WordPress.Utils.I18nTextDomainFixer old_text_domain a2-optimized
+
+//  Plugin Name: Text domain which should be fixed, multi-line //-style comment.
+//  Plugin URI: https://wordpress.org/plugins/a2-optimized/
+//  Version: 2.0.1
+//  Author: A2 Hosting
+//  Author URI: https://www.a2hosting.com/
+//  Description: A2 Optimized - WordPress Optimization Plugin
+//  Text Domain: a2-optimized
```
GaryJones: Can one of these tests have more or less whitespace after the `Text Domain:`, just to show it is whitespace-agnostic?

WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
jrfnl: > ngettext, _nc (deprecated), __ngettext (deprecated), and maybe others, should be included?

Good call on the last two. Updating the PR as we speak.
Thralling through the `wp-includes/deprecated.php` file which contained these, yielded me two more: [`translate_with_context()`](https://developer.wordpress.org/reference/functions/translate_with_context/) and [`__ngettext_noop`](https://developer.wordpress.org/reference/functions/__ngettext_noop/).

If anyone can find yet more, please comment.

As for the first - that is filter, not a function and I would not expect a plugin/theme to contain a call to such a specific WP core native filter, so I'm not sure we should account for that.
What do you think ?

WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
jrfnl: Should I also check for this in the `I18n` sniff and the documentation, i.e. changelog and such ?

WordPress/Tests/Utils/I18nTextDomainFixerUnitTest.3.inc
jrfnl: I'm not 100% sure what you mean. The varying comments in both this file as well as the `.css` test file, have varying layouts which is how this is tested.

The _adding_ of a missing header is not whitespace agnostic: it will basically take the previous line, copy it, change it and add it to the file, whitespace warts and all.

As for replacing - do you mean tabs vs spaces within the `Text Domain` line ? That is something I could possibly improve upon.
All the same, for multi-line function calls, the sniff also had to make a judgement call and if needed adds a tab indentation.

jrfnl:
@GaryJones I've added some additional commits which hopefully address your concerns.

WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
GaryJones: > As for the first - that is filter

Yep, good catch. Ignore that part of my suggestion :-)

WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
GaryJones: Yes please. Looking through the WP core functions (like [`__`](https://developer.wordpress.org/reference/functions/__/)), they all use two words for non-code instances (variables, etc.), so I think we should use the same too.

(Aside: "meta box" is often mistakenly written as "metabox" as well.)

WordPress/Tests/Utils/I18nTextDomainFixerUnitTest.3.inc
GaryJones: For the reading of the existing text domain, to know whether it needs replacing or not.

All I see right now, is `Text Domain: Foo` - that is, exactly one space between the colon and the text domain.

Dropping in some variants would show it's whitespace-agnostic and possibly case insensitive (if WP supports that as well):

```
Text Domain:foo
Text Domain:     foo
Text Domain:[tab]foo
text domain: FOO
```

WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
jrfnl: Well, I've adjusted it for this sniff in one of the new commits.

I suggest either one of us does a check for both `text[-]domain` as well as `meta[-]box` for the rest of the files and send that in as a separate PR to not pollute this PR with unrelated changes.

WordPress/Tests/Utils/I18nTextDomainFixerUnitTest.3.inc
jrfnl: IIRC correctly, `text-domains` are case-sensitive, so I don't think we need to allow for that (and the sniff contains a check that the `new_text_domain` is lowercase, dash-separated.

The last new commit I added, already contains variants of the headers with inline tabs.

This is mostly relevant for the _adding_ of a new text domain header, not so much for the replacing, as replacing doesn't touch the whitespace, it just replaces the `text-domain` string within the larger comment string.

PR #1527, opened 2018-11-06
Change log for WPCS version 1.2.0

jrfnl:
* Release date set at this Monday **November 12th**.
* Includes all currently merged changes.

PR #1528, opened 2018-11-06
Release version 1.2.0

jrfnl:
:warning: **DO NOT MERGE (YET)** :warning: 

PR for tracking changes for the 1.2.0 release. Suggested target release date: **Monday November 12**.
- [x] Add changelog for 1.2.0 - PR #1527
- [ ] Merge this PR
- [ ] Add release tag & copy & paste the changelog to it
- [ ] Close the 1.2.0 milestone
- [x] Open a new milestone for the next release
- [x] If any open PRs which were milestoned for 1.2.0 do not make it into the release, update their milestone.
- [ ] Tweet about the release.
- [ ] Submit for "Month in WordPress": https://make.wordpress.org/community/month-in-wordpress-submissions/


WordPress-Core/ruleset.xml
```diff
@@ -393,7 +415,7 @@
 	</rule>
 
 	<!-- Rule: create_function() function, which internally performs an eval(),
-		 is deprecated in PHP 7.2. Both of these must not be used. -->
+		 is deprecated in PHP 7.2. ... these must not be used. -->
```
GaryJones: This doesn't look right?

jrfnl: Wrong PR for that comment and it would need another PR to change it.

All the same, this was just a way to try and quote only the relevant bit of the handbook paragraph in the ruleset, similar to how it was done for the section above it on `eval()` (line 388/410).
> `<!-- Rule: The eval() construct is very dangerous, and is impossible to secure. ... these must not be used. -->`

The original paragraph is:
> The eval() construct is very dangerous, and is impossible to secure. Additionally, the create_function() function, which internally performs an eval(), is deprecated in PHP 7.2. Both of these must not be used.

https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#clever-code

jrfnl: As a side-note: I'd be happy for that text to be improved, but as it's nothing user-facing, I don't think it's a reason to delay the release.

PR #1531, opened 2018-11-10
Fix missing check for infinite posts_per_page

### LINKED ISSUE #1379: check of posts_per_page isn't working properly when passing the value from array variable.
As per WordPress-VIP standards, we must not set "-1" to posts_per_page, it show give me an proper error when passing -1 in any way. I expect to see error message like **"Disabling pagination is prohibited in VIP context, do not set `posts_per_page` to `-1` ever."** when I run `phpcs --standard=WordPress-VIP` 

But I'm not getting any error message when passing -1 passing/assigning value to posts_per_page using below syntax:

```
$args = array( 'posts_per_page' => -1 );
get_posts( $args);

```

```
function custom_get_posts(){
	$args = array( 'posts_per_page' => -1 );
	get_posts( $args);
}

```

```
$args = array( 'posts_per_page' => -1 );

function custom_get_posts(){
	get_posts( $args);
}

```
Working syntax is (means, giving me an error in terminal):
```
get_posts( 'posts_per_page=-1');
```
```
$args4 = array();
$args4['posts_per_page'] = -1;

get_posts( $args4 ); // Bad.
```
Please correct me if I missed anything.

LC43:
Hi, 

the class description says "Flag returning high or infinite posts_per_page.". It was missing this last check:  `'-1' === $val`, which i added it here:

```php
if ( $val > $this->posts_per_page || '-1' === $val ) {
	return 'Detected high pagination limit, `%s` is set to `%s`';
}
```

This infinite (-1) limit is worse that high fixed limit. Like [other have mentioned](https://10up.github.io/Engineering-Best-Practices/php/), this fix should be reintroduced.

This will fix #1379 in cases where the value is not a variable.

I also introduced a new test:

```php
$query_args['posts_per_page'] = '-1'; // Bad.
```

Not sure if i should add more cases, as they are missing.

Also, i didn't touch the wrong indentation on the last tests. There's a space at the beginning of each line. Can i remove it?

```php
// @codingStandardsChangeSetting WordPress.WP.PostsPerPage posts_per_page 50
 $query_args['posts_per_page'] = 50; // OK.
 $query_args['posts_per_page'] = 100; // Bad.
 $query_args['posts_per_page'] = 200; // Bad.
 $query_args['posts_per_page'] = 300; // Bad.
// @codingStandardsChangeSetting WordPress.WP.PostsPerPage posts_per_page 200
 $query_args['posts_per_page'] = 50; // OK.
 $query_args['posts_per_page'] = 100; // OK.
 $query_args['posts_per_page'] = 200; // OK.
 $query_args['posts_per_page'] = 300; // Bad.
```

All tests look good:
```script

$phpunit --bootstrap="./Test/phpcs3-bootstrap.php" --filter WordPress ./vendor/squizlabs/php_codesniffer/tests/AllTests.php
PHPUnit 6.5.5 by Sebastian Bergmann and contributors.

................................................................. 65 / 80 ( 81%)
...............                                                   80 / 80 (100%)

Tests generated 580 unique error codes; 52 were fixable (8.97%)

Time: 2.58 seconds, Memory: 36.00MB

OK (80 tests, 0 assertions)
```

thanks, let me know if i missed something.

### DIFF
--- a/WordPress/Sniffs/WP/PostsPerPageSniff.php
+++ b/WordPress/Sniffs/WP/PostsPerPageSniff.php
@@ -68,7 +68,7 @@ class PostsPerPageSniff extends AbstractArrayAssignmentRestrictionsSniff {
 	public function callback( $key, $val, $line, $group ) {
 		$this->posts_per_page = (int) $this->posts_per_page;
 
-		if ( $val > $this->posts_per_page ) {
+		if ( $val > $this->posts_per_page || '-1' === $val ) {
 			return 'Detected high pagination limit, `%s` is set to `%s`';
 		}
 
--- a/WordPress/Tests/WP/PostsPerPageUnitTest.inc
+++ b/WordPress/Tests/WP/PostsPerPageUnitTest.inc
@@ -2,7 +2,7 @@
 
 $args = array(
 	'posts_per_page' => 999, // Bad.
-	'posts_per_page' => -1, // OK.
+	'posts_per_page' => -1, // Bad.
 	'posts_per_page' => 1, // OK.
 	'posts_per_page' => '1', // OK.
 );
@@ -13,7 +13,8 @@ _query_posts( 'numberposts=999' ); // Bad.
 $query_args['posts_per_page'] = 999; // Bad.
 $query_args['posts_per_page'] = 1; // OK.
 $query_args['posts_per_page'] = '1'; // OK.
-$query_args['numberposts'] = '-1'; // OK.
+$query_args['posts_per_page'] = '-1'; // Bad.
+$query_args['numberposts'] = '-1'; // Bad.
 
 $query_args['my_posts_per_page'] = -1; // OK.
 
--- a/WordPress/Tests/WP/PostsPerPageUnitTest.php
+++ b/WordPress/Tests/WP/PostsPerPageUnitTest.php
@@ -41,13 +41,16 @@ class PostsPerPageUnitTest extends AbstractSniffUnitTest {
 	public function getWarningList() {
 		return array(
 			4  => 1,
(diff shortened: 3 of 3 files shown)


WordPress/Sniffs/WP/PostsPerPageSniff.php
```diff
@@ -68,7 +68,7 @@ public function getGroups() {
 	public function callback( $key, $val, $line, $group ) {
 		$this->posts_per_page = (int) $this->posts_per_page;
 
-		if ( $val > $this->posts_per_page ) {
+		if ( $val > $this->posts_per_page || '-1' === $val ) {
```
GaryJones: This is checking a value that is a string, not a negative integer.

LC43: hi, i checked it, and $val is always converted to string. not sure why or where, didn't look into it deeply.

thanks why the tests pass, either with  `-1` and with `'-1'`

LC43: the token is parse into a string here:

```php
// this->tokens
23: - // "PHPCS_T_MINUS"
24: 1 // "T_LNUMBER"
25: , // "PHPCS_T_COMMA"
```

```php
// call: WordPress/AbstractArrayAssignmentRestrictionsSniff.php L170
// valStart: 23
// valEnd: 25
$val            = $this->phpcsFile->getTokensAsString( $valStart, ( $valEnd - $valStart ) );
```

LC43:
hi @jrfnl, no problem, thanks for taking your time to review this.

I'll open a ticket as soon as i have better understanding of of these deprecations.

PR #1532, opened 2018-11-12
PrecisionAlignment: bug fix

jrfnl:
This issue was discovered through a report upstream in issue squizlabs/PHP_CodeSniffer#2179#issuecomment-437634372.

What it comes down to is that if the `ignoreAlignmentTokens` property would be set, the `merge_custom_array()` call in the `process_token()` would override it with a flipped array with values `false` and for the next file, would override it again, flipping the array again etc etc.

This in effect made the custom property only work for the first file, not for subsequent files.

This has now been fixed.

Includes unit tests to guard against the issue.

Additionally, I've improved the inline comments to clarify what's being tested in test case file 2.

### DIFF
--- a/WordPress/Sniffs/WhiteSpace/PrecisionAlignmentSniff.php
+++ b/WordPress/Sniffs/WhiteSpace/PrecisionAlignmentSniff.php
@@ -91,7 +91,7 @@ class PrecisionAlignmentSniff extends Sniff {
 		}
 
 		// Handle any custom ignore tokens received from a ruleset.
-		$this->ignoreAlignmentTokens = $this->merge_custom_array( $this->ignoreAlignmentTokens );
+		$ignoreAlignmentTokens = $this->merge_custom_array( $this->ignoreAlignmentTokens );
 
 		$check_tokens  = array(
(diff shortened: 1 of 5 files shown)


jrfnl:
FYI: I've done a quick code review of all other uses of the `merge_custom_array()` method and I don't believe any other sniffs are susceptible to this issue as they currently are.

DaveLak:
I've tested this PR with [the repo I created to reproduce the issue](https://github.com/DaveLak/phpcs-poc) and can confirm the proposed change does fix the problem.

A note to others that end up in this thread directly: The issue being fixed in this PR only seems to exist on macOS. Linux and Windows seem to be fine.

jrfnl:
@DaveLak Thank you for testing and confirming the validity of this fix.

> A note to others that end up in this thread directly: The issue being fixed in this PR only seems to exist on macOS. Linux and Windows seem to be fine.

FYI: With the unit test cases I've added in this PR, I did manage to reproduce the issue on Windows as well, so I'm honestly not sure what was causing the issue not to show up with the original POC.

PR #1390, opened 2018-06-26
Abstracts: change $exclude property from string to array

jrfnl:
The `$exclude` property which is included in most abstract sniffs and by extension in all sniffs which use these abstracts, used to expect a comma-delimited list.
This comma-delimited list was then exploded to an array from within the sniff.

In other words, in reality the sniff expected an `array`.

This has now been formalized. All `$exclude` properties now expect `array` input.

As outlined in #1368, this is not a BC-break as properties passed as a comma-delimited string will still be supported (for now). Support for this may be removed at some point in the future.

The upside of making this change is that users who use WPCS 3.3.0, can start using the new property array format in their custom rulesets which makes for a more easily readable ruleset.

- [ ] Update the wiki section about the `exclude` property in custom rulesets.

Fixes #1368

### Testing this PR

Testing this PR is a little finicky as it needs to be done manually, so here are the steps to do so:

1 - Start by saving the below code to a temporary test file.
<details>
    <summary>Test file content</summary>

```php
<?php

// AbstractArrayAssignmentRestrictions
$args = array(
	'orderby' => 'rand', // Bad.
);

// AbstractFunctionRestrictions
parse_url( 'http://example.com/' ); // Warning.

$json = json_encode( $thing ); // Warning, use wp_json_encode instead.

file_get_contents(); // Warning.

readfile(); // Warning.
fopen(); // Warning.

// AbstractClassRestrictions
$a = new WP_User_Search;

// AbstractFunctionParameter
in_array( 1, array( '1', 1, true ) ); // Warning.
array_search( 1, $array, false ); // Warning.

// AbstractVariableRestrictions
$query = "SELECT * FROM $wpdb->usermeta"; // Error.
$wp_db->update( $wpdb->users, array( 'displayname' => 'Kanobe!' ), array( 'ID' => 1 ) ); // Error.
```
</details>

2 - Next, create a custom ruleset with the below content and save it to a file
<details>
    <summary>Initial ruleset</summary>

```xml
<?xml version="1.0"?>
<ruleset name="TEST 1368">
	<description>TEST 1368</description>

	<autoload>/path/to/WordPress/PHPCSAliases.php</autoload>

	<!-- AbstractArrayAssignmentRestrictions -->
	<rule ref="WordPress.VIP.OrderByRand"/>

	<!-- AbstractFunctionRestrictions -->
	<rule ref="WordPress.WP.AlternativeFunctions"/>

	<!-- AbstractClassRestrictions -->
[...]

PR #1584, opened 2018-12-21
Sniff::merge_custom_array(): Remove support for incorrectly set custom array properties

jrfnl:
Array properties passed in from a custom ruleset need to have a `type="array"` attribute.

Until now WPCS was lenient when this attribute was missing and would convert an array passed as a comma delimited string to an array.

This functionality was deprecated for the only property which still "officially" allowed for a comma-delimited string list in WPCS 1.0.0 and will now be removed.

The only exception are the two array properties which can also be set via the command-line, i.e. the `prefixes` property for the `WordPress.NamingConventions.PrefixAllGlobals` sniff and the `text_domain` property for the `WordPress.WP.I18n` sniff.
If - and only if - these properties are set from the command-line, the sniffs will still convert the received string to an array.

Refs:
* #1390

This PR also removes the leniency for spaces surrounding array values.

PHPCS 3.2.0 includes a similar fix as was previously made in WPCS via PR #1185, which means that properties passed as arrays will already be trimmed, so no need to do it again.

Refs:
* #1185
* squizlabs/PHP_CodeSniffer#1689

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -943,10 +943,8 @@ abstract class Sniff implements PHPCS_Sniff {
 	}
 
 	/**
-	 * Merge a pre-set array with a ruleset provided array or inline provided string.
+	 * Merge a pre-set array with a ruleset provided array.
 	 *
-	 * - Will correctly handle custom array properties which were set without
-	 *   the `type="array"` indicator.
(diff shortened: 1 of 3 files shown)


GaryJones:
This might catch folks out in a more subtle way than some of the more obvious fatal errors, so sufficient attention should be drawn to it in the release notes.

PR #1590, opened 2018-12-22
PHPCS native ruleset: don't require exact array double arrow alignment

jrfnl:
When the item with the _longest_ array key is removed, WPCS requires the double arrows in the whole array to be re-aligned.
Setting the [`exact` property](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#array-alignment-allow-non-exact-alignment) to `false` removes this requirement and allows for cleaner diffs when this happens.

As PR #1547 removed a large number of items from the `Sniff::$autoEscapedFunctions` array, PHPCS was currently throwing warnings about the WPCS codebase.

This small tweaks gets rid of those.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -26,6 +26,7 @@
 	<rule ref="WordPress.Arrays.MultipleStatementAlignment">
 		<properties>
 			<property name="alignMultilineItems" value="!=100"/>
+			<property name="exact" value="false" phpcs-only="true"/>
 		</properties>
 	</rule>
 


GaryJones:
> This small tweaks gets rid of those.

But does that stop other mis-alignments from being reported?

jrfnl:
> But does that stop other mis-alignments from being reported?

From [the wiki](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#array-alignment-allow-non-exact-alignment):
> When set to false, the largest index key + 1 space is taken as a minimum and if a predominant number of items is already aligned at a position which is more than the minimum, this predominant alignment position is taken as the preferred alignment.
>
> This allows for cleaner diffs as the array operators do not have to be realigned when the array item with the largest key is taken out.

So, the short answer is: no, not really, the arrows still have to be aligned, it's just the position at which they are aligned for which we are more lenient.

What about if I set it to `phpcs-only` ? I.e. if the auto-fixer is run, they will still get fixed.

GaryJones:
> What about if I set it to `phpcs-only` ? I.e. if the auto-fixer is run, they will still get fixed.

I'm open to that.

jrfnl:
Done ;-)

PR #1591, opened 2018-12-22
Composer: some tweaks

jrfnl:
* Make the `scripts` section more readable by setting each script up as an array.
* `require-dev` the DealerDirect Composer plugin to make life easier on WPCS devs using Composer for their dev environment.
    Includes:
    - Adjusting the `install-codestandards` script to use the plugin.
        This script should be considered _deprecated_, but is left in place for devs who are used to having it available in their workflow.
    - Removing the `post-install-cmd` and `post-update-cmd` scripts which shouldn't be needed anymore as the plugin automatically takes care of installing the standards.
* Includes updating the inline documentation in the Travis script to be inline with these changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ before_install:
     - |
       if [[ "$SNIFF" == "1" ]]; then
           composer install --dev --no-suggest
-          # The post-install-cmd script takes care of the installed_paths.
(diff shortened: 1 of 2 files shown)


jrfnl:
@GaryJones Good point, happy to hear it still has function.

PR #1602, opened 2018-12-25
DiscouragedConstants: make the error codes more modular

jrfnl:
This sniff is based on an array of constants the sniff looks for.
Previously, the sniff had two distinct error codes for the different use-cases in which the constant could be found (use of vs declaration).

This change makes the error codes unique to the constant found, making it easier to silence the error message(s) for just one particular constant.

This is a breaking change as `<exclude>`s for the old errorcodes currently in custom rulesets will be invalidated by it, as well as inline annotations using the error codes.


GaryJones:
It appears `$content` is not defined for the second instance.

PR #1604, opened 2018-12-25
WPCS native PHPCS ruleset: minor tweaks

jrfnl:
The `basepath` and `parallel` options were introduced in PHPCS 3.0 and can now be enabled for the WPCS native PHPCS checking.

And as WPCS 2.0.0 drops the `VIP` ruleset, we can now include the `WordPress` ruleset instead of including the `WordPress-Extra` + `WordPress-Docs` rulesets separately.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -7,18 +7,18 @@
 
 	<arg value="sp"/>
 	<arg name="extensions" value="php"/>
+	<arg name="basepath" value="."/>
+	<arg name="parallel" value="8"/>
 
 	<exclude-pattern>/bin/class-ruleset-test.php</exclude-pattern>
 	<!-- Exclude Composer vendor directory. -->
 	<exclude-pattern>*/vendor/*</exclude-pattern>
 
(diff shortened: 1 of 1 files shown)


jrfnl:
> Does this not still include three less-than-ideal sniffs that aren't currently included in any of the named rulesets?

Good point, I'd forgotten about that.

I don't think those sniffs are problematic for our own ruleset though (the build passes after all), but it would be good to have a think about what to do with those sniffs.

Adding them to `Extra` seems most appropriate, though my remark about these sniffs is still valid.

Refs:
* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1440
* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1261

jrfnl:
Is there any reason why this PR cannot be merged ?

The above mentioned "three sniffs in `WordPress`, not in `Extra`" discussion should probably be moved to a separate issue and is not problematic for the WPCS native ruleset (which is what this PR targets).

PR #1635, opened 2019-01-23
ValidateSanitizedInput: allow for validation using array_key_exists()

jrfnl:
Just like, `isset()` and `empty()`, `array_key_exists()` is a way to validate a variable exists and should therefore be recognized by this function.

Note the difference between the functions:
* `isset( $foo['bar'] )` - checks whether the key `bar` exists in the array `$foo` _and has a value assigned to it_.
* `! empty( $foo['bar'] )` - checks whether the key `bar` exists in the array `$foo` _and has a _non-falsy_ value assigned to it_.
* `array_key_exists( 'bar', $foo )` - checks whether the key `bar` exists in the array `$foo`. I.e. when using `array_key_exists()` a `null` value is acceptable.

Which validation method is used, should be left up to the judgement of a developer.

All three will prevent _Undefined variable_ notices, so in my opinion, `array_key_exists()` should be regarded as a valid way of validating a variable.

Includes unit tests covering the change in `Sniff::is_validated()`.

The change in `Sniff::is_in_isset_or_empty()` is currently not (sufficiently) unit tested as the only sniff which uses the method, `ValidatedSanitizedInput`, requires for a superglobal to have an array key set and therefor doesn't throw an error for the use of `$_POST` in a call to `array_key_exists()`.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -1703,7 +1703,7 @@ abstract class Sniff implements PHPCS_Sniff {
 	}
 
 	/**
-	 * Check if the existence of a variable is validated with isset() or empty().
+	 * Check if the existence of a variable is validated with isset(), empty() or array_key_exists().
 	 *
 	 * When $in_condition_only is false, (which is the default), this is considered
 	 * valid:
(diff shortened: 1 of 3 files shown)


jrfnl:
Any reason why this PR is not merged ?

GaryJones:
Just tripped up on this myself in client-related code, so a belated +1 for getting this merged ;-)

PR #1647, opened 2019-02-18
PrefixAllGlobals: improve error message clarity

jrfnl:
When people run the sniffs without the `-s` option, the messages for some of the errors in this sniff were not specific enough as the "Prefix**AllGlobals**" part would not be seen.

I considered adjusting the error message template `ERROR_MSG`, however, this would cause more confusion as, for instance, namespace declarations will **always** be in the global namespace.

So, I've opted to adjust select error messages instead.

Issue identified by @justintadlock in https://github.com/WPTRT/WPThemeReview/issues/200#issuecomment-464789010


jrfnl:
Just checking, is this waiting for something ?

PR #1648, opened 2019-02-20
Fixes #1447, adds new sniff for blacklisted ini_set directives.

### LINKED ISSUE #1447: DiscouragedPHPFunctions: whitelist allow ini_set directives
The `PHP.DiscouragedPHPFunctions` sniff warns about using `ini_set()`, however, there are some ini directives which if set from within a plugin can be considered valid usage.

I'd like to suggest adding a limited whitelist of ini directives to this sniff and to stop throwing warnings when `ini_set()` is setting one of these directives.
This could for some directives be combined with a check of the value being set.

For some ini directives, there are also alternatives available which should possibly be recommended instead.

Depending on the input received on this ticket and the amount of exceptions to be made, this should possibly be extracted to a separate sniff dedicated to `ini_set()`.

In that case, we could also add a _blacklist_ with ini directives which when a plugin tries to set them should throw an error instead of a warning, such as [`output_buffering`](http://php.net/manual/en/outcontrol.configuration.php#ini.output-buffering) or any of the session related ini directives.

List of ini directives in PHP: http://php.net/manual/en/ini.list.php

Proposed initial whitelist:
* [`auto_detect_line_endings`](http://php.net/manual/en/filesystem.configuration.php#ini.auto-detect-line-endings) - This defaults to `off`, but turning this on when reading a file improves OS-interoperability, so if this is being set, it will normally be a valid use-case.
* [`highlight.bg`, `highlight.comment`, `highlight.default`, `highlight.html`, `highlight.keyword`, `highlight.string`](http://php.net/manual/en/misc.configuration.php#ini.syntax-highlighting) - set the color coding for the PHP build-in syntax highlighting.
    I'm not sure if a plugin exists which uses this, but adjusting these color codes is quite innocent, so can be ignored.
* [`max_execution_time`](http://php.net/manual/en/info.configuration.php#ini.max-execution-time) - WP core itself changes this for certain operations as well and for plugins which, for instance, import data from an external source or perform system scans, it is perfectly valid to change this value, though advising to use the PHP `set_time_limit()` function instead would also be an option.
* [`short_open_tag`](http://php.net/manual/en/ini.core.php#ini.short-open-tag) - turning this on for PHP < 5.4 when using short echo tags is a valid use case. Turning it off should be prohibited.

Additional candidates:
* [`error_log`](http://php.net/manual/en/errorfunc.configuration.php#ini.error-log) - This sets the location of the error log. WP sets this value to `wp-content`, but it would be quite valid for a plugin to set this to a different, user-defined location.

There are also a number of ini directives which really shouldn't be used as there are either PHP native or WP native alternatives for setting these from within a plugin. For these we could consider throwing a custom error message (+ error code):
* [`bcmath.scale`](http://php.net/manual/en/bc.configuration.php#ini.bcmath.scale) - use `bcscale()` instead.
* [`display_errors`](http://php.net/manual/en/errorfunc.configuration.php#ini.display-errors) - use `WP_DEBUG_DISPLAY` instead.
* [`error_reporting`](http://php.net/manual/en/errorfunc.configuration.php#ini.error-reporting) - use `WP_DEBUG` instead.
* [`filter.default`, `filter.default_flags`](http://php.net/manual/en/filter.configuration.php#ini.filter.default) - use the filter flag constants when calling the functions instead (as you will possibly break other plugins if you change this).
* [`iconv.input_encoding`, `iconv.internal_encoding`, `iconv.output_encoding`](http://php.net/manual/en/iconv.configuration.php) - PHP < 5.6 only - use `iconv_set_encoding()` instead.
* [`ignore_user_abort`](http://php.net/manual/en/misc.configuration.php#ini.ignore-user-abort) - use `ignore_user_abort()` instead.
* [`log_errors`](http://php.net/manual/en/errorfunc.configuration.php#ini.log-errors) - use `WP_DEBUG_LOG` instead.
* [`memory_limit`](http://php.net/manual/en/ini.core.php#ini.memory-limit) - use `wp_raise_memory_limit()` or hook into the filters in that function.

Opinions ?

NielsdeBlaauw:
I've decided to create a new sniff for checking proper ini_set directives. With this feature the original sniff from `DiscouragedPHPFunctionsSniff.php` is removed.

- It adds a whitelist of options that are valid to set by a plugin (in the case of `short_open_tag` it is only valid with some limited values).
- It adds a blacklist of options that have better alternatives, or might influence other plugins. (`short_open_tag` is checked for values turning it off).
- All ini_set directives that are not either in the whitelist or blacklist are marked as risky with a warning. This mimics original behaviour.

For now I've used the following insights from the issue:
- `max_execution_time` is an error, as a better alternative exists that should be preferred.
- I've left `error_log` as a warning for now, as I personally would prefer that plugins do not move the `error_log` around. I could only see some security plugins doing this. They can use an ignore statement if they really need it.

phpcbf did not work for me locally, so this probably needs some formatting initially.

Fixes #1447


WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+			'message' => 'Use `bcscale()` instead.',
+		),
+		'display_errors' => array(
+			'message' => 'Use `WP_DEBUG_DISPLAY` instead.',
+		),
+		'error_reporting' => array(
+			'message' => 'Use `WP_DEBUG` instead.',
+		),
+		'filter.default' => array(
+			'message' => 'Use the filter flag constants when calling the functions instead (as you will possibly break other plugins if you change this).',
```
jrfnl: For this and the one below it:

```suggestion
			'message' => 'Changing the option value can break other plugins. Use the filter flag constants when calling the Filter functions instead.',
```

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+	 *         )
+	 *     );
+	 */
+	protected $blacklisted_options = array(
+		'max_execution_time' => array(
+			'message' => 'Use `set_time_limit()` instead.',
+		),
+		'short_open_tag' => array(
+			'invalid_values' => array( 'false', '0', 'Off' ),
+			'message'        => 'Turning off short_open_tag is prohibited as it might possibily break other plugins.',
```
jrfnl:
```suggestion
			'message'        => 'Turning off short_open_tag is prohibited as it can break other plugins.',
```

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+	 * @var array Multidimensional array with parameter details.
+	 *     $blacklisted_options = array(
+	 *         (string) option name. = array(
+	 *             (string[]) 'invalid_values' = array()
+	 *             (string) 'message'
+	 *         )
+	 *     );
+	 */
+	protected $blacklisted_options = array(
+		'max_execution_time' => array(
```
jrfnl: Safe for these first two entries, this list is in alphabetic order. Might be an idea to make it completely so ?

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+		'log_errors' => array(
+			'message' => 'Use `WP_DEBUG_LOG` instead.',
+		),
+		'memory_limit' => array(
+			'message' => 'Use `wp_raise_memory_limit()` or hook into the filters in that function.',
+		),
+	);
+
+	/**
+	 * Process the parameter of a matched function. Errors if an option
```
jrfnl: Suggestion: Start the phrase from `Errors if ..` on a new line as the long description (with a blank line between the short and the long description).

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+	 *
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		$ini_set_function = $this->tokens[ $stackPtr ];
```
jrfnl: This line can be removed. The variable is only used to get the content (function name) of the matched function, which is already available through the `$matched_content` parameter.

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+
+		if ( array_key_exists( $option_name, $this->blacklisted_options ) ) {
+			$blacklisted_option = $this->blacklisted_options[ $option_name ];
+			if ( ! isset( $blacklisted_option['invalid_values'] ) || in_array( $option_value, $blacklisted_option['invalid_values'], true ) ) {
+				$this->phpcsFile->addError(
+					'%s(%s, %s) found. %s',
+					$stackPtr,
+					'Blacklisted',
+					array(
+						$ini_set_function['content'],
```
jrfnl:
```suggestion
						$matched_content,
```

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+				);
+				return;
+			}
+		}
+		$this->phpcsFile->addWarning(
+			'%s(%s, %s) found. Changing these configuration values at runtime is rarely necessary.',
+			$stackPtr,
+			'Risky',
+			array(
+				$ini_set_function['content'],
```
jrfnl:
```suggestion
				$matched_content,
```

WordPress/Tests/PHP/IniSetUnitTest.inc
```diff
@@ -0,0 +1,28 @@
...
+ini_set('iconv.input_encoding', ''); // Error.
+ini_set('iconv.internal_encoding', ''); // Error.
+ini_set('iconv.output_encoding', ''); // Error.
+ini_set('iconv.output_encoding', ''); // Error.
+ini_set('ignore_user_abort', true); // Error.
+ini_set('log_errors', true); // Error.
+ini_set('memory_limit', -1); // Error.
+
+ini_set('report_memleaks', true); // Warning.
+ini_set('short_open_tag', 1230); // Warning.
```
jrfnl: I think it would be a good idea to add an additional test case documenting the behaviour when variables are passed to the function.
```php
ini_set( $name, $new_value ); // Warning
```

WordPress/Tests/PHP/IniSetUnitTest.inc
```diff
@@ -0,0 +1,28 @@
+<?php
+
+ini_set('auto_detect_line_endings', true); // Ok.
+ini_set('highlight.bg', '#FFFFFF'); // Ok.
+ini_set('highlight.comment', '#FFFFFF'); // Ok.
+ini_set('highlight.default', '#FFFFFF'); // Ok.
+ini_set('highlight.html', '#FFFFFF'); // Ok.
+ini_set('highlight.keyword', '#FFFFFF'); // Ok.
+ini_set('highlight.string', '#FFFFFF'); // Ok.
+ini_set('short_open_tag', true); // Ok.
```
jrfnl: It might be an idea to also add tests with the other "valid" values.

WordPress/Tests/PHP/IniSetUnitTest.inc
```diff
@@ -0,0 +1,28 @@
...
+
+ini_set('max_execution_time', 60); // Error.
+ini_set('short_open_tag', false); // Error.
```
jrfnl: It might be an idea to also add tests with the other "invalid" values.

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		$ini_set_function = $this->tokens[ $stackPtr ];
+		$option_name      = $this->strip_quotes( $parameters[1]['raw'] );
+		$option_value     = $this->strip_quotes( $parameters[2]['raw'] );
+		if ( array_key_exists( $option_name, $this->whitelisted_options ) ) {
```
jrfnl: Is there any particular reason why you are using the function call `array_key_exists()` rather than the (normally faster) language construct `isset()` ?

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+						$ini_set_function['content'],
+						$parameters[1]['raw'],
+						$parameters[2]['raw'],
+						$blacklisted_option['message'],
+					)
+				);
+				return;
+			}
+		}
+		$this->phpcsFile->addWarning(
```
jrfnl: Nitpick: Missing blank line above this line.

jrfnl:
Darn... one more thing and this doesn't necessarily have to be solved now:

It might be an idea to make the error codes modular, at least for the `Blacklist` code, like so `$this->string_to_errorcode( $parameters[1]['raw'] . '_Blacklisted' ),`

That way, if a plugin has a very good reason to change one of these ini directives, they could `<exclude ...>` just the errorcode for that particular directive, without disabling the whole blacklist.

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+			'%s(%s, %s) found. Changing these configuration values at runtime is rarely necessary.',
```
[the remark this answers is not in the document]
GaryJones: I'd go with `...Changing configuration values at runtime is strongly discouraged.`

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Sniffs\PHP;
+
+use WordPressCS\WordPress\AbstractFunctionParameterSniff;
+use \PHP_CodeSniffer\Util\Tokens;
```
jrfnl: One more... just noticed you're not actually using the `Tokens` class in this sniff, so this `use` statement should be removed.

WordPress/Sniffs/PHP/IniSetSniff.php
NielsdeBlaauw: Started originally as a flat array before adding the 'valid_values' dimension. No particular preference, so will edit this.

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+	 *
+	 * @since x.x.x
+	 *
+	 * @var array Multidimensional array with parameter details.
+	 *     $target_functions = array(
+	 *         (string) Function name.
+	 *     );
+	 */
+	protected $target_functions = array(
+		'ini_set' => array(),
```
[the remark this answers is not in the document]
NielsdeBlaauw: The parameter value sniff was in part inspired by logic in `DeprecatedParametersSniff.php`. As it checks for the array key, I had to set it to something. To me it made more sense to make this an empty array, consistent with the other parameter checking sniffs, then to make the value `true`, or `null`. Which might imply it actually has a function or value. Also this seems more flexible if future edits are required, but this might be classic YAGNI.

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+	 * Array of options that are allowed to be manipulated
+	 *
+	 * @since x.x.x
+	 *
+	 * @var array Multidimensional array with parameter details.
+	 *     $whitelisted_options = array(
+	 *         (string) option name. = array(
+	 *             (string[]) 'valid_values' = array()
+	 *         )
+	 *     );
```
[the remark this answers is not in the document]
NielsdeBlaauw: > Just wondering: is there a particular reason you've chosen to make this a three, rather than a two dimensional array with the valid value directly in the `option_name` array?

The array key makes the intent explicit. Also is more flexible this way.

NielsdeBlaauw:
Thanks for the feedback. All should be fixed now. Let me know.

WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -0,0 +1,173 @@
...
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		$ini_set_function = $this->tokens[ $stackPtr ];
+		$option_name      = $this->strip_quotes( $parameters[1]['raw'] );
```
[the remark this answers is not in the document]
jrfnl: Just marking this as "unresolved" to make it easier to find for other people as this comment documents why using the `raw` parameter value for this sniff is ok.

WordPress/Sniffs/PHP/IniSetSniff.php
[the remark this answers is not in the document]
jrfnl: @NielsdeBlaauw Looks like you might have missed @GaryJones's remark ?

WordPress/Sniffs/PHP/IniSetSniff.php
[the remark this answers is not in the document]
jrfnl: The `AbstractFunctionParameterSniff` which is underneath only requires the array keys to be the function names and a value to be set. It doesn't define the type of the value on purpose to give child classes maximum flexibility.
See: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/AbstractFunctionParameterSniff.php#L35-L38

> To me it made more sense to make this an empty array, consistent with the other parameter checking sniffs

Generally speaking, for WPCS, if an array property is used where only the keys are relevant, the values are set to `true`.

This type of array format allows for using the faster `isset()` instead of `in_array()`, which would be needed if the array just contained values, or `array_key_exist()` if the value would be set to `null`.

See, for instance, the large number of array properties set up like this in the [`Sniff` class](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniff.php).

For the sniffs based on the `AbstractFunctionParameterSniff` class, the values vary wildly depending on the use case of the child class:
* `PrefixAllGlobals` - uses the `Sniff::$hookInvokeFunctions` property where the values are set to `true` without meaning.
* `ValidHookName` - same as `PrefixAllGlobals`
* `PregQuoteDelimiter` - value set to `true` without meaning.
* `StrictInArray` - value set to boolean _with_ meaning. The meaning of the boolean is documented in the property documentation within the child sniff.
* `PluginMenuSlug` - value set to an array containing integers indicating the parameter positions which should be examined. Meaning documented in the property documentation.
* `I18nTextDomainFixer` - value set to the integer property position of the parameter to be examined.
* `DeprecatedParameters` - value set to a multi-dimensional array of parameter positions combined with info on that particular parameter.
* `DeprecatedParameterValues` - value set to a multi-dimensional array of parameter positions combined with info on that particular parameter.
* `DiscouragedConstants` - value set to the integer property position of the parameter to be examined.
* `EnqueuedResourceParameters` - value set to `true` without meaning.

As the value in the case of this sniff has no meaning, the most consistent choice based on the other code in WPCS, would be to use `true` as the value.

WordPress/Sniffs/PHP/IniSetSniff.php
jrfnl: This same change still needs to be made for the `filter.default_flags` message.

WordPress/Sniff.php
```diff
@@ -175,65 +175,65 @@ abstract class Sniff implements PHPCS_Sniff {
 	 * @var array
 	 */
 	protected $autoEscapedFunctions = array(
-		'allowed_tags'              => true,
```
jrfnl: Hmm... this change shouldn't be here.... (doesn't belong in this PR. You've probably run phpcbf, but this was fine as it was.)

WordPress/Sniffs/PHP/IniSetSniff.php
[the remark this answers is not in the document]
NielsdeBlaauw: I did, fixed it and also replaced the message in the `WordPress/Sniffs/PHP/DiscouragedPHPFunctionsSniff.php` to be consistent.

WordPress/Sniffs/PHP/IniSetSniff.php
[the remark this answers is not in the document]
NielsdeBlaauw: Very valid, thanks for the great explanation. I've changed this part to be consistent with the other ones now.

NielsdeBlaauw:
@jrfnl Thanks for taking the time for such extensive feedback.

Travis seems to be having some issues on the php5.4 test at the moment (see https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/497198381)

> Would you like to squash the commits before that time or shall I do that when I merge it ?

If you can squash it in the merge that'd be great.

jrfnl:
> Thanks for taking the time for such extensive feedback.

You're welcome & thanks for the quick turn-around, makes things so much easier :smile: 

> Travis seems to be having some issues on the php5.4 test at the moment (see https://travis-ci.org/WordPress-Coding-Standards/WordPress-Coding-Standards/jobs/497198381)

I restarted the build and it's fine now.

> If you can squash it in the merge that'd be great.

Will do.

 ---

@GaryJones Did you want to have another look or shall I merge it ?

WordPress/Tests/PHP/IniSetUnitTest.inc
```diff
@@ -0,0 +1,37 @@
...
+ini_set('auto_detect_line_endings', true); // Ok.
+ini_set('highlight.bg', '#FFFFFF'); // Ok.
+ini_set('highlight.comment', '#FFFFFF'); // Ok.
+ini_set('highlight.default', '#FFFFFF'); // Ok.
+ini_set('highlight.html', '#FFFFFF'); // Ok.
+ini_set('highlight.keyword', '#FFFFFF'); // Ok.
+ini_set('highlight.string', '#FFFFFF'); // Ok.
+ini_set('short_open_tag', true); // Ok.
+ini_set('short_open_tag', 1); // Ok.
+ini_set('short_open_tag', 'On'); // Ok.
```
GaryJones: How about `'on'`? Is this lowercase a valid value?

WordPress/Tests/PHP/IniSetUnitTest.inc
```diff
@@ -0,0 +1,37 @@
+<?php
+
+ini_set('auto_detect_line_endings', true); // Ok.
```
GaryJones: Might be good to see at least one entry with some unusual whitespace around and inside the parentheses, just to show it's not affecting the result.

WordPress/Tests/PHP/IniSetUnitTest.inc
```diff
@@ -0,0 +1,37 @@
...
+ini_set('filter.default_flags', 0); // Error.
+ini_set('iconv.input_encoding', ''); // Error.
+ini_set('iconv.internal_encoding', ''); // Error.
+ini_set('iconv.output_encoding', ''); // Error.
+ini_set('iconv.output_encoding', ''); // Error.
+ini_set('ignore_user_abort', true); // Error.
+ini_set('log_errors', true); // Error.
+ini_set('max_execution_time', 60); // Error.
+ini_set('memory_limit', -1); // Error.
+ini_set('short_open_tag', false); // Error.
```
GaryJones: Is `FALSE` or `FaLsE` allowed for `ini_set()`? Will it still trigger the error here?

NielsdeBlaauw:
Hi @GaryJones 

Added the additional tests and changed the value verification to be case insensitive.

PR #1652, opened 2019-02-27
Update PrefixAllGlobalsSniff with new core hook

dingo-d:
Added `the_content` to the list of whitelisted core hooks.

Related: https://github.com/WPTRT/theme-sniffer/issues/104

EDIT: We should also wait a bit, because there are extra hooks that we are missing (I'm still waiting on the issue report from people using the Theme Sniffer)


jrfnl:
@dingo-d Just checking: have you had any more input yet ?

dingo-d:
To be honest, I'm not sure at this point. Since we released the Theme Sniffer we get so many false positives that we cannot discern what should be and what shouldn't be prefixed.

In addition to that, I think that `the_content` filter shouldn't be used by the themes directly, so I'll close this PR for now. Once we get a good solution regarding the prefix situation I'll see if we need to add something to the WPCS or if we can just modify the WPThemeReview standards.

Issue #1465, opened 2018-08-25
Extract re-usable code from Sniff.php into new package


GaryJones:
[Sniff.php](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniff.php) contains a lot of helper methods: re-usable bits of code that could be used in one or more sniffs, e.g.:

- `is_foreach_as()` (part of #1457)
- `is_token_in_test_method()`
- `is_test_class()`
- `is_assignment()`
- `is_in_isset_or_empty()`
- `get_array_access_key()`
- `is_comparison()`
- etc.

All of the methods are non-WPCS / non-WP specific (or could be made to be). It would be even nicer if some of these could be moved into PHPCS itself (cc @gsherwood), but if not, then having a separate package, perhaps including the WordPress-specific methods as well, would allow the helper methods to be re-used by other WP-specific coding standards (Neutron, VIP CS, etc.)

`Sniff.php` is _very_ long, and while that's not a bad thing on its own, it does make maintenance trickier, and re-using all of that non-WPCS goodness harder too.

jrfnl:
@GaryJones I'd been thinking about this myself and my take on this is slightly different.

For a number of "groups" of methods which are in the `Sniff` class, I have been thinking of moving them to Traits once we drop PHPCS 2.x support.
* Traits were introduced in PHP 5.4, so once PHPCS 3.x is the minimum we can start using them.
* Traits would allow for smaller groupings of related methods in contrast to having everything in the `Sniff` class, making maintenance simpler.
* Traits would allow these methods to be used, even when sniffs do not extend the base `Sniff` class (if the traits are set up properly that is).

I'd need to look into how this would work with PHPCS autoloading, but with the `<autoload>` directive which was added in one of the early PHPCS 3.x versions, if PHPCS autoloading would not handle this correctly, this will be solvable anyway.

I don't think the other standards is a consideration as such. They can just declare WPCS as a dependency - which most already do anyway as they use WPCS sniffs - and either, at this moment, use the WPCS `Sniff` base class, or if we'd move things around to traits, just use the traits.

At this moment, there are only very few methods in the WPCS `Sniff` class which I think are good enough, code style independent enough + generic enough to be considered for pulling to PHPCS.

[...]

Issue #295, opened 2015-01-05
fputcsv() flagged when writing to stream


paulschreiber:
I have some code that creates a CSV for download:

``` php
$output_stream = fopen( 'php://output', 'w' );

foreach ( $list as $row ) {
    fputcsv( $output_stream, $row );
}
```

PHPCS flags this as an error:

```
  88 | ERROR   | Filesystem writes are forbidden, you should not be using
     |         | fputcsv() (WordPress.VIP.FileSystemWritesDisallow)
```

I'm writing to a stream, not the filesystem.

grappler:
There is no problem in the sniff. Perhaps the Wording should be improved. This could be labeled with `vip`

paulschreiber:
Filed https://github.com/Automattic/VIP-Coding-Standards/issues/404.

PR #1655, opened 2019-03-06
WP/AlternativeFunctions: allow for more input streams with file related functions

jrfnl:
Similar to #1649 which allowed for using `php://input` with `file_get_contents()` and surprisingly inspired by the closing of issue #295.

This PR expands on the earlier work done in relation to the PHP native input streams by:
* recognizing more PHP native input streams;
* recognizing the PHP input stream constants;
* allowing for these in a number of the `file_system_read` group functions as well as for the `file_get_contents` function.

Refs:
* http://php.net/manual/en/wrappers.php.php
* http://php.net/manual/en/features.commandline.io-streams.php

Includes unit tests.

Related #1649
Related #295

Notes:
* I have not checked the WP FileSystem to see if it even could handle these input streams. If it can, we may need to discuss what is the preferred option in that case.
    Personally, this to me seems like something for which the WP FileSystem would be overkill/superfluous.
* At a later point in time, the new method + properties could be a candidate for moving to `Sniff` or a separate utility class.
    As no other sniffs currently need them though, this is not necessary at this moment and could possible be combined with/actioned when #1465 comes into play.

### DIFF
--- a/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
+++ b/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
@@ -26,6 +26,47 @@ use WordPressCS\WordPress\AbstractFunctionRestrictionsSniff;
 class AlternativeFunctionsSniff extends AbstractFunctionRestrictionsSniff {
 
 	/**
+	 * Local input streams which should not be flagged for the file system function checks.
+	 *
+	 * @link http://php.net/manual/en/wrappers.php.php
+	 *
+	 * @var array
+	 */
+	protected $allowed_local_streams = array(
+		'php://input'  => true,
+		'php://output' => true,
+		'php://stdin'  => true,
+		'php://stdout' => true,
+		'php://stderr' => true,
+	);
+
+	/**
+	 * Local input streams which should not be flagged for the file system function checks if
+	 * the $filename starts with them.
+	 *
+	 * @link http://php.net/manual/en/wrappers.php.php
+	 *
+	 * @var array
+	 */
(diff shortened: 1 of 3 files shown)


WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -223,4 +285,29 @@ public function process_matched_token( $stackPtr, $group_name, $matched_content
...
 
+	/**
+	 * Determine based on the "raw" parameter value, whether a file parameter points to
+	 * a local data stream.
+	 *
+	 * @param string $raw_param_value Raw parameter value.
+	 *
+	 * @return bool True if this is a local data stream. False otherwise.
+	 */
+	protected function is_local_data_stream( $raw_param_value ) {
```
GaryJones: It's possible to create [custom stream wrappers](https://secure.php.net/manual/en/function.stream-wrapper-register.php) (we have one for VIP: `vip://`), so I'd like to see a public property that would allow for this to be added at the ruleset level, and then added as part of a check here.

jrfnl: I think those kind of wrappers would be extremely rare in the context of WP and depending on how they can be used shouldn't necessarily be whitelisted - similar to the PHP native `php://filter` wrapper which I didn't include in this sniff on purpose.

As for VIPCS, this sniff could just be extended by a VIP native sniff and the extra wrapper added to the existing properties in a `__construct()` method.

jrfnl:
P.s.: @GaryJones you can ignore the separate review request, that was just GH being weird.

PR #1661, opened 2019-03-27
Core: forbid assignments in conditions

jrfnl:
Move the `WordPress.CodeAnalysis.AssignmentInCondition` from the `Extra` ruleset to the `Core` ruleset.

As per the proposal in https://make.wordpress.org/core/2019/03/26/coding-standards-updates-for-php-5-6/

Open actions:
* [x] Adjust Core PHP handbook to mention this rule in the `Clever code` section

/cc @pento


pento:
Handbook is updated.

jrfnl:
Thanks @pento !

I've updated the PR to use the exact text from the handbook in the inline documentation, as well as changed the placement of the rule to match the order of the rules in the handbook.

Issue #1624, opened 2019-01-13
Core: Ban Yoda conditions, loose comparisons and assignments in conditions


jrfnl:
## Is your feature request related to a problem?

_Yoda conditions_ are the practice to reverse how comparisons in conditions are written, having the value being compared to on the left hand of the condition and the variable on the right hand, to prevent accidentally assigning something to a variable in a condition.
```php
// This condition:
if ( $variable === 'value' ) {}

// ...would be written like this in Yoda style:
if ( 'value' === $variable ) {}

// ... to prevent this:
if ( $variable = 'value' ) {}
```

**The problem with this is that code readability is decreased significantly to prevent a beginner's mistake which shouldn't be made in the first place and the chances of which are small when using strict comparisons anyhow.**

Now, the WordPress coding standards contains a few sniffs related to this:
* `Core`: `WordPress.PHP.YodaConditions` (`error`) - demands the use of Yoda conditions as per [the handbook](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#yoda-conditions).
    However, this sniff only looks at _comparison operators_, i.e. `==`, `===`, `!=` and `!==`, **not** at _assignment operators_, like `=`, so in effect, **it doesn't actually prevent assignments in conditions in the first place**.
* `Extra`: `WordPress.CodeAnalysis.AssignmentInCondition` (`warning`) - warns against assignments found in conditions of control structures as well as, in as far possible, assignments in the condition part of ternaries.
* `Extra`: `WordPress.PHP.StrictComparisons` (`warning`) - warns against using loose comparisons.
* `Extra`: `WordPress.PHP.StrictInArray` (`warning`) - warns against using loose comparisons in array comparison functions.

Note: none of these sniffs currently contain (auto-)fixers.

## Describe the solution you'd like

At this moment, I'd like to propose the following:
* Drop the handbook rule to demand Yoda conditions.
* Remove the `WordPress.PHP.YodaConditions` sniff - both from the `Core` ruleset and from the repo (initially deprecated and emptied out, to be removed in the next major release).
* Move the `WordPress.CodeAnalysis.AssignmentInCondition` sniff to the `Core` ruleset and change it from a `warning` to an `error`.
* Move the `WordPress.PHP.StrictComparisons` and `WordPress.PHP.StrictInArray` sniffs to the `Core` ruleset, but leave these as a `warning`.

### Impact
[...]

PR #1662, opened 2019-03-27
Core: warn against the use of loose comparisons

jrfnl:
As per the proposal in #1624, this moves the sniffs which `warn` against the use of loose comparisons from `Extra` to the `Core` ruleset.

Open actions:
* [x] Adjust Core PHP handbook to mention this rule in the `Clever code` section (or wherever else it should go).

/cc @pento

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -392,6 +392,11 @@
 	<rule ref="Squiz.PHP.DisallowMultipleAssignments"/>
 	<rule ref="Generic.Formatting.DisallowMultipleStatements"/>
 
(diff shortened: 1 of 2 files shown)


pento:
Handbook is updated.

jrfnl:
Thanks @pento !

I've updated the PR to use the exact text from the handbook in the inline documentation, as well as changed the placement of the rule to match the order of the rules in the handbook.

If there are no objections, this is ready for merge now.

GaryJones:
There's a merge conflict for this PR.

jrfnl:
@GaryJones Yes, I expected that one ;-) I've rebased the PR, should be okay now.

PR #1668, opened 2019-03-29
New utility method Sniff::is_in_function_call() + implement use of it

jrfnl:
## New utility method `Sniff::is_in_function_call()`

Allows for checking whether a token is (part of) a parameter passed to a specific (set of) function(s).

Optionally checks whether the function call is to a global function.
Optionally only checks the "deepest" function call.

Returns the stack pointer to the `T_STRING` token for the function call or `false`.

Note: most of the time, it is more efficient to sniff from an outward structure-in, however, for some checks going from inside a structure-out works better.

This utility method is specifically for those situations.

### Sniff::is_in_isset_or_empty(): implement new `is_in_function_call()` method

This fixes potential false negatives when a method/namespaced function mirroring the name of the PHP array_key_exists() function would be used.

### Sniff::is_sanitized(): implement new `is_in_function_call()` method

This fixes potential false negatives when a method/namespaced function mirroring the name of the WP unslashing/sanitization functions would be used.

Includes unit test.

### CronInterval: implement new `Sniff::is_in_function_call()` method

This fixes some potential false positives.

Includes unit tests.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -1703,9 +1703,16 @@ abstract class Sniff implements PHPCS_Sniff {
 			return true;
 		}
 
-		// If this isn't a call to a function, it sure isn't a sanitizing function.
-		if ( \T_STRING !== $this->tokens[ $functionPtr ]['code'] ) {
-			if ( $require_unslash ) {
+		$valid_functions               = $this->sanitizingFunctions;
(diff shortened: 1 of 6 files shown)


jrfnl:
@GaryJones I've added some additional documentation to the docblock now.

PR #1673, opened 2019-03-31
Sniff::$unslashingSanitizingFunctions: add doubleval() and count()

### LINKED ISSUE #1659: add count() to the list of functions not requiring sanitization
This:

```
$variable = count( $_POST['my-array'] );
```

should not trigger:
Detected usage of a non-sanitized input variable: $_POST
Missing wp_unslash() before sanitization.

because this is safe.

jrfnl:
While `doubleval()` is an alias of `floatval()` and shouldn't be used, for the purposes of the `ValidatedSanitizedInput` sniff, both functions should be recognized.

And as `count()` doesn't actually access the data in the variable, but only counts the number of elements, it is also safe to use without unslashing/sanitizing the variable beforehand.
Same goes for the `sizeof()` alias of `count()`.

Includes unit tests.

Fixes #1659

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -309,10 +309,13 @@ abstract class Sniff implements PHPCS_Sniff {
 	protected $unslashingSanitizingFunctions = array(
 		'absint'       => true,
 		'boolval'      => true,
+		'count'        => true,
+		'doubleval'    => true,
 		'floatval'     => true,
 		'intval'       => true,
 		'is_array'     => true,
 		'sanitize_key' => true,
+		'sizeof'       => true,
 	);
 
 	/**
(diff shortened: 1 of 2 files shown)


jrfnl:
@GaryJones Can I merge this ?

GaryJones:
Yes - when I approved, the tests hadn't finished running.

kkmuffme:
Not sure if it can be added here or needs a separate issue:
strpos and stripos also do not have to have sanitized variables, since they do not modify either (and its just useless to sanitize e.g. the HTTP_HOST when its only used in strpos)

jrfnl:
@kkmuffme This PR has been merged already. Please open an issue.

PR #1674, opened 2019-03-31
Update default minimum_supported_version to WP 4.8

jrfnl:
The minimum version should be three versions behind the latest WP release, so what with 5.1 being out, it should now be 4.8.

Includes updating the list of deprecated functions. Input for this based on @JDGrimes's WP deprecated code scanner.


GaryJones:
We're already in WP 5.2-beta1, so while I'll approve this, depending on the timing of WPCS 2.1 releases against WP 5.2, it may be worth bumping it again.

PR #1676, opened 2019-04-01
ValidatedSanitizedInput/Sniff::is_validated(): allow for multi-level array key check

jrfnl:
This PR builds onto earlier improvements to this sniff via PR #1634 and #1635

It adds a new `Sniff::get_array_access_keys()` utility method to retrieve _all_ array keys for a multi-level array access.
The original `Sniff::get_array_access_key()` method defers to the new method under the hood, however, the behaviour of that method has not changed.

It then implements the use of the new `get_array_access_keys()` method in the `WordPress.Security.ValidatedSanitizedInput` sniff and the `Sniff::is_validated()` method.

This allow for more precise checking of whether the correct variable has been validated properly before use.

This should prevent some false positives as well as false negatives.

Includes unit tests.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -1916,17 +1916,20 @@ abstract class Sniff implements PHPCS_Sniff {
 	 *
 	 * @since 0.5.0
 	 * @since 2.1.0 Now recognizes array_key_exists() and key_exists() as validation functions.
+	 * @since 2.1.0 Stricter check on whether the correct variable and the correct
+	 *              array keys are being validated.
 	 *
-	 * @param int    $stackPtr          The index of this token in the stack.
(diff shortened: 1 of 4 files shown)


jrfnl:
> Might be worth adding a note at the top of the test file and Sniff DocBlock that "validated" in the context of this sniff means "check it exists", to help remove some ambiguity until we can a major release and can consider renaming it all.

There's another PR coming up to clarify the error message ;-)

jrfnl:
See #1677

GaryJones:
> There's another PR coming up to clarify the error message ;-)

In this case, I meant for someone looking at the WPCS code, not running PHPCS.

jrfnl:
I understand, but the message clarification should help with that as well.

PR #1678, opened 2019-04-01
Security: allow for type testing superglobals

jrfnl:
When a superglobal variable is being tested with, for instance, `is_numeric()`, unslashing or sanitization are not needed and it's ok for the nonce check to be done after it. This is completely safe.
Ref: https://php.net/manual/en/ref.var.php

Using these type testing functions should however not be regarded as a way of sanitizing/unslashing the variable and the variable does still need to be validated before being passed to one of these functions.

To test whether a variable is used in one of these functions, a new `is_in_type_test()` method has been added to the `WordPressCS\Sniff` class, along with a `$typeTestFunctions` property containing the names of the functions this applies to.

Notes:
* The `is_array()` function which was previously, incorrectly, listed in the `$unslashingSanitizingFunctions` list has been moved to the new property.
    `is_array()` does not unslash or sanitize the contents of a variable, it only checks the variable type.
* Implemented the use of the new `Sniff::is_in_type_test()` method  in both the `ValidatedSanitizedInput` sniff, as well as in the `Sniff:has_nonce_check()` method for the `NonceVerification` sniff.

Includes unit tests via the sniffs.

### DIFF
--- a/WordPress/Tests/Security/NonceVerificationUnitTest.inc
+++ b/WordPress/Tests/Security/NonceVerificationUnitTest.inc
@@ -17,7 +17,7 @@ function ajax_process() {
 }
 add_action( 'wp_ajax_process', 'ajax_process' );
 
-// It's also OK to check with isset() before the the nonce check.
+// It's also OK to check with isset() before the nonce check.
 function foo() {
 	if ( ! isset( $_POST['test'] ) || ! wp_verify_nonce( 'some_action' ) ) {
 		exit;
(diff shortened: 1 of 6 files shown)


WordPress/Tests/Security/NonceVerificationUnitTest.inc
```diff
@@ -160,3 +160,21 @@ function foo_8() {
 	sanitize_pc( $_POST['bar'] ); // Bad.
 	my_nonce_check( sanitize_twitter( $_POST['tweet'] ) ); // Bad.
 }
+
+/*
+ * Using a superglobal in a is_...() function is ok as long as it is not otherwise
```
GaryJones: Can this "not ... without" double-negative be rewritten as a positive, to make it easier to understand?

jrfnl: I've adjusted the comment.

PR #1679, opened 2019-04-01
Sniff::is_sanitized(): allow for map_deep() to sanitize arrays

### LINKED ISSUE #1618: array_walk_recursive
Detected usage of a non-sanitized input variable

```php
$example = isset( $_POST['example'] ) ? ( wp_unslash( $_POST['example'] ) ) : array();
array_walk_recursive( $example, 'sanitize_text_field' );
```

but when I do this, which is essentially the same, I don't get that error:

```php
$example = array();
if ( isset( $_POST['example'] ) ) {
	foreach( $_POST['example'] as $array ) {
		$example[] = array_map( 'sanitize_text_field', wp_unslash( $array ) );
	}	
}
```

### LINKED ISSUE #1660: Sanitizing a multi-dimensional array
How can I sanitize a multi-dimensional array in POST?

`$_POST['my-example']` is a multi-dimensional array like:

```php
array(
	array( 
		'someValue' => 123,
		'hello' => 'world',
	),
	array(
		'someValue' => 456,
		'hello' => 'mars',	
	)
);
```

This doesn't work (since it only works for 1 level of arrays)
```php
$my_example = array_map( 'sanitize_text_field', wp_unslash( $_POST['my-example'] ) );
```

and using:

```php
$my_example = array();
foreach( $_POST['my-example'] as $value ) {
	$my_example[] = array_map( 'sanitize_text_field', wp_unslash( $value ) );
}
```

throws an error that I am using unsanitized POST

Is there any way without using phpcs:ignore?

Thanks for your help!

jrfnl:
Array sanitization via `array_map()` was already recognized. This PR adds recognition of array sanization using `map_deep()`.

**Important note:**
Functions which alter the array by reference, such as the PHP native `array_walk()` and `array_walk_recursive()` are not listed in the new `$arrayWalkingFunctions` property on purpose.

* These cannot easily be used for sanitization as they can't be combined with unslashing, so could only be used in combination with the `$unslashingSanitizingFunctions`.
    Additionally as they alter the array by reference, at this moment, accessing the array after the sanitization through one of these function would _still_ trigger errors.
    If at some point in the future, the above combination would be accounted for, preventing these additional errors would also need to be addressed.
* Similarly, they cannot be used for late escaping as the return value is a boolean, not the altered array.

Also see my comment in the original issue where I go into more detail: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1660#issuecomment-478254403

Includes unit test.

Fixes #1660
Fixes #1618

Once this has been merged, the [Sanitizing array input data](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Sanitizing-array-input-data) wiki page should probably be updated (or an issue be opened saying that it should be with a link to this PR and the tickets).

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -362,6 +362,25 @@ abstract class Sniff implements PHPCS_Sniff {
 	);
 
 	/**
+	 * List of array functions which apply a callback to the array.
+	 *
+	 * These are often used for sanitization/escaping an array variable.
+	 *
+	 * Note: functions which alter the array by reference are not listed here on purpose.
+	 * These cannot easily be used for sanitization as they can't be combined with unslashing.
+	 * Similarly, they cannot be used for late escaping as the return value is a boolean, not
+	 * the altered array.
+	 *
+	 * @since 2.1.0
+	 *
+	 * @var array <string function name> => <int parameter position of the callback parameter>
+	 */
+	protected $arrayWalkingFunctions = array(
+		'array_map' => 1,
+		'map_deep'  => 2,
+	);
+
+	/**
 	 * Functions that format strings.
 	 *
 	 * These functions are often used for formatting values just before output, and
@@ -1778,8 +1797,8 @@ abstract class Sniff implements PHPCS_Sniff {
 
 		$valid_functions               = $this->sanitizingFunctions;
 		$valid_functions              += $this->unslashingSanitizingFunctions;
+		$valid_functions              += $this->arrayWalkingFunctions;
 		$valid_functions['wp_unslash'] = true;
-		$valid_functions['array_map']  = true;
 
 		$functionPtr = $this->is_in_function_call( $stackPtr, $valid_functions );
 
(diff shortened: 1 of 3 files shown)


GaryJones:
Merged, but not amended any wiki page.

TukuToi:
@jrfnl - sorry the ping on this, I read 
> Array sanitization via array_map() was already recognized. 

But I am confused, because...

`$test = array_map( 'esc_url_raw', $_POST['array_member'] );`
==> This will flag as $_POST not unslashed before sanitisation

Thus, I do:
`array_map( 'esc_url_raw', array_map( 'wp_unslash', $_POST['array_member'] ) )`
==> This not only flags the missing unslash but also complains it is not sanitised at all

I don't think my approach is wrong, since `array_map` returns an array of whatever array values we passed thru the callback, this would mean `array_map( 'wp_unslash', $_POST['array_member'] )` gives us an array of unslashed, and thus we can now array_map that thru the sanitisation or escaping function of choice.

Is there a chance that this can be improved? Should I open a ticket for this, or do you consider it rather edge-case and prefer to leave it as is?

jrfnl:
@TukuToi Please don't respond to closed two year old PRs with new questions. They will not be answered. So yes, always open a new issue if there isn't one open already and make sure you show a *complete* code sample as I have a feeling something else is going on.

Issue #924, opened 2017-04-08
Update needed to the WP global variables list


jrfnl:
As discussed in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/904#pullrequestreview-30382018, the list with WP global variables needs to be updated.

The list is currently used by two sniffs:
* `Variables.GlobalVariables` to check for overwriting WordPress native global variables.
* `NamingConventions.PrefixAllGlobals` - to verifies whether a prefix is needed for variables.

For the `NamingConventions.PrefixAllGlobals` sniff the list needs to be complete.
For the `Variables.GlobalVariables` we'll need to allow for a number of variables which plugins/themes are allowed to override.
> we should probably have a small $overrule_allowed list in the Global Var Override sniff and do an array_diff() to prime the list for usage in that sniff.

@JDGrimes has created an initial diff between WP core and the list currently in WPCS which is a good starting point:
> I've just checked over trunk and come up with a new list of WordPress globals, and there are some changes from the old list: https://www.diffchecker.com/jOhQZYHl

But take note of:
> While searching for global and $GLOBALS gives us a good starting point, any variable created in the global namespace and not unset should be on the list.

Some more references about WP global vars (which may well be out of date):
* https://codex.wordpress.org/Global_Variables
* https://codex.wordpress.org/WPMU_Global_Variables

mundschenk-at:
Not sure if this is the right ticket, but `$wp_cockneyreplace` should be on the whitelist to be overridden (because that's what it's for). Of course this might be more or less dead code, I've not seen a plugin in the wild that actually puts something in that global (only some that honor it in the same way as `wp_texturize`does).

dingo-d:
Hi! I just got the report of a false positive for the core `the_content` string in the Theme Sniffer repo: https://github.com/WPTRT/theme-sniffer/issues/104

> Error msg: Hook names invoked by a theme/plugin should start with the theme/plugin prefix. Found: "the_content".

Example code:

```php
        /**
	 * Display post content.
	 *
	 * @since 1.0.0
	 *
	 * @param str $style  Current display post style.
	 */
	public function excerpt( $style ) {

		// Short circuit filter.
		$check = apply_filters( 'bayleaf_display_posts_excerpt', false, $style );
		if ( false !== $check ) {
			return;
		}

		$text = get_the_content( '' );
		$text = wp_strip_all_tags( strip_shortcodes( $text ) );

[...]

PR #1693, opened 2019-04-20
Add missing WP global variable `$current_blog`

schlessera:
I encountered a missing global variable called `$current_blog`.

It is used for Multisite installations and referes to the current site, just as `$current_site` (which is included in the list) refers to the current network.


schlessera:
One of the references in WP Core: https://github.com/WordPress/wordpress-develop/blob/4de03a542e91b92db01e0c5be4583541dcfbd278/src/wp-includes/ms-load.php#L286

jrfnl:
Loosely related to #924

PR #1697, opened 2019-05-05
Accept tag_ID as variable name

Chouby:
WordPress has a global variable `$tag_ID`, used for example in `wp-admin/edit-tags.php` or `wp-admin/term.php`. It should be accepted as an exception for the valid variable name sniff.


Chouby:
Here come the requested changes :)

PR #1711, opened 2019-05-22
Travis: run the code style related and ruleset checks in separate stages

jrfnl:
Travis now offers stages. Using stages we can:
- Run the code style related checks before running any unit tests and stop the build early if any are detected.
- Remove the duplicate unit test runs - i.e. previously we had an extra (third) build against PHP 7.2 (now changed to 7.3) which would run the code style related checks, but would also re-run the unit tests. This extra build will now no longer run the unit tests.

While this does mean that the unit tests will run with a slight delay (the `Sniff` and `Rulesets` stages have to finish before they start), it also means that we:
* Get code style errors reported earlier as it's been moved to be the first stage and the build will just stop if any are found.
* We won't be wasting Travis's resources on builds which will have to be re-run anyway.
* The output of the Travis jobs will be a _lot_ easier to decipher as it only shows the output related to that particular stage.

Ref: https://docs.travis-ci.com/user/build-stages/

Note that `Allowed failures` is no longer listed as a separate block in the Travis result overview, but is _is_ respected.

For more discussion about this:
* https://github.com/travis-ci/travis-ci/issues/7789
* https://travis-ci.community/t/always-show-allow-failures-allowed-failures-when-build-stages-are-used/217
* https://travis-ci.community/t/work-out-kinks-in-interactions-between-stages-allow-fail-and-fast-finish/1090
* https://github.com/travis-ci/travis-ci/issues/9677

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,7 @@ jobs:
         - diff -B --tabsize=4 ./WordPress-Extra/ruleset.xml <(xmllint --format "./WordPress-Extra/ruleset.xml")
         - diff -B --tabsize=4 ./phpcs.xml.dist.sample <(xmllint --format "./phpcs.xml.dist.sample")
 
-    # Make sure the rulesets don't thrown unexpected errors or warnings.
+    # Make sure the rulesets don't throw unexpected errors or warnings.
     # This check needs to be run against a high PHP version to prevent triggering the syntax error check.
     # It also needs to be run against all PHPCS versions WPCS is tested against.
     - stage: rulesets
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,16 +27,67 @@ env:
   # Lowest supported release in the 3.x series with which WPCS is compatible.
   - PHPCS_BRANCH="3.3.1"
 
-matrix:
+stages:
+  - sniff
+  - rulesets
+  - test
+
+jobs:
   fast_finish: true
   include:
-    # Run PHPCS against WPCS. I just picked to run it against 7.2.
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -27,16 +27,67 @@ env:
...
+        - xmllint --noout --schema ./vendor/squizlabs/php_codesniffer/phpcs.xsd ./phpcs.xml.dist.sample
+
+        # Check the code-style consistency of the xml files.
+        - diff -B --tabsize=4 ./WordPress/ruleset.xml <(xmllint --format "./WordPress/ruleset.xml")
+        - diff -B --tabsize=4 ./WordPress-Core/ruleset.xml <(xmllint --format "./WordPress-Core/ruleset.xml")
+        - diff -B --tabsize=4 ./WordPress-Docs/ruleset.xml <(xmllint --format "./WordPress-Docs/ruleset.xml")
+        - diff -B --tabsize=4 ./WordPress-Extra/ruleset.xml <(xmllint --format "./WordPress-Extra/ruleset.xml")
+        - diff -B --tabsize=4 ./phpcs.xml.dist.sample <(xmllint --format "./phpcs.xml.dist.sample")
+
+    # Make sure the rulesets don't thrown unexpected errors or warnings.
```
ntwb:
```suggestion
    # Make sure the rulesets don't throw unexpected errors or warnings.
```

jrfnl:
@ntwb Ah, that explains why I couldn't force-push the fix myself. We were fixing at the same time ;-)

PR #1712, opened 2019-05-22
ValidHookName: bug fix  / skip over array keys

jrfnl:
String array keys for variable array access would also be examined as part of the hook name, while those should be skipped over as they are not part of the _hook_name.

This fixes that.

Includes unit test.

This commit also lowers the nesting level of the loop by continuing early whenever possible.


jrfnl:
@GaryJones Any particular reason not to merge this (yet) ?

PR #1714, opened 2019-05-28
RestrictedPHPFunctions: add error for use of date()

### LINKED ISSUE #1713: Discourage date() in favor of gmdate()

WordPress core sets PHP time zone to UTC during boot. However it cannot guarantee that timezone isn't changed by third party code, which is relatively common in the wild.

`date()` calls rely on current PHP time zone. If time zone is changed from UTC it causes invalid results through the core and extensions code.

`date()` function should be discouraged in favor of `gmdate()`, which is functionally equivalent, but always operates in UTC and is not affected by runtime time zone changes.

## Additional context

WordPress core code had been recently switched from `date()` to `gmdate()`, see [#46438](https://core.trac.wordpress.org/ticket/46438)

jrfnl:
As this sniff is already included in the WP Core ruleset and the enhancement request is about PHP native functions, this seemed liked a logical sniff to add this to.

I've elected to make this an `error` instead of a `warning` as the code within WP Core has been cleaned up for this issue now and it would be good to prevent new instances of `date()` creeping in.

Fixes #1713

### DIFF
--- a/WordPress/Sniffs/PHP/RestrictedPHPFunctionsSniff.php
+++ b/WordPress/Sniffs/PHP/RestrictedPHPFunctionsSniff.php
@@ -17,6 +17,7 @@ use WordPressCS\WordPress\AbstractFunctionRestrictionsSniff;
  * @package WPCS\WordPressCodingStandards
  *
  * @since   0.14.0
+ * @since   2.2.0  New group `date` added.
  */
 class RestrictedPHPFunctionsSniff extends AbstractFunctionRestrictionsSniff {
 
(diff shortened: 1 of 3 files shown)


jrfnl:
Question: Should this be added to the handbook ? /cc @ntwb @pento

jrfnl:
Rebased to get a passing build after the move. @GaryJones Feel free to merge this if all is good.

> Question: Should this be added to the handbook ?

@ntwb @pento Had a chance to have a look at this yet ?

Issue #1715, opened 2019-05-31
wp_sanitize_redirect missing from sanitizing list


losnappas:
Hello,

as title, I believe https://developer.wordpress.org/reference/functions/wp_sanitize_redirect/ is missing from the validations. [possibly this line](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniff.php#L252)

At least search finds nothing:
>  We couldn’t find any code matching 'wp_sanitize_redirect' in WordPress-Coding-Standards/WordPress-Coding-Standards
You could search all of GitHub or try an advanced search.

jrfnl:
@losnappas Nice catch! Would you like to submit a PR to fix this ?

losnappas:
Sure thing, so long as adding it to that list is all it takes? I also came across `wp_kses_no_null` on that page and, as it seems to be missing, I'll add that too.

jrfnl:
@losnappas 

> Sure thing, so long as adding it to that list is all it takes

Well, that + adding a unit test covering the change to the https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Tests/Security/ValidatedSanitizedInputUnitTest.inc file.

The unit test only needs to be a very small snippet of code which would trigger the sniff if this function would not have been added to the list.

> I also came across `wp_kses_no_null` on that page and, as it seems to be missing, 

I'm not sure that using `wp_kses_no_null()` on its own, without other sanitization, is sufficient to consider a variable sanitized as that would, for instance, still allow for arbitrary HTML injection.

What do you think ?

losnappas:
You're right. Got ahead of myself there. Took it out and added a line in the text.inc file.

PR #1716, opened 2019-05-31
Add wp_sanitize_redirect

losnappas:
#1715


GaryJones:
@losnappas The failure is due to the fact that by adding your extra line in the middle of the test file, all of the other line numbers are now out by one.

jrfnl:
Thanks @losnappas !

Issue #1722, opened 2019-06-15
Add documentation for each sniff


jrfnl:
All sniffs should be accompanied by documentation in the form of a `SniffNameStandard.xml` file.
A `SniffNameStandard.xml` file contains a (short) description of each rule(s) the sniff checks and a code comparison for each check.

Examples can be found in the PHP_CodeSniffer repository:
* https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Generic/Docs/Formatting/MultipleStatementAlignmentStandard.xml
* https://github.com/squizlabs/PHP_CodeSniffer/tree/master/src/Standards/Generic/Docs

The sniff documentation for the complete standard can be shown using the following command:
```bash
vendor/bin/phpcs --standard=WordPress --generator=text
```

The sniff documentation for an individual sniff can be shown using the following command:
```bash
vendor/bin/phpcs --standard=WordPress --generator=text --sniffs=WordPress.Category.SniffName
```

Aside from the `text` generator, there are also generators available for HTML and Markdown, which will allow us to use these documentation files to auto-generate documentation about the sniffs to be placed in, for instance, the wiki.

At this time, none of the WPCS native sniffs have this kind of documentation available.

This issue is intended to track progress for adding this documentation.

**Adding this documentation is a focus-task for the WCEU contributor day 2019.**
During the WCEU Contributor day Gary (@GaryJones ) and Juliette (@jrfnl) will be available to help contributors get set up and answer any questions they may have.

### Process
* When you start working on the documentation for one of these sniffs, please mention it in the WP Slack `#core-coding-standards` channel and mention your GitHub nickname (during the WCEU contributor day) or leave a comment in this issue (after WCEU Contributor day).
* Gary or Juliette will update this issue to show you have claimed that sniff.

#### Preparation
* Create a [GitHub user account](https://www.wikihow.com/Create-an-Account-on-GitHub) (if you haven't got one already).
* [Fork this repository on GitHub](https://help.github.com/en/articles/fork-a-repo).
* Make sure you have both [git](https://git-scm.com/downloads) as well as [Composer](https://getcomposer.org/) installed on your machine.
* [Clone your fork to your local machine](https://help.github.com/en/articles/cloning-a-repository) using your preferred git client.
* Run `composer install` from the command line in the root directory of your local clone of this repo.

#### For each sniff:
* Create a new branch off `develop` and name it `docs/sniffname`.
    ```bash
[...]

Issue #76, opened 2013-10-11
Flag manipulation of server (default) timezone


westonruter:
- [x] Flag any use of `date_default_timezone_set`

See also: [Do not change the default timezone from UTC in WordPress](https://weston.ruter.net/2013/04/02/do-not-change-the-default-timezone-from-utc-in-wordpress/)

> <h4 id="manipulating-the-timezone-server-side">Manipulating the timezone server-side <a href="http://vip.wordpress.com/documentation/code-review-what-we-look-for/#manipulating-the-timezone-server-side" class="anchor">#</a></h4>
> 
> <p>Using date_default_timezone_set() and similar isn’t allowed because it conflicts with stats and other systems. Developers instead should use WordPress’s internal timezone support. <a href="http://vip.wordpress.com/documentation/use-current_time-not-date_default_timezone_set/">More information</a>.</p>

PR #1731, opened 2019-06-20
Docs: Add TimezoneChange XML doc

GaryJones:
See #1722.


WordPress/Docs/WP/TimezoneChangeStandard.xml
```diff
@@ -0,0 +1,22 @@
+<documentation title="Disallow the changing of timezone.">
+    <standard>
+    <![CDATA[
+      Using date_default_timezone_set() and similar isn't allowed, instead use WP internal timezone support.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using WP internal timezone support.">
```
GaryJones: The description doesn't match up with the example. The description says WP, but the example is pure PHP!

(Pulled from the unit test file.)

WordPress/Docs/WP/TimezoneChangeStandard.xml
```diff
@@ -0,0 +1,22 @@
+<documentation title="Disallow the changing of timezone.">
+    <standard>
+    <![CDATA[
+      Using date_default_timezone_set() and similar isn't allowed, instead use WP internal timezone support.
```
jrfnl:
```suggestion
<documentation title="Disallow Changing the Timezone">
```
See: https://github.com/WordPress/WordPress-Coding-Standards/pull/1732#discussion_r301523885

WordPress/Docs/WP/TimezoneChangeStandard.xml
```diff
@@ -0,0 +1,22 @@
+<documentation title="Disallow the changing of timezone.">
```
jrfnl:
```suggestion
<documentation title="Disallow Changing the Timezone">
```
See: https://github.com/WordPress/WordPress-Coding-Standards/pull/1732#discussion_r301523885

WordPress/Docs/WP/TimezoneChangeStandard.xml
jrfnl: I've done some tracing back of this and it looks like the sniff has not had any significant changes (other than sniff technical improvements) since it was first added to the library:
* Issue #76 requested it.
* PR #103 added it.

The issue and the inline docs refer to the following links as relevant:
* https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/#manipulating-the-timezone-server-side
* https://wpvip.com/documentation/vip-go/use-current_time-not-date_default_timezone_set/
* https://weston.ruter.net/2013/04/02/do-not-change-the-default-timezone-from-utc-in-wordpress/

Based on the information there, the more recent issue around the use of `gmdate()` instead of `date()` - #1713 - is very much related to this as well.

Now there are two things which are confusing here:
1. The unit test "good" case.
    That case shouldn't really be there as it doesn't test the sniff at all.
    The sniff is not looking for anything in the code of the "good" case, so would never trigger on it.
    It could be replaced with something like `My_Class::date_default_timezone_set( 'Foo/Bar' );` which would be testing the proper functioning of the sniff against false positives.
2. The error message refers to "_WP internal timezone support_" which is an unclear phrase to begin with.
    This is IMO where the docs you are creating come in and can help hugely to clarify what is meant by this, though if someone can come up with a simple phrase to improve the error message, that could be helpful too.
    I think that summarizing the information found in the last two docs links I mention above, to come up with a better description and better code samples would be the way forward.
    I can also imagine some additional issues may need to be opened, sooner rather than later, to get WPCS to detect additional cases of _doing it wrong_ regarding retrieving/displaying date/time information. /cc @Rarst 

On a related note - PR #1714 added a check for the use of `date()` to the `WordPress.PHP.RestrictedPHPFunctions` sniff.
This change has not yet been released and based on the above, I'm now wondering if we should move that check to this sniff instead.

What do you think ?

Rarst: Yeah, `gmdate()` change is purely to make timezone issue more foolproof, so if it can be here.

> Using date_default_timezone_set() and similar isn't allowed, instead use WP internal timezone support.

Now this line is a mess. There is no "similar", there aren't other things that would mess up default time zone. And this doesn't really come up in _WordPress_ context, this comes up when people do _PHP_ in WordPress in incompatible way.

It should be something like this (probably in sniff itself as well?):

> Using date_default_timezone_set() isn't allowed, because WP core needs default time zone set to UTC. Use DateTime object or core API functions instead.

PS https://wpvip.com/documentation/vip-go/use-current_time-not-date_default_timezone_set/ this is highly outdated rec, local timestamps are a huge problem (I am close to getting rid of `current_time('timestamp')` in core) and `timezone_string` is highly preferable to `gmt_offset` if available (that mess is about to get abstracted as `wp_timezone()` in core).

jrfnl:
@GaryJones Now there is clarity about the future of the sniff - this sniff has been deprecated and the check moved to the `WordPress.DateTime.RestrictedFunctions` sniff, would you like to update this PR to cover the new `WordPress.DateTime.RestrictedFunctions` sniff instead ?

GaryJones:
I'm not going to get to this any time soon - happy for someone else to take it over.

jrfnl:
Keeping my fingers crossed that there might be someone at contributor day WCUS who'd like to.

GaryJones:
Noting that I've refreshed the PR on my local branch, but my account is having difficulty pushing to this repo at the moment.

GaryJones:
Manually copying it here for visibility:

<details>
<summary>WordPress/Docs/DateTime/RestrictedFunctionsStandard.xml</summary>

```xml
<documentation title="Restricted Date and Time Functions">
    <standard>
    <![CDATA[
      The restricted functions date_default_timezone_set() and date() should not be used.

      Using date_default_timezone_set() isn't allowed, because WordPress core needs default time zone set to UTC. Use DateTime object or WP core API functions instead.

      Using date() isn't allowed, as it is affected by runtime timezone changes which can cause date/time to be incorrectly displayed. Use gmdate() instead.
    ]]>
    </standard>
    <code_comparison>
        <code title="Valid: Using DateTime object.">
        <![CDATA[
$date = new DateTime();
$date->setTimezone(
    new DateTimeZone( 'Europe/Amsterdam' )
);
        ]]>
        </code>
        <code title="Invalid: Using date_default_timezone_set().">
        <![CDATA[
date_default_timezone_set( 'Europe/Amsterdam' );
        ]]>
        </code>
    </code_comparison>
    <code_comparison>
        <code title="Valid: Using gmdate().">
            <![CDATA[
$last_updated = gmdate(
    'Y-m-d\TH:i:s',
    strtotime( $plugin['last_updated'] )
);
        ]]>
        </code>
        <code title="Invalid: Using date().">
            <![CDATA[
$last_updated = date(
    'Y-m-d\TH:i:s',
    strtotime( $plugin['last_updated'] )
);
        ]]>
        </code>
    </code_comparison>
</documentation>
```
</details>

Commit message: 

> Docs: Refresh restricted datetime functions docs
>
> Kept the descriptions for individual code comparisons simpler, and populated the `<standard>` element with more description instead.

GaryJones:
Fixed my pushing issue, so now rebased onto `develop`, and it is now ready for review (again).

Rarst:
I have no memory of this or mental bandwidth for it, but from quick look at last screenshot I only don't really like `gmdate()` bit.

It would be a little hard to succinctly document, but `gmdate()` is kinda "technical" 1:1 replacement, _but_ in many cases code should be written with DateTime anyway. That "wrong" example in practice it should be changed to create from format or something (depending on what's in that input).

WordPress/Docs/DateTime/RestrictedFunctionsStandard.xml
```diff
@@ -0,0 +1,44 @@
...
+
+      Using date() isn't allowed, as it is affected by runtime timezone changes which can cause date/time to be incorrectly displayed. Use gmdate() instead.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using DateTime object.">
+            <![CDATA[
+$date = new DateTime();
+$date->setTimezone(
+    new DateTimeZone( 'Europe/Amsterdam' )
```
jrfnl, line 16: Would it be an idea to "WordPressify" this code sample a little more and use something like `get_option()` with `gmt_offset` or `timezone_string` to retrieve the timezone to be passed to `new DateTimeZone()` ?

jrfnl (changes requested):
Verified:
- [x] File is in the right place and uses the correct file name.
- [x] Displays correctly within character limits when displayed on CLI.
- [x] Uses space indentation.
- [x] Code samples comply with WPCS (except for the bit which is being described).
- [ ] **Missing**: Uses `<em>` tags to emphasize the "good" vs "bad" code.
- [ ] The start of the "open" and "close" part of the `<![CDATA[` markers do not align.

When looking at the docs, I'm wondering if this is one of those where it would be good to have multiple `<standard>` and `<code comparison> blocks.

At this moment, there is one `<standard>` block at the top with two associated code comparisons, but parts of the text in the `<standard>` block apply specifically to each code sample, so I'm wondering if a structure like the below would work better ?
```
<standard>Generic info about the sniff</standard>
<standard>Specific info about the date_default_timezone_set() restriction</standard>
<code comparison>date_default_timezone_set() code sample</code comparison>
<standard>Specific info about the date() restriction</standard>
<code comparison>date() code sample</code comparison>
```

WordPress/Docs/DateTime/RestrictedFunctionsStandard.xml
Rarst: I think the basic PHP example is fine there, people tend to change default time zone to something that is _not_ normal WP time zone.

Again, there is a problem of choice between a technical 1:1 code equivalent and idiomatic rewrite, that is hard to decide on for brief out-of-context example.

dingo-d (approved):

PR #1741, opened 2019-06-20
Fixes #1733 - Error on short prefixes

### LINKED ISSUE #1733: PrefixAllGlobals: sniff should check for two-letter prefixes
The sniffer for PrefixAllGlobals does not check for two letter prefixes. This is a problem as there are roughly 80,000 (yes thousand) plugins, and a large number 

Prefixes should be checked for length. They should be at least 3 characters.

INVALID: function mp_save_post()

VALID: function myplugin_save_post()

This should be recommended by default

NielsdeBlaauw:
Fixes #1733

### DIFF
--- a/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
+++ b/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
@@ -933,6 +933,14 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 					array( $prefix )
 				);
 				continue;
+			} elseif ( ! function_exists( 'iconv_strlen' ) && strlen( $prefix, $this->phpcsFile->config->encoding ) < self::MIN_PREFIX_LENGTH ) {
+				$this->phpcsFile->addError(
+					'The "%s" prefix is too short. Short prefixes are not unique enough and may cause name collisions with other code.',
+					0,
+					'ShortPrefixPassed',
+					array( $prefix )
+				);
+				continue;
 			}
 
 			// Validate the prefix against characters allowed for function, class, constant names etc.
--- a/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
+++ b/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
@@ -34,6 +34,17 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	const ERROR_MSG = '%s by a theme/plugin should start with the theme/plugin prefix. Found: "%s".';
 
 	/**
+	 * Minimal number of characters the prefix needs in order to be valid.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @link https://github.com/WordPress/WordPress-Coding-Standards/issues/1733 Issue 1733.
+	 *
+	 * @var int
+	 */
+	const MIN_PREFIX_LENGTH = 3;
+
+	/**
 	 * Target prefixes.
 	 *
 	 * @since 0.12.0
@@ -909,6 +920,16 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 					'The "%s" prefix is not allowed.',
 					0,
 					'ForbiddenPrefixPassed',
+					array( $prefix )
+				);
+				continue;
+			}
+
+			if ( function_exists( 'iconv_strlen' ) && iconv_strlen( $prefix, $this->phpcsFile->config->encoding ) < self::MIN_PREFIX_LENGTH ) {
+				$this->phpcsFile->addError(
+					'The "%s" prefix is too short. Short prefixes are not unique enough and may cause name collisions with other code.',
+					0,
+					'ShortPrefixPassed',
 					array( $prefix )
 				);
 				continue;
--- a/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
+++ b/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
@@ -925,15 +925,12 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 				continue;
 			}
 
-			if ( function_exists( 'iconv_strlen' ) && iconv_strlen( $prefix, $this->phpcsFile->config->encoding ) < self::MIN_PREFIX_LENGTH ) {
-				$this->phpcsFile->addError(
-					'The "%s" prefix is too short. Short prefixes are not unique enough and may cause name collisions with other code.',
-					0,
-					'ShortPrefixPassed',
-					array( $prefix )
-				);
-				continue;
-			} elseif ( ! function_exists( 'iconv_strlen' ) && strlen( $prefix, $this->phpcsFile->config->encoding ) < self::MIN_PREFIX_LENGTH ) {
+			$prefix_length = strlen( $prefix );
+			if ( function_exists( 'iconv_strlen' ) ) {
+				$prefix_length = iconv_strlen( $prefix, $this->phpcsFile->config->encoding );
+			}
+
+			if ( $prefix_length < self::MIN_PREFIX_LENGTH ) {
 				$this->phpcsFile->addError(
 					'The "%s" prefix is too short. Short prefixes are not unique enough and may cause name collisions with other code.',
 					0,
--- a/WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.1.inc
+++ b/WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.1.inc
@@ -425,3 +425,28 @@ define(
 );
 
 // phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[]
+
+/*
+ * Bad: Issue https://github.com/WordPress/WordPress-Coding-Standards/issues/1733.
+ * 
+ * Short prefixes are not allowed. The errors are triggered
+ * on LINE 1 for the unit-test, because it's the phpcs:set command that is
+ * wrong, not the implementing code.
+ */
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] a
+function a_do_something(){}
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] aa
+function aa_do_something(){}
+
+// The following line mimicks an empty prefix value.
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] ,
+function aa_do_something(){}
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] 😊
+function 😊_do_something(){}
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] 😊😊
+function 😊😊_do_something(){}
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[]
--- a/WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.php
+++ b/WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.php
@@ -32,7 +32,7 @@ class PrefixAllGlobalsUnitTest extends AbstractSniffUnitTest {
 		switch ( $testFile ) {
 			case 'PrefixAllGlobalsUnitTest.1.inc':
 				return array(
-					1   => 2, // 1 x error for blacklisted prefix passed.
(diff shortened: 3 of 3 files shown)


WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -33,6 +33,14 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	 */
 	const ERROR_MSG = '%s by a theme/plugin should start with the theme/plugin prefix. Found: "%s".';
 
+	/**
+	 * Minimal number of characters the prefix needs in order to be valid.
+	 *
+	 * @see Issue #1733
```
GaryJones: Let's change this issue number into a standard URL, so that it becomes clickable in some editors, and doesn't presume that the person reading this snippet of code knows where the correct repo lives.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -914,6 +922,16 @@ private function validate_prefixes() {
 				continue;
 			}
 
+			if ( strlen( $prefix ) < self::MIN_PREFIX_LENGTH ) {
```
GaryJones: This should be `mb_strlen()` since multibyte characters could be used as prefixes - see https://3v4l.org/qTg7h as a quick example.

WordPress/Tests/Arrays/ArrayDeclarationSpacingUnitTest.php
```diff
@@ -62,7 +62,7 @@ public function getErrorList( $testFile = '' ) {
 					104 => 2,
 					108 => 1,
 					109 => 1,
-					121 => 1,
+					121 => 1
```
GaryJones: Why the removal of this comma?

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.1.inc
```diff
@@ -425,3 +425,18 @@ define(
...
+ */
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] a
+function a_do_something(){}
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] aa
+function aa_do_something(){}
+
+// The following line mimicks an empty prefix value.
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] ,
+function aa_do_something(){}
```
GaryJones: Clear line needed at the end of the file.

GaryJones: I'd like to see some multibyte prefixes tested for as well.

NielsdeBlaauw:
I've implemented `mb_strlen()` instead of `strlen()` and added tests. `PHP <=5.5` tests now break because `mb_strlen()` has a different outcome. Is there a timeline to remove the php environment tests for versions not supported by WordPress anymore?

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.1.inc
```diff
@@ -439,4 +441,10 @@ function aa_do_something(){}
...
 // phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] ,
-function aa_do_something(){}
\ No newline at end of file
+function aa_do_something(){}
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] 😊
+function 😊_do_something(){}
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] 😊😊
+function 😊😊_do_something(){}
```
GaryJones: Clear line space needed at end of file.

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.php
```diff
@@ -32,7 +32,7 @@ public function getErrorList( $testFile = 'PrefixAllGlobalsUnitTest.1.inc' ) {
 		switch ( $testFile ) {
 			case 'PrefixAllGlobalsUnitTest.1.inc':
 				return array(
-					1   => 2, // 1 x error for blacklisted prefix passed.
+					1   => 8, // 1 x error for blacklisted prefix passed. 4 x error for short prefixes. 2 x no prefix.
```
GaryJones: The description only totals 7 errors.

GaryJones:
> I've implemented `mb_strlen()` instead of `strlen()` and added tests. `PHP <=5.5` tests now break because `mb_strlen()` has a different outcome.

That's odd. I bet @jrfnl will know why :-)

> Is there a timeline to remove the php environment tests for versions not supported by WordPress anymore?

We tend to use the same versions as PHPCS itself, since that's the dependency of WPCS, and not WordPress itself. Requirements for tooling doesn't necessarily have to match what the project requires / uses. That said, feel free to open a new Issue where we can discuss this in more detail.

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.php
NielsdeBlaauw: Good catch. This was apparently introduced quite a while ago in this commit: https://github.com/WordPress/WordPress-Coding-Standards/commit/c23e34299d68b645168105f132f3f6013d94afa5#diff-4d42ce9e9b1e9d6a152b2fe03837da0eR401

But fixed it now.

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.1.inc
NielsdeBlaauw: This should've been added in this commit: https://github.com/WordPress/WordPress-Coding-Standards/pull/1741/commits/bbe5ef96d6601a0de159ab20f42dfb48c63c0a53

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -33,6 +33,14 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	 */
 	const ERROR_MSG = '%s by a theme/plugin should start with the theme/plugin prefix. Found: "%s".';
 
+	/**
+	 * Minimal number of characters the prefix needs in order to be valid.
+	 *
+	 * @see Issue https://github.com/WordPress/WordPress-Coding-Standards/issues/1733
```
jrfnl: As this is pointing to a URI, not a structural element, the `@link` tag would be more appropriate.
Also, the link should be first, with an optional description after it.

Refs:
* https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc-tags.md#56-link
* https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc-tags.md#512-see

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -914,6 +922,16 @@ private function validate_prefixes() {
 				continue;
 			}
 
+			if ( mb_strlen( $prefix ) < self::MIN_PREFIX_LENGTH ) {
+				$this->phpcsFile->addError(
+					'The "%s" prefix is too short. Short prefixes are not unique enough and may still cause name collisions with other plugins.',
```
jrfnl: Prefixes is not just about plugins, but also about themes and about colliding with WP itself or even PHP.

Can we make this message a little more generic ?

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -914,6 +922,16 @@ private function validate_prefixes() {
 				continue;
 			}
 
+			if ( mb_strlen( $prefix ) < self::MIN_PREFIX_LENGTH ) {
```
jrfnl: PHP_CodeSniffer itself uses `iconv` instead `MBString` when examining potentially multi-byte text. IIRC, this is because the `iconv` extension is more widely installed.
The results of using `iconv` would be the same: https://3v4l.org/bejLK

For the `iconv` equivalent, see: https://www.php.net/manual/en/function.iconv-strlen.php

Whichever is used, you cannot rely on the `iconv` or `MBstring` extension being available, so a function call to either should always be accompanied by a call to `function_exists()` or `extension_loaded()`.

Similarly, whichever is used, you will need to pass the `$encoding`.
Since PHPCS 3.0, the default encoding presumed is `utf-8`, but projects can change this using `--encoding=...`.

To get the correct encoding, use `$phpcsFile->config->encoding`.

> I've implemented mb_strlen() instead of strlen() and added tests. PHP <=5.5 tests now break because mb_strlen() has a different outcome.

Setting the `$encoding` should fix this. The default `encoding` used by PHP for MbString changed in PHP 5.6 to `utf-8` which explains why the unit tests pass on PHP 5.6 and higher and fail below.

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.1.inc
```diff
@@ -425,3 +425,26 @@ define(
...
+
+// The following line mimicks an empty prefix value.
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] ,
+function aa_do_something(){}
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] 😊
+function 😊_do_something(){}
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] 😊😊
+function 😊😊_do_something(){}
```
jrfnl: Please reset the prefixes to an empty array at the end of the file as otherwise the unit tests in the next test case file may be influenced.

```php
// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[]
```

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -33,6 +33,14 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	 */
 	const ERROR_MSG = '%s by a theme/plugin should start with the theme/plugin prefix. Found: "%s".';
 
+	/**
+	 * Minimal number of characters the prefix needs in order to be valid.
+	 *
```
jrfnl: Missing the `@since` tag for this constant.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
jrfnl: Unrelated to this PR, but I see a potential issue with the `strtolower()` on line 913 in this context as it doesn't take multibyte characters into account either.

jrfnl: FYI: https://github.com/PHPCompatibility/PHPCompatibility/issues/839 :wink:

NielsdeBlaauw: Very interesting. I've converted the check to use the`iconv_strlen` function, use the correct encoding given by the `phpcsFile` and check for availability of the function.

The `strtolower` shouldn't cause to much of an issue, as it's only used to check against the blacklist, and for now there are no non-ascii characters found in the blacklist.

jrfnl:
@NielsdeBlaauw Thank you for making those changes.

I'd like to have one last look before approving, but as something seems to have gone wrong with the rebase, this PR has now become really hard review as it contains changes from already merged PRs.

Do you need help with doing a rebase on `develop` ? If so, please let me know and I can talk you through it on Slack or in a call if that would help.

Independently of that, I think you may want to have another look at the string length condition.
If I read things correctly, the way it is done now, if `iconv_strlen()` is not available, the length of the prefix is not checked at all.
Should this maybe fall-back to using the generic `strlen()` function if `iconv` is not available ?

NielsdeBlaauw:
Hi @jrfnl 

I've squashed the whole thing to make it a bit cleaner.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -914,6 +925,24 @@ private function validate_prefixes() {
 				continue;
 			}
 
+			if ( function_exists( 'iconv_strlen' ) && iconv_strlen( $prefix, $this->phpcsFile->config->encoding ) < self::MIN_PREFIX_LENGTH ) {
```
jrfnl: @NielsdeBlaauw What do you think: could the `if` and the `elseif` conditions be combined to get rid of the duplicate error block code ?

Also: the plain `strlen()` function does not take an `$encoding` parameter.

PR #1746, opened 2019-06-26
Fixes #1696, errors on invalid post type

### LINKED ISSUE #1696: Check for post type key length in register_post_type
I noticed that while registering a custom post type, my post type key was more than 20 characters and therefore failing to show up on the admin panel. I had to check the developer handbook to see that the post type key must be 20 characters or less

Adding a check for a max of 20 characters would be great.

For example: 
`register_post_type( 'post_type', $args );`
would succed without error.

However, 
`register_post_type( 'post_type_key_is_too_long', $args );`
would throw an error and a suggestion for the user to shorten the title to meet the WordPress Standards

See the link below for the reference to the specific character length. 
[Register Post Type](https://developer.wordpress.org/reference/functions/register_post_type/)

NielsdeBlaauw:
Adds a new sniff. Checks if the first parameters given to a `register_post_type()` call is actually a valid value.

Fixes #1696

### DIFF
--- a/WordPress/Docs/NamingConventions/ValidPostTypeSlugStandard.xml
+++ b/WordPress/Docs/NamingConventions/ValidPostTypeSlugStandard.xml
@@ -1,0 +1,117 @@
+<documentation title="Validate post type slugs">
+    <standard>
+    <![CDATA[
+    The post type slug used in register_post_type() must be between 1 and 20 characters.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: short post type slug.">
+        <![CDATA[
+register_post_type( 
+    <em>'my_short_slug'</em>, 
+    array() 
+);
+        ]]>
+        </code>
+        <code title="Invalid: too long post type slug.">
+        <![CDATA[
+register_post_type( 
+    <em>'my_own_post_type_too_long'</em>, 
+    array() 
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    The post type slug used in register_post_type() can only contain lowercase alphanumeric characters, dashes and underscores.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: no special characters in post type slug.">
+        <![CDATA[
+register_post_type( 
+    <em>'my_post_type_slug'</em>, 
+    array() 
+);
+        ]]>
+        </code>
+        <code title="Invalid: invalid characters in post type slug.">
+        <![CDATA[
+register_post_type( 
+    <em>'my/post/type/slug'</em>, 
+    array() 
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    One should be careful with passing dynamic slug names to "register_post_type()", as the slug may become too long and could contain invalid characters.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: static post type slug.">
+        <![CDATA[
+register_post_type( 
+    <em>'my_post_active'</em>, 
+    array() 
+);
+        ]]>
+        </code>
+        <code title="Invalid: dynamic post type slug.">
+        <![CDATA[
+register_post_type( 
+    <em>"my_post_{$status}"</em>, 
+    array() 
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    The post type slug used in register_post_type() can not use reserved keywords, such as the ones used by WordPress itself.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: prefixed post slug.">
+        <![CDATA[
+register_post_type( 
+    <em>'prefixed_author'</em>, 
+    array() 
+);
+        ]]>
+        </code>
+        <code title="Invalid: using a reserved keyword as slug.">
+        <![CDATA[
+register_post_type( 
+    <em>'author'</em>, 
+    array() 
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    The post type slug used in register_post_type() can not use reserved prefixes, such as 'wp_', which is used by WordPress itself.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: custom prefix post slug.">
+        <![CDATA[
+register_post_type( 
+    <em>'prefixed_author'</em>, 
+    array() 
+);
+        ]]>
+        </code>
+        <code title="Invalid: using a reserved prefix.">
+        <![CDATA[
+register_post_type( 
+    <em>'wp_author'</em>, 
+    array() 
+);
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>
--- a/WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
@@ -1,0 +1,208 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Sniffs\NamingConventions;
+
+use WordPressCS\WordPress\AbstractFunctionParameterSniff;
+use PHP_CodeSniffer\Util\Tokens;
+
+/**
+ * Validates post type names.
+ *
+ * Checks the post type slug for invalid characters, long function names
+ * and reserved names.
+ *
+ * @link https://developer.wordpress.org/reference/functions/register_post_type/
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since 2.2.0
+ */
+class ValidPostTypeSlugSniff extends AbstractFunctionParameterSniff {
+
+	/**
+	 * Max length of a post type name is limited by the SQL field.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @var int
+	 */
+	const POST_TYPE_MAX_LENGTH = 20;
+
+	/**
+	 * Regex that whitelists characters that can be used as the post type slug.
+	 *
+	 * @link https://developer.wordpress.org/reference/functions/register_post_type/
+	 * @since 2.2.0
+	 *
+	 * @var string
+	 */
+	const POST_TYPE_CHARACTER_WHITELIST = '/^[a-z0-9_-]+$/';
+
+	/**
+	 * Array of functions that must be checked.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @var array List of function names as keys. Value irrelevant.
+	 */
+	protected $target_functions = array(
+		'register_post_type' => true,
+	);
+
+	/**
+	 * Array of reserved post type names which can not be used by themes and plugins.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @var array
+	 */
+	protected $reserved_names = array(
+		'post'                => true,
+		'page'                => true,
+		'attachment'          => true,
+		'revision'            => true,
+		'nav_menu_item'       => true,
+		'custom_css'          => true,
+		'customize_changeset' => true,
+		'oembed_cache'        => true,
+		'user_request'        => true,
+		'wp_block'            => true,
+		'action'              => true,
+		'author'              => true,
+		'order'               => true,
+		'theme'               => true,
+	);
+
+	/**
+	 * All valid tokens for in the first parameter of register_post_type().
+	 *
+	 * Set in `register()`.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @var string
+	 */
+	private $valid_tokens = array();
+
+	/**
+	 * Returns an array of tokens this test wants to listen for.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @return array
+	 */
+	public function register() {
+		$this->valid_tokens = Tokens::$textStringTokens + Tokens::$heredocTokens + Tokens::$emptyTokens;
+		return parent::register();
+	}
+
+	/**
+	 * Process the parameter of a matched function.
+	 *
+	 * Errors on invalid post type names when reserved keywords are used,
+	 * the post type is too long, or contains invalid characters.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+
+		$string_pos         = $this->phpcsFile->findNext( Tokens::$textStringTokens, $parameters[1]['start'], ( $parameters[1]['end'] + 1 ) );
+		$has_invalid_tokens = $this->phpcsFile->findNext( $this->valid_tokens, $parameters[1]['start'], ( $parameters[1]['end'] + 1 ), true );
+		if ( false !== $has_invalid_tokens || false === $string_pos ) {
+			// Check for non string based slug parameter (we cannot determine if this is valid).
+			$this->phpcsFile->addWarning(
+				'The post type slug is not a string literal. It is not possible to automatically determine the validity of this slug. Found: %s.',
+				$stackPtr,
+				'NotStringLiteral',
+				array(
+					$parameters[1]['raw'],
+				),
+				3
+			);
+			return;
+		}
+
+		$post_type = $this->strip_quotes( $this->tokens[ $string_pos ]['content'] );
+
+		if ( strlen( $post_type ) === 0 ) {
+			// Error for using empty slug.
+			$this->phpcsFile->addError(
+				'register_post_type() called without a post type slug. The slug must be a non-empty string.',
+				$parameters[1]['start'],
+				'Empty'
+			);
+			return;
+		}
+
+		$data = array(
+			$this->tokens[ $string_pos ]['content'],
+		);
+
+		// Warn for dynamic parts in the slug parameter.
+		if ( 'T_DOUBLE_QUOTED_STRING' === $this->tokens[ $string_pos ]['type'] || ( 'T_HEREDOC' === $this->tokens[ $string_pos ]['type'] && strpos( $this->tokens[ $string_pos ]['content'], '$' ) !== false ) ) {
+			$this->phpcsFile->addWarning(
+				'The post type slug may, or may not, get too long with dynamic contents and could contain invalid characters. Found: %s.',
+				$string_pos,
+				'PartiallyDynamic',
+				$data
+			);
+			$post_type = $this->strip_interpolated_variables( $post_type );
+		}
+
+		if ( preg_match( self::POST_TYPE_CHARACTER_WHITELIST, $post_type ) === 0 ) {
+			// Error for invalid characters.
+			$this->phpcsFile->addError(
+				'register_post_type() called with invalid post type %s. Post type contains invalid characters. Only lowercase alphanumeric characters, dashes, and underscores are allowed.',
+				$string_pos,
+				'InvalidCharacters',
+				$data
+			);
+		}
+
+		if ( isset( $this->reserved_names[ $post_type ] ) ) {
+			// Error for using reserved slug names.
+			$this->phpcsFile->addError(
+				'register_post_type() called with reserved post type %s. Reserved post types should not be used as they interfere with the functioning of WordPress itself.',
+				$string_pos,
+				'Reserved',
+				$data
+			);
+		} elseif ( stripos( $post_type, 'wp_' ) === 0 ) {
+			// Error for using reserved slug prefix.
+			$this->phpcsFile->addError(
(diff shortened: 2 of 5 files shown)


WordPress/Sniffs/WP/PostTypeSniff.php
```diff
@@ -0,0 +1,117 @@
...
+				$this->string_to_errorcode( $post_type . '_TooLong' ),
+				array(
+					$matched_content,
+					$parameters[1]['raw'],
+					self::MAX_POST_TYPE_LENGTH,
+				)
+			);
+		}
+
+		if ( preg_match( '/[A-Z _]/', $post_type ) === 1 ) {
```
GaryJones: Are emoji or other Unicode characters (think umlauts in German or something) allowed? Right now, you're _blacklisting_ uppercase, space and underscore; a _whitelisting_ of what is allowed might be easier to manage.

NielsdeBlaauw: The blacklist comes from this comment 

> max. 20 characters, cannot contain capital letters, underscores or spaces
found here https://codex.wordpress.org/Function_Reference/register_post_type#Parameters

I don't see why using unicode characters would be wrong, as they can correctly be saved to the database. They could possibly cause issues with template loading. The whole idea of using non-ascii characters as a post type is probably ill advised. 

I also have no idea why underscores would be disallowed apart from the comment in the documentation. 

Maybe we can add a whitelist, and post types with non-whitelisted characters trigger a warning/risky notice?

jrfnl: I think the documentation was written from an American English point of view and didn't even consider that people would want to use non-ascii characters.

The post type name is used/can be used in all sorts of places, including query variables, theme templates file names, hook names etc. I imagine there may even be callbacks which use the post type name as the PHP/JS function name.
I dread to start researching this in depth as I'm very afraid of what we would encounter....

With that in mind, I have a feeling that the _real_ rule is "_only use lowercase a-z and possibly a `-` dash_".

> Maybe we can add a whitelist, and post types with non-whitelisted characters trigger a warning/risky notice?

Sounds like a very good idea :+1: , a stern warning at that if it were up me.
This would need additional unit tests, both with unicode post type names, as well as some simple ascii names which I think would be problematic anyway, even though based on the existing documentation they would be allowed...
Think: `q&a` (query vars), `questions?` (query vars), `my/post/type` (template file names), `prefix\post-type`, `comma,separated,list`

jrfnl: Just looked at the plugin handbook;

> To ensure forward compatibility, do not use wp_ as your identifier — it is being used by WordPress core.
>
> Make sure your custom post type identifier does not exceed 20 characters as the post_type column in the database is currently a VARCHAR field of that length.
> 
> If your identifier is too generic — for example: product. It may conflict with other plugins or themes.

That blacklists one more name (or is it a prefix ?) as reserved + explains the why of the hard 20 char limit.
Also: as the database charset is configurable in `wp-config.php` using the `DB_CHARSET` constant, we should presume that not everyone uses `utf8`.

jrfnl: Hmm... just noticed that the example in the plugin handbook uses an `_` underscore in the post type name.... _sigh_

jrfnl: Related: https://wordpress.org/support/article/post-types/

jrfnl: Just a though - I wonder if the "_no underscores_" rule is a documentation error.

Even just looking at the WP native post types, underscores are used.
All Custom Post Type generators I've seen also support them (and I would have expected them to have received bug reports if the code didn't work).

jrfnl: Confirmed - underscores are allowed and only alphanumeric characters are supported:

> Must not exceed 20 characters and may only contain lowercase alphanumeric characters, dashes, and underscores.

Source: https://developer.wordpress.org/reference/functions/register_post_type/

WordPress/Sniffs/WP/PostTypeSniff.php
```diff
@@ -0,0 +1,117 @@
...
+				$stackPtr,
+				$this->string_to_errorcode( $post_type . '_Reserved' ),
+				array(
+					$matched_content,
+					$parameters[1]['raw'],
+				)
+			);
+		}
+
+		if ( strlen( $post_type ) > self::MAX_POST_TYPE_LENGTH ) {
```
jrfnl: Is `strlen()` correct here or should this use `iconv_strlen()` ?

WordPress/Sniffs/WP/PostTypeSniff.php
jrfnl: As per the above discussion, this regex should be adjusted. You may also want to consider moving the regex to a class constant and documenting it there.

WordPress/Sniffs/WP/PostTypeSniff.php
```diff
@@ -0,0 +1,117 @@
...
+					$matched_content,
+					$parameters[1]['raw'],
+					self::MAX_POST_TYPE_LENGTH,
+				)
+			);
+		}
+
+		if ( preg_match( '/[A-Z _]/', $post_type ) === 1 ) {
+			$this->phpcsFile->addError(
+				'%s() called with invalid post type %s. Post type contains invalid characters.',
```
jrfnl: 1. No need to use variable replacement for the function name.
2. Might it be an idea to be explicit about what characters are allowed in the error message (as per the above discussion) ?

WordPress/Sniffs/WP/PostTypeSniff.php
NielsdeBlaauw: I don't think we need to check for multibyte characters, as we are already warning for disallowed characters as part of this rule:

> Must not exceed 20 characters and may only contain lowercase alphanumeric characters, dashes, and underscores.

WordPress/Sniffs/WP/PostTypeSniff.php
```diff
@@ -0,0 +1,117 @@
...
+	const MAX_POST_TYPE_LENGTH = 20;
+
+	/**
+	 * Array of functions that must be checked.
+	 *
+	 * @since X.X.X
+	 *
+	 * @var array Multidimensional array with parameter details.
+	 *     $target_functions = array(
+	 *         (string) Function name.
```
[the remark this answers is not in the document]
NielsdeBlaauw: Made this a simple array type with a description.

WordPress/Docs/NamingConventions/ValidPostTypeSlugStandard.xml
```diff
@@ -0,0 +1,70 @@
...
+    The post type slug used in register_post_type() can not be longer then 20 characters.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: short post type slug.">
+        <![CDATA[
+register_post_type( 'my_short_slug', array() );
+        ]]>
+        </code>
+        <code title="Invalid: long post type slug.">
```
jrfnl:
```suggestion
        <code title="Invalid: too long post type slug.">
```

WordPress/Docs/NamingConventions/ValidPostTypeSlugStandard.xml
```diff
@@ -0,0 +1,70 @@
...
+        </code>
+        <code title="Invalid: invalid characters in post type slug.">
+        <![CDATA[
+register_post_type( 'my/post/type/slug', array() );
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Register post type should be carefull with dynamic slug names, as they can not be verified to be correct.
```
jrfnl:
```suggestion
    One should be careful with passing dynamic slug names to "register_post_type()", as they may be too long and can contain invalid characters.
```

WordPress/Docs/NamingConventions/ValidPostTypeSlugStandard.xml
```diff
@@ -0,0 +1,70 @@
...
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: static post slugs.">
```
jrfnl:
```suggestion
        <code title="Valid: static post type slug.">
```

WordPress/Docs/NamingConventions/ValidPostTypeSlugStandard.xml
```diff
@@ -0,0 +1,70 @@
...
+        <![CDATA[
+register_post_type( 'prefixed_author', array() );
+        ]]>
+        </code>
+        <code title="Invalid: dynamic post slug.">
```
jrfnl:
```suggestion
        <code title="Invalid: dynamic post type slug.">
```

jrfnl:
@NielsdeBlaauw I saw the updates. Will you leave a comment when the PR is ready for another review ?

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
```diff
@@ -0,0 +1,175 @@
...
+					),
+					3
+				);
+				return;
+			}
+		}
+
+		// Handling string interpolation.
+		$stringPos = $this->phpcsFile->findNext( Tokens::$textStringTokens, $parameters[1]['start'], $parameters[1]['end'] + 1, false, null, true );
+		if ( false === $stringPos ) {
```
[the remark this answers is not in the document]
NielsdeBlaauw: I've caught this in the dynamic warning. Open to suggestions for changing the message.

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
```diff
@@ -0,0 +1,175 @@
...
+	 *
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		// Check for non string based slug parameter (we cannot determine if this is valid).
+		$valid_tokens = Tokens::$textStringTokens + Tokens::$emptyTokens + Tokens::$heredocTokens;
+		for ( $i = $parameters[1]['start']; $i <= $parameters[1]['end']; $i++ ) {
+			if ( ! isset( $valid_tokens[ $this->tokens[ $i ]['code'] ] ) ) {
+				$this->phpcsFile->addWarning(
+					'Dynamic post type slug found. It is not possible to determine validity of this slug. Found: %s.',
```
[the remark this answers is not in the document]
NielsdeBlaauw: I'm open for suggestions to change the message here. I'm not quite sure what to make of this myself.

NielsdeBlaauw:
@jrfnl Latest changes are now all pushed.

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
```diff
@@ -0,0 +1,208 @@
...
+				'register_post_type() slug may or may not get too long with dynamic contents and could contain invalid characters. Found: %s.',
+				$stringPos,
+				'PartiallyDynamic',
+				$data
+			);
+			$post_type = $this->strip_interpolated_variables( $post_type );
+		}
+
+		// Error for using reserved slug names.
+		if ( strlen( $post_type ) === 0 ) {
```
jrfnl: As all the other checks are irrelevant when the post type is empty, I'd like to suggest moving this up to above the "dynamic contents" check and adding a `return` to it.
For this specific case, we know that none of the other errors can happen if this error gets triggered, so a `return` here after the error is thrown, is perfectly fine.

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
```diff
@@ -0,0 +1,208 @@
...
+				'register_post_type() called with reserved post type %s. Reserved post types should not be used as they interfere with the functioning of WordPress itself.',
+				$stringPos,
+				'Reserved',
+				$data
+			);
+		}
+
+		if ( preg_match( self::POST_TYPE_WP_PREFIX, $post_type ) !== 0 ) {
+			$this->phpcsFile->addError(
+				'register_post_type() called with invalid prefix. Found: %s.',
```
jrfnl:
```suggestion
				'The post type passed to register_post_type() uses a prefix reserved for WordPress itself. Found: %s.',
```

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
```diff
@@ -0,0 +1,208 @@
...
+		if ( isset( $this->reserved_names[ $post_type ] ) ) {
+			$this->phpcsFile->addError(
+				'register_post_type() called with reserved post type %s. Reserved post types should not be used as they interfere with the functioning of WordPress itself.',
+				$stringPos,
+				'Reserved',
+				$data
+			);
+		}
+
+		if ( preg_match( self::POST_TYPE_WP_PREFIX, $post_type ) !== 0 ) {
```
jrfnl: See my remark on the constant declaration.

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
```diff
@@ -0,0 +1,208 @@
...
+			$this->phpcsFile->addError(
+				'register_post_type() called with invalid prefix. Found: %s.',
+				$stringPos,
+				'InvalidPrefix',
```
jrfnl:
```suggestion
				'ReservedPrefix',
```

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
jrfnl: This should probably be an `elseif`, otherwise `wp_block` would give two errors instead of just the one.

```php
register_post_type( 'wp_block', array() );
```

WordPress/Docs/NamingConventions/ValidPostTypeSlugStandard.xml
```diff
@@ -0,0 +1,117 @@
...
+register_post_type( 
+  <em>"my_post_{$status}"</em>, 
+  array() 
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    The post type slug used in register_post_type() can not use reserved keywords.
```
jrfnl: I'm just wondering if "reserved keywords" should be expanded a bit to explain what that means ?

WordPress/Docs/NamingConventions/ValidPostTypeSlugStandard.xml
```diff
@@ -0,0 +1,117 @@
...
+register_post_type( 
+  <em>'author'</em>, 
+  array() 
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    The post type slug used in register_post_type() can not use reserved prefixes.
```
jrfnl: Same here: should this list the single reserved prefix which the sniff is checking for to make it explicit ?

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
NielsdeBlaauw: replaced by `stripos`.

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
```diff
@@ -0,0 +1,208 @@
...
+	);
+
+	/**
+	 * All valid tokens for in the first parameter of register_post_type().
+	 *
+	 * Set in `register()`.
+	 *
+	 * @var string
+	 */
+	public $valid_tokens = array();
```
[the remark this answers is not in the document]
NielsdeBlaauw: Setting this to `private` to be consistent with others like `ArrayIndentationSniff`.

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
```diff
@@ -0,0 +1,208 @@
...
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		// Check for non string based slug parameter (we cannot determine if this is valid).
+		$contains_text_string_token = $this->phpcsFile->findNext( $this->valid_tokens, $parameters[1]['start'], $parameters[1]['end'] + 1 );
+		if ( ! $contains_text_string_token || ( $contains_text_string_token && $this->phpcsFile->findNext( Tokens::$functionNameTokens, $parameters[1]['start'], $parameters[1]['end'] + 1 ) !== false ) ) {
```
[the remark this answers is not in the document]
NielsdeBlaauw: The case I was trying to catch is 

`register_post_type( sprintf( 'something_%s', 'something_else' ), array() )` where 'something_%s' and 'something else' are found as valid tokens, but the result is invalid.

NielsdeBlaauw:
> However, I'm wondering if for the test cases where we are certain of the value and it is an invalid token type, we should throw an error at severity 5 ? I.e for the `1000`, `null` etc.

I attempting this during the previous iteration, but it becomes complicated when something like `register_post_type( 'test' . 1000, array() );` happens. I want to think a bit more about how I want to handle this (if I want to make an exception at all).

NielsdeBlaauw:
So I played around with this for a bit and don't really see a way to reliably determine if something is partially a string and partially dynamic, or just a constant that might result in a valid name. I rewrote the warning to indicate that the code should be manually inspected as it is not a string literal. Phrasing could probably be improved a bit.

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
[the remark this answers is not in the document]
NielsdeBlaauw: Changed this anyway and am looking for non-valid tokens now instead with new wording from here: https://github.com/WordPress/WordPress-Coding-Standards/pull/1746#issuecomment-536213620

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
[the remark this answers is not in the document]
NielsdeBlaauw: See https://github.com/WordPress/WordPress-Coding-Standards/pull/1746#issuecomment-536213620

NielsdeBlaauw:
@jrfnl I think I've addressed most/all of the latest feedback round now. I'd kindly take suggestions for the phrasing for the non string literal case.

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
jrfnl: Please add the above test case to the unit test file so we can safeguard that this only throws one error, not two.

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
jrfnl: Changing this back to "Unresolved".

Sorry, my previous remark may not have been clear enough, I meant _moving this check up above the `PartiallyDynamic`_ error, not to above the `NotStringLiteral` error.

This check has now been moved too far up, will never be triggered and isn't tested either.

If the function call doesn't have any parameter, the `process_parameters() method won't be called, the `process_no_parameters()` method will be called instead.
See: https://github.com/WordPress/WordPress-Coding-Standards/blob/058be1bc5b2c5f20bcb0fb44603e43abe6315be4/WordPress/AbstractFunctionParameterSniff.php#L69-L78

And if it's an empty string, the check still will not be triggered as it is now above the point in the sniff where `$this->strip_quotes()` is called, so the string length wouldn't be `0`, but `2`.

You can verify this by running the sniff against the unit tests with `--report=full,source`, the `Empty` code is not triggered by the tests.

jrfnl:
> @jrfnl I think I've addressed most/all of the latest feedback round now. I'd kindly take suggestions for the phrasing for the non string literal case.

@NielsdeBlaauw Thanks again for the updates. I've "unresolved" two issues and left comments with them as they still need a little tweak. Other than that, I'm ok with merging this sniff.

I'm still struggling a bit with only throwing the `warning` at severity `3` for clearly invalid values for the post type like `1000` or `true`, but we can address that at a later point in time.

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
NielsdeBlaauw: Additional test has been added.

WordPress/Sniffs/NamingConventions/ValidPostTypeSlugSniff.php
NielsdeBlaauw: Right, didn't know about the `process_no_parameters()`, and was thinking about catching the empty case as early as possible.

It now actually correctly checks for an empty first param after the latest commit.

WordPress/Tests/NamingConventions/ValidPostTypeSlugUnitTest.php
```diff
@@ -0,0 +1,76 @@
...
+namespace WordPressCS\WordPress\Tests\NamingConventions;
+
+use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
+
+/**
+ * Unit test class for the PostType sniff.
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   X.X.X
```
jrfnl:
```suggestion
 * @since   2.2.0
```

jrfnl:
Thank you @NielsdeBlaauw for all your work on this and congratulations on yet another new sniff created :tada:

PR #1754, opened 2019-06-27
Add `$plugin` to the list of global variables that shouldn't be overridden

johnbillion:
If a plugin overrides the `$plugin` global, things can break. See https://github.com/Rarst/laps/issues/35.

This variable is missing from the list of global WP variables.


Rarst:
Note that the same issue extends to `$network_plugin` and `$mu_plugin`. This is more noticeable now since starting with WP 5.1 these pass through new hooks, though they were never a good idea to override.

jrfnl:
@Rarst Thanks for the heads-up. PRs for this are welcome :hugs:

Issue #1762, opened 2019-07-10
Proposal: Add rules about formatting of the spread operator


jrfnl:
## Is your feature request related to a problem?
PHP 5.6 introduced the spread operator `...` for packing arguments in function declarations (variadic functions) and unpacking them in function calls.

PHP 7.4 will introduce unpacking of arrays using the spread operator.

Refs:
* https://www.php.net/manual/en/migration56.new-features.php#migration56.new-features.variadics
* https://www.php.net/manual/en/migration56.new-features.php#migration56.new-features.splat
* https://wiki.php.net/rfc/spread_operator_for_array

At this moment WPCS has no formatting rules for the spread operator.

## Describe the solution you'd like
I would like to propose adding the following rules regarding the spread operator to the WP Coding Standards handbook:
* Rule 1: There should be one space, or a new line + indentation, before the spread operator.
* Rule 2: There should be no space (or comments for that matter) between the spread operator and the variable/function call it applies to.
* Rule 3: When combining the spread operator with the reference operator, there should be no space between them.

```php
// Correct.
function foo( &...$spread ) {
	bar( ...$spread );

	bar(
		[ ...$foo ],
		...array_values( $keyed_array )
	);
}

// Incorrect.
function fool( &   ... $spread ) {
	bar(...
             $spread );

	bar(
		[...  $foo ],.../*comment*/array_values( $keyed_array )
	);
}
```

If this proposal is accepted, we can then add sniff(s) to WPCS to check these rules automatically and - except for issues with comments - auto-fix them.

## Notes for the WPCS Implementation:
* Rule 1 is generally already checked via other sniffs, such as sniffs checking the spacing of function declaration parameters, function call parameters, array items.
* Rule 2 warrants a new sniff.
    For the function declaration part, PSR-12 demands the same - no space between the spread operator and the variable/function call -. For usage within array declarations and function calls, PSR-12 has no explicit opinion.
    With that in mind, I have pulled a comprehensive sniff to PHPCS itself which - once merged - could be used to cover this rule.
    See: https://github.com/squizlabs/PHP_CodeSniffer/pull/2548
* Rule 3: AFAICS there is no sniff available currently to cover this, but PSR12 demands the same, so an upstream sniff to cover this can be expected in the near future.

/cc @pento

dingo-d:
Should we put in this sniff the check, in the case of multiple arguments used with the spread operator, that the variadic argument is always placed on the last place?

```php
// Correct.
[...]

Issue #1524, opened 2018-11-06
PHPCS 3.4.0: Deprecate the WordPress.WhiteSpace.CastStructureSpacing sniff


jrfnl:
## Rationale

PHPCS 3.4.0 will introduce a new `Generic.Formatting.SpaceBeforeCast` sniff which checks that there is exactly one space before a type cast and includes fixers for no space and too much space.

This sniff can replace the last remaining check in the `WordPress.WhiteSpace.CastStructureSpacing` sniff,

## References
* Upstream (merged) PR: https://github.com/squizlabs/PHP_CodeSniffer/pull/2217

## Action Checklist
Once PHPCS 3.4.0 has been released and WPCS is ready/willing to up the minimum required PHPCS version:
- [ ] Update the PHPCS version requirements in the `composer.json` file.
- [ ] Update references to the WPCS minimum PHPCS version requirements in the Readme and wherever else needed.
- [ ] Deprecate the `WordPress.WhiteSpace.CastStructureSpacing` sniff.
     - [ ] Make the sniff extend the upstream `Generic.Formatting.SpaceBeforeCast` sniff and turn it into an empty class.
    - [ ] Mark the sniff as deprecated.
- [ ] Once this is done, open a new issue as a reminder to remove the WP native sniff in the next major version of WPCS and to replace the ruleset reference in the `WordPress-Core` ruleset.

jrfnl:
The upstream sniff does not make an exception for the spread operator, while based on #1762 we would want that exception. PR #1767 adds that exception to the existing WP specific sniff.

As this is quite specific, I'm not sure a similar PR would be accepted upstream. For now, this issue is superseded by PR #1767.

PR #1767, opened 2019-07-20
CastStructureSpacing: allow for spread operator

jrfnl:
Allow for no whitespace before a cast when used in combination with a spread operator.

This pre-emptively prevents a fixer conflict between this sniff and the upstream sniff we intend to add to address #1762.

Includes updated docs.

Related #1762
Related #1524


jrfnl:
Anything I can do to move this forward ?

PR #1768, opened 2019-07-20
New DisallowShortTernary sniff

jrfnl:
This new sniff addresses the new "_The short ternary operator must not be used._" rule which was recently added to the handbook.

The sniff has been added to the `WordPress-Core` ruleset.

Refs:
* https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#ternary-operator
* https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/

Includes unit tests.
Includes documentation.

Loosely related to #764

### DIFF
--- a/WordPress/Docs/PHP/DisallowShortTernaryStandard.xml
+++ b/WordPress/Docs/PHP/DisallowShortTernaryStandard.xml
@@ -1,0 +1,19 @@
+<documentation title="Disallow Short Ternaries">
+    <standard>
+    <![CDATA[
+    The short ternary operator must not be used.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: long ternary.">
+        <![CDATA[
+$height = ! empty( $data['height'] ) <em>?</em> $data['height'] <em>:</em> 0;
+        ]]>
+        </code>
+        <code title="Invalid: short ternary.">
+        <![CDATA[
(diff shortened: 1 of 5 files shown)


WordPress/Docs/PHP/DisallowShortTernaryStandard.xml
```diff
@@ -0,0 +1,19 @@
...
+    </standard>
+    <code_comparison>
+        <code title="Valid: long ternary.">
+        <![CDATA[
+$height = ! empty( $data['height'] ) <em>?</em> $data['height'] <em>:</em> 0;
+        ]]>
+        </code>
+        <code title="Invalid: short ternary.">
+        <![CDATA[
+$height = $data['height'] <em>? :</em> 0;
```
dingo-d: The space between `?` and `:` is intentional?

jrfnl: Yes. WPCS has no specific rules about the formatting of short ternaries and considering they will be banned, specifying such rules has low priority.

In the absence of specific rules, WPCS will treat the `?` and `:` the same way as other operators and demands whitespace on both sides.

dingo-d: Oh this makes sense, thanks!

PR #1769, opened 2019-07-20
OperatorSpacing: add insteadof keyword

jrfnl:
The PHP `insteadof` ~~operator~~ keyword is used to resolve conflicts when importing multiple traits (PHP 5.4+).

This adds this ~~operator~~ keyword to the list of operators checked by the `WordPress.WhiteSpace.OperatorSpacing` sniff.

Includes unit test.

Also see: https://www.php.net/manual/en/language.oop5.traits.php

Related #764

### DIFF
--- a/WordPress/Tests/WhiteSpace/OperatorSpacingUnitTest.inc.fixed
+++ b/WordPress/Tests/WhiteSpace/OperatorSpacingUnitTest.inc.fixed
@@ -68,3 +68,14 @@ if ( $a === $b or
 // Instanceof
 if ( MyClass instanceof SomeOtherClass ) {}
 if ( MyClass instanceof SomeOtherClass ) {}
+
+// Insteadof
+class Foo {
+    use A, B {
+        A::Talk insteadof B; // OK.
+        B::smallTalk insteadof A;
+        A::bigTalk
+			insteadof
+			B; // OK.
+    }
+}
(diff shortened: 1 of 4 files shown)


jrfnl:
Note: `insteadof` is not actually an operator, but rather a PHP keyword. All the same, for the purposes of spacing, it fits well in this sniff.

dingo-d:
Since you mentioned traits and `insteadof`, what about `as` keyword? Should we include it as well in the spacing sniff, and if yes, should we include it here as well?

Would it benefit from separating keyword and operator spacing sniffs? Both have special meaning in PHP, but I guess the semantics are different.

jrfnl:
> Since you mentioned traits and insteadof, what about as keyword? Should we include it as well in the spacing sniff, and if yes, should we include it here as well?
> 
> Would it benefit from separating keyword and operator spacing sniffs? Both have special meaning in PHP, but I guess the semantics are different.

I was just thinking the same. Haven't made up my mind yet, but if we'd add more keywords, I do think a separate sniff would be better.

I'd welcome to hear some more opinions on this.

jrfnl:
I'm closing this PR for the time being while having a think about this. Please do feel free to leave an opinion in a comment anyway.

WordPress/Tests/WhiteSpace/OperatorSpacingUnitTest.inc.fixed
```diff
@@ -68,3 +68,14 @@ if ( $a === $b or
 // Instanceof
 if ( MyClass instanceof SomeOtherClass ) {}
 if ( MyClass instanceof SomeOtherClass ) {}
+
+// Insteadof
```
GaryJones: Is the whitespace meant to be inconsistent like this?

Screenshot 2019-07-21 at 18 25 49

PR #1770, opened 2019-07-20
Core: Add upstream Generic.Arrays.DisallowShortArraySyntax sniff

jrfnl:
Recently a new section has been added to the handbook which forbids the use of short arrays.

> Using long array syntax ( array( 1, 2, 3 ) ) for declaring arrays is generally more readable than short array syntax ( [ 1, 2, 3 ] ), particularly for those with vision difficulties. Additionally, it’s much more descriptive for beginners.
>
> Arrays must be declared using long array syntax.

https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#declaring-arrays

This PR add an existing upstream sniff which addresses this.
Includes auto-fixer.

Also see: https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/

Loosely related to #764

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -101,6 +101,16 @@
 
 	<!--
 	#############################################################################
+	Handbook: PHP - Declaring Arrays.
+	Ref: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#declaring-arrays
+	#############################################################################
+	-->
+	<!-- Covers rule: Arrays must be declared using long array syntax. -->
+	<rule ref="Generic.Arrays.DisallowShortArraySyntax"/>
+
+
+	<!--
+	#############################################################################
 	Handbook: PHP - Use elseif, not else if.
 	Ref: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#use-elseif-not-else-if
 	#############################################################################


GaryJones:
I presume to override this, folks can do something like:

```xml
<rule ref="Generic.Arrays.DisallowShortArraySyntax">
	<severity>0</severity>
</rule>
```

...with an optional:

```xml
<rule ref="Generic.Arrays.DisallowLongArraySyntax"/>
```

jrfnl:
@GaryJones Correct. Alternatively for the first code snippet, you can use:
```xml
<rule ref="WordPresss">
    <exclude name="Generic.Arrays.DisallowShortArraySyntax"/>
</rule>
```
(within WPCS we use `severity` to make it clear to people how to re-enable an excluded rule, in a "end-user" custom ruleset you normally would use the above `exclude`).

GaryJones:
Since https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/ explicitly says these changes are only intended for WP core itself:

> While reading these changes, it’s important to keep in mind that they only apply to WordPress Core:

I'm wondering whether the addition could have been made to https://core.trac.wordpress.org/browser/trunk/phpcs.xml.dist instead of to the standards that all of the community would be using.

jrfnl:
@GaryJones I hear you, but I believe that discussion has been had numerous times already and the conclusion was always that the rulesets in WPCS follow the `Core` rules.

Issue #1291, opened 2018-01-25
No unneeded ternary check?


dingo-d:
When using ternary can we have a check that will see if the ternary is needed or not? [ESLint](https://eslint.org/docs/rules/no-unneeded-ternary) uses this rule. This would be thrown when we want to store boolean in a variable. For instance

```php
<?php

$var = 'a';
// Example of an unneeded ternary:
$bla = $var === 'a' ? true : false;

// Can be written as:
$bla = $var === 'a';
```

Don't know if we have this or not (since I'm not getting warnings I'm guessing we don't have it).

Also, I've stumbled into this: https://www.drupal.org/project/drupal/issues/1838368

Would this be useful to have as a rule?

jrfnl:
> Also, I've stumbled into this: https://www.drupal.org/project/drupal/issues/1838368

As stated in various comments in that thread, the memory leak issue which that ticket refers to was fixed in PHP 5.4, so IMO is quite irrelevant.

dingo-d:
We'll move this to the PHPCSExtra ruleset in the future: https://github.com/PHPCSStandards/PHPCSExtra/issues/157

PR #1771, opened 2019-07-21
New unneeded ternary sniff

dingo-d:
This sniff addresses #1291 issue.

When doing an assignment to a boolean with a ternary, if the ternary is only containing boolean values, this is an overhead, as a simple test will yield a boolean.

```php
<?php

$var = 'a';
// Example of an unneeded ternary:
$is_true = $var === 'a' ? true : false;

// Can be written as:
$is_true = $var === 'a';
```

Includes unit tests.
Includes documentation.

I have added some unit tests, but I'd like to see if some more extra cases need to be addressed.

### DIFF
--- a/WordPress/Sniffs/PHP/NoUnneededTernarySniff.php
+++ b/WordPress/Sniffs/PHP/NoUnneededTernarySniff.php
@@ -1,0 +1,77 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Sniffs\PHP;
+
+use WordPressCS\WordPress\Sniff;
+use PHP_CodeSniffer\Util\Tokens;
+
+/**
+ * Discourage the usage of the unneeded ternary operator.
+ *
+ * This sniff is influenced by the same rule set in the ESLint rules.
+ * When we want to store a boolean value in a variable, using a ternary operator
+ * is not needed, because a test will return a boolean value.
+ *
+ * @link https://eslint.org/docs/rules/no-unneeded-ternary
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   2.2.0
+ */
+class NoUnneededTernarySniff extends Sniff {
+
+	/**
+	 * Boolean tokens array.
+	 *
+	 * @since   2.2.0
+	 *
+	 * @var array
+	 */
+	private $boolean_tokens = array(
+		'T_TRUE'  => true,
+		'T_FALSE' => true,
+	);
+
+	/**
+	 * Returns an array of tokens this test wants to listen for.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @return array
+	 */
+	public function register() {
+		return array( \T_INLINE_THEN );
+	}
+
+	/**
+	 * Processes this test, when one of its tokens is encountered.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @param int $stackPtr The position of the current token in the stack.
+	 */
+	public function process_token( $stackPtr ) {
+		$first_token      = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $stackPtr + 1 ), null, true );
+		$end_of_statement = $this->phpcsFile->findNext( array( \T_SEMICOLON, \T_CLOSE_TAG ), $stackPtr );
+		$last_token       = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, ( $end_of_statement - 1 ), null, true );
+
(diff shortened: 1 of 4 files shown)


dingo-d:
Seems like travis is failing because of the addition of this sniff 😂 

https://github.com/WordPress/WordPress-Coding-Standards/blob/80e82a2b48fa02df57d2bd0a3ce24dc28de89278/bin/class-ruleset-test.php#L60-L62

WordPress/Sniffs/PHP/NoUnneededTernarySniff.php
```diff
@@ -0,0 +1,77 @@
...
+namespace WordPressCS\WordPress\Sniffs\PHP;
+
+use WordPressCS\WordPress\Sniff;
+use PHP_CodeSniffer\Util\Tokens;
+
+/**
+ * Discourage the usage of the unneeded ternary operator.
+ *
+ * This sniff is influenced by the same rule set in the ESLint rules.
+ * When we want to store a boolean value in a variable, using a ternary operator
```
GaryJones: Should this sniff also cover `return ... ? true : false;`?

WordPress/Sniffs/PHP/NoUnneededTernarySniff.php
```diff
@@ -0,0 +1,77 @@
...
+		$last_token       = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, ( $end_of_statement - 1 ), null, true );
+
+		$first_token_type = $this->tokens[ $first_token ]['type'];
+		$last_token_type  = $this->tokens[ $last_token ]['type'];
+
+		if ( isset( $this->boolean_tokens[ $first_token_type ] ) && isset( $this->boolean_tokens[ $last_token_type ] ) ) {
+			$this->phpcsFile->addError(
+				'Don\'t use ternary epression to store a boolean value in a variable.',
+				$stackPtr,
+				'Detected'
```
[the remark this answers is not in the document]
dingo-d: Wasn't aware of the `Found` vs `Detected`. I guess I saw the error message of some other sniff having `Detected` so I used it. Will fix this :+1:

WordPress/Sniffs/PHP/NoUnneededTernarySniff.php
dingo-d: I think this is the point of the sniff. Wherever there is an unneeded ternary detected, it can be removed. This includes conditionals, return statements or assignment to a variable 👍

WordPress/Sniffs/PHP/NoUnneededTernarySniff.php
```diff
@@ -0,0 +1,77 @@
...
+
+use WordPressCS\WordPress\Sniff;
+use PHP_CodeSniffer\Util\Tokens;
+
+/**
+ * Discourage the usage of the unneeded ternary operator.
+ *
+ * This sniff is influenced by the same rule set in the ESLint rules.
+ * When we want to store a boolean value in a variable, using a ternary operator
+ * is not needed, because a test will return a boolean value.
```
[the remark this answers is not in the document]
dingo-d: I'll rephrase this description so that it's not only limited to variable assignments (as is the initial intention of the sniff) 👍

WordPress/Sniffs/PHP/NoUnneededTernarySniff.php
GaryJones: And yet, there are no unit tests which shows that return statements are covered by this sniff.

WordPress/Sniffs/PHP/NoUnneededTernarySniff.php
```diff
@@ -0,0 +1,77 @@
...
+		$first_token      = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $stackPtr + 1 ), null, true );
+		$end_of_statement = $this->phpcsFile->findNext( array( \T_SEMICOLON, \T_CLOSE_TAG ), $stackPtr );
+		$last_token       = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, ( $end_of_statement - 1 ), null, true );
+
+		$first_token_type = $this->tokens[ $first_token ]['type'];
+		$last_token_type  = $this->tokens[ $last_token ]['type'];
+
+		if ( isset( $this->boolean_tokens[ $first_token_type ] ) && isset( $this->boolean_tokens[ $last_token_type ] ) ) {
+			$this->phpcsFile->addError(
+				'Don\'t use ternary epression to store a boolean value in a variable.',
```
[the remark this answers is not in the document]
dingo-d: The ESLint will throw `Unnecessary use of boolean literals in conditional expression`, can we use this instead of the proposed message, or make something better?

WordPress/Sniffs/PHP/NoUnneededTernarySniff.php
dingo-d: Will add those when I update the sniff :+1:

dingo-d:
I'll close this PR here as the sniff is not WordPress specific, so I'll open an issue in the PHPCSExtra to add this sniff there (and I'll probably start working on it there).

Issue #1315, opened 2018-03-01
Suggestion: new sniff to detect code inefficiencies


jrfnl:
What's the general opinion on having a sniff to detect the following type of code patterns and throw `warning`s when detected ? Useful or not ?

```php
echo sprintf( 'string %s', 'replacement'); // Should use `printf()` instead.
echo esc_html__( 'string', 'domain' ); // Should use `esc_html_e()` instead.
echo selected( $selected, $current, false ); // Should use `selected( $selected, $current )` instead.
echo get_search_form( false ); // Should use `get_search_form()` instead.
```
Of course, it would need to be verified that such a call is not followed by a concatenation adding more text, but that shouldn't be too hard to do.

```php
echo sprintf( 'string %s', 'replacement') . 'something else'; // This is fine.
```

For a list of functions which have a `$echo = true` toggle as one of the parameters, [Google Search](https://www.google.nl/search?dcr=0&ei=0ZyXWuDiB8nMwQL0ha7wDA&q=developer.wordpress.org%3A+%24echo+%3D+true&oq=developer.wordpress.org%3A+%24echo+%3D+true&gs_l=psy-ab.3...58702.58702.0.59346.1.1.0.0.0.0.57.57.1.1.0....0...1.1.64.psy-ab..0.0.0....0.mu7c3a0PLtw) will probably work better than the wp.org search.

Oh... and if you like the idea, more examples of code patterns to listen for would be welcome!

jrfnl:
@GaryJones I'd initially like to focus this ticket on PHP / WP Core constructs.

> I'd like the option to be able to extend the list of suggested replacements.

I will consider that for a later stage as that will not be easy of their are various variations of code patterns to consider and I'm not all that sure, we could allow for that when accepting input from the ruleset.

jrfnl:
Related #346

kkmuffme:
I think the use of (s)printf when no argument is used could be added.

e.g.
```
sprintf( __( 'hello', 'textdomain' ) );
should be:
__( 'hello', 'textdomain' );
```

also applies to printf or when esc_html__ (or esc_html_x,...) is used

jrfnl:
For PHP functions, I believe this should be handled by a (new) sniff to be added to PHPCSExtra. For WordPress specific situations, a new sniff in WPCS is needed.
The sniff added in #1777 is a good example of that type of sniff.

PR #1777, opened 2019-07-24
New CodeAnalysis/EscapedNotTranslated sniff

jrfnl:
At times, people will accidentally forget to add the `__` when they intend to use one of the "translate + escape" functions.
I've run into this a number of times now when reviewing/fixing code and found that they aren't that easy to spot visually when you're focussed on other things.

So as it was such an easy sniff to write, I figured I may as well.

AFAICS there are only two escaping functions in core which have direct "translate + escape" sister-functions.
All the same, the sniff has been set up to allow for more similar function-combis to be added.

Includes unit tests.
Includes documentation.

Sniff has been added to the `WordPress-Extra` ruleset.

### DIFF
--- a/WordPress-Extra/ruleset.xml
+++ b/WordPress-Extra/ruleset.xml
@@ -168,6 +168,11 @@
 	</rule>
 
 
+	<!-- Prevent some typical mistakes people make accidentally.
+	     https://github.com/WordPress/WordPress-Coding-Standards/pull/1777 -->
+	<rule ref="WordPress.CodeAnalysis.EscapedNotTranslated"/>
+
+
 	<!--
 	#############################################################################
 	Code style sniffs for more recent PHP features and syntaxes.
--- a/WordPress/Docs/CodeAnalysis/EscapedNotTranslatedStandard.xml
+++ b/WordPress/Docs/CodeAnalysis/EscapedNotTranslatedStandard.xml
@@ -1,0 +1,20 @@
+<documentation title="Escaped Not Translated Text">
+    <standard>
+    <![CDATA[
+    Text intended for translation needs to be wrapped in a localization function call.
+    This sniff will help you find instances where text is escaped for output, but no localization function is called, even though an (unexpected) text domain argument is passed to the escape function.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: esc_html__() used to translate and escape.">
+        <![CDATA[
+echo <em>esc_html__</em>( 'text', 'domain' );
+        ]]>
+        </code>
+        <code title="Invalid: esc_html() used to only escape a string intended to be translated as well.">
+        <![CDATA[
+echo <em>esc_html</em>( 'text', 'domain' );
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>
--- a/WordPress/Sniffs/CodeAnalysis/EscapedNotTranslatedSniff.php
+++ b/WordPress/Sniffs/CodeAnalysis/EscapedNotTranslatedSniff.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Sniffs\CodeAnalysis;
+
+use WordPressCS\WordPress\AbstractFunctionParameterSniff;
+use PHP_CodeSniffer\Util\Tokens;
+
+/**
+ * Flag calls to escaping functions which look like they may have been intended
+ * as calls to the "translate + escape" sister-function due to the presence of
+ * more than one parameter.
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   2.2.0
+ */
+class EscapedNotTranslatedSniff extends AbstractFunctionParameterSniff {
+
+	/**
+	 * The group name for this group of functions.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @var string
+	 */
+	protected $group_name = 'escapednottranslated';
+
+	/**
+	 * List of functions to examine.
+	 *
+	 * @link https://developer.wordpress.org/reference/functions/esc_html/
+	 * @link https://developer.wordpress.org/reference/functions/esc_html__/
+	 * @link https://developer.wordpress.org/reference/functions/esc_attr/
+	 * @link https://developer.wordpress.org/reference/functions/esc_attr__/
+	 *
+	 * @since 2.2.0
+	 *
+	 * @var array <string function_name> => <string alternative function>
+	 */
+	protected $target_functions = array(
+		'esc_html' => 'esc_html__',
+		'esc_attr' => 'esc_attr__',
+	);
+
+	/**
+	 * Process the parameters of a matched function.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 *
+	 * @return void
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		if ( \count( $parameters ) === 1 ) {
+			return;
+		}
+
+		/*
+		 * We already know that there will be a valid open+close parenthesis, otherwise the sniff
+		 * would have bowed out long before.
+		 */
+		$opener = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $stackPtr + 1 ), null, true );
+		$closer = $this->tokens[ $opener ]['parenthesis_closer'];
+
+		$data = array(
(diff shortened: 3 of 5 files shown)


WordPress/Docs/CodeAnalysis/EscapedNotTranslatedStandard.xml
```diff
@@ -0,0 +1,20 @@
+<documentation title="Escaped Not Translated Text">
+    <standard>
+    <![CDATA[
+    Text intended for translation needs to be wrapped in a localization function call.
+    This sniff will help you find instances where text is escaped for output, but no localization function is called.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: long ternary.">
```
GaryJones: These titles seem incorrect.

jrfnl: Thanks for catching that! Fixed now.

WordPress-Extra/ruleset.xml
```diff
@@ -168,6 +168,11 @@
 	</rule>
 
 
+	<!-- Prevent some typical mistakes people make accidentally.
+	     .. -->
```
GaryJones: This bit seems incomplete?

WordPress/Docs/CodeAnalysis/EscapedNotTranslatedStandard.xml
```diff
@@ -0,0 +1,20 @@
+<documentation title="Escaped Not Translated Text">
+    <standard>
+    <![CDATA[
+    Text intended for translation needs to be wrapped in a localization function call.
+    This sniff will help you find instances where text is escaped for output, but no localization function is called.
```
GaryJones: I'd like to see some sort of extra clarification in here re number of args - something like:

> This sniff will help you find instances where text is escaped for output, but not with a localization function, despite there appearing to be a text domain argument.

WordPress/Docs/CodeAnalysis/EscapedNotTranslatedStandard.xml
```diff
@@ -0,0 +1,20 @@
...
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: esc_html__() used to translate and escape.">
+        <![CDATA[
+echo <em>esc_html__</em>( 'text', 'domain' );
+        ]]>
+        </code>
+        <code title="Invalid: esc_html__() used to only escape a string intended to be translated as well.">
```
GaryJones:
```suggestion
        <code title="Invalid: esc_html() used to only escape a string intended to be translated as well.">
```

WordPress/Sniffs/CodeAnalysis/EscapedNotTranslatedSniff.php
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+namespace WordPressCS\WordPress\Sniffs\CodeAnalysis;
+
+use WordPressCS\WordPress\AbstractFunctionParameterSniff;
+use PHP_CodeSniffer\Util\Tokens;
+
+/**
+ * Flag calls to escaping functions which look like they may have been intended
+ * as calls to the "translate + escape" sister-function.
```
GaryJones:
```suggestion
 * as calls to the "translate + escape" sister-function due to the presence of a second parameter.
```

```suggestion
 * as calls to the "translate + escape" sister-function due to the presence of more than one parameter.
```

WordPress/Sniffs/CodeAnalysis/EscapedNotTranslatedSniff.php
```diff
@@ -0,0 +1,89 @@
...
+		$closer = $this->tokens[ $opener ]['parenthesis_closer'];
+
+		$data = array(
+			$matched_content,
+			$this->target_functions[ $matched_content ],
+			$this->phpcsFile->getTokensAsString( $stackPtr, ( $closer - $stackPtr + 1 ) ),
+		);
+
+		$this->phpcsFile->addWarning(
+			'%s() expects only one parameter. Did you mean to use %s__() ? Found: %s',
```
GaryJones:
```suggestion
			'%s() expects only one parameter. Did you mean to use %s() ? Found: %s',
```

I think the lookup of `$this->target_functions[ $matched_content ]` will already give a string value that includes the trailing underscores.

WordPress-Extra/ruleset.xml
jrfnl: Oh darn! Yes, I needed the PR number for that, but obviously didn't have it when I opened the PR.
And then I promptly forgot about it :joy:

WordPress/Docs/CodeAnalysis/EscapedNotTranslatedStandard.xml
jrfnl: I've adjusted the text along the lines of your suggestion. Better now ?

WordPress/Docs/CodeAnalysis/EscapedNotTranslatedStandard.xml
jrfnl: Good catch. Fixed now.

WordPress/Sniffs/CodeAnalysis/EscapedNotTranslatedSniff.php
jrfnl: Correct. I change this around from a semi-hard-coded suggestion to an array retrieval to allow for other additions to the sniff where the alternative function may not have a direct name relation to the escape function.

Note to self: don't _"let's quickly finish this as I need to go get some food"_ and pull :confounded:

jrfnl:
@GaryJones Thanks for being thorough!

GaryJones:
> Thanks for being thorough!

It's rare that I get a chance to find more than one valid thing wrong with your PRs. I suggest you take the rest of the night off as you're clearly not up to your high standards today 😛

WordPress/Docs/CodeAnalysis/EscapedNotTranslatedStandard.xml
```diff
@@ -0,0 +1,20 @@
...
+    <standard>
+    <![CDATA[
+    Text intended for translation needs to be wrapped in a localization function call.
+    This sniff will help you find instances where text is escaped for output, but no localization function is called, even though an (unexpected) text domain argument is passed to the escape function.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: esc_html__() used to translate and escape.">
+        <![CDATA[
+echo <em>esc_html__</em>( 'text', 'domain' );
```
dingo-d: While this is a correct example, should we maybe add another correct example?

`esc_html_e( 'text', 'domain' );` ?

As this is just the shorthand for `echo esc_html__()` function.

https://developer.wordpress.org/reference/functions/esc_html_e/

WordPress/Tests/CodeAnalysis/EscapedNotTranslatedUnitTest.inc
```diff
@@ -0,0 +1,8 @@
+<?php
+
+esc_html( 'text' );
+esc_html( $var );
+
+esc_html( 'text', 'domain' ); // Warning.
+esc_html( $foo, $bar ); // Warning.
```
dingo-d: This should be an error because one should never use a variable for text domain name substitution.

Should it be included in the unit test?

https://pascalbirchler.com/text-domain-wordpress-internationalization/
https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#text-domains

WordPress/Docs/CodeAnalysis/EscapedNotTranslatedStandard.xml
jrfnl: Sorry, no. That is not a function this sniff even looks at, so the example would be misleading.

WordPress/Tests/CodeAnalysis/EscapedNotTranslatedUnitTest.inc
jrfnl: No. That is covered by the `I18n` sniff.
A unit test case file should just and only test the sniff it is related to and can contain errors and problems which are forbidden by other sniffs to make sure the _sniff being tested_ handles this correctly.

The principle is that sniffs should function independently of each other. When one sniff is turned off, this shouldn't impact the functioning of any of the other sniffs.

dingo-d: Oh, I get it! Thanks for the explanation. This would fall under testing in isolation when doing unit tests, right?

WordPress/Docs/CodeAnalysis/EscapedNotTranslatedStandard.xml
dingo-d: Ok, I guess I looked at it from a reviewer perspective - I always say to my colleagues to use appropriate functions instead 😄

jrfnl: Oh, absolutely for reviewing "normal code".
There's actually a separate issue open to (at some point) create a sniff for that:#1315

WordPress/Tests/CodeAnalysis/EscapedNotTranslatedUnitTest.inc
jrfnl: Correct :+1:

PR #1782, opened 2019-07-28
Fixing return of ignore_annotations()

szepeviktor:
While doing static analysis, @phpstan discovered a _jungle of return type problems_ in `return` statements, missing `return`-s, incorrect `@return` in PHPDoc blocks.

phpstan.neon.dist

```yaml
# Add this to composer.json
#       "autoload-dev": {
#               "psr-4": {
#                       "WordPressCS\\WordPress\\": "WordPress/"
#               }
#       },

# Start: phpstan analyze

includes:
    - phar://phpstan.phar/conf/bleedingEdge.neon
parameters:
    #level: max
    level: 4
    inferPrivatePropertyTypeFromConstructor: true
    paths:
        - %currentWorkingDirectory%/WordPress/
    excludes_analyse:
        - %currentWorkingDirectory%/WordPress/Tests/
    autoload_files:
        - %currentWorkingDirectory%/vendor/squizlabs/php_codesniffer/autoload.php
    ignoreErrors:
#        - '#^Result of method WordPressCS\S+ \(void\) is used\.$#'
#        - '#^PHPDoc tag @var has invalid value \(array <.*> => <.*>\): Unexpected token#'
```

What do you think about it?

### DIFF
--- a/WordPress/PHPCSHelper.php
+++ b/WordPress/PHPCSHelper.php
@@ -98,12 +98,14 @@ class PHPCSHelper {
 	public static function ignore_annotations( File $phpcsFile = null ) {
 		if ( isset( $phpcsFile, $phpcsFile->config->annotations ) ) {
 			return ! $phpcsFile->config->annotations;
-		} else {
-			$annotations = Config::getConfigData( 'annotations' );
-			if ( isset( $annotations ) ) {
-				return ! $annotations;
-			}
 		}
+
+		$annotations = Config::getConfigData( 'annotations' );
+		if ( isset( $annotations ) ) {
+			return ! $annotations;
+		}
+
+		return false;
 	}
 
 }


jrfnl:
@szepeviktor Thank you for this PR. I appreciate you taking the time to check over the code with tooling like this.

While I agree the code you are pulling would be an improvement, that method is deprecated - or rather: only still used by the `Sniff::has_whitelist_comment()` method which is deprecated and both will be removed in the next major release.

To be honest, that whole file could be removed as it is no longer needed at the moment. We only left it in place to allow for accounting for other upstream changes in the future.

I'm happy enough to merge this, just though I'd let you know.

Aside from that, I've had a play with PHPStan in combination with the code in this repo and aside from some documentation copy/paste snafu's, I don't really see anything useful reported, just false positives.

The change to the `composer.json` would not be an option no matter what, there have been bug reports regarding adding an `autoload` directive in external PHPCS standard numerous times already. I don't remember the exact details, but I can look it up if you like.
Luckily, not a problem as it can be set in the PHPStan config anyway.

I ended up with the config below which ignores some issues regarding mostly array comments where PHPStan doesn't recognize the format used correctly, PHPCS native constants not being recognized and some things regarding non-explicit void `return`s.

What I'm left with after that are, AFAICS, false positives, including some to do with documentation issues in PHPCS upstream which can't be fixed in WPCS.

```neon
includes:
    - phar://phpstan.phar/conf/bleedingEdge.neon
parameters:
    #level: max
    level: 5
    inferPrivatePropertyTypeFromConstructor: true
    paths:
        - %currentWorkingDirectory%/WordPress/
    excludes_analyse:
        - %currentWorkingDirectory%/WordPress/Tests/
    autoload_files:
        - %currentWorkingDirectory%/vendor/squizlabs/php_codesniffer/autoload.php
    autoload_directories:
        - %currentWorkingDirectory%/WordPress/
    ignoreErrors:
        - '#^Result of method WordPressCS\S+ \(void\) is used\.$#'
        - '#^PHPDoc tag @var has invalid value \(array <.*> => <.*>\): Unexpected token#'
        - '#^Constant T_[A-Z_]+ not found\.$#'
        - '#^Method [a-zA-Z0-9\\_]+::[a-z_]+() should return int|void but return statement is missing\.$#'
        - '#^Property [a-zA-Z0-9\\_]+::\$[a-zA-Z0-9_]+ \(array<.*>\) does not accept default value of type array<.*>\.$#'
```

jrfnl:
PR #1784 should take care of the actual documentation issues within WPCS. The others being reported are IMO incorrect/false positives.

szepeviktor:
All right, I am happy that you took a look!!

PR #1783, opened 2019-07-28
PrefixAllGlobals/GlobalVariablesOverride: detect variables being set via list()

### LINKED ISSUE #1774: PrefixAllGlobals/GlobalVariablesOverride: detect variables being set via list()
## Bug Description
Variables being assigned a value via the `list()` construct - including PHP 7.1 short list syntax - are not recognized as variable assignments for the purposes of the `GlobalVariablesOverride` and the `PrefixAllGlobals` sniffs.

This causes false negatives.

## Minimal Code Snippet

The below code doesn't produce any errors for the above mentioned sniffs, while it should.

```php
// PREFIX ALL GLOBALS.
// Detect variables being set via the list() construct.
list( $var1, $var2 )                 = $array; // Bad x 2.
list( $acronym_var1, $acronym_var2 ) = $array;

[ $var1, $var2 ]                 = $array; // Bad x 2.
[ $acronym_var1, $acronym_var2 ] = $array;

/**
 * GLOBAL VARS OVERRIDE.
 * Detect var override via list construct.
 */
function acronym_prepare_items() {
	global $wp_query, $post_mime_types, $avail_post_mime_types, $mode;
	list( $post_mime_types, $avail_post_mime_types ) = get_an_array(); // Bad x2.
	[ $post_mime_types, $avail_post_mime_types ]     = get_an_array(); // PHP 7.1 short list syntax, bad x2.
}
```

jrfnl:
Global variables to which an assignment is made via the long/short `list()` construct should also be prefixed (`PrefixAllGlobals`), unless they are WordPress native global variables, in which case the `GlobalVariablesOverride` sniff should kick in and throw an error about overwriting WP globals.

Includes unit tests for both sniffs.

Fixes #1774 
Loosely related to #764 (short lists, keyed lists)

---

This PR includes adding two new utility methods to the `Sniff` class:
* `Sniff::find_list_open_close()` - Sister-method to the `find_array_open_close()` utility method to find the opener and closer for `list()` constructs, including short lists.
* `Sniff::get_list_variables()`  - A utility method which will retrieve an array with the token pointers to the variables which are being assigned to in a `list()` construct, whether short or long list.
    This utility method takes all currently supported list features in PHP into account and handles the following correctly:
    * Nested lists
    * Empty list items
    * Trailing comma's in lists
    * Empty lists (no longer allowed as of PHP 7.0.0)
    * Short lists (PHP 7.1.0+)
    * Keyed lists (PHP 7.1.0+)

### DIFF
--- a/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
+++ b/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
@@ -21,6 +21,7 @@ use PHP_CodeSniffer\Util\Tokens;
  * @since   0.12.0
  * @since   0.13.0 Class name changed: this class is now namespaced.
  * @since   1.2.0  Now also checks whether namespaces are prefixed.
+ * @since   2.2.0  Now also checks variables assigned via the list() construct.
  *
  * @uses    \WordPressCS\WordPress\Sniff::$custom_test_class_whitelist
  */
@@ -200,11 +201,13 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 
 		// Set the sniff targets.
 		$targets  = array(
-			\T_NAMESPACE => \T_NAMESPACE,
-			\T_FUNCTION  => \T_FUNCTION,
-			\T_CONST     => \T_CONST,
-			\T_VARIABLE  => \T_VARIABLE,
-			\T_DOLLAR    => \T_DOLLAR, // Variable variables.
+			\T_NAMESPACE        => \T_NAMESPACE,
+			\T_FUNCTION         => \T_FUNCTION,
+			\T_CONST            => \T_CONST,
+			\T_VARIABLE         => \T_VARIABLE,
+			\T_DOLLAR           => \T_DOLLAR, // Variable variables.
+			\T_LIST             => \T_LIST,
+			\T_OPEN_SHORT_ARRAY => \T_OPEN_SHORT_ARRAY,
 		);
 		$targets += Tokens::$ooScopeTokens; // T_ANON_CLASS is only used for skipping over test classes.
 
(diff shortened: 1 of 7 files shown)


WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -716,6 +728,35 @@ protected function process_variable_assignment( $stackPtr ) {
...
+	 * each list part can only contain one variable being written to.
+	 *
+	 * @since 2.2.0
+	 *
+	 * @param int $stackPtr The position of the current token in the stack.
+	 *
+	 * @return int|void Integer stack pointer to skip forward or void to continue
+	 *                  normal file processing.
+	 */
+	protected function process_list_assignment( $stackPtr ) {
```
dingo-d: Should this be added as a util method in the `Sniff.php` like `get_list_variables` is? Because it's used in two places.

jrfnl: Fair question. IMO no.

`process_...()` methods are not utility methods, but sniff processing specific. IMO they can either be in the sniff itself or in an abstract sniff - or possibly a sniff helper `trait`.

As these two sniffs are basically the polar opposite of each other, a case can be made to refactor them and either work with an abstract class, a trait or let one sniff extend the other.
(and yes, I've been thinking about that myself as well)

That was not what this PR is about though. This PR adds certain missing functionality. To keep things clean and reviewable, the refactor should - again IMO - be done in a separate PR.

dingo-d: Thanks for the explanation, this makes more sense now. And I agree, implementing either a trait or an abstract class seems like overkill for now.

:+1:

PR #1787, opened 2019-08-01
EscapeOutput: add highlight_string() to escaping functions

jrfnl:
While intended for code highlighting of PHP code, based on some tests I've run, the output of the PHP native `highlight_string()` function does appear to be safe, so I'm proposing to add this to the list of `$escapingFunctions`.

Note: I'd appreciate some scrutiny of this PR. I wouldn't want to inadvertently add an unsafe function to the list.

Refs:
* https://3v4l.org/mYK5A
* https://www.php.net/manual/en/function.highlight-string.php

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -149,6 +149,7 @@ abstract class Sniff implements PHPCS_Sniff {
 		'filter_input'         => true,
 		'filter_var'           => true,
 		'floatval'             => true,
+		'highlight_string'     => true,
 		'intval'               => true,
 		'json_encode'          => true,
 		'like_escape'          => true,


dingo-d:
I was worried that you could add some xss when changing settings during runtime but I wasn't successful at reproducing this (all my attempts failed):

```php
ini_set('highlight.string', '#ccc;font-size:1em" onload="<script>alert(\'XSS!\')</script>');

highlight_string('<?php echo \'Hi!\'; ?>');
```

This didn't trigger anything, so I guess it's safe?

I did found [this](https://github.com/ethicalhack3r/DVWA/blob/master/vulnerabilities/view_source_all.php), not sure what to make of it 🤷‍♂

jrfnl:
> I did found this, not sure what to make of it :man_shrugging: 

That looks to me like an educational page to display code which would be vulnerable and they use `highlight_string` to make the code look pretty.

PR #1794, opened 2019-09-11
[ON HOLD] Move `date` function group from `RestrictedPHPFunctions` to `TimezoneChange`

### LINKED ISSUE #1713: Discourage date() in favor of gmdate()

jrfnl:
Follow up on PR #1714 which fixed #1713.

As per the discussion in https://github.com/WordPress/WordPress-Coding-Standards/pull/1731#discussion_r301840793. moving the check for use of the `date()` to the `TimezoneChange` sniff.

As the `date` function group has not been in a released version (yet), we can still safely make this change without concerns about introducing breaking changes (error code change).

As that sniff is not part of the `Core` ruleset, I'm explicitly adding this particular error code to the `Core` ruleset.

Note: we may want to consider renaming this sniff as the name does not seem to be 100% correct anymore.

Suggestions for a better name welcome. If we go down that road, I'd deprecate the `TimezoneChange` sniff to be removed in the next major release.

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -519,4 +519,8 @@
 	<!-- Check for correct spelling of WordPress. -->
 	<rule ref="WordPress.WP.CapitalPDangit"/>
 
+	<!-- Use gmdate() not date().
+		 See: https://github.com/WordPress/WordPress-Coding-Standards/issues/1713 -->
+	<rule ref="WordPress.WP.TimezoneChange.date_date"/>
+
 </ruleset>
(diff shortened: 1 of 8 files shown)


jrfnl:
Closing in favour of PR #1807 which addresses the proposal as put forward in #1805.

PR #1800, opened 2019-09-20
DisallowShortTernary: improve docs

jrfnl:
The valid example line was a bit too long... oops...

### DIFF
--- a/WordPress/Docs/PHP/DisallowShortTernaryStandard.xml
+++ b/WordPress/Docs/PHP/DisallowShortTernaryStandard.xml
@@ -7,7 +7,8 @@
     <code_comparison>
         <code title="Valid: long ternary.">
         <![CDATA[
-$height = ! empty( $data['height'] ) <em>?</em> $data['height'] <em>:</em> 0;
+$height = ! empty( $data['height'] ) <em>?</em>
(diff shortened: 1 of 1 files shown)


jrfnl:
Before:
image

After:
image

jrfnl:
> We don't have a rule about where the break for the conditional should be, right?

Correct, no such rule currently exists for Core or otherwise.

If we _would_ create such a rule, I would advocate for it to be a separate sniff and to sniff for all multi-line statements with an operator at the end/start of a line.

My personal preference if such a rule would be suggested would be to always have the operator at the start of the next line (and yes, I know the above example does not comply and in a way, ternaries may be the exception anyway).

Reasoning behind that preference: for longer multi-line statements, this will often make for smaller and cleaner diffs and it improves readability as the operator tells you how the code directly after it will be handled, instead of having to "switch lines" with your eyes.

Some examples:
```php
// If the second or third line would be removed, only that line will show in the diff.
if ( $a === $b
    && $b < $c
    || $c !== $d
) { ... }

// Readability
$var = 10
    + 20
    . 'text';

// versus
$var = 10 +
    20 .
    'text';
```

Anyways, not something which should be handled in this PR either way.

If you want to propose such a sniff, I suggest opening an issue for it.

PR #1802, opened 2019-09-22
OneObjectStructurePerFile: move from `Extra` to `Core`

jrfnl:
Pulling this after @GaryJones brought this up as [a suggestion on Slack](https://wordpress.slack.com/archives/C5VCTJGH3/p1568550568000300).
WP Core currently has 27 violations against this rule - 3 in `src`, the rest in `tests`.

All the same, it would be a good thing to move this sniff to the `Core` ruleset. /cc @pento 

Open questions:
* The `Extra` configuration downgraded this to a `warning`. Should this be an `error` or a `warning` now it's moving to `Core` ?
* The `Extra` configuration also adjusted the error message. Is this still necessary or should we just use the default message from the sniff ?

Location | Message
--------- | -------
Original sniff error message | `Only one object structure is allowed in a file`.
Message as it was in `Extra` | `Best practice suggestion: Declare only one class/interface/trait in a file.`
Message proposed for `Core` | `Best practices: Declare only one class/interface/trait in a file.`

To do before this can be merged:
- [ ] Add a rule to this effect to the handbook.
- [ ] Adjust the `Core` ruleset to place the sniff in the right place with the text from the handbook.

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -517,6 +517,12 @@
 		<severity>0</severity>
 	</rule>
 
+	<!-- Encourage having only one class/interface/trait per file.
+		 Moved from Extra to Core after discussion on Slack. -->
+	<rule ref="Generic.Files.OneObjectStructurePerFile">
+		<message>Best practices: Declare only one class/interface/trait in a file.</message>
+	</rule>
+
 	<!--
 	#############################################################################
 	Not in the coding standard handbook: WP specific sniffs.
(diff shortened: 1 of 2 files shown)


GaryJones:
I'd say leave it as an error. I think the proposed message for Core is friendlier than the default upstream.

jrfnl:
> I'd say leave it as an error. I think the proposed message for Core is friendlier than the default upstream.

I've updated the PR to leave the message at an `error`.

This rule still needs to be added to the WP/Core CS Handbook. Any volunteers ? /cc @pento

dingo-d:
I'd merge this after the handbook is updated, is this ok?

jrfnl:
Rebased to have the GH Actions checks run.

dingo-d:
I've opened a PR for the handbook, so once that is merged we can merge this PR as well.

PR #1806, opened 2019-10-04
PrefixAllGlobals + ValidFunctionName: ignore deprecated methods by design

### LINKED ISSUE #1797: ValidFunctionName: ignore deprecated methods by design

Functions and methods should comply with the WP Coding Standards of being `snake_case`.

When a (`public`/`protected`) non-snake-case method/function is renamed, however, it breaks backward-compatibility, so it is common to deprecate the old method name and add a new method with a name which complies to the coding standards.

I'm proposing that if a function/method is marked as `deprecated` in the function docblock, that the sniff should not throw an error.

```php
class Example {
    public function new_name() {
    }

    /**
     * @deprecated x.x.x Use Example::new_name() instead.
     */
    public function oldName() {
        return self::new_name();
    }
}
```

Currently the sniff would throw an error for the `oldName()` method. Once this enhancement has been implemented, it no longer should.

jrfnl:
## Sniff: add new `is_function_deprecated()` utility function

This new function:
* Tries to find a function docblock, if it exists.
* If found, checks if the docblock contains at least one `@deprecated` tag.

Returns boolean true/false.

Note: this method is `static` to allow the `ValidFunctionName` sniff which extends an upstream sniff to use the method as well.

## ValidFunctionName: ignore deprecated functions

Check the function docblock for a `@deprecated` tag and if found, bow out.

Includes unit tests.

## PrefixAllGlobals: ignore deprecated functions

Check the function docblock for a `@deprecated` tag and if found, bow out.

Includes unit tests.

Note: the same should probably also be done for classes/interfaces/traits/constants marked as deprecated, but that's for another PR.

---

Fixes #1797

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -3385,4 +3385,38 @@ abstract class Sniff implements PHPCS_Sniff {
 		return $var_pointers;
 	}
 
+	/**
+	 * Check whether a function has been marked as deprecated via a @deprecated tag
+	 * in the function docblock.
+	 *
+	 * {@internal This method is static to allow the ValidFunctionName class to use it.}}
+	 *
+	 * @since 2.2.0
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+	 * @param int                         $stackPtr  The position of a T_FUNCTION
+	 *                                               token in the stack.
+	 *
+	 * @return bool
+	 */
+	public static function is_function_deprecated( File $phpcsFile, $stackPtr ) {
+		$tokens = $phpcsFile->getTokens();
+		$find   = Tokens::$methodPrefixes;
+		$find[] = \T_WHITESPACE;
+
+		$comment_end = $phpcsFile->findPrevious( $find, ( $stackPtr - 1 ), null, true );
+		if ( \T_DOC_COMMENT_CLOSE_TAG !== $tokens[ $comment_end ]['code'] ) {
+			// Function doesn't have a doc comment or is using the wrong type of comment.
+			return false;
+		}
+
+		$comment_start = $tokens[ $comment_end ]['comment_opener'];
+		foreach ( $tokens[ $comment_start ]['comment_tags'] as $tag ) {
+			if ( '@deprecated' === $tokens[ $tag ]['content'] ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
 }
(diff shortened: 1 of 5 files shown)


WordPress/Sniff.php
```diff
@@ -3385,4 +3385,38 @@ protected function get_list_variables( $stackPtr, $list_open_close = array() ) {
 		return $var_pointers;
 	}
 
+	/**
+	 * Check whether a function has been marked as deprecated via a @deprecated tag
+	 * in the function docblock.
+	 *
+	 * {@internal This method is static to allow the ValidFunctionName class to use it.}}
```
dingo-d:
```suggestion
	 * {@internal This method is static to allow the ValidFunctionName class to use it.}
```

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -21,7 +21,8 @@
  * @since   0.12.0
  * @since   0.13.0 Class name changed: this class is now namespaced.
  * @since   1.2.0  Now also checks whether namespaces are prefixed.
- * @since   2.2.0  Now also checks variables assigned via the list() construct.
+ * @since   2.2.0  - Now also checks variables assigned via the list() construct.
+ *                 - Will now ignore globals functions which are marked as @deprecated.
```
dingo-d: This sounds a bit odd. Maybe it would be better to say: `Ignores global functions marked as @deprecated`?

WordPress/Sniff.php
GaryJones: https://docs.phpdoc.org/references/phpdoc/tags/internal.html

> The inline version of this tag may, contrary to other inline tags, contain text but also other inline tags. To increase readability and ease parsing should the tag be terminated with a double closing brace, instead of a single one.

dingo-d: Oh wasn't aware of this. Thanks for the link with the explanation :+1:

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
jrfnl: To me, the difference between our phrasing is "what has changed" and "what it does", with the first being what is intended as it is a changelog, not part of the generic class description.

Happy to change it, but in that case, I'd like to see a suggestion which reads like a changelog still.

dingo-d: Yeah, I guess I had class description in mind vs changelog which this is. When I look at it that way the `Sniff will now ignore global functions which are marked as @deprecated` sounds better.

In that case, I think that just the `globals functions` should be changed with `global functions` no?

jrfnl: Thanks for catching that. I've fixed that and made a minor tweak to the text. Hope it's ok now.

PR #1808, opened 2019-10-05
New DateTime.CurrentTimeTimestamp sniff

### LINKED ISSUE #1791: Discourage use `current_time( 'timestamp' )`

`current_time()` with `timestamp` or `U` format produces a "WordPress timestamp", a sum of Unix timestamp and current time zone offset.

While historically entrenched in core, the use of such timestamps is deeply problematic and we are working on eliminating them in favor of real Unix timestamps and full interoperability with PHP.

- `current_time( 'timestamp', true )` / `current_time( 'U', true )`  are equivalent to `time()` and should be _replaced_ with it
- `current_time( 'timestamp' )` / `current_time( 'U' )` should be _discouraged_ in favor of requesting non-timestamp format, or otherwise refactoring the usage

The use of `current_time( 'timestamp' )` had been mostly eliminated from core in [#40657](https://core.trac.wordpress.org/ticket/40657). Minimal usage in unit tests and older functions that _must_ operate with WP timestamps for backwards compatibility reasons remains.

jrfnl:
This new sniff adds a check for use of current_time() to retrieve a timestamp.

A (fixable) `error` will be thrown when the `$gmt` parameter is set to `true` or `1`, a `warning` when it is not.

Includes unit tests.
Includes fixer.
Includes documentation.

This new sniff has been added to the `Core` ruleset.

Fixes #1791

/cc @Rarst

### DIFF
--- a/WordPress/Sniffs/DateTime/CurrentTimeTimestampSniff.php
+++ b/WordPress/Sniffs/DateTime/CurrentTimeTimestampSniff.php
@@ -1,0 +1,174 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Sniffs\DateTime;
+
+use WordPressCS\WordPress\AbstractFunctionParameterSniff;
+use PHP_CodeSniffer\Util\Tokens;
+
+/**
+ * Don't use current_time() to get a (timezone corrected) "timestamp".
+ *
+ * Disallow using the current_time() function to get "timestamps" as it
+ * doesn't produce a *real* timestamp, but a "WordPress timestamp", i.e.
+ * a Unix timestamp with current timezone offset, not a Unix timestamp ansich.
+ *
+ * @link https://developer.wordpress.org/reference/functions/current_time/
+ * @link https://make.wordpress.org/core/2019/09/23/date-time-improvements-wp-5-3/
+ * @link https://core.trac.wordpress.org/ticket/40657
+ * @link https://github.com/WordPress/WordPress-Coding-Standards/issues/1791
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   2.2.0
+ */
+class CurrentTimeTimestampSniff extends AbstractFunctionParameterSniff {
+
+	/**
+	 * The group name for this group of functions.
+	 *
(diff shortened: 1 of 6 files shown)


jrfnl:
@Rarst Looking at [your blogpost at Make](https://make.wordpress.org/core/2019/09/23/date-time-improvements-wp-5-3/), should this sniff also check for `get_post_time( 'U' )` ?

Rarst:
> should this sniff also check for get_post_time( 'U' )

Uh... Maybe later. We hadn't scraped _those_ out of core yet. Want to just let release go out and dust settle for a bit right now. :)

jrfnl:
> Uh... Maybe later. We hadn't scraped those out of core yet. 

That shouldn't really be a consideration.
If that code pattern shouldn't be used, we can already start throwing errors/warnings for it.

As long as we don't make the error/warning auto-fixable, the Core builds won't fail on it (and it will help detect the issue in Core).

Rarst:
> That shouldn't really be a consideration.
> If that code pattern shouldn't be used, we can already start throwing errors/warnings for it.

Well, it is for me. Getting these things out of core is a necessary legwork to figure out what depends on it and what needs to be done to dump it in all (or at least a lot of) cases.

My opinion is that we hadn't done equivalent amount of work to get rid of that one and are not ready to start sniffing it out. Soft recommendation to start using replacements takes a lower level of confidence than a hard sniff for me.

jrfnl:
> Soft recommendation to start using replacements takes a lower level of confidence than a hard sniff for me.

Ok, fair enough. We can always add it at a later point in time.

WordPress/Sniffs/DateTime/CurrentTimeTimestampSniff.php
```diff
@@ -0,0 +1,174 @@
...
+
+use WordPressCS\WordPress\AbstractFunctionParameterSniff;
+use PHP_CodeSniffer\Util\Tokens;
+
+/**
+ * Don't use current_time() to get a (timezone corrected) "timestamp".
+ *
+ * Disallow using the current_time() function to get "timestamps" as it
+ * doesn't produce a *real* timestamp, but a "WordPress timestamp", i.e.
+ * a Unix timestamp with current timezone offset, not a Unix timestamp ansich.
```
GaryJones: What is `ansich`?

jrfnl: https://en.wikipedia.org/wiki/Thing-in-itself ;-)

jrfnl:
Rebased for merge conflict.

jrfnl:
Anything I can do to move this PR forward ?

PR #1810, opened 2019-10-14
DeprecatedFunctions: update function list

jrfnl:
... with functions which will be deprecated in WP 5.3.

Not updating the `Sniff::$minimum_supported_version` property yet as WP 5.3 has not yet been released.


jrfnl:
Any particular reason why this PR has not been merged ? Happy to adjust if needed.

dingo-d:
I thought that the second person who approves it merges it. I've merged it now, no problems 🙂

Issue #867, opened 2017-03-07
Extra: Sniff to check for overruling of WP included scripts


jrfnl:
WP ships with a number of scrips by default. If a theme or plugin wants to use one of those scripts they should use the version included with WP and not ship their own version.

Ref: https://developer.wordpress.org/reference/functions/wp_register_script/#core-registered-scripts

While this will never be something for which we can get a 100% detection - as people can rename the script files and use arbitrary handle names -, should we try to detect if a theme/plugin tries to enqueue their own script instead of the WP version ?

I imagine we could sniff the function parameters of the `wp_enqueue_scripts` function and:
* throw an error if one of the reserved `handle`s is used in combination with a value for the `scr` parameter
* throw a warning if the filename in the `scr` parameter closely resembles the expected filename of one of the scripts WP ships with.

Additionally we could throw a warning if a `deps` parameter with one of the default values for a reserved `handle` is found as there is no need to register that dependency (already done by WP core).

Opinions ?

I imagine that the Theme Review theme and the VIP theme might be interested in including such a sniff as well ? /cc @grappler @sboisvert @david-binda

grappler:
In the WPTRT for we have a PR open that disallows deregistering core scripts that is used to overwrite the scripts. https://github.com/WPTRT/WordPress-Coding-Standards/pull/26

The most common case that I have seen this happening is with Masonry and and imagesloaded.

The problem that I see with a partial match is that have seen scripts that extend masonry have masonry in the handle and file name. I think we would have to live with a warning instead of an error unless the file matches `masonry.pkgd.js` or `masonry.pkgd.min.js`.

I don't fully understand what you mean here:

> Additionally we could throw a warning if a deps parameter with one of the default values for a reserved handle is found as there is no need to register that dependency (already done by WP core).

jrfnl:
@grappler I have seen too many plugins to count which ship with their own version of jQuery causing havoc for other plugins and themes. That's what this sniff would try to prevent.

The Theme sniff to prevent deregistering of core scripts could be used as an additional check, it is related to this, but does not address the same issue.

> The problem that I see with a partial match is that have seen scripts that extend masonry have masonry in the handle and file name. I think we would have to live with a warning instead of an error unless the file matches masonry.pkgd.js or masonry.pkgd.min.js.

Example code showing this in more detail would be very welcome so it can be used for the unit tests !

> I don't fully understand what you mean here:

[...]

PR #1824, opened 2019-10-31
Adds documentation for WordPress.WP.EnqueuedResources.

NielsdeBlaauw:
Related to #1722

### DIFF
--- a/WordPress/Docs/WP/EnqueuedResourcesStandard.xml
+++ b/WordPress/Docs/WP/EnqueuedResourcesStandard.xml
@@ -1,0 +1,53 @@
+<documentation title="Enqueued Resources">
+    <standard>
+    <![CDATA[
+    Scripts must be registered/enqueued via wp_enqueue_script().
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Script registered and enqueued correctly.">
+        <![CDATA[
+<em>wp_enqueue_script</em>( 
+    'someScript-js', 
+    $path_to_file, 
+    array( 'jquery' ),
+    '1.0.0',
+    true
+);
+        ]]>
+        </code>
+        <code title="Invalid: Script is directly embedded in HTML.">
+        <![CDATA[
+printf( 
+    '<em><script src="%s"></script></em>',
+    esc_url( $path_to_file )
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Stylesheets must be registered/enqueued via wp_enqueue_style().
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Stylesheet registered and enqueued correctly.">
+        <![CDATA[
+<em>wp_enqueue_style</em>( 
+    'style-name', 
+    $path_to_file, 
+    array(), 
+    '1.0.0'
+);
+        ]]>
+        </code>
+        <code title="Invalid: Stylesheet is directly embedded in HTML.">
+        <![CDATA[
+printf( 
+    '<em><link rel="stylesheet" href="%s" /></em>',
+    esc_url( $path_to_file )
+);
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>


WordPress/Docs/WP/EnqueuedResourcesStandard.xml
```diff
@@ -0,0 +1,50 @@
+<documentation title="Enqueued Resources">
+    <standard>
+    <![CDATA[
+    Scripts must be registered/enqueued via wp_enqueue_script.
```
jrfnl:
```suggestion
    Scripts must be registered/enqueued via wp_enqueue_script().
```

WordPress/Docs/WP/EnqueuedResourcesStandard.xml
```diff
@@ -0,0 +1,50 @@
...
+        ]]>
+        </code>
+        <code title="Invalid: Stylesheet is directly embedded in HTML.">
+        <![CDATA[
+<link rel="stylesheet" 
+      <em>href="http://someurl/somefile.css"</em> />
+        ]]>
+        </code>
+    </code_comparison>
+
```
jrfnl: Nitpick: (blank) line contains trailing whitespace.

WordPress/Docs/WP/EnqueuedResourcesStandard.xml
```diff
@@ -0,0 +1,50 @@
...
+        <code title="Invalid: Script is directly embedded in HTML.">
+        <![CDATA[
+<script <em>src="http://someurl/somefile.js"</em>>
+</script>
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Stylesheets must be registered/enqueued via wp_enqueue_style.
```
jrfnl:
```suggestion
    Stylesheets must be registered/enqueued via wp_enqueue_style().
```

WordPress/Docs/WP/EnqueuedResourcesStandard.xml
```diff
@@ -0,0 +1,50 @@
...
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Stylesheet registered and enqueued correctly.">
+        <![CDATA[
+<em>wp_enqueue_style</em>( 
+    'style-name', 
+    'https://example.com/somefile.css', 
+    false,
```
jrfnl: The third parameter of `wp_enqueue_style()` is expected to be an array.

Ref: https://developer.wordpress.org/reference/functions/wp_enqueue_style/

NielsdeBlaauw:
> Aside from that, I'm wondering if it would be better for the example code to reference local files instead of external URLs, as that is (or should be) the more common use-case.

I'm worried the code for the invalid example might be too far from what the user is actually doing, or what the sniff will be picking up. It's not about what resource is loaded, it is more about the way it is being done. IMHO this example is easier and more accessible to people with limited knowledge of WP. Also it avoids a discussion of where the actual resource is coming from.

jrfnl:
> > Aside from that, I'm wondering if it would be better for the example code to reference local files instead of external URLs, as that is (or should be) the more common use-case.
> 
> I'm worried the code for the invalid example might be too far from what the user is actually doing, or what the sniff will be picking up. It's not about what resource is loaded, it is more about the way it is being done. IMHO this example is easier and more accessible to people with limited knowledge of WP. Also it avoids a discussion of where the actual resource is coming from.

True enough. Maybe using a `$path_to_file` variable would simplify it in that case ?

The reason I'm not keen to leave the examples with the URLs is two-fold and revolves around other problems which this sniff is not sniffing for:
1. People using a CDN-version of script which WP natively ships with, often at a different version than the one which ships with WP.
    This breaks other plugins/themes as those expect the WP native version and their code may break when a different version is used. Also see #867 
2. People using just a CDN-version of a script without fall-back to local.
    This can break sites pretty badly when:
    - the CDN is down;
    - the CDN stops hosting a particular script or the particular version being pointed to;
    - the pointer is to a version-less script name and the version of the script on the CDN changes.

These are issues outside the scope of the sniff, so I agree that we shouldn't address them here, but I do believe showing code which we may well forbid at a later stage, is not a good idea either.

NielsdeBlaauw:
Agree, I going to replace it with a variable.

WordPress/Docs/WP/EnqueuedResourcesStandard.xml
```diff
@@ -0,0 +1,53 @@
...
+    '1.0.0',
+    true
+);
+        ]]>
+        </code>
+        <code title="Invalid: Script is directly embedded in HTML.">
+        <![CDATA[
+printf( 
+    '<em><script src="%s"></script></em>',
+    esc_attr( $path_to_file )
```
jrfnl:
```suggestion
    esc_url( $path_to_file )
```

WordPress/Docs/WP/EnqueuedResourcesStandard.xml
```diff
@@ -0,0 +1,53 @@
...
+    array(), 
+    '1.0.0'
+);
+        ]]>
+        </code>
+        <code title="Invalid: Stylesheet is directly embedded in HTML.">
+        <![CDATA[
+printf( 
+    '<em><link rel="stylesheet" href="%s" /></em>',
+    esc_attr( $path_to_file )
```
jrfnl:
```suggestion
    esc_url( $path_to_file )
```

PR #1825, opened 2019-10-31
Adds documentation for WordPress.WP.EnqueueResourceParameters.

NielsdeBlaauw:
Related to #1722

### DIFF
--- a/WordPress/Docs/WP/EnqueuedResourceParametersStandard.xml
+++ b/WordPress/Docs/WP/EnqueuedResourceParametersStandard.xml
@@ -56,7 +56,7 @@ wp_enqueue_script(
     </code_comparison>
     <standard>
     <![CDATA[
-    You must explicitly set a javascript resource to load in either the header or the footer of your page. It is recommended to load these resources in the footer by setting the `$in_footer` parameter to `true`.
+    You must explicitly set a JavaScript resource to load in either the header or the footer of your page. It is recommended to load these resources in the footer by setting the `$in_footer` parameter to `true`.
 
     Loading scripts in the header blocks parsing of the page and has a negative impact on load times. However, loading in the footer may break compatibility when other scripts rely on the resource to be available at any time.
     In that case, you can pass `false` to make it explicit that the script should be loaded in the header of the page.
--- a/WordPress/Docs/WP/EnqueuedResourceParametersStandard.xml
+++ b/WordPress/Docs/WP/EnqueuedResourceParametersStandard.xml
@@ -1,15 +1,15 @@
 <documentation title="Enqueued Resource Parameters">
     <standard>
     <![CDATA[
-    The resource version must be set, to prevent the browser from retrieving an outdated, cached version after the resource has been updated.
+    The resource version must be set, to prevent the browser from using an outdated, cached version after the resource has been updated.
     ]]>
     </standard>
     <code_comparison>
         <code title="Valid: Resource has a version number.">
         <![CDATA[
-wp_register_style( 
-    'someStyle-css', 
-    $path_to_local_file, 
+wp_register_style(
+    'someStyle-css',
+    $path_to_local_file,
     array(),
     <em>'1.0.0'</em>
 );
@@ -17,9 +17,9 @@ wp_register_style(
         </code>
         <code title="Invalid: Resource has no version number set.">
         <![CDATA[
-wp_register_style( 
-    'someStyle-css', 
-    $path_to_local_file, 
+wp_register_style(
+    'someStyle-css',
+    $path_to_local_file,
     array()
 );
         ]]>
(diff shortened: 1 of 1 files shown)


WordPress/Docs/WP/EnqueuedResourceParametersStandard.xml
```diff
@@ -0,0 +1,88 @@
+<documentation title="Enqueued Resource Parameters">
+    <standard>
+    <![CDATA[
+    The resource version must be set, to prevent the browser cache from retrieving and outdated version after updates.
```
jrfnl:
```suggestion
    The resource version must be set, to prevent the browser from retrieving an outdated, cached version after the resource has been updated.
```

WordPress/Docs/WP/EnqueuedResourceParametersStandard.xml
```diff
@@ -0,0 +1,88 @@
...
+);
+        ]]>
+        </code>
+        <code title="Invalid: Resource has no version number set.">
+        <![CDATA[
+wp_register_script( 
+    'someScript-js', 
+    'https://example.com/someScript.js' , 
+    array( 'jquery' ),
+    <em>null</em>,
```
jrfnl: I'm wondering if the example would be clearer if the last two (optional) parameters in this _invalid_ example were not passed at all ?

I realize that it would make it awkward, or rather impossible for the `<em>` tags to be used in this example, but so be it.

WordPress/Docs/WP/EnqueuedResourceParametersStandard.xml
```diff
@@ -0,0 +1,88 @@
...
+    array( 'jquery' ),
+    <em>false</em>,
+    true
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    You must explicitly set a javascript resource to load in the header or footer of your page. It is recommended to load these resources in the footer. Loading scripts in the header blocks parsing of the page and has a negative impact on load times. However, loading in the footer may break compatibility when other scripts rely on the resource to be available at any time.
```
jrfnl:
```suggestion
    You must explicitly set a javascript resource to load in the header or footer of your page.
    It is recommended to load these resources in the footer. Loading scripts in the header blocks parsing of the page and has a negative impact on load times. However, loading in the footer may break compatibility when other scripts rely on the resource to be available at any time. In that case, you can pass `false` to make it explicit that the script should be loaded in the header of the page.
```

(and yes, you can use line breaks in the `standard` description)

jrfnl:
He @NielsdeBlaauw Just checking in to see what the status is of this PR. Will you have a chance to finish this in the near future ?

NielsdeBlaauw:
Hey @jrfnl 

At the moment I don't have the spare time to work on this. I'm fine with someone else picking this up, or will be working on it again in january.

WordPress/Docs/WP/EnqueuedResourceParametersStandard.xml
NielsdeBlaauw: I've removed the null, indeed seems clearer this way, and the `<em>` is already in the valid example. Should be clear enough.

WordPress/Docs/WP/EnqueuedResourceParametersStandard.xml
```diff
@@ -0,0 +1,86 @@
+<documentation title="Enqueued Resource Parameters">
+    <standard>
+    <![CDATA[
+    The resource version must be set, to prevent the browser from retrieving an outdated, cached version after the resource has been updated.
```
jrfnl, line 4:
```suggestion
    The resource version must be set, to prevent the browser from using an outdated, cached version after the resource has been updated.
```

The whole point is that the browser _won't_ be retrieving the new version, but will use the locally cached version.

jrfnl (dismissed):
Hi @NielsdeBlaauw Thanks for the update and sorry it took so long for me to get to this.

I've had another look and just have some minor remarks:

* As the sniff looks at both the `register` and the `enqueue` variants of the functions, I feel it would be useful to show an example with those variants too.
* The description for the third part is very long, breaking it up into two paragraphs would improve the readability.
* Most example lines have trailing spaces.

As discussed via Slack DM and with your permission, I will update the PR to fix the above.

@GaryJones @dingo-d For whoever reviews next and merges this: please squash the commits on merge.

jrfnl (approved):

WordPress/Docs/WP/EnqueuedResourceParametersStandard.xml
```diff
@@ -0,0 +1,88 @@
...
+    array( 'jquery' ),
+    <em>false</em>,
+    true
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    You must explicitly set a javascript resource to load in either the header or the footer of your page. It is recommended to load these resources in the footer by setting the `$in_footer` parameter to `true`.
```
GaryJones, line 59:
```suggestion
    You must explicitly set a JavaScript resource to load in either the header or the footer of your page. It is recommended to load these resources in the footer by setting the `$in_footer` parameter to `true`.
```

GaryJones (changes requested):
Small typo fix needed.

dingo-d (approved):
I'll approve it so that, once the typo is fixed, Gary can merge it 🙂

GaryJones (approved):

PR #1827, opened 2019-10-31
Adds documentation for the WordPress.WP.CapitalPDangit sniff.

NielsdeBlaauw:
Note I used 'Discouraged:' in the second standard code example.
This is because it is not necesserily invalid, but I still want
to mention the opposite of the 'correct'/'valid' use.

Opinions on this?

Related to #1722

### DIFF
--- a/WordPress/Docs/WP/CapitalPDangitStandard.xml
+++ b/WordPress/Docs/WP/CapitalPDangitStandard.xml
@@ -1,0 +1,41 @@
+<documentation title="Capital P Dangit">
+    <standard>
+    <![CDATA[
+The correct spelling of `WordPress` should be used in text strings, comments and object names.
+
+When part of an identifier or a URL, WordPress does not have to be capitalized.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: WordPress is correctly capitalized.">
+        <![CDATA[
+class <em>WordPress</em>Example {
+
+        /**
+         * This function is about <em>WordPress</em>.
+         */
+        public function explain() {
+                $is_wordpress_awesome = true;
+                echo 'This is an explanation 
+                      about <em>WordPress</em>.';
+        }
+}
+        ]]>
+        </code>
+        <code title="Invalid: WordPress is not correctly capitalized.">
+        <![CDATA[
+class <em>WordpressExample</em> {
+
+        /**
+         * This function is about <em>Wordpress</em>.
+         */
+        public function explain() {
+                $is_wordpress_awesome = true;
+                echo 'This is an explanation 
+                      about <em>wordpress</em>.';
+        }
+}
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>
--- a/WordPress/Docs/WP/CapitalPDangitStandard.xml
+++ b/WordPress/Docs/WP/CapitalPDangitStandard.xml
@@ -1,39 +1,37 @@
 <documentation title="Capital P Dangit">
     <standard>
     <![CDATA[
-The correct spelling of `WordPress` should be used in text strings, comments and object names.
+The correct spelling of "WordPress" should be used in text strings, comments and object names.
 
-When part of an identifier or a URL, WordPress does not have to be capitalized.
+In select cases, when part of an identifier or a URL, WordPress does not have to be capitalized.
     ]]>
     </standard>
     <code_comparison>
         <code title="Valid: WordPress is correctly capitalized.">
         <![CDATA[
-class <em>WordPress</em>Example {
+class <em>WordPress</em>_Example {
 
-        /**
-         * This function is about <em>WordPress</em>.
-         */
-        public function explain() {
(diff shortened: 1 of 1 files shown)


WordPress/Docs/WP/CapitalPDangitStandard.xml
```diff
@@ -0,0 +1,36 @@
...
+        </code>
+        <code title="Invalid: WordPress is not correctly capitalized.">
+        <![CDATA[
+echo 'This is an explanation about wordpress.';
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    When part of and identifier or a URL, WordPress does not have to be capitalized.
```
jrfnl:
```suggestion
    When part of an identifier or a URL, WordPress does not have to be capitalized.
```

jrfnl:
> Note I used 'Discouraged:' in the second standard code example.
> This is because it is not necesserily invalid, but I still want
> to mention the opposite of the 'correct'/'valid' use.
> 
> Opinions on this?

I understand what you are trying to do, but that's basically the sniff just being "intelligent".

I think just adding the "standard" phrase to the first "standard" should suffice (without a code sample).

Adding some extra code samples to the first code comparison maybe useful though. Think: a class declaration with `WordPress` in the class name and a short docblock above it with `WordPress` in the description text.

The `echo` you currently have as code sample, could then be in a simple method in that class.

WordPress/Docs/WP/CapitalPDangitStandard.xml
```diff
@@ -0,0 +1,36 @@
+<documentation title="Capital P Dangit">
+    <standard>
+    <![CDATA[
+    Verify the correct spelling of `WordPress` in text strings, comments and class names.
```
jrfnl:
```suggestion
    The correct spelling of `WordPress` should be used in text strings, comments and object names.
```

jrfnl:
He @NielsdeBlaauw Just checking in to see what the status is of this PR. Will you have a chance to finish this in the near future ?

NielsdeBlaauw:
Hey @jrfnl 

At the moment I don't have the spare time to work on this. I'm fine with someone else picking this up, or will be working on it again in january.

jrfnl (dismissed):
Hi @NielsdeBlaauw Thanks for the update and sorry it took so long for me to get to this.

I've had another look and there are still some issues:
1. Other than for the issue the documentation is about, the code samples should comply with WPCS. They currently don't.
    1. The indentation is off. WPCS demands 1-tab indentation. As these XML files need spaces, that should be 4-space indentation for the code samples. They are using 8 space indentation.
    2. Precision alignment is used for the second part of the text string.
    3. Classnames should use _Camel_Caps_ with words separated by underscores, so `WordPress_Example`, not `WordPressExample`.
2. The class name without the underscore did not generate an error in WPCS, so the "incorrect" example contained an "incorrect" sample which wouldn't trigger an error. With an underscore it will trigger the error. 
3. The example with the variable name which does not use a capitalized form - `$is_wordpress_awesome` - can be seen as confusing. I understand it is an example of "identifier names" as mentioned in the standard description, but identifier names is broader than that and could also cover objectnames, in which we don't allow incorrect spelling.
    To prevent confusion, I'd remove the example. I'd be happy to revisit this and further improve the phrasing of the "identifier" part of the description too.

As discussed via Slack DM and with your permission, I will update the PR to fix the above.

@GaryJones @dingo-d For whoever reviews next and merges this: please squash the commits on merge.

jrfnl (approved):

GaryJones:
The description mentions "select cases" when WordPress doesn't need to be correctly capitalised, but then there is no example of them?

jrfnl:
> The description mentions "select cases" when WordPress doesn't need to be correctly capitalised, but then there is no example of them?

That's what the "when part of an identifier or a URL" is about. To list all the exceptions would become a very long and technical list.

GaryJones:
> To list all the exceptions would become a very long and technical list.

It doesn't need all exceptions - just an example or two, just the same as we don't list all text strings, comments and object names that it could possibly be.

jrfnl:
> just an example or two

@GaryJones Help me out, I'm not sure what you mean. To me `when part of an identifier or a URL` is an example or two.

GaryJones:
Valid: WordPress used as part of an identifier or a URL
```php
$wordpress_org_url = 'https://wordpress.org';
```

(i.e. a simple example to show that neither of these `wordpress` instances would complain about needing to be capitalized.)

jrfnl:
@GaryJones I see what you mean, but then there would be no "invalid" and those examples need both. Also, it would be showing what the sniff is *ignoring*, not what the sniff is looking for, so I'd consider that confusing. @NielsdeBlaauw actually had an example like that in the original and I removed it because of this.

GaryJones (approved):

dingo-d (approved):

PR #1852, opened 2020-01-17
Let sanitize_hex_color() count as an escape function

### LINKED ISSUE #1846: Let sanitize_hex_color() count as an escape function
Related: Comments in #780 

## Bug Description
When PHP outputs some CSS, you might see:

```php
color: <?php echo sanitize_hex_color($text_color); ?>
```

However, that gives a violation of:

> Error: All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks), found 'sanitize_hex_color' (WordPress.Security.EscapeOutput.OutputNotEscaped).

The [`sanitize_hex_color()` function](https://github.com/WordPress/WordPress/blob/f545bb3f634c8e949db9051bd88924df592c3337/wp-includes/formatting.php#L5905) returns either a valid 3 or 6-character hexadecimal colour code, or an empty string. In typical usage (putting the value into a CSS output) there is nothing to escape.

As such, I propose that `sanitize_hex_color()` is considered to already be escaped, and the violation is not shown. The alternative would be something like `esc_attr()`, which has a less strict allowance of values.

## Environment

| PHP version             | 7.3
| PHP_CodeSniffer version | 3.5
| WPCS version            | 2.2.0
| WPCS install type       | Composer

We might also consider `sanitize_hex_color_no_hash()` and possibly `maybe_hash_hex_color()` as well, since they are ultimately wrappers for `sanitize_hex_color()` as well.

paulschreiber:
Add sanitize_hex_color, sanitize_hex_color_no_hash and maybe_hash_hex_color to the list of escaping functions. Fixes #1846


paulschreiber:
@jrfnl updated the PR.

jrfnl:
@paulschreiber P.S.: for the next time - could you please create your PR and pull it against `develop` ?

I was scratching my head for five minutes before I noticed it was pulled against `master` which meant merging was not allowed.

Issue #1419, opened 2018-07-10
Sniff for HTML in translatable strings


jrfnl:
> Avoid HTML in strings if possible

Ref: https://make.wordpress.org/core/handbook/best-practices/internationalization/

HTML in translatable strings should be avoided as it puts an extra burden on translators in that they'd have to know HTML.

It also makes output escaping translatable strings much easier if there is no HTML in the string.

The HTML should be moved out of the translatable string and placeholders should be used instead.
```php
// Bad:
echo wp_kses( __( 'This is a <strong>bold</strong> text', 'text-domain' ) );

// Good:
printf (
    /* translators: 1: <strong> open tag 2: close tag. */
    esc_html__( 'This is a %1$sbold%2$s text', 'text-domain' ),
    '<strong>',
    '</strong>'
);
```

I think this should be quite easily sniffable and could be a good addition to the existing `WP.I18n` sniff.

I propose for this to be a `warning`.

I also propose *not* to include an auto-fixer. While for a simple example as the above, this could be auto-fixed, in general, it will need the critical look of a human being to determine the best course of action.

aristath:
Just a question out of curiosity

```php
// Good:
printf (
    /* translators: 1: <strong> open tag 2: close tag. */
    esc_html__( 'This is a %1$sbold%2$s text', 'text-domain' ),
    '<strong>',
    '</strong>'
);
```

Would this not be better like this since it's easier for translators?
```
// Good:
printf (
    /* translators: The "bold" string, wrapped in <strong> tags. */
    esc_html__( 'This is a %s text', 'text-domain' ),
    '<strong>' . esc_html__( 'bold', 'text-domain' ) . '</strong>',
);
```

Strings like `%1$sbold%2$s` are something that I've seen confuse lots of people and over the years I've spotted many wrong translations because of strings like that

jrfnl:
@aristath No. You should always avoid fragmenting a sentence as the sentence build up may be (very) different in different languages.

See: https://make.wordpress.org/core/handbook/best-practices/internationalization/#avoid-fragmented-or-patchwork-messages

aristath:
Ah I don't know how I missed that... Thanks! I learned something new today. 👍

GaryJones:
I'm -1 for this.

[...]

PR #1856, opened 2020-01-30
I18n: Add check to detect string wrapped in HTML

ockham:
Fixes the non-controversial part of #1419 by adding a sniff that detects translated strings wrapped in HTML tags, and a corresponding unit test. Quoting https://github.com/WordPress/WordPress-Coding-Standards/issues/1419#issuecomment-403797241:

> Examples where the markup _should_ be removed from inside of the string would be:
>
> ```
> <?php __( '<h1>Settings Page</h1>', 'text-domain ); ?>
> should be
> <h1><?php __( 'Settings Page', 'text-domain' ); ?></h1>
> because the markup only wraps the string.
> ```

Note that I had to add code to `I18nUnitTest.php` to reset the text domain as set via `setConfigData()` for one test file, as it would otherwise persist for every test file after that one.

Based on a WordPress.com-internal check. /cc @akirk

### DIFF
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -639,7 +639,7 @@ class I18nSniff extends AbstractFunctionRestrictionsSniff {
 		}
 
 		/*
-		 * NoHtmlWrappedStrings.
+		 * NoHtmlWrappedStrings and NoAHrefWrappedStrings.
 		 *
 		 * Strip surrounding quotes.
 		 */
@@ -659,7 +659,15 @@ class I18nSniff extends AbstractFunctionRestrictionsSniff {
 			}
 		}
 
-		// Is this all there is?
+		// Dedicated rule for cases where strings are wrapped in '<a href="...">...</a>'.
+		// The link target might actually need localization.
+		// Using a separate rule here allows codebases to disable it independently of the NoHtmlWrappedStrings rule.
+		if ( $reader->name === 'a' && $reader->getAttribute( 'href' ) ) {
+			$this->phpcsFile->addWarning( 'Strings should not be wrapped in <a href="...">...</a>', $stack_ptr, 'NoAHrefWrappedStrings' );
+			return;
+		}
+
+		// Does the entire string only consist of this HTML node?
 		if ( $reader->readOuterXml() === $content_without_surrounding_quotes ) {
 			$this->phpcsFile->addWarning( 'Strings should not be wrapped in HTML', $stack_ptr, 'NoHtmlWrappedStrings' );
 		}
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -639,7 +639,7 @@ class I18nSniff extends AbstractFunctionRestrictionsSniff {
 		}
 
 		/*
-		 * NoHtmlWrappedStrings and NoAHrefWrappedStrings.
+		 * NoHtmlWrappedStrings
 		 *
 		 * Strip surrounding quotes.
 		 */
@@ -659,11 +659,8 @@ class I18nSniff extends AbstractFunctionRestrictionsSniff {
 			}
 		}
 
-		// Dedicated rule for cases where strings are wrapped in '<a href="...">...</a>'.
-		// The link target might actually need localization.
-		// Using a separate rule here allows codebases to disable it independently of the NoHtmlWrappedStrings rule.
+		// We don't flag strings wrapped in `<a href="...">...</a>`, as the link target might actually need localization.
 		if ( 'a' === $reader->name && $reader->getAttribute( 'href' ) ) {
-			$this->phpcsFile->addWarning( 'Strings should not be wrapped in <a href="...">...</a>', $stack_ptr, 'NoAHrefWrappedStrings' );
 			return;
 		}
 
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -662,7 +662,7 @@ class I18nSniff extends AbstractFunctionRestrictionsSniff {
 		// Dedicated rule for cases where strings are wrapped in '<a href="...">...</a>'.
 		// The link target might actually need localization.
 		// Using a separate rule here allows codebases to disable it independently of the NoHtmlWrappedStrings rule.
-		if ( $reader->name === 'a' && $reader->getAttribute( 'href' ) ) {
+		if ( 'a' === $reader->name && $reader->getAttribute( 'href' ) ) {
 			$this->phpcsFile->addWarning( 'Strings should not be wrapped in <a href="...">...</a>', $stack_ptr, 'NoAHrefWrappedStrings' );
 			return;
 		}
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -630,11 +630,38 @@ class I18nSniff extends AbstractFunctionRestrictionsSniff {
 		 *
 		 * Strip placeholders and surrounding quotes.
 		 */
-		$non_placeholder_content = trim( $this->strip_quotes( $content ) );
-		$non_placeholder_content = preg_replace( self::SPRINTF_PLACEHOLDER_REGEX, '', $non_placeholder_content );
+		$content_without_surrounding_quotes = trim( $this->strip_quotes( $content ) );
+		$non_placeholder_content = preg_replace( self::SPRINTF_PLACEHOLDER_REGEX, '', $content_without_surrounding_quotes );
 
 		if ( '' === $non_placeholder_content ) {
 			$this->phpcsFile->addError( 'Strings should have translatable content', $stack_ptr, 'NoEmptyStrings' );
+			return;
+		}
+
+		/*
+		 * NoHtmlWrappedStrings.
+		 *
+		 * Strip surrounding quotes.
+		 */
+		$reader = new \XMLReader();
+		$reader->XML( $content_without_surrounding_quotes, 'UTF-8', LIBXML_NOERROR | LIBXML_ERR_NONE | LIBXML_NOWARNING );
+
+		// Is the first node an HTML element?
+		if ( ! $reader->read() || \XMLReader::ELEMENT !== $reader->nodeType ) {
+			return;
+		}
+
+		// If the opening HTML element includes placeholders in its attributes, we don't warn.
+		// E.g. '<option id="%1$s" value="%2$s">Translatable option name</option>'.
+		for ( $i = 0; $attr = $reader->getAttributeNo( $i ); $i++ ) {
+			if ( preg_match( self::SPRINTF_PLACEHOLDER_REGEX, $attr ) ) {
+				return;
+			}
+		}
+
+		// Is this all there is?
+		if ( $reader->readOuterXml() === $content_without_surrounding_quotes ) {
+			$this->phpcsFile->addWarning( 'Strings should not be wrapped in HTML', $stack_ptr, 'NoHtmlWrappedStrings' );
 		}
 	}
 
--- a/WordPress/Tests/WP/I18nUnitTest.1.inc
+++ b/WordPress/Tests/WP/I18nUnitTest.1.inc
@@ -206,9 +206,10 @@ __( '<<>>123</<>', 'my-slug' );
 
 // Strings wrapped in '<a href="...">...</a>'. These get a different warning, as the link target might actually be localized.
 // phpcs:disable WordPress.WP.I18n.NoHtmlWrappedStrings
-__( '<a href="http://wordpress.org">WordPress</a>', 'my-slug' ); // Bad, strings shouldn't be wrapped in <a href="...">...</a>
-__( '<a id="anchor">translatable text</a>', 'my-slug' ) // Good, since we have disabled the NoHtmlWrappedStrings rule here
-__( '<a>translatable text</a>', 'my-slug' ) // Good, since we have disabled the NoHtmlWrappedStrings rule here
+__( '<a href="http://wordpress.org">WordPress</a>', 'my-slug' ); // Bad, strings shouldn't be wrapped in <a href="...">...</a>.
+__( '<a id="anchor">translatable text</a>', 'my-slug' ) // Good, since we have disabled the NoHtmlWrappedStrings rule here.
+__( '<a>translatable text</a>', 'my-slug' ) // Good, since we have disabled the NoHtmlWrappedStrings rule here.
+__( '<a href="%s">translatable text</a>', 'my-slug' ) // Good, contains a placeholder in the href attribute.
 // phpcs:enable WordPress.WP.I18n.NoHtmlWrappedStrings
 
 // phpcs:set WordPress.WP.I18n text_domain[]
--- a/WordPress/Tests/WP/I18nUnitTest.1.inc
+++ b/WordPress/Tests/WP/I18nUnitTest.1.inc
@@ -196,6 +196,7 @@ __( '<div>text to translate', 'my-slug' );
 __( 'text to <div>translate', 'my-slug' );
 __( 'text < to translate', 'my-slug' );
 __( 'text to > translate', 'my-slug' );
+__( '<div data-type=">foo">my address</div>', 'my-slug' );
 
 // phpcs:set WordPress.WP.I18n text_domain[]
 // phpcs:set WordPress.WP.I18n check_translator_comments true
--- a/WordPress/Tests/WP/I18nUnitTest.1.inc
+++ b/WordPress/Tests/WP/I18nUnitTest.1.inc
@@ -195,9 +195,9 @@ __( '<foo>Foo</foo>', 'my-slug' ); // Good
 
 // Strings wrapped in `<a href="...">...</a>` aren't flagged, since the link target might require localization.
 __( '<a href="http://wordpress.org">WordPress</a>', 'my-slug' ); // Good
-__( '<a id="anchor">translatable text</a>', 'my-slug' ) // Bad, since no href.
-__( '<a>translatable text</a>', 'my-slug' ) // Bad, since no href.
-__( '<a href="%s">translatable text</a>', 'my-slug' ) // Good, contains a placeholder in the href attribute.
+__( '<a id="anchor">translatable text</a>', 'my-slug' ) // Bad, since no href
+__( '<a>translatable text</a>', 'my-slug' ) // Bad, since no href
+__( '<a href="%s">translatable text</a>', 'my-slug' ) // Good
 
 // Strings containing malformed XML, to make sure we're able to parse them. No warnings should be thrown.
 __( '<div>text to translate', 'my-slug' );
--- a/WordPress/Tests/WP/I18nUnitTest.1.inc
+++ b/WordPress/Tests/WP/I18nUnitTest.1.inc
@@ -191,5 +191,11 @@ __( '<span>Text</span> More text <span>Text</span>', 'my-slug' ); // Good, we're
 __( '<div class="my-class">Translatable content</div>', 'my-slug' ); // Bad
 __( '<option id="%1$s" value="%2$s">Translatable option name</option>', 'my-slug' ); // Wrapping is okay, since there are placeholders
 
+// Strings containing malformed XML, to make sure we're able to parse them. No warnings should be thrown.
+__( '<div>text to translate', 'my-slug' );
+__( 'text to <div>translate', 'my-slug' );
+__( 'text < to translate', 'my-slug' );
+__( 'text to > translate', 'my-slug' );
+
 // phpcs:set WordPress.WP.I18n text_domain[]
 // phpcs:set WordPress.WP.I18n check_translator_comments true
--- a/WordPress/Tests/WP/I18nUnitTest.1.inc
+++ b/WordPress/Tests/WP/I18nUnitTest.1.inc
@@ -204,5 +204,12 @@ __( '<div><div>', 'my-slug' );
 __( '<<>>text to translate<<>', 'my-slug' );
 __( '<<>>123</<>', 'my-slug' );
 
+// Strings wrapped in '<a href="...">...</a>'. These get a different warning, as the link target might actually be localized.
+// phpcs:disable WordPress.WP.I18n.NoHtmlWrappedStrings
+__( '<a href="http://wordpress.org">WordPress</a>', 'my-slug' ); // Bad, strings shouldn't be wrapped in <a href="...">...</a>
+__( '<a id="anchor">translatable text</a>', 'my-slug' ) // Good, since we have disabled the NoHtmlWrappedStrings rule here
+__( '<a>translatable text</a>', 'my-slug' ) // Good, since we have disabled the NoHtmlWrappedStrings rule here
+// phpcs:enable WordPress.WP.I18n.NoHtmlWrappedStrings
+
 // phpcs:set WordPress.WP.I18n text_domain[]
 // phpcs:set WordPress.WP.I18n check_translator_comments true
--- a/WordPress/Tests/WP/I18nUnitTest.1.inc
+++ b/WordPress/Tests/WP/I18nUnitTest.1.inc
@@ -209,13 +209,11 @@ __( '<b><i>Foo</b></i>', 'my-slug' );
 __( '<b><i>Foo</b></i></b>', 'my-slug' );
 __( '<foo>Foo</bar>', 'my-slug' );
 
-// Strings wrapped in '<a href="...">...</a>'. These get a different warning, as the link target might actually be localized.
-// phpcs:disable WordPress.WP.I18n.NoHtmlWrappedStrings
-__( '<a href="http://wordpress.org">WordPress</a>', 'my-slug' ); // Bad, strings shouldn't be wrapped in <a href="...">...</a>.
-__( '<a id="anchor">translatable text</a>', 'my-slug' ) // Good, since we have disabled the NoHtmlWrappedStrings rule here.
-__( '<a>translatable text</a>', 'my-slug' ) // Good, since we have disabled the NoHtmlWrappedStrings rule here.
+// Strings wrapped in `<a href="...">...</a>` aren't flagged, since the link target might require localization.
+__( '<a href="http://wordpress.org">WordPress</a>', 'my-slug' ); // Good
+__( '<a id="anchor">translatable text</a>', 'my-slug' ) // Bad, since no href.
+__( '<a>translatable text</a>', 'my-slug' ) // Bad, since no href.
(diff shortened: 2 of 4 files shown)


WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -80,6 +80,8 @@ class I18nSniff extends AbstractFunctionRestrictionsSniff {
 		[bcdeEfFgGosuxX]           # Type specifier.
 	)/x';
 
+	const WRAPPED_IN_HTML_REGEX = '#^(<[^>]+>).*(<[^>]+>)$#';
```
chriskmnds: Should this also include the forward slash `/` in the closing tag, so it doesn't match `<foo>123<foo>`? i.e. `^(<[^>]+>).*(<\/[^>]+>)$` 

Also, not sure if it matters, but this will also match empty content (e.g. `<foo><foo>`) or non-alphanumeric tags (e.g. `<<>>123<<>`). Maybe using `\w` instead of `[^>]` would handle these? 🤔

Sorry if I'm just adding noise, only tested briefly here: https://regex101.com/r/rIp4dV/3/tests

chriskmnds: A bit more noise: here's same tests with `\w` https://regex101.com/r/rIp4dV/4/tests :D

chriskmnds: Ah yeah definitely some noise. HTML has optional closing slashes :D

jrfnl: Constant is missing a docblock.

WordPress/Tests/WP/I18nUnitTest.4.inc
```diff
@@ -0,0 +1,10 @@
+<?php
```
jrfnl: There is no need for a new unit test case file. These tests could (and should) just have been added to the `I18nUnitTest.1.inc` file. Just add them to the end of the file just above the property reset.

WordPress/Tests/WP/I18nUnitTest.4.inc
```diff
@@ -0,0 +1,10 @@
...
+/*
+ * Test sniffing for strings wrapped in HTML.
+ */
+// phpcs:set WordPress.WP.I18n text_domain[] my-slug
+
+ __( '<address>123 Fake Street</address>', 'my-slug' ); // Bad, string shouldn't be wrapped in HTML.
+ __( 'I live at <address>123 Fake Street</address>', 'my-slug' ); // Okay, no consensus on HTML tags withing strings.
```
jrfnl: I would like to see some more unit tests. This feels very minimal for a new feature.

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -622,12 +624,30 @@ protected function check_text( $context ) {
...
+		 *
+		 * Strip surrounding quotes.
+		 */
+		if ( preg_match( self::WRAPPED_IN_HTML_REGEX, $content_without_surrounding_quotes, $tags ) ) {
+			// Starts with < and ends with >
+			$starting_tag = substr( $tags[1], 1, -1 );
+			// Starts with </ and ends with >
+			$ending_tag = substr( $tags[2], 2, -1 );
+
+			// Don't wrap in tags if the starting and ending tag are different
```
jrfnl:
```suggestion
			// Don't wrap in tags if the starting and ending tag are the same.
```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -622,12 +624,30 @@ protected function check_text( $context ) {
...
+			if ( $starting_tag === $ending_tag ) {
+				// Wrapped in HTML.
+				$this->phpcsFile->addError( 'Strings should not be wrapped in HTML', $stack_ptr, 'NoHtmlWrappedStrings' );
```
jrfnl: IMO a `warning` would be more appropriate.

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -622,12 +624,30 @@ protected function check_text( $context ) {
...
 		}
+
+		/*
+		 * NoHtmlWrappedStrings.
+		 *
+		 * Strip surrounding quotes.
+		 */
+		if ( preg_match( self::WRAPPED_IN_HTML_REGEX, $content_without_surrounding_quotes, $tags ) ) {
+			// Starts with < and ends with >
+			$starting_tag = substr( $tags[1], 1, -1 );
```
jrfnl: The regex can be improved to prevent the need for these two lines, as well as the `if` comparison. See my remarks at the regex.

WordPress/Sniffs/WP/I18nSniff.php
jrfnl: I agree with @chriskmnds that this regex needs improvement.

Some suggestions:
* Don't remember what doesn't need to be remembered.
* Match the (first part of the) content matched in the first tag with the content in the last tag.
* Keep in mind that HTML tags can have attributes. The regex in combination with the code below as is, would break straight away on this.
    ```php
    __( '<div class="my-class">Translatable content</div>', 'domain' );
    ```
* Keep in mind that there may be placeholders in the HTML attributes. If so, the HTML should not be removed and the warning not thrown.
    Without making the regex overly complicated, you could just check the full tag results of the tag at the start of the string with the same after running it through the `SPRINTF_PLACEHOLDER_REGEX`. If these are not the same, again, don't throw the message.
    ```php
    __( '<option id="%s" value="%s">Translatable option name</option>', 'domain' );
    ```

Let me know if you need help with this. I'm happy to provide you with a regex.

WordPress/Tests/WP/I18nUnitTest.4.inc
ockham: 84f5173

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -622,12 +624,30 @@ protected function check_text( $context ) {
 		 *
 		 * Strip placeholders and surrounding quotes.
 		 */
-		$non_placeholder_content = trim( $this->strip_quotes( $content ) );
-		$non_placeholder_content = preg_replace( self::SPRINTF_PLACEHOLDER_REGEX, '', $non_placeholder_content );
+		$content_without_surrounding_quotes = trim( $this->strip_quotes( $content ) );
+		$non_placeholder_content = preg_replace( self::SPRINTF_PLACEHOLDER_REGEX, '', $content_without_surrounding_quotes );
 
 		if ( '' === $non_placeholder_content ) {
 			$this->phpcsFile->addError( 'Strings should have translatable content', $stack_ptr, 'NoEmptyStrings' );
```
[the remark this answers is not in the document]
ockham: Thanks! 0bfe8cf1

WordPress/Sniffs/WP/I18nSniff.php
ockham: 101ae2bc

WordPress/Sniffs/WP/I18nSniff.php
ockham: Thanks, those are great points. I'll try tweaking the regex a bit.

WordPress/Sniffs/WP/I18nSniff.php
ockham: Obsolete, since the if clause was removed.

WordPress/Sniffs/WP/I18nSniff.php
ockham: Fixed in 43c80e3

WordPress/Sniffs/WP/I18nSniff.php
ockham: Fixed in 43c80e3.

WordPress/Sniffs/WP/I18nSniff.php
ockham: 43c80e3

ockham: As you've pointed out in Slack, it's really hard to parse HTML with a RegEx -- even if that HTML is supposedly simply, there can be nuance such as e.g. an attribute containing `>`, such as in `<address id="%s" data-type=">foo" value="%s">my address</address>`

See this famous Stack Overflow post: https://stackoverflow.com/a/1732454

The correct way would probably be to use one of [PHP's included HTML parsing facilities](https://stackoverflow.com/questions/3577641/how-do-you-parse-and-process-html-xml-in-php), but OTOH, that seems a bit overblown. 

Some of these cases are arguably rather pathological, so I'm going to stop tweaking the RegEx for now. I think it might be fine for starters, and it can be extended to include use cases that other people might run into over time.

ockham:
I think this might be ready for another look :blush:

WordPress/Sniffs/WP/I18nSniff.php
chriskmnds: Yep, agree 100%. 😃

jrfnl:
I was just talking about this PR with a friend of mine and he suggested (rightly) to also consider the following case for which the HTML tags should not be removed:
```php
__( '<span>Text</span> More text <span>Text</span>', 'domain' );
```

ockham:
Thanks, that's all valuable feedback.

However, I'm starting to think that using RegExes here is an uphill battle, as discussed [here](https://github.com/WordPress/WordPress-Coding-Standards/pull/1856#discussion_r373473501).

I'll give one of PHP's integrated HTML parsers a shot -- [XMLReader](https://www.php.net/manual/en/book.xmlreader.php) seems to have the smallest memory, resource-wise.

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -80,6 +80,29 @@ class I18nSniff extends AbstractFunctionRestrictionsSniff {
 		[bcdeEfFgGosuxX]           # Type specifier.
 	)/x';
 
+	/**
+	 * Match translatable strings wrapped in HTML, e.g. '<address>123 Fake Street</address>'
+	 */
+	const WRAPPED_IN_HTML_REGEX = '/(?:
```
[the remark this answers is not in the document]
ockham: Thanks! This is quite insightful, and I'll keep it handy as a reference for the future :slightly_smiling_face: 

For now, I've decided to use an actual HTML Parser rather than RegExps, see https://github.com/WordPress/WordPress-Coding-Standards/pull/1856#issuecomment-581280385.

WordPress/Tests/WP/I18nUnitTest.1.inc
```diff
@@ -177,5 +177,10 @@ $offset_or_tz = _x( '0', 'default GMT offset or timezone string', 'my-slug' );
 $test = __( '%1$s %2$s', 'my-slug' ); // OK(ish), placeholder order may change depending on language.
 $test = __( '   %s    ', 'my-slug' ); // Bad, no translatable content.
 
+__( '<address>123 Fake Street</address>', 'my-slug' ); // Bad, string shouldn't be wrapped in HTML.
+__( 'I live at <address>123 Fake Street</address>', 'my-slug' ); // Okay, no consensus on HTML tags withing strings.
+__( '<div class="my-class">Translatable content</div>', 'my-slug' ); // Bad
+__( '<option id="%1$s" value="%2$s">Translatable option name</option>', 'my-slug' ); // Wrapping is okay, since there are placeholders
+
```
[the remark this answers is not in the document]
ockham: I added [your example](https://github.com/WordPress/WordPress-Coding-Standards/pull/1856#issuecomment-581033790) and another variation of an existing one, but seem to struggle to come up with meaningful ones overall. Could you give me some pointers as to what you're thinking of?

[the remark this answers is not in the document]
jrfnl: Well, try this a as a good starting point for finding more tests: try running the sniff over WP core and see which false positives it throws up for this new error code. Those cases will all need to be accounted for and the patterns found which caused the false positives should be covered by unit tests.

Also see the link in the original issue to Core trac where there are also quite some examples.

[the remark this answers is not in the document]
ockham: > Well, try this as a good starting point for finding more tests: try running the sniff over WP core and see which false positives it throws up for this new error code. Those cases will all need to be accounted for and the patterns found which caused the false positives should be covered by unit tests.

Ah yeah, makes sense, thanks. `wp-includes` doesn't yield any occurrences, `wp-content` has the following:

https://github.com/WordPress/WordPress/blob/c289bb59acdd3217cdb81e15b580dc3d7cd6a007/wp-content/themes/twentythirteen/image.php#L30

https://github.com/WordPress/WordPress/blob/c289bb59acdd3217cdb81e15b580dc3d7cd6a007/wp-content/themes/twentyeleven/inc/theme-options.php#L121-L122

Those seem legit, but I think we've covered that class of issue with our current unit tests, no?

`wp-admin` has quite a few, e.g.

https://github.com/WordPress/WordPress/blob/c289bb59acdd3217cdb81e15b580dc3d7cd6a007/wp-admin/comment.php#L56-L57

https://github.com/WordPress/WordPress/blob/c289bb59acdd3217cdb81e15b580dc3d7cd6a007/wp-admin/edit-comments.php#L194-L197

https://github.com/WordPress/WordPress/blob/c289bb59acdd3217cdb81e15b580dc3d7cd6a007/wp-admin/edit-form-advanced.php#L313-L314

and so on.

A lot of these are `<a />` links where only the link text needs translating, and the actual `<a />` could be moved out of the `__()` call.

> Also see the link in the original issue to Core trac where there are also quite some examples.

Uh, not sure which issue you mean :sweat_smile: Maybe [this list](https://core.trac.wordpress.org/search?q=%22Avoid+using+HTML%22&noquickjump=1&changeset=on&ticket=on) (which https://github.com/WordPress/WordPress-Coding-Standards/issues/1419#issuecomment-403957076 refers to)? I checked a few of those.

```
<span>Previewing:</span> %s
Plugin <strong>resumed</strong>.
<span class="screen-reader-text">link to view ratings opens in a new tab</span>
```

These also seem pretty much covered I'd say :thinking:

ockham:
Rewrote the sniff using `XMLReader` (which seems much more precise than the RegExps could ever be for this purpose), added two more test cases (including [this](https://github.com/WordPress/WordPress-Coding-Standards/pull/1856#issuecomment-581033790)), and squashed the commits.

Also ran the sniff on the WordPress codebase (results are [here](https://github.com/WordPress/WordPress-Coding-Standards/pull/1856#discussion_r374155781)), which seems to indicate no false positives so far, and that our current unit tests are at least representative of the kind of code encountered in a real-world codebase.

Think this is in good enough shape to be considered for merging, and where potential refinements could be addressed in follow-up PRs? :sweat_smile:

WordPress/Tests/WP/I18nUnitTest.1.inc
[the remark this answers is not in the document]
jrfnl: > A lot of these are <a /> links where only the link text needs translating, and the actual <a /> could be moved out of the __() call.

Sorry, but that's not true. Those are legit translation calls including the tag as there may be a localized page available for that URL and part of translating is changing the URL to the localized version.

I think it would be a good idea to allow for those, i.e. if the tag is an `a` AND has a `href` with a hard-coded URL, there should be no warning.

jrfnl:
> A lot of these are <a /> links where only the link text needs translating, and the actual <a /> could be moved out of the __() call.

I'll have another look tomorrow but this PR will definitely not go into WPCS 2.2.1, which is due to be released tomorrow. I'd want this feature to be in place for beta testing for at least a little while so we can see if any more issues are reported.

WordPress/Tests/WP/I18nUnitTest.1.inc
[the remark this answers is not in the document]
ockham: > > A lot of these are  links where only the link text needs translating, and the actual  could be moved out of the __() call.
> 
> Sorry, but that's not true. Those are legit translation calls including the tag as there may be a localized page available for that URL and part of translating is changing the URL to the localized version.

Right, I didn't think of that.

> I think it would be a good idea to allow for those, i.e. if the tag is an `a` AND has a `href` with a hard-coded URL, there should be no warning.

I'm not totally sure it should be generalized that much though? This seems like a fairly 'semantic' addition to the sniff that implies quite some background information (there can be localized link targets), which might not be true for all codebases alike. One could argue that the localized link targets are more the exception than the rule, that the sniff doesn't necessarily need to know about them, and that it'd be fine to require a sniff ignore to be added by the author in case it's needed :thinking:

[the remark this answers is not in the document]
jrfnl: If the warning would be kept for link tags with hard-coded URLs, I'd like to see a separate error code for those.
That would at least allow codebases - including WP Core - which do have localized URLs and therefore do this with good reason to exclude the warning for those particular text strings from their custom ruleset.

ockham:
:wave: @jrfnl Would you mind giving this another look? :slightly_smiling_face:

jrfnl:
@ockham Has my remark in https://github.com/WordPress/WordPress-Coding-Standards/pull/1856#discussion_r374186666 been addressed ?

jrfnl:
Oh and please also include some tests with malformed/incomplete HTML and verify that the code handles this gracefully. I seem to remember that XMLreader does not handle that well.

Some examples of the top of my head (please try and find some more/more specific ones for known issues):
```php
__( '<div>text to translate', 'domain' );
__( 'text to <div>translate', 'domain' );
__( 'text < to translate', 'domain' );
__( 'text to > translate', 'domain' );
```

WordPress/Tests/WP/I18nUnitTest.1.inc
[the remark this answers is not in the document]
ockham: Added per 1294413a2c59e5c170a0e2666c29bd45415fc2a5

ockham:
I added a few commits to add your malformed HTML examples, plus some that Christos had in his regex tests. I also added a rule for `<a href="...">...</a>`. Please give it another look!

jrfnl:
I'd like to get some more eyes on this. @dingo-d @GaryJones Would you please have a look ?

dingo-d:
Looks good just by looking at it, I have started to run the i18n sniff on themes repo that I have locally (this will take a while), just to see if anything interesting pops up, since themes can get _creative_ when it comes to what they add in the translation strings 😄

dingo-d:
I ran this over my local themes repo and from what I see it caught everything correctly

https://gist.github.com/dingo-d/aecd13368ec1a91e1bd40772d0d94a21

I ran over the themes using 
```bash
vendor/bin/phpcs --sniffs=WordPress.WP.i18n ../WordPress-Themes/themes/ --report-file=../phpcs-result.txt --error-severity=0 --extensions=php --report=code
```

This caught both `NoHtmlWrappedStrings` and `MismatchedPlaceholders` warnings, so there is a lot of information in the results.

WordPress/Tests/WP/I18nUnitTest.1.inc
```diff
@@ -183,5 +183,34 @@ _n_noop($singular); // Bad x 3.
...
+__( '<div>text to translate<div>', 'my-slug' );
+__( '<>text to translate</>', 'my-slug' );
+__( '<>text to translate<>', 'my-slug' );
+__( '<div><div>', 'my-slug' );
+__( '<<>>text to translate<<>', 'my-slug' );
+__( '<<>>123</<>', 'my-slug' );
+
+// Strings wrapped in '<a href="...">...</a>'. These get a different warning, as the link target might actually be localized.
+// phpcs:disable WordPress.WP.I18n.NoHtmlWrappedStrings
+__( '<a href="http://wordpress.org">WordPress</a>', 'my-slug' ); // Bad, strings shouldn't be wrapped in <a href="...">...</a>.
```
GaryJones: This is the one that would concern me. By adding a warning, we're effectively recommending that URLs shouldn't be localized, especially when the comments say that this is "bad".

I'd rather see this one as good, and the placeholder test below possibly as bad.

WordPress/Tests/WP/I18nUnitTest.1.inc
```diff
@@ -183,5 +183,34 @@ _n_noop($singular); // Bad x 3.
...
+__( 'text < to translate', 'my-slug' );
+__( 'text to > translate', 'my-slug' );
+__( '<div data-type=">foo">my address</div>', 'my-slug' );
+__( '<div>text to translate<div>', 'my-slug' );
+__( '<>text to translate</>', 'my-slug' );
+__( '<>text to translate<>', 'my-slug' );
+__( '<div><div>', 'my-slug' );
+__( '<<>>text to translate<<>', 'my-slug' );
+__( '<<>>123</<>', 'my-slug' );
+
```
GaryJones: More tests please:

```php
__( '<b><i>Foo</i></b>', 'my-slug' );
__( '<b><i>Foo</b></i>', 'my-slug' );
__( '<b><i>Foo</b></i></b>', 'my-slug' );
__( '<foo>Foo</foo>', 'my-slug' );
__( '<foo>Foo</bar>', 'my-slug' );
```

WordPress/Tests/WP/I18nUnitTest.1.inc
ockham: See [this comment](https://github.com/WordPress/WordPress-Coding-Standards/pull/1856#discussion_r374180209), and [the one below](https://github.com/WordPress/WordPress-Coding-Standards/pull/1856#discussion_r374186666).

I ended up adding a separate warning for this case (`NoAHrefWrappedStrings` vs `NoHtmlWrappedStrings`) so that users would be able to separately disable the `<a href=".">...</a>` rule (while retaining the more 'generic' one about wrapping strings in HTML tags), to effectively get the behavior that you're describing. Does that sounds okay?

WordPress/Tests/WP/I18nUnitTest.1.inc
ockham: Thanks! Added in c56f18a9.

WordPress/Tests/WP/I18nUnitTest.1.inc
GaryJones: I had read that, and Juliette has the same concerns that I do.

My concern, is that the person writing the code, may not _know_ of any localised link, but is being a good citizen by making the URL available for translation. Yet, the code in this PR is penalising them for that.

If instead of: 

```php
__( '<a href="http://wordpress.org">WordPress</a>', 'my-slug' );
```

It was:

```php
__( '<a href="http://wordpress.org">WordPress</a> Rocks!', 'my-slug' );
```

then this sniff would have no issue with the hard-coded URL.

I'm all for the rest, and I know this currently has a different violation code, but right now, I'd rather let any strings wrapped in `<a ...>...</a>` to not be flagged, regardless of attribute values or placeholders, and then maybe tighten it up in the future if needed.

akirk: > This is the one that would concern me. By adding a warning, we're effectively recommending that URLs shouldn't be localized, especially when the comments say that this is "bad".

This is incorrect. The recommended approach would be this: `<a href="<?php echo esc_url( __( 'https://wordpress.org/ ' ) ); ?>"><?php echo __( 'WordPress Rocks' ); ?></a>`, and not "not localizing" the URL.

In the approach you recommend (including the `<a>` tag incl the `href` value), translation systems like GlotPress (on translate.wordpress.org) will actually throw a warning if you want to submit a translation with a changed tag (and changing the URL in a tag is changing the tag), so I'd consider this discouraged as well.

This is why I think it's preferrable to wrap the URL separately in gettext and have the sniff advise so (i.e. change the message `Strings should not be wrapped in <a href="...">...</a>` to something like `Strings should not be wrapped in <a href="...">...</a>, please wrap the URL separately`).

GaryJones: > The recommended approach would be this: `<a href="<?php echo esc_url( __( 'https://wordpress.org/' ) ); ?>"><?php _e( 'WordPress Rocks' ); ?></a>`, and not "not localizing" the URL.

Your example doesn't match what I put. In mine, `WordPress` was linked, while `Rocks` was not. By following your approach, the string would be split.

https://developer.wordpress.org/apis/handbook/internationalization/internationalization-guidelines/#best-practices-for-writing-strings says "Do not leave URLs for translation, unless they could have a version in another language.".

In the case of my example, I'd want to switch it to `https://en-gb.wordpress.org`.

An examples where it wouldn't be switched, would be when it points to an internal settings page.

At this stage, I wouldn't want to get into analysing the URL itself to decide whether to flag it or not, and so, would still like to see this not flagged for this first pass. With more discussion, and consensus, maybe a variant of this check for a href could be added.

akirk: > Your example doesn't match what I put. In mine, WordPress was linked, while Rocks was not. By following your approach, the string would be split.

This was for the sake of an argument. The whole point of this PR is to encourage leaving out as much as HTML possbile from Gettext string. Your string of argument is distracting from this approach.

> Correct, along with warnings for other validation failures as well - that's just a way to highlight it to Translation Editors, rather than let potentially spam URLs to make it through to production translation sets. It's not to discourage valid localised URLs from being submitted.

We can disagree about the reasons behind the validation failure. In general your approach encourages translators to change HTML tags which in my opinion should not be their business and they should actually be protected. It's easy to introduce invalid HTML and just dismiss the warning. So actually the fact that you can override warnings in GlotPress is not helpful in this case.

ockham:
Removed the `NoAHrefWrappedStrings` rule; strings wrapped in `<a href="...">...</a>` now pass the check.

WordPress/Tests/WP/I18nUnitTest.1.inc
GaryJones: > The whole point of this PR is to encourage leaving out as much as HTML possbile from Gettext string.

Your definition of "as possible" and mine clearly differ. I'll refer to my [previous comment](https://github.com/WordPress/WordPress-Coding-Standards/issues/1419#issuecomment-403797241) on the issue as for why.

> In general your approach encourages translators to change HTML tags which in my opinion should not be their business and they should actually be protected.

_Your opinion_ disagrees with the `gettext` manual, WP best practices document, and one or two handbooks.

Besides that, the prior art here is that the translators using GlotPress have been handling (wrapping) markup in strings from developers who didn't know better for a number of years now. That, on its own, isn't necessarily reason enough to continue to allow it, but I can't see what major issue has happened that would warrant the overly strict set of checks that you want to see imposed in the first pass.

As I've said, I'm happy to continue this discussion for wrapping anchor links, but by moving it outside of this PR, we can unblock the majority of the benefit that the rest of it brings.

jrfnl:
Sorry I've been quiet over the last week. I'm in full conference prep mode and will be traveling for the next two weeks, so won't have a chance to have another look until after that time.

Though I haven't reviewed the latest version of the technical implementation, I do believe my concerns about the actual functionality have been addressed.

I'll leave it up to the other committers to decide whether they feel confident merging this as it is now or whether they want to wait until I'm back. They have my blessing.

PR #1857, opened 2020-02-01
Travis: various tweaks

jrfnl:
Travis: various tweaks

* Run against the highest stable PHP version for the sniff/rulesets build (+ the `dev-master` `quicktest` build).
* Only run `composer validate` on the `sniff` stage.
    The `composer.json` file will be validated on the install for the all builds anyway. This just adds the `strict` checking.
* Add an extra `test` build against PHPCS 4.x for which development has just started.
    This build is allowed to fail for now.
* Remove the DealerDirect plugin and PHPCompatibility for the build against `nightly` and the PHPCS 4.x build.
    Neither are needed anyway as a `--no-dev` install is done for the `test` stages and the `installed_paths` is set directly.
    The DealerDirect plugin currently won't install on PHP 8.x, so the build would fail on "failure to install" before doing anything.
    And both PHPCompatibility as well as the DealerDirect plugin won't install/will block `--no-dev` installs with PHPCS 4.x.
    This way, the linting and unit tests should still run on `nightly` and with PHPCS 4.x.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ jobs:
   include:
     #### SNIFF STAGE ####
     - stage: sniff
-      php: 7.3
+      php: 7.4
       env: PHPCS_BRANCH="dev-master"
       addons:
         apt:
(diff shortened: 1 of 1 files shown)


jrfnl:
The changes I made in this PR exposed a PHP 7.4 related bug. PR #1858 should fix that bug. I will rebase this PR after that one has been merged so we can see a passing build (or not in which case I will look at what else is failing :confused: )

jrfnl:
Rebased after the merge of #1858 and did some more tweaking and we now have a passing build (including running builds for PHP 8/nightly and PHPCS 4.x-dev, though the PHPCS 4.x-dev one currently doesn't pass, but that's unrelated to the Travis script) :tada:

PR #1858, opened 2020-02-01
I18n: bug fix - trying to access array offset on null (PHP 7.4)

jrfnl:
When a function call to one of the singular/plural text translation functions - like `_n_noop()` - is missing an argument, the sniff would still try an execute the `compare_single_and_plural_arguments()` check and on PHP 7.4 would generate a "_Trying to access array offset on value of type null_" error which would stop the PHPCS run dead with an `Internal.Exception` error code.

Fixed by skipping the check in that case. The missing argument should be reported by the preceding `check_argument_tokens()` check anyhow.

Includes unit test.

**Note: I've set this PR to priority high as this bug stops the PHPCS run dead on PHP 7.4, so this fix should be included in the upcoming WPCS 2.2.1 release.**

### DIFF
--- a/WordPress/Tests/WP/I18nUnitTest.1.inc
+++ b/WordPress/Tests/WP/I18nUnitTest.1.inc
@@ -177,5 +177,11 @@ $offset_or_tz = _x( '0', 'default GMT offset or timezone string', 'my-slug' );
 $test = __( '%1$s %2$s', 'my-slug' ); // OK(ish), placeholder order may change depending on language.
 $test = __( '   %s    ', 'my-slug' ); // Bad, no translatable content.
 
+// Missing plural argument.
+_n_noop($singular); // Bad x 3.
+
+// This test is needed to verify that the missing plural argument above does not cause an internal error, stopping the run.
+_n_noop( 'I have %d cat.', 'I have %d cats.' ); // Bad.
+
 // phpcs:set WordPress.WP.I18n text_domain[]
 // phpcs:set WordPress.WP.I18n check_translator_comments true
--- a/WordPress/Sniffs/WP/I18nSniff.php
+++ b/WordPress/Sniffs/WP/I18nSniff.php
@@ -390,8 +390,16 @@ class I18nSniff extends AbstractFunctionRestrictionsSniff {
 			$this->check_argument_tokens( $argument_assertion_context );
 		}
 
-		// For _n*() calls, compare the singular and plural strings.
-		if ( false !== strpos( $this->i18n_functions[ $matched_content ], 'number' ) ) {
+		/*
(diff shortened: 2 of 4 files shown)


dingo-d:
I knew there were some issues with 7.4 because my colleague got the internal error message, but I couldn't find any info where this could come from, and just told him to use 7.3, which did solve the issues.

I didn't want to open any issues, because I was unsure of the cause 😅 .

jrfnl:
> I knew there were some issues with 7.4 because my colleague got the internal error message, but I couldn't find any info where this could come from, and just told him to use 7.3, which did solve the issues.
> 
> I didn't want to open any issues, because I was unsure of the cause sweat_smile .

@dingo-d For the next time: please do report it so it's on our radar. Even if we don't know the cause (yet), if there is a code sample - even if it's a link to a large file - with which to reproduce the issue, someone would be able to track it down eventually.

And in case you want to track it down yourself, here are helpful hints and tips:
1. An internal error will stop the run on that file dead. So the problem line will be on a line of the last reported error for that file or after it.
    Not helpful if the file is otherwise clean, but for "dirty" files, it helps track down the line of code triggering the error.
2. The error will be reported and the report will include the name of the sniff causing the problem as well as the error itself, like this:
```
   1 | ERROR   | [ ] An error occurred during processing; checking has been aborted. The error
     |         |     message was: Trying to access array offset on value of type null in
     |         |     path/to/WPCS/WordPress/Sniffs/WPI18nSniff.php on line 534
     |         |     (Internal.Exception)
```

With those two things combined it's normally possible to figure out what going on.

dingo-d:
Yeah, something like this was shown, and I never saw it before so suspected the JIT issue on 7.4 that was reported in PHPCS earlier, wasn't aware that it was something else instead.

I'll ask the colleague to try to reproduce the bug again to check if it had to do with the i18 sniff (can't remember what it was exactly).

WordPress/Tests/WP/I18nUnitTest.1.inc
```diff
@@ -177,5 +177,11 @@ $offset_or_tz = _x( '0', 'default GMT offset or timezone string', 'my-slug' );
 $test = __( '%1$s %2$s', 'my-slug' ); // OK(ish), placeholder order may change depending on language.
 $test = __( '   %s    ', 'my-slug' ); // Bad, no translatable content.
 
+// Missing plural argument.
+_n_noop($singular); // Bad x 3.
+
+// This test is needed to verify that the missing plural argument above does not cause an internal error, stopping the run.
+_n_noop( 'I have %d cat.', 'I have %d cats.' ); // Bad.
```
GaryJones: What's bad about this line?

Issue #1616, opened 2019-01-06
PHPCS 3.4.1: Use the Generic.WhiteSpace.LanguageConstructSpacing sniff


jrfnl:
## Rationale

The `Squiz.WhiteSpace.LanguageConstructSpacing` sniff has been deprecated in PHPCS 3.3.0 and will be removed in PHPCS 4, in favour of the newly introduced `Generic.WhiteSpace.LanguageConstructSpacing` sniff.

WPCS should start using that (much more comprehensive) sniff as soon as possible, but the PEAR release of PHPCS 3.3.0 up to, at least, PHPCS 3.4.0 contains a bug causing this sniff not to be available on PEAR at this moment.

## References
* Introduction of the sniff in PHPCS: squizlabs/PHP_CodeSniffer#1337
* Original issue: #1328
* Original PR: #1585
* Bug report/PEAR: #1614
* Revert PR: #1615
* Upstream bugfix PR: squizlabs/PHP_CodeSniffer#2340 (merged in what will be released as PHPCS 3.4.1)

## Action Checklist
Once the upstream hotfix PR has been merged, a new PHPCS version has been released and WPCS is ready/willing to up the minimum required PHPCS version:
- [ ] Update the PHPCS version requirements in the `composer.json` file.
- [ ] Update references to the WPCS minimum PHPCS version requirements in the Readme and wherever else needed.
- [ ] Make the ruleset change to swop out the old sniff for the new.

PR #1862, opened 2020-02-06
Move the `include`/`require` parenthesis rules

desrosj:
This is a follow-up to [WP-Core 48082](https://core.trac.wordpress.org/ticket/48082) and [WP-Core 49376](https://core.trac.wordpress.org/ticket/49376).

This PR moves the rules that check and enforce the following from `WordPress-Extra` to `WordPressCore`
- No parenthesis when using the `include`/`require` language constructs.
- One and only one space immediately following the use of a language construct.

[Related Slack conversation](https://wordpress.slack.com/archives/C5VCTJGH3/p1580999225029300).

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -98,6 +98,16 @@
 	<!-- Covers rule: Braces should always be used, even when they are not required. -->
 	<rule ref="Generic.ControlStructures.InlineControlStructure"/>
 
+	<!-- Covers rule: parenthesis should not be used when using include, include_once, require, require_once. -->
+	<rule ref="PEAR.Files.IncludingFile.BracketsNotRequired">
+		<type>warning</type>
+	</rule>
+	<rule ref="PEAR.Files.IncludingFile.UseRequire">
+		<type>warning</type>
+	</rule>
+	<rule ref="PEAR.Files.IncludingFile.UseRequireOnce">
+		<type>warning</type>
+	</rule>
 
 	<!--
 	#############################################################################
@@ -235,6 +245,10 @@
 		<severity>0</severity>
 	</rule>
 
+	<!-- Check correct spacing of language constructs. This also ensures that the
+	     above rule for not using brackets with require is fixed correctly.
+		 https://github.com/WordPress/WordPress-Coding-Standards/issues/1153 -->
+	<rule ref="Squiz.WhiteSpace.LanguageConstructSpacing"/>
 
 	<!-- Covers rule: Call a function, like so: my_function( $param1, func_param( $param2 ) ); -->
 	<rule ref="PEAR.Functions.FunctionCallSignature">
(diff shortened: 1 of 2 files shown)


WordPress-Core/ruleset.xml
```diff
@@ -235,6 +245,10 @@
 		<severity>0</severity>
 	</rule>
 
+	<!-- Check correct spacing of language constructs. This also ensures that the
+	     above rule for not using brackets with require is fixed correctly.
+		 https://github.com/WordPress/WordPress-Coding-Standards/issues/1153 -->
+	<rule ref="Squiz.WhiteSpace.LanguageConstructSpacing"/>
```
GaryJones: Note that this sniff is deprecated and [removed in PHPCS 4.0](https://github.com/squizlabs/PHP_CodeSniffer/blob/077ebabafe34c7bc3c214bc8c9f591580707914b/CHANGELOG.md#L43-L44), but fixing it should be done in a different PR.

jrfnl: Issue #1616 is already open about that and as soon as we decide to up the minimum supported PHPCS version to 3.4.1 or higher, it should be swopped out.

WordPress-Core/ruleset.xml
```diff
@@ -98,6 +98,16 @@
 	<!-- Covers rule: Braces should always be used, even when they are not required. -->
 	<rule ref="Generic.ControlStructures.InlineControlStructure"/>
 
+	<!-- Covers rule: parenthesis should not be used when using include, include_once, require, require_once. -->
+	<rule ref="PEAR.Files.IncludingFile.BracketsNotRequired">
+		<type>warning</type>
```
jrfnl: As the rule is now moved to `Core`, this is no longer a "best practice" advise, but a hard rule, so I'd like to suggest removing the `type` change for this specific error code.

For the other two error codes, I'd leave the `type` in place as this is more contentious.
You could even consider not moving the other two error codes from this sniff and just moving the `BracketsNotRequired` one to `Core`.
In that case, you may need to add `<severity>5</severity>` to these rules in the `Extra` ruleset (as including the one code in `Core` sets all the other error codes in the sniff to severity `0` IIRC).

Also take note of the fact that WPCS explicitly does not include the other two error codes from this sniff `UseIncludeOnce` and `UseInclude`.

WordPress-Core/ruleset.xml
```diff
@@ -235,6 +245,10 @@
 		<severity>0</severity>
 	</rule>
 
+	<!-- Check correct spacing of language constructs. This also ensures that the
+	     above rule for not using brackets with require is fixed correctly.
```
jrfnl: Use tabs not spaces for indentation.

jrfnl:
@desrosj Just checking: do you still intend to finish this PR (i.e. update it for the feedback given) ?

jrfnl:
Looked this over again. Actions needed to make this mergable:

- [ ] Add the rule to the [Coding Standards handbook page](https://github.com/WordPress/wpcs-docs/blob/master/wordpress-coding-standards/php.md).
- [ ] Rebase the PR to fix the conflict after #1921 
- [ ] Update the PR for the feedback left above.

Issue #1865, opened 2020-02-12
Namespaced cron_schedules callbacks not detected


ocean90:
## Bug Description

The `WordPress.WP.CronInterval` sniff is producing a warning when a namespaced callback is used: "Detected changing of cron_schedules, but could not detect the interval value."

## Minimal Code Snippet

```php
namespace Foo {
	function my_add_every_hour( $schedules ) {
		$schedules['every_hour'] = [
			'interval' => HOUR_IN_SECONDS,
			'display' => __( 'Once every hour' )
		];
		return $schedules;
	}
	add_filter( 'cron_schedules', __NAMESPACE__ . '\my_add_every_hour'); // OK: > 10 min.
}
```

## Environment

| Question               | Answer
| ------------------------| -------
| PHP version             | 7.2.23
| PHP_CodeSniffer version | 3.5.4
| WPCS version            | 2.2.1
| WPCS install type       | Composer global

## Tested Against `develop` branch?
- [x] I have verified the issue still exists in the `develop` branch of WPCS.

ocean90:
Added a (failing) test in #1866.

I'm happy to work on a patch, though I'd need some hints. So far I noticed that the value of [`$callback` here](https://github.com/WordPress/WordPress-Coding-Standards/blob/dd86ccddab77f4631f005893a8b7786b5311efe3/WordPress/Sniffs/WP/CronIntervalSniff.php#L99) is
```
array(3) {
  ["start"]=>
  int(82)
  ["end"]=>
  int(87)
  ["raw"]=>
  string(36) "__NAMESPACE__ . '\my_add_every_hour'"
}
```

And `token_name( $this->tokens[ $callbackArrayPtr ]['code'] )` around [here](https://github.com/WordPress/WordPress-Coding-Standards/blob/dd86ccddab77f4631f005893a8b7786b5311efe3/WordPress/Sniffs/WP/CronIntervalSniff.php#L110) returns `T_NS_C`.

jrfnl:
@ocean90 Thanks for reporting this.

A lot of the WPCS sniffs at this moment are not adapted yet to handle namespaced code correctly in all situations.

Fixing this for this sniff will not be easy as, in most cases, a file based namespace will be used not a scoped (curly braces) namespace.
And with a file based namespace, the function could be:
- declared in another file in the same namespace;
- could be imported via a `use function` statement from elsewhere;
- could be in another namespace all together.

[...]

PR #1866, opened 2020-02-12
Add test for namespaced cron_schedules callback

ocean90:
See https://github.com/WordPress/WordPress-Coding-Standards/issues/1865.


jrfnl:
I agree it would be nice to address this, so leaving the associated ticket #1865 open to do so. However, this PR as-is is not actionable as it would introduce a failing test and doesn't fix it, so I'm closing the PR.

PR #1876, opened 2020-04-01
PreparedSQL: quick fix for namespace separator

### LINKED ISSUE #1875: PreparedSQLSniff: False-positive for `array_fill()` with `\count()`
## Bug Description
When `count()` is prefixed with the root namespace `\` in the pattern `implode( ',', array_fill( 0, count( $something ), '%s' ) )` you'll get a false-positive for a not prepared query.

## Minimal Code Snippet
```php
$wpdb->prepare(
	"SELECT * FROM $wpdb->something WHERE id IN(" . implode( ',', array_fill( 0, \count( $ids ), '%d' ) ) . ')',
	$ids
);
```

## Error Code
`Use placeholders and $wpdb->prepare(); found \ (WordPress.DB.PreparedSQL.NotPrepared)`

## Environment

| PHP version             | 7.3.16
| PHP_CodeSniffer version | 7.3.16
| WPCS version            | 2.2.1
| WPCS install type       | Composer global
| IDE (if relevant)       | VS Code

jrfnl:
Reported in #1875: using a namespace separator with function calls in the `WordPress.DB.PreparedSQL` triggers a false positive on the namespace separator as "not being escaped".

While this really should be handled differently with proper FQN function name determination for function calls etc, for now, let's just ignore the namespace separator as that is not something which would need to be "prepared".

Fixes #1875

### DIFF
--- a/WordPress/Sniffs/DB/PreparedSQLSniff.php
+++ b/WordPress/Sniffs/DB/PreparedSQLSniff.php
@@ -69,6 +69,7 @@ class PreparedSQLSniff extends Sniff {
 		\T_INT_CAST                 => true,
 		\T_DOUBLE_CAST              => true,
 		\T_BOOL_CAST                => true,
+		\T_NS_SEPARATOR             => true,
 	);
 
 	/**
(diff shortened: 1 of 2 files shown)


jrfnl:
The build failure is unrelated to this PR - see open PR #1874.

jrfnl:
Rebased to get a passing build.

PR #1883, opened 2020-04-09
Clean up the I18n sniff

jrfnl:
While running some tests for something else, I noticed that the I18n sniff was throwing all sorts of CS warnings.

Fixed now.

#### I18n: rename variable

Rename a very long variable name and fix a CS warning for incorrect alignment.

#### I18n: assignment in condition is only allowed in a while loop

The code as was, was throwing two warnings for the assignment in the for loop. Let's not set a bad example.


jrfnl:
Thanks for pointing it out. That was the usual Travis erroring out on installs. Just a case of restarting the jobs (which I just did). Build should pass now.

PR #1885, opened 2020-04-14
Add missing auto-escaped function

zzap:
New(ish) privacy features added new auto-escaped function `the_privacy_policy_link` - [docs](https://developer.wordpress.org/reference/functions/the_privacy_policy_link/).

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -218,6 +218,7 @@ abstract class Sniff implements PHPCS_Sniff {
 		'term_description'          => true,
 		'the_author'                => true,
 		'the_date'                  => true,
+		'the_privacy_policy_link'   => true,
 		'the_title_attribute'       => true,
 		'walk_nav_menu_tree'        => true,
 		'wp_dropdown_categories'    => true,


zzap:
Yes, I was playing with `get_the_privacy_policy_link()` in my local and it is potential hole. I'd escape it with `wp_kses_post`. At least to avoid Javascript injection.

Thank you for the welcome :)

GaryJones:
So there's a way for unsafe content to be output, but we want to make it as auto-escaped and therefore not flag it? What am I missing here?

jrfnl:
@GaryJones Yes, there is a way, but that's no different with nearly every other function in the list. So either we should accept this PR or we should remove all the other functions with similar weaknesses from the list.

GaryJones:
By itself, I'd say No for this change, but, I also don't like inconsistencies. I'd reluctantly approve this, if needed.

Perhaps as a separate task we could look at what functions really do provide solid escaping at the last moment, and which make an effort before allowing a filter at the last moment instead.

jrfnl:
Based on the above discussion, I'm closing this PR. I've opened a separate issue to review the existing list of "safe functions". See #2119.

PR #1892, opened 2020-05-08
First pass at sniff to confirm inline docs for filters are present

### LINKED ISSUE #424: Check that hooks follow documentation standards
Forked from https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/241#issuecomment-133562703.

kraftbj:
Fixes #424

First pass at sniff to confirm inline docs for hooks are present.

I'm adding this into Jetpack via https://github.com/Automattic/jetpack/pull/15693 since I wanted to check for some Jetpack-specific things. Per suggestion, wanted to offer the "base" to core WPCS for everyone's benefit.

In the initial commit, it finds a matching the various hook function (L27) using WPCS' `AbstractFunctionRestrictionsSniff` and checks for a preceding comment. The preceding comment need to be on the previous line (noting this is an open question from the Jetpack PR https://github.com/Automattic/jetpack/pull/15693#pullrequestreview-406406725) and confirm it is using the docblock format.

For the initial commit, it is verifying it has a `@since` tag within the docblock and that the value starts with a X.Y.Z version number. I'm not sure about WP itself, but Jetpack requires three levels to both confirm the exact version (e.g. 8.5 vs 8.5.0) and ensure that all hooks in the same major release (X.Y) will have the same value to avoid our docs parser tagging it as two separate versions.

If Core WPCS doesn't need that, I can remove that to just verify a `@since` tag is present and move the version check itself to Jetpack as I did with our `@module` tag in https://github.com/Automattic/jetpack/pull/15693/files#diff-eb7bc4afcdf23e6f99ae1f793afd119bR57

### DIFF
--- a/WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
+++ b/WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
@@ -8,7 +8,9 @@
         <code title="Valid: using a DocBlock to describe a hook.">
             <![CDATA[
 /**
- * Description of hook.
+ * Summary of hook.
+ *
+ * This is a longer description of a hook.
  *
  * @since 1.0.0
  */
@@ -18,7 +20,7 @@ do_action( 'hook' );
         <code title="Invalid: not using a DocBlock comment">
             <![CDATA[
 /*
- * This is not a docblock.
+ * This is not a DocBlock.
  */
 do_action( 'hook' );
         ]]>
--- a/WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
+++ b/WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
@@ -1,11 +1,11 @@
-<documentation title="Hooks Inline Docblock Standard">
+<documentation title="Hooks Inline DocBlock Standard">
     <standard>
         <![CDATA[
-    All WordPress hook execution functions should be preceded by a docblock explaining the hook.
+    All WordPress hook execution functions should be preceded by a DocBlock explaining the hook.
     ]]>
     </standard>
     <code_comparison>
-        <code title="Valid: using a docblock to describe a hook.">
+        <code title="Valid: using a DocBlock to describe a hook.">
             <![CDATA[
 /**
  * Description of hook.
@@ -15,7 +15,7 @@
 do_action( 'hook' );
         ]]>
         </code>
-        <code title="Invalid: using a non-docblock comment">
+        <code title="Invalid: not using a DocBlock comment">
             <![CDATA[
 /*
  * This is not a docblock.
@@ -28,7 +28,7 @@ do_action( 'hook' );
         <code title="Valid: including a @since tag with a valid X.Y.Z version.">
             <![CDATA[
 /**
- * Description of hook.
+ * Summary of hook.
  *
  * <em>@since 1.0.0</em>
  */
--- a/WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
+++ b/WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
@@ -1,0 +1,49 @@
+<documentation title="Hooks Inline Docblock Standard">
+    <standard>
+        <![CDATA[
+    All WordPress hook execution functions should be preceded by a docblock explaining the hook.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: using a docblock to describe a hook.">
+            <![CDATA[
+/**
+ * Description of hook.
+ *
+ * @since 1.0.0
+ */
+do_action( 'hook' );
+        ]]>
+        </code>
+        <code title="Invalid: using a non-docblock comment">
+            <![CDATA[
+/*
+ * This is not a docblock.
+ */
+do_action( 'hook' );
+        ]]>
+        </code>
+    </code_comparison>
+    <code_comparison>
+        <code title="Valid: including a @since tag with a valid X.Y.Z version.">
+            <![CDATA[
+/**
+ * Description of hook.
+ *
+ * <em>@since 1.0.0</em>
+ */
+do_action( 'hook' );
+        ]]>
+        </code>
+        <code title="Invalid: including a @since tag with a different undeclared version.">
+            <![CDATA[
+/**
+ * Description of hook.
+ *
+ * <em>@since Initial release.</em>
+ */
+do_action( 'hook' );
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>
--- a/WordPress/Sniffs/Commenting/HooksInlineDocsSniff.php
+++ b/WordPress/Sniffs/Commenting/HooksInlineDocsSniff.php
@@ -95,7 +95,6 @@ class HooksInlineDocsSniff extends AbstractFunctionRestrictionsSniff {
 					$stack_ptr,
 					'NoDocblockFound'
 				);
-				// return; Do we need to return here? Can we keep going?
 			}
 
 			/*
@@ -121,8 +120,8 @@ class HooksInlineDocsSniff extends AbstractFunctionRestrictionsSniff {
 					// If it is false, there is no text or if the text is on the another line, error.
 					if ( false === $string || $this->tokens[ $string ]['line'] !== $this->tokens[ $tag ]['line'] ) {
 						$this->phpcsFile->addError( 'Since tag must have a value.', $tag, 'EmptySince' );
-					} elseif ( ! preg_match( '\'/^\d+\.\d+\.\d+/\'', $string ) ) { // Requires X.Y.Z. Trailing 0 is needed for a major release.
-						$this->phpcsFile->addError( 'Since tag must have a X.Y.Z. version number.', $tag, 'InvalidSince' );
+					} elseif ( ! preg_match('/^\d+\.\d+\.\d+/', $this->tokens[ $string ]['content'] ) ) { // Requires X.Y.Z. Trailing 0 is needed for a major release.
+						$this->phpcsFile->addError( 'Since tag must have a X.Y.Z version number.' . $this->tokens[ $string ]['content'], $tag, 'InvalidSince' );
 					}
 				}
 			}
@@ -160,11 +159,19 @@ class HooksInlineDocsSniff extends AbstractFunctionRestrictionsSniff {
 	 *
 	 * @param int $start       The position of the starting token in the stack.
 	 * @param int $end       The position of the ending token in the stack.
+	 *
+	 * @return bool If docblock matches the previously documented convention.
 	 */
 	protected function is_previously_documented( $start, $end ) {
 		$string = $this->phpcsFile->findNext( T_DOC_COMMENT_STRING, $start, $end );
+
+		$content = $this->tokens[ $string ]['content'];
 		// If the call is documented elsewhere, stop here.
-		if ( 0 === strpos( $this->tokens[ $string ]['content'], 'This filter is documented in' ) ) {
+		if ( 0 === strpos( $content, 'This filter is documented' ) ) {
+			return true;
+		}
+
+		if ( 0 === strpos( $content, 'This action is documented' ) ) {
 			return true;
 		}
 
--- a/WordPress/Sniffs/Commenting/HooksInlineDocsSniff.php
+++ b/WordPress/Sniffs/Commenting/HooksInlineDocsSniff.php
@@ -40,13 +40,13 @@ class HooksInlineDocsSniff extends AbstractFunctionRestrictionsSniff {
 	 * they can be passed in the ruleset XML file via:
 	 * <rule ref="WordPress.Commenting.HooksInlineDocs">
 	 *  <properties>
-	 *   <property name="allowed_extra_versions" type="array">
-	 *    <element key="0.71" value="0.71"/>
-	 *    <element key="MU (3.0.0)" value="MU (3.0.0)"/>
-	 *   </property>
+	 *   <property name="allowed_extra_versions" type="array" value="0.71"/>
+	 *   <property name="allowed_extra_versions" type="array" value="MU (3.0.0)"/>
 	 * </rule>
 	 *
 	 * The key values are used to determine if a version is allowed outside of the X.Y.Z scheme. The value is not considered.
+	 *
+	 * @var array Array of allowed exceptional values.
 	 */
 	public $allowed_extra_versions = array();
 
@@ -137,7 +137,7 @@ class HooksInlineDocsSniff extends AbstractFunctionRestrictionsSniff {
 					// If it is false, there is no text or if the text is on the another line, error.
 					if ( false === $string || $this->tokens[ $string ]['line'] !== $this->tokens[ $tag ]['line'] ) {
 						$this->phpcsFile->addError( 'Since tag must have a value.', $tag, 'EmptySince' );
-					} elseif ( ! preg_match('/^\d+\.\d+\.\d+/', $this->tokens[ $string ]['content'], $matches ) ) { // Requires X.Y.Z. Trailing 0 is needed for a major release.
+					} elseif ( ! preg_match( '/^\d+\.\d+\.\d+/', $this->tokens[ $string ]['content'], $matches ) ) { // Requires X.Y.Z. Trailing 0 is needed for a major release.
 						if ( empty( $this->allowed_extra_versions ) || ! $this->array_begins_with( $this->tokens[ $string ]['content'], $this->allowed_extra_versions ) ) {
 							$this->phpcsFile->addError( 'Since tag must have a X.Y.Z version number.', $tag, 'InvalidSince' );
 						}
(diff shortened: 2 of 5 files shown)


jrfnl:
Hi @kraftbj Welcome to WPCS ;-)

Thanks for this PR. I haven't looked at the code yet, but would like to ask you whether you have run the sniff over WP Core to see what it throws up.

Any false positives from WP Core should be handled, or when in doubt whether something is a false positive, let's discuss it.

> For the initial commit, it is verifying it has a `@since` tag within the docblock and that the value starts with a X.Y.Z version number. I'm not sure about WP itself, but Jetpack requires three levels to both confirm the exact version (e.g. 8.5 vs 8.5.0) and ensure that all hooks in the same major release (X.Y) will have the same value to avoid our docs parser tagging it as two separate versions.

Core requires three levels too as per the [Documentation guidelines](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#since-section-changelogs), so checking for three levels is fine.

FYI: I expect WPCS 3.0.0 to contain a trait for version number checking (written for another sniff), so that can probably be re-used.

With that in mind and the last 2.x release expected soon, after which the pulls for WPCS 3.0.0 will start, I'd like to earmark this PR for WPCS 3.0.0.

kraftbj:
Sounds good. New to PHPCS overall, so appreciate the guidance on naming scheme. Took a guess that `WP` seem to fit best based on the existing directories.

I'll get some unit tests worked up early next week since it sounds like this is a sniff that would be considered for inclusion.

jrfnl:
> New to PHPCS overall, so appreciate the guidance ...

Please feel free to ask for as much guidance as you need from me and the others in the team.

In my case, my remarks will often presume you know what you're doing with PHPCS (as why would you send in a sniff otherwise?), please don't let that intimidate you and ask for as much clarification as you need.

> it sounds like this is a sniff that would be considered for inclusion.

Definitely. There are long standing open issues about this check missing from WPCS, though I also have to caution you that it may not be the easiest issue to tackle. Can be a great way to learn more about writing sniffs though ;-)

kraftbj:
Sounds good. I appreciate the tip on running it against WP to find false positives. I'll make a couple of tweaks when sending up the unit tests.

GaryJones:
> For the initial commit, it is verifying it has a `@since` tag within the docblock and that the value starts with a X.Y.Z version number.

Maybe `@since Unknown` should also be permitted? I believe that has/had some instances in core function/methods DocBlocks as well at one point ([example in SimplePie package](https://github.com/WordPress/WordPress/search?q=%22%40since+Unknown%22&unscoped_q=%22%40since+Unknown%22)).

jrfnl:
> Maybe @since Unknown should also be permitted?

I don't think we should allow that. Both SVN as well as Git allow to trace things through history, so I can't think of a valid reason to allow `@since Unknown`.

kraftbj:
I added unit tests and documentation, but I'm not sure how to unit test when the ruleset would be passing a property as an array.

With the X.Y.Z. format, there are two instances currently in Core which does not follow this which I am not keen at automatically changing:
* `0.71` -- the initial release did not follow the X.Y.Z format.
* `MU (3.0.0)` -- to denote items that came in when MU was merged into Core. We could modify these to be `3.0.0 (MU)` as an alternative.

To allow these, I set a rule property of an array that could be defined in the ruleset XML for any "exceptional" versions. I just don't know how to pass that to the unit test via `phpcs:set`. 

Any pointers?

kraftbj:
> Maybe `@since Unknown` should also be permitted?

For Core, I would disagree, but for implementing projects, I did add a way to set an `exceptional` version. If an implementing project wanted to allow that, they could in their own `phpcs.xml.dist` file as the PR stands now.

jrfnl:
> I'm not sure how to unit test when the ruleset would be passing a property as an array.

See: https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.1.inc#L6 for an example of setting an array property.

Please also remember to reset the property after the test using it and/or at the end of the test file (see bottom of the above linked test file for an example).

kraftbj:
@jrfnl I really appreciate the pointers and direction. I think this is at a state where you can ~rip it to shreds~ review it.

jrfnl:
@kraftbj Wowie! I'll try and have a look some time this week ;-)

WordPress/Sniffs/Commenting/HooksInlineDocsSniff.php
```diff
@@ -0,0 +1,236 @@
...
+ * @package WPCS\WordPressCodingStandards
+ */
+class HooksInlineDocsSniff extends AbstractFunctionRestrictionsSniff {
+
+	/**
+	 * Array of WordPress hook execution functions.
+	 *
+	 * @var array WordPress hook execution function name => filter or action.
+	 */
+	protected $hook_functions = array(
```
GaryJones, line 27: We have a lot of arrays of functions names in the `Sniff` class that all of the Abstract (and therefore concrete) classes extend from - might be worth renaming the property to a bit more explicitly (`$hook_executing_functions`?) and then moving them to there so then other sniffs that need this collection can use them.

WordPress/Sniffs/Commenting/HooksInlineDocsSniff.php
```diff
@@ -0,0 +1,236 @@
...
+
+		$previous_comment = $this->return_previous_comment( $stack_ptr );
+
+		if ( false !== $previous_comment ) {
+			/*
+			 * Check to determine if there is a comment immediately preceding the function call.
+			 */
+			if ( ( $this->tokens[ $previous_comment ]['line'] + 1 ) !== $this->tokens[ $stack_ptr ]['line'] ) {
+				$this->phpcsFile->addError(
+					'The inline documentation for a hook must be on the line immediately before the function call.',
```
GaryJones, line 100: As noted previously, this may not always be possible. Consider this example from Juliette:

```
/**
 * Hook docblock
 */
if ( $a === true
    && apply_filter('some_filter', $value) > 10
) {
    // Do something.
}
```

Likewise, this should also be valid IMO, since it matches the PSR-5 definition of associating a DocBlock with a Structural Element with an optional number of clear lines in between:

```
/**
 * Hook docblock
 */

$value = apply_filters( 'some_filter', $value );
```

WordPress/Tests/Commenting/HooksInlineDocsUnitTest.inc
```diff
@@ -0,0 +1,108 @@
...
+
+// Has a since tag with an incomplete value.
+/**
+ * This is a description.
+ *
+ * @since 8.5
+ */
+apply_filters( 'hook', 'function' );
+
+// Has a since tag with an wrong value.
```
GaryJones, line 53:
```suggestion
// Has a since tag with a wrong value.
```

WordPress/Tests/Commenting/HooksInlineDocsUnitTest.inc
```diff
@@ -0,0 +1,108 @@
...
+apply_filters( 'hook', 'function' );
+
+// "Already documented" example of an action.
+/** This action is documented in file.php */
+do_action( 'hook ' );
+
+// "Already documented" example of a filter.
+/** This filter is documented in file.php */
+apply_filters(' hook', 'function' );
+
```
GaryJones, line 107: These are good, but some real-world edge cases will be needed!

Consider:

```php
/**
 * This is my function DocBlock.
 *
 * @since 0000.00000.0000000000
 */
function my_function_has_a_hook() {
    do_action( 'foo' );
}
```

Likewise:

```php
/**
 * This is my function DocBlock.
 *
 * @since 0000.00000.0000000000
 */
function my_function_has_a_hook() {
    echo 'foo';
}
do_action( 'foo' );
```

If these both already get reported as missing the hook DocBlock, then great! Add them in as unit tests, and then try:

```php
/**
 * This is my function DocBlock.
 *
 * @since 0000.00000.0000000000
 */
function my_function_has_a_hook() { do_action( 'foo' ); }
```

😈

WordPress/Tests/Commenting/HooksInlineDocsUnitTest.inc
```diff
@@ -0,0 +1,108 @@
+<?php
+// phpcs:set WordPress.Commenting.HooksInlineDocs allowed_extra_versions[] MU
```
GaryJones, line 2: This doesn't have to be specified at the top - you could move it to just before it's needed, so that you can check that the behaviour works correctly without any allowed_extra_versions having been set.

WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
```diff
@@ -0,0 +1,49 @@
+<documentation title="Hooks Inline Docblock Standard">
```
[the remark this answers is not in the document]
kraftbj: b50e8e3d

WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
```diff
@@ -0,0 +1,49 @@
...
+    <standard>
+        <![CDATA[
+    All WordPress hook execution functions should be preceded by a docblock explaining the hook.
```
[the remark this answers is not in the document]
kraftbj: b50e8e3d

WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
```diff
@@ -0,0 +1,49 @@
...
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: using a docblock to describe a hook.">
```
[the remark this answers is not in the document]
kraftbj: b50e8e3d

WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
```diff
@@ -0,0 +1,49 @@
...
+            <![CDATA[
+/**
+ * Description of hook.
```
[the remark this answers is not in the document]
kraftbj: b50e8e3d

WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @since 1.0.0
+ */
+do_action( 'hook' );
+        ]]>
+        </code>
+        <code title="Invalid: using a non-docblock comment">
```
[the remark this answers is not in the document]
kraftbj: b50e8e3d

WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * <em>@since 1.0.0</em>
+ */
+do_action( 'hook' );
+        ]]>
+        </code>
+        <code title="Invalid: including a @since tag with a different undeclared version.">
+            <![CDATA[
+/**
+ * Description of hook.
```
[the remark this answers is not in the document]
kraftbj: b50e8e3d

WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
```diff
@@ -0,0 +1,49 @@
...
+ */
+do_action( 'hook' );
+        ]]>
+        </code>
+    </code_comparison>
+    <code_comparison>
+        <code title="Valid: including a @since tag with a valid X.Y.Z version.">
+            <![CDATA[
+/**
+ * Description of hook.
```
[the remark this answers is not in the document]
kraftbj: 56960143

WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @since 1.0.0
+ */
+do_action( 'hook' );
+        ]]>
+        </code>
+        <code title="Invalid: using a non-docblock comment">
+            <![CDATA[
+/*
+ * This is not a docblock.
```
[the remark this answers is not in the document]
kraftbj: 56960143

WordPress/Tests/Commenting/HooksInlineDocsUnitTest.inc
kraftbj: b50e8e3d

WordPress/Tests/Commenting/HooksInlineDocsUnitTest.inc
kraftbj: b2855e67

WordPress/Sniffs/Commenting/HooksInlineDocsSniff.php
kraftbj: Nice pointer. In 11e3a8fc, I removed my array. There is already a `$hookInvokeFunctions` array in the Sniff class.

The difference is the value of each key in the existing one is `true` whereas mine had `filter|action`. Originally, I was going to ensure that the "previously documented" check (`This filter is documented`) used `filter` or `action` correctly, but opted against being that nit-picky.

Since I wasn't used the key, let's just use what's already there.

WordPress/Tests/Commenting/HooksInlineDocsUnitTest.inc
kraftbj: The first two did catch "correctly", but probably wouldn't after I resolve the preceding line issue mentioned in https://github.com/WordPress/WordPress-Coding-Standards/pull/1892#discussion_r424599130

But, the third example of a single-line function fails the unit test since it doesn't see that as an error.

I'm going to push up what I have (and mark this as resolved since the tests are added) and think on it a bit.

I'm thinking for the third example, we can search for a comment previous to the hook but stop at a function declaration at least. May be others.

Thanks!

kraftbj: ad220fdb

WordPress/Sniffs/Commenting/HooksInlineDocsSniff.php
GaryJones: If the existing values were all `true`, then chances are the values weren't being used elsewhere. If there is benefit in having them be either 'filter' or 'action' so that _this_ sniff can check a little more finely (would you propose to throw a different violation if they didn't match?), then I'm all for that.

cc @jrfnl for consideration. Unresolving this comment so it doesn't get missed.

WordPress/Tests/Commenting/HooksInlineDocsUnitTest.inc
GaryJones: Either here or PHPCSUtils, I bet @jrfnl has examples / helper functions to find something within a specific scope i.e. find a DocComment within the current scope.

WordPress/Docs/Commenting/HooksInlineDocsStandard.xml
```diff
@@ -15,7 +15,7 @@
 do_action( 'hook' );
         ]]>
         </code>
-        <code title="Invalid: using a non-docblock comment">
+        <code title="Invalid: not using a DocBlock comment">
```
GaryJones, line 20: A "DocBlock" is made up of two parts:

- the `/**  */` structure, known as a _DocComment_.
- the contents with Summary, optional description, tags, etc. known as the _PHPDoc_.

As such, the "DocBlock comment" here, which is referencing whether one asterisk or two is used, is really checking if it's a multi-line comment (`/*`) or a _DocComment_ (`/**`), hence my earlier suggestion of `Invalid: not using a DocComment."

(Note that the period is still missing at the end from your version.)

WordPress/Sniffs/Commenting/HooksInlineDocsSniff.php
jrfnl: I'm pretty sure that the values in the array aren't currently used, but a search of the codebase can verify this for sure.

I'm all for it to have `filter|action` as the values. The error messages could then potentially also use that term if it makes the error message clearer.

WordPress/Tests/Commenting/HooksInlineDocsUnitTest.inc
jrfnl: I do have docblock finder helper functions lined up to go into PHPCSUtils at some point, but as hook docblocks are a bit of a different beast compared to most structural elements I don't think that will help here.

I do have a draft `HooksTrait` lined up for WPCS 3.0.0 with logic for finding hook docblocks (for another sniff).

When I pull that, I will combine the logic for finding hook docblocks from this sniff with the logic in the trait to get the best of both brains in the trait (and switch this sniff over to using the trait).

kraftbj:
This is wonderful. Thank you so much for the review. Today is a standard holiday in the U.S. so a short week for me. I'll try to make a good dent on this during the week, but may get bumped to next. If so, my delay doesn't mean I'm discouraged! 🙂

jrfnl:
While it would be nice to have a sniff for this, the PR as-is is not mergable with way too many open review comments and no action taken on them in over two years, so I'm going to close this PR.

PR #1907, opened 2020-06-30
3.0: Tests: add a bootstrap file

jrfnl:
While not strictly necessary, it makes life easier to have this bootstrap in place.

The bootstrap file will:
* Load the PHPCS autoloader (was previously done from the command-line);
* Load the Composer autoload file.
* Will automatically set the `PHPCS_IGNORE_TESTS` environment variable to the correct value to prevent running tests from other standards.

For locally running the tests after having run `composer install`, all should work fine.

For locally running the tests with a _git clone_ of PHPCS, a `phpunit.xml` overload file can be used with a `env` variable `PHPCS_DIR` to point the bootstrap to that install. Do make sure that PHPCSUtils is installed and registered in the `installed_paths` of that install if you do so.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,6 +9,7 @@
 /phpunit.xml.dist export-ignore
 /.github export-ignore
 /bin export-ignore
+/Tests export-ignore
 /WordPress/Tests export-ignore
 
 #
--- a/Tests/bootstrap.php
+++ b/Tests/bootstrap.php
@@ -1,0 +1,85 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * Bootstrap file for running the tests.
+ *
+ * - Load the PHPCS PHPUnit bootstrap file providing cross-version PHPUnit support.
+ *   {@link https://github.com/squizlabs/PHP_CodeSniffer/pull/1384}
+ * - Load the Composer autoload file.
+ * - Automatically limit the testing to the WordPressCS tests.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+if ( ! defined( 'PHP_CODESNIFFER_IN_TESTS' ) ) {
+	define( 'PHP_CODESNIFFER_IN_TESTS', true );
+}
+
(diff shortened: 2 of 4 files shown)


jrfnl:
@dingo-d I fully agree, though I think it should go in the `CONTRIBUTING.md` doc as running unit tests is typically something for contributors.

I already have this written up, or rather, I have a local commit with:
* Updated Readme
* Updated Contributing
* Update bug report template
* End-user upgrade guide -> to be added the changelog
* Dev upgrade guide (i.e. for standards extending WPCS) -> to be added to the wiki

As the `develop` branch is the default branch in this repo, if I would pull those docs now, those would start showing straight away, while all of that will not apply yet until 3.0 has a (RC) release, so it would be really confusing for people who try to install the latest (2.x) release.

So as per my note in #1905:

> Note: the install/upgrade instructions in the Readme and other documentation will be addressed in a separate PR late in the 3.0 cycle, so as not to confuse people using the latest release, while 3.0.0 development is ongoing.

GaryJones:
I wonder if PHPCSExtra / PHPCSUtils could handle a generic version of this logic in a function - find PHPCS directory, autoload PHPCS files, ignore all but a standard passed in as a string - so that WPCS and other standards packages could simply call `\PHPCSExtra\bootstrap_tests('My-Cool-Standard');` or whatever.

jrfnl:
> Sorry for the delayed review - there was a bit more going on in this PR, and I wanted to have the time to get my head around what was actually happening.

@GaryJones No worries, I just wanted to make sure (with the review request) it had not been overlooked/escaped attention.

> I wonder if PHPCSExtra / PHPCSUtils could handle a generic version of this logic in a function - find PHPCS directory, autoload PHPCS files, ignore all but a standard passed in as a string - so that WPCS and other standards packages could simply call \PHPCSExtra\bootstrap_tests('My-Cool-Standard'); or whatever.

I like the idea - would be something for PHPCSUtils. I will need to have a think about it. Might be good to open an issue about it in that repo.

There is a related issue open already to create a PHP/PHPCS/PHPUnit cross-version compatible alternative to the PHPCS native base test class `AbstractSniffUnitTest`, which would take priority and would partially solve this already anyhow.
See: https://github.com/PHPCSStandards/PHPCSUtils/issues/72

PR #1908, opened 2020-06-30
3.0: Remove deprecated `Sniff::has_whitelist_comment()` method and all references to it

### LINKED ISSUE #1583: WPCS 3.0: Remove the deprecated Sniff::has_whitelist_comment() method
PR #1580 for WPCS 2.0.0 deprecated the WPCS native whitelist comments and the associated `Sniff::has_whitelist_comment()` method.

In WPCS 3.0.0 this method and all calls to it should be removed.

>  Note: all unit tests testing that the WPCS native whitelist comments are being respected should also be removed at that time. Or alternatively, they can remain and it should be verified that the WPCS native whitelist comment make no difference anymore.

jrfnl:
1. Remove the method which was deprecated in WPCS 2.0.0 (PR #1580).
2. Remove all function calls to the method.
    Includes:
    * Removing the `SlowQuerySniff::process_token()` method.
        This will now fall through automatically to the method in the parent class.
    * Removing the `GlobalVariablesOverrideSniff::maybe_add_error()` method.
        This method no longer has a function.
3. Adjusting the unit tests.
    * Where tests could be removed without having to renumber a lot, especially when the test was more about testing the whitelist functionality than testing the sniff, I've removed them.
    * In all other cases, I've changed the previous deprecation warning in the lines lists to a "normal" error/warning and annotated the change in behaviour.
    * Note: in the `PrefixAllGlobals` tests, there were two testcases which would previously throw a warning for the whitelist comment, but where the whitelist comment had no function anyway as the sniff had already been adjusted to no longer throw an error for those cases. For those, I've removed the whitelist comments.

Includes:
* Removing an exclusion for a whitelist comment deprecation notice from the WPCS native PHPCS ruleset.
* Updating the instructions regarding the old whitelist comments in the CONTRIBUTING file.
   :point_right:  :question: **Question**: As an alternative, the reference to the old-style whitelist comments could be removed completely from the CONTRIBUTING file. Should we ?

Fixes #1583


dingo-d:
In some places, I've seen comments in tests like `Bad - whitelisted via old-style whitelist comment`. For old code explanation (the `whitelist comment` part) it's reasonable to stay (as it was that way before), but should we change the `Bad - whitelisted via` part maybe? To something like: 

```Bad - included via old-style whitelist comment```

Does this make sense? It's just with everything going on in the world atm, maybe we should change this?

> Updating the instructions regarding the old whitelist comments in the CONTRIBUTING file.
👉 ❓ Question: As an alternative, the reference to the old-style whitelist comments could be removed completely from the CONTRIBUTING file. Should we?

I'm for this. The change will be written in the changelog so no information is lost.

jrfnl:
@dingo-d Good point, I fully agree.

I created this commit way before that discussion got started (april 2020 - see commit date) and have to admit that I didn't review it with that in mind before pulling it.

With this commit, the issue is largely gone from the sniff code, though the wording in `CONTRIBUTING` should probably still be improved, though if we remove that section, the issue is gone anyhow.

For the test files, I could change the wording from something like `// Old-style WPCS whitelist comments are no longer supported.` to something like this: `// Old-style WPCS-native ignore comments are no longer supported.`

I'll work on changing this and add that as a separate commit to this PR (which can be squashed on merge).

:point_right:  IIRC, there are some more places where the whitelist/blacklist terminology is used - PrefixAllGlobals comes to mind. Maybe we should open a separate issue as a reminder to address those too ? And while we're at it, we may want to open an issue about renaming the `master` branch too.

jrfnl:
Oh and we should probably also open an issue as a reminder to remove/update the wiki page about these type of "whitelist" comments to use "ignore comments" instead. That update can be done ASAP and would not need to wait till the release.

jrfnl:
PR updated for language used - see the three new commits (please squash on merge).

GaryJones:
> (please squash on merge)

Argh, sorry!

PR #1913, opened 2020-06-30
3.0: Tests/RestrictedClasses: compatibility with PHPCS 4.x

jrfnl:
PHPCS 4.x is set up to use PHPUnit 8+, including using the `void` return type for the `setUp[BeforeClass]()` and `tearDown[AfterClass]()` fixtures.

This commit makes our tests compatible with both PHPCS 3.x as well as 4.x and with PHPUnit < 8 and 8+.

This is done by:
* Renaming the fixtures in use in our test suite to use a different method name ( = only this class).
* And using the PHPUnit `@before` and `@after` annotations, which translate to `setUp()` and `tearDown()` to get round the `void` return type requirement.

Refs:
* https://phpunit.readthedocs.io/en/7.5/annotations.html#before
* https://phpunit.readthedocs.io/en/7.5/annotations.html#after

### DIFF
--- a/WordPress/Tests/DB/RestrictedClassesUnitTest.php
+++ b/WordPress/Tests/DB/RestrictedClassesUnitTest.php
@@ -19,6 +19,7 @@ use WordPressCS\WordPress\AbstractFunctionRestrictionsSniff;
  *
  * @since   0.10.0
  * @since   0.13.0 Class name changed: this class is now namespaced.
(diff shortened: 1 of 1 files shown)


jrfnl:
@GaryJones Thanks ;-)

Yes, I came up with this a while ago and for a codebase where the tests are largely based on a few abstract `TestCase  classes containing the fixtures and/or have few `setUp[BeforeClass]()`/`tearDown[AfterClass]()` methods, this is a workable solution and I verified with Sebastian Bergmann that these annotations will stay supported (at least there's no intention at this time to remove them).

PR #1922, opened 2020-07-06
3.0: Remove the PHPCSHelper class

jrfnl:
... in favour of the PHPCSUtils version of that class which adds:
* Improved PHPCS cross-version compatibility.
* More utility methods.

For full details on the upstream class, see the [PHPCSHelper class documentation for PHPCSUtils](https://phpcsutils.com/phpdoc/classes/PHPCSUtils-BackCompat-Helper.html).

The removal is done in three steps, which correspond to the commits in this PR:
* Deprecate the old code
* Switch over any uses of the old code
* Remove the old code

Doing this removal in steps allowed for testing in between, making sure every use of the old class was caught.

#### PHPCSHelper: deprecate the class and all the methods in it

As everything here is now contained within PHPCSUtils, including updated and/or improved versions of some of these.

#### PHPCSHelper::get_version(): remove call to the function

This function call is no longer needed now support for PHPCS < 3.5.0 has been dropped.

#### Remove the PHPCSHelper.php file

All of the methods contained herein are now provided via PHPCSUtils and all code has been switched over to use the PHPCSUtils versions of the methods instead.


GaryJones:
Not merged, in case this was preferred to be squashed or not.

jrfnl:
@GaryJones As these commits "tell a story", I'm fine with merging as-is, without squashing. If would squash this, we'd need to make sure that the context if maintained in the new commit message.

PR #1929, opened 2020-07-09
3.0: CapitalPDangit: implement PHPCSUtils and check namespace names

jrfnl:
This PR will be easiest to review by looking at the individual commits.

### CapitalPDangit: implement PHPCSUtils and other improvements

* The `Tokens::$textStringTokens` property was introduced in PHPCS 2.9.0 and is the same as the `$text_string_tokens` property in this sniff.
    Replacing this was missed in an earlier review round after the minimum PHPCS version changed.
* Use the PHPCSUtils [`ObjectDeclarations::getName()`](https://phpcsutils.com/phpdoc/classes/PHPCSUtils-Utils-ObjectDeclarations.html#method_getName) method instead of the PHPCS native `File::getDeclarationName()` method for a slightly more stable result.

### CapitalPDangit: check namespace names

Up to now, class names, trait names and interface names were already checked, but namespace names were not checked for the correct spelling of `WordPress`.

This new feature has now been added.

Includes unit tests.

### CapitalPDangit: minor efficiency tweak

No need to call `findPrevious()` twice.

### CapitalPDangit: minor code layout tweak

Minor code readability improvement.

### DIFF
--- a/WordPress/Sniffs/WP/CapitalPDangitSniff.php
+++ b/WordPress/Sniffs/WP/CapitalPDangitSniff.php
@@ -113,9 +113,13 @@ class CapitalPDangitSniff extends Sniff {
 		 * The return values skip to the end of the array.
 		 * This prevents the sniff "hanging" on very long configuration arrays.
 		 */
-		if ( \T_OPEN_SHORT_ARRAY === $this->tokens[ $stackPtr ]['code'] && isset( $this->tokens[ $stackPtr ]['bracket_closer'] ) ) {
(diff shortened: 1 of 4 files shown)


jrfnl:
> The misspelled variable named should be fixed, but the rest is otherwise OK.

:joy: LOL, that's actually a perpetuation of the misspelling everywhere else in the sniff. Kind of ironic for "misspelled" to be misspelled ;-)

Either way, happy to fix it throughout the sniff via a new commit.

PR #1931, opened 2020-07-09
3.0: ValidFunctionName: implement PHPCSUtils and more

jrfnl:
This PR will be easiest to review by looking at the individual commits.

### ValidFunctionName: implement PHPCSUtils

Use the improved PHPCSUtils versions of the PHPCS native methods.

### ValidFunctionName: use PHPCSUtils `isMagicFunctionName()`

Refs:
* [isMagicFunctionName()](https://phpcsutils.com/phpdoc/classes/PHPCSUtils-Utils-FunctionDeclarations.html#method_isMagicFunctionName)

### ValidFunctionName: use PHPCSUtils `isMagicMethodName()`

The magic method name list in PHPCSUtils is more complete than the one in the PEAR sniff.

Note: As this sniff bows out for methods declared in extended classes anyway, the PHP native double underscore methods from the SOAPClient class do not need to be taken into account, which is why the `isMagicMethodName()` method is used instead of the `isSpecialMethodName()` method.

Refs:
* [isMagicMethodName()](https://phpcsutils.com/phpdoc/classes/PHPCSUtils-Utils-FunctionDeclarations.html#method_isMagicMethodName)
* [isSpecialMethodName()](https://phpcsutils.com/phpdoc/classes/PHPCSUtils-Utils-FunctionDeclarations.html#method_isSpecialMethodName)

### ValidFunctionName: decouple from the PEAR sniff

This sniff extended the upstream `PEAR.NamingConventions.ValidFunctionName` sniff to allow using the properties related to magic functions/methods.

As handling of the magic functions/methods is now done via the PHPCSUtils `isMagicFunctionName()`/`isMagicMethodName()` methods, these properties are no longer used, so there is no need to extend the upstream PEAR sniff anymore.

In this commit:
* The parent sniff is changed from the upstream PEAR sniff to the WPCS base sniff.
* The `register()` and `process_token()` methods have been added.
* The `processTokenOutsideScope()` method has been renamed to `process_function_declaration()` to make the method name more descriptive.
* The `processTokenWithinScope()` method has been renamed to `process_method_declaration()` to make the method name more descriptive.
* The "method vs global function" determination is now done using the PHPCSUtils `Scopes::validDirectScope()` method which has higher accuracy.
    This fixes a bug where global functions declared nested within a method would not be examined at all.
* As the WPCS base sniff is now used, the `$this->phpcsFile` and `$this->tokens` properties are now available and use of them has been implemented.

Includes unit tests for the bug fix.

### ValidFunctionName: (re)move some code duplication

Move some code which was used in both the `process_function_declaration()` as well as the `process_method_declaration()` method to the `process_token()` method to remove some code duplication.

### ValidFunctionName: don't throw "DoubleUnderscore" error for triple underscore

This change was made in various upstream sniffs a while back as _triple_ underscore prefixed method names are not reserved by PHP.

Includes unit tests.

### ValidFunctionName: small efficiency fix

No need to call the `ObjectDeclarations::getName()` method for an anonymous class.

### DIFF
--- a/WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
@@ -151,6 +151,11 @@ class ValidFunctionNameSniff extends PHPCS_PEAR_ValidFunctionNameSniff {
 			return;
 		}
 
+		// PHP magic methods are exempt from our rules.
+		if ( FunctionDeclarations::isMagicMethodName( $methodName ) === true ) {
+			return;
+		}
+
 		$extended   = ObjectDeclarations::findExtendedClassName( $phpcsFile, $currScope );
 		$interfaces = ObjectDeclarations::findImplementedInterfaceNames( $phpcsFile, $currScope );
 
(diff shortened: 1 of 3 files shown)


jrfnl:
@dingo-d I'm not sure I understand your remark. As you saw already, both variables are used for the PHP 4 constructor/destructor checks.

Removing the variables would mean we'd need to have duplicate function calls to `strtolower()` for both these comparisons.

Did you, by chance, miss the second (destructor) comparison ?

dingo-d:
Oh yeah 🤦🏼‍♂️ GH folded that part of the code. My bad.

jrfnl:
@dingo-d Happens to the best of us ;-)

PR #1932, opened 2020-07-10
CI: use Parallel-Lint

jrfnl:
Parallel-Lint, as the name implies, runs PHP lint in parallel, making it faster than native linting.
It also provides more helpful output with a syntax highlighter code snippet showing the context of syntax errors.
And lastly, it allows for easily running PHP linting over the files in a local dev environment.

So, as we're now doing a full `composer install` on every Travis build anyway, we may as well use this tool.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -169,7 +169,7 @@ install:
 
 script:
   # Lint the PHP files against parse errors.
-  - if [[ "$LINT" == "1" ]]; then if find . -path ./vendor -prune -o -path ./bin -prune -o -name "*.php" -exec php -l {} \; | grep "^[Parse error|Fatal error]"; then exit 1; fi; fi
+  - if [[ "$LINT" == "1" ]]; then composer lint; fi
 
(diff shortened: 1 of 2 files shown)


GaryJones:
Looks fine but job https://travis-ci.com/github/WordPress/WordPress-Coding-Standards/jobs/359617749 needs rerunning

jrfnl:
@gmponos What we had in place originally was also non-fancy and about 10 times faster than running this via PHPCS, so why would we ever even consider that ?

The current solution is 4x faster again and more user-friendly for running locally etc.

jrfnl:
@GaryJones Build is passing now. Was the usual Travis hickups only had to restart the build 3 times...

gmponos:
> was also non-fancy and about 10 times faster

It might be depending the number of files and processes (I use 8 processes).. 
in my projects that have lots of files the total time is shorter

Anyway.. if you count it and it is faster then ok.. 🤷

PR #1934, opened 2020-07-20
3.0: EnqueuedResources: PHPCSUtils implementation

dingo-d:
This PR is a part of the next milestone where we are adding [PHPCSUtils](https://phpcsutils.com/) into existing sniffs.

### Changes

- Added extra tokens to the check, to account for multiline text strings and inline HTML as well
- Fixed the handling of multi-line text strings:
  - As it was, if the text the sniff is looking for was split over multiple lines, the sniff would incorrectly not throw an error (false negative).
  By using the PHPCSUtils [`TextStrings::getCompleteTextString()`](https://phpcsutils.com/phpdoc/classes/PHPCSUtils-Utils-TextStrings.html#method_getCompleteTextString) method, we can get access to the complete content of a multi-line text string and prevent the false negative.
  This does create two minor complications though:
    1. The complete text string may contain multiple link/script tags, so we need to use `preg_match_all()` instead of `preg_match()` to be able to report on each of the matches.
    2. We need to figure out on which line of the multi-line text string the matched content occurred so we can throw the error on the correct line.
    Luckily, as multi-line text strings have only one token per line, that's easy enough to work out, especially if we let `preg_match_all()` remember the offsets at which a match was found, so we also won't confuse one match for another.

Thanks to @jrfnl for doing the majority of the job and walking me through the changes needed to happen and explaining why 😊 .

### DIFF
--- a/WordPress/Sniffs/WP/EnqueuedResourcesSniff.php
+++ b/WordPress/Sniffs/WP/EnqueuedResourcesSniff.php
@@ -9,8 +9,11 @@
 
 namespace WordPressCS\WordPress\Sniffs\WP;
 
-use WordPressCS\WordPress\Sniff;
+use PHP_CodeSniffer\Exceptions\RuntimeException;
 use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Tokens\Collections;
+use PHPCSUtils\Utils\TextStrings;
+use WordPressCS\WordPress\Sniff;
 
 /**
  * Makes sure scripts and styles are enqueued and not explicitly echo'd.
@@ -22,6 +25,8 @@ use PHP_CodeSniffer\Util\Tokens;
  * @since   0.3.0
  * @since   0.12.0 This class now extends the WordPressCS native `Sniff` class.
  * @since   0.13.0 Class name changed: this class is now namespaced.
+ * @since   3.0.0  Added a check for a complete text string in the case where it
+ *                 spans two lines.
  */
 class EnqueuedResourcesSniff extends Sniff {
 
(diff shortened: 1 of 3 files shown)


WordPress/Sniffs/WP/EnqueuedResourcesSniff.php
```diff
@@ -22,6 +25,10 @@
  * @since   0.3.0
  * @since   0.12.0 This class now extends the WordPressCS native `Sniff` class.
  * @since   0.13.0 Class name changed: this class is now namespaced.
+ * @since   3.0.0  Extended tokens array to include inline HTML and
```
jrfnl, line 28: The target token array has actually not been extended, just changed.

The difference is that in the old logic, the sniff would look at the tokens containing the content of heredoc/nowdoc structures - `T_HEREDOC`, `T_NOWDOC` -, while now it looks for the start tokens of those.
Other than that, the same tokens are being sniffed (`T_CONSTANT_ENCAPSED_STRING`, `T_DOUBLE_QUOTED_STRING` and `T_INLINE_HTML`.

As a side-effect, this makes the sniff more efficient as the heredoc/nowdoc start tokens exist a lot less frequently than the content tokens for the same.

Either, that means the changelog entry for that can be removed. It's just an implementation detail.

```suggestion
 * @since   3.0.0  Added a check for a complete text string in the case where it
 *                 spans two lines.
```

As for what's left in the class changelog entry - I don't really think this entry is needed.
The sniff does nothing new, it just does what it was supposed to do _better_.

WordPress/Sniffs/WP/EnqueuedResourcesSniff.php
```diff
@@ -22,6 +25,11 @@
  * @since   0.3.0
  * @since   0.12.0 This class now extends the WordPressCS native `Sniff` class.
  * @since   0.13.0 Class name changed: this class is now namespaced.
+ * @since   3.0.0  Added a check for a complete text string in the case where it
```
jrfnl, line 28: Looks like the update didn't quite work as expected...

PR #1935, opened 2020-07-21
PHP 8 | ControlStructureSpacing: bug fix - $data should be an array

jrfnl:
The `$data` parameter of the `$phpcsFile->add[Fixable]Error|Warning()` methods is expected to be an array, not a string.

Without this fix, PHP 8 will throw an (uncaught) `TypeError: vsprintf(): Argument #2 ($args) must be of type array, string given`.


stingray21:
I'm trying to set phpcs and keep getting this error. When will this be merged into master? It was fixed quite a while ago.

jrfnl:
@stingray21 When we're ready to release version 3.0.0.

kraftbj:
Howdy! Is there anything we can do to help get this shipped out?

jrfnl:
@kraftbj If you can get PHP Core to slow down with the deprecations, so I can focus back on CS work instead of spending all my time on PHP 8.1 fixes, that would be lovely!

PR #1936, opened 2020-07-21
Move "minimum WP version" related utilities to dedicated `MinimumWPVersionTrait`

jrfnl:
### Move "minimum WP version" related utilities to dedicated `MinimumWPVersionTrait`

The "minimum WP version" related utilities are only used by a small set of sniffs, so are better placed in a dedicated trait.

This moves the `$minimum_supported_version` property and the `get_wp_version_from_cl()` method to a new `WordPressCS\WordPress\Helpers\MinimumWPVersionTrait` and starts using that trait in the relevant sniffs.

Related to #1465

### MinimumWPVersionTrait: use PHPCSUtils Helper::getCommandLineData()

The `get_wp_version_from_cl()` method now uses the PHPCSUtils `Helper::getCommandLineData()` method instead of the `Helper::getConfigData()` for more reliable results.

Because of this change, the `$phpcsFile` object is now a required parameter for the `get_wp_version_from_cl()` method.
All calls to the method have been adjusted for this change.

--- [EDIT] New commit --- 

### MinimumWPVersionTrait: rename the `get_wp_version_from_cl()` method

... to `get_wp_version_from_cli()` as per the review suggestion.

Includes adjusting all relevant call to the method.

### DIFF
--- a/WordPress/Helpers/MinimumWPVersionTrait.php
+++ b/WordPress/Helpers/MinimumWPVersionTrait.php
@@ -86,12 +86,12 @@ trait MinimumWPVersionTrait {
 	 * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
 	 */
 	protected function get_wp_version_from_cli( File $phpcsFile ) {
-		$cl_supported_version = trim( Helper::getCommandLineData( $phpcsFile, 'minimum_supported_wp_version' ) );
+		$cli_supported_version = trim( Helper::getCommandLineData( $phpcsFile, 'minimum_supported_wp_version' ) );
 
-		if ( ! empty( $cl_supported_version )
-			&& filter_var( $cl_supported_version, \FILTER_VALIDATE_FLOAT ) !== false
+		if ( ! empty( $cli_supported_version )
+			&& filter_var( $cli_supported_version, \FILTER_VALIDATE_FLOAT ) !== false
 		) {
-			$this->minimum_supported_version = $cl_supported_version;
+			$this->minimum_supported_version = $cli_supported_version;
 		}
 	}
 
--- a/WordPress/Helpers/MinimumWPVersionTrait.php
+++ b/WordPress/Helpers/MinimumWPVersionTrait.php
@@ -1,0 +1,95 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Helpers;
+
+use PHP_CodeSniffer\Files\File;
+use PHPCSUtils\BackCompat\Helper;
+
+/**
+ * Helper utilities for sniffs which take the minimum supported WP version of the
+ * code under examination into account.
+ *
+ * Usage instructions:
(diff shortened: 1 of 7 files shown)


jrfnl:
@GaryJones I don't have a problem with that, though it makes the BC-break slightly bigger in case there are external standards which use this WPCS logic. Then again, they'd have the function calls anyway to add the `$phpcsFile` parameter, so we may as well.

WordPress/Helpers/MinimumWPVersionTrait.php
```diff
@@ -0,0 +1,98 @@
...
+
+use PHP_CodeSniffer\Files\File;
+use PHPCSUtils\BackCompat\Helper;
+
+/**
+ * Helper utilities for sniffs which take the minimum supported WP version of the
+ * code under examination into account.
+ *
+ * Usage instructions:
+ * - Add a trait `use` statement to this trait in the class which intends to use this functionality.
```
dingo-d, line 20: This sentence reads a little odd to me:

```
Add a trait `use` statement to this trait...
```

Shouldn't this be 

```
Add a trait `use` statement in the class which intends to use this functionality.
```

?

WordPress/Helpers/MinimumWPVersionTrait.php
```diff
@@ -0,0 +1,98 @@
...
+	 *
+	 * @since 0.14.0
+	 * @since 3.0.0  - Moved from the Sniff class to this dedicated Trait.
+	 *               - Renamed from `get_wp_version_from_cl()` to `get_wp_version_from_cli()`.
+	 *               - Now requires the $phpcsFile object to be passed in.
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
+	 */
+	protected function get_wp_version_from_cli( File $phpcsFile ) {
+		$cl_supported_version = trim( Helper::getCommandLineData( $phpcsFile, 'minimum_supported_wp_version' ) );
```
dingo-d, line 89: Should we name this `$cli_supported_version`? Just to have consistency with the method name.

WordPress/Helpers/MinimumWPVersionTrait.php
jrfnl: It should probably have read "referring to this trait" or something, but I agree, the text can/should be made clearer.

In reality, we'd normally want:
- An import `use` statement to the trait in the file using it
- A trait `use` statement using the unqualified name in the class using it

Now, how to say that in a clear mar, while still being concise and not presuming that everyone codes like we do...

dingo-d: I'm fine with having this broken down into two separate points. I like to think that people who will use this will know how to import traits into their classes 👀

jrfnl: > I like to think that people who will use this will know how to import traits into their classes

That's exactly why I tried to address it in one sentence... having it broken down into the minutia feels like talking down to people.

dingo-d: In that case, I think just saying that one should use it should suffice. The actual how-to will be known to the person using it.

jrfnl: Updated now to "Add appropriate `use` statement(s) to the file/class which intends to use this functionality." Does that work for you ?

jrfnl:
Gentlemen: just checking - is this PR waiting for something or can it be merged ?

PR #1938, opened 2020-07-21
ArrayIndentation: bug fix - prevent fixer conflict

jrfnl:
If an array item has something else on the same line before it, the indentation should be left alone.

This fixes a potential fixer conflict with the `DisallowInlineTabs` sniff which could occur when the sniffs are being run in a unconventional order.

Discovered when making a change to the Travis file and not reproducible locally, but consistently reproducible with the alternative running order of sniffs on select platforms, as per the [Travis build on Xenial](https://travis-ci.com/github/WordPress/WordPress-Coding-Standards/builds/176411794).


jrfnl:
@dingo-d Oh hell yes ;-)

For the record: that code sample came from another sniff in this codebase and was added in response to an issue, so while it's really ugly code, the sniff should still be able to handle it correctly.

PR #1940, opened 2020-07-22
Travis: change from "trusty" to "xenial"

jrfnl:
As the "trusty" environment is no longer officially supported by Travis, they decided in their wisdom to silently stop updating the PHP "nightly" image, which makes it next to useless as the last image apparently is from January....

This updates the Travis config to:
* Use the `xenial` distro, which at this time is the default.
* Sets the distro for low PHP versions explicitly to `trusty`.
* Makes the expected OS explicit (linux).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-dist: trusty
+os: linux
+dist: xenial
 
 cache:
   apt: true
@@ -11,8 +12,6 @@ cache:
 language: php
 
 php:
-    - 5.4
-    - 5.5
     - 5.6
     - 7.0
     - 7.1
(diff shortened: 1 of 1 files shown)


jrfnl:
> Approving, though it does now show that there are some failures for PHP 8.

@GaryJones Yes, sorry, I should have mentioned that.

1. All the more reason for the switch over as otherwise we wouldn't be notified of those impending failures.
2. I did look at those failures briefly, and while I haven't tested to confirm, I expect these all to be related to the PHP 8 change in how comments are tokenized.
    There is a [PR open](https://github.com/squizlabs/PHP_CodeSniffer/pull/3027) in PHPCS itself to fix the tokenization to be stable across versions. That PR is actually how I discovered the issue with the `trusty` images on Travis ;-)
    I expect the build will pass (for `dev-master`) again once that upstream PR has been merged.

jrfnl:
FYI: the upstream PR for the PHP 8 comment tokenization has been merged, I restarted the last `develop` / `nightly` / PHPCS `dev-master` build and as expected, that build now passes for the comment issues, though it does show there is still another issue to investigate (might be an upstream issue, I'll investigate): https://travis-ci.com/github/WordPress/WordPress-Coding-Standards/jobs/364009318

jrfnl:
Pulled a PR upstream to fix that last issue: https://github.com/squizlabs/PHP_CodeSniffer/pull/3033

PR #1950, opened 2020-09-08
 Travis: retry composer install on failure

jrfnl:
Initially I was only going to pull the first commit, but Composer `1.10.11` has just come out and is breaking builds, so the second commit is a temporary fix for that.

Once this PR has been merged, I will rebase @NielsdeBlaauw's two Docs PRs so we can have a passing build for those.

### Travis: retry composer install on failure

The builds are failing a little too often for my taste on the below error.
```
[Composer\Downloader\TransportException]
Peer fingerprint did not match
```

I'm prefixing the `composer install` commands now with `travis_retry` in an attempt to get round this problem.

Ref:
* https://docs.travis-ci.com/user/common-build-problems/#timeouts-installing-dependencies

### ~~Travis: roll-back Composer to v 1.10.10~~

~~... as there is an issue with Composer 1.10.11, which prevents the `install` from working.~~

~~For more information: composer/composer#9191~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -171,15 +171,15 @@ install:
       else
         composer config preferred-install.squizlabs/php_codesniffer auto
       fi
-    - composer require squizlabs/php_codesniffer:"${PHPCS_BRANCH}" --no-update --no-suggest --no-scripts
+    - travis_retry composer require squizlabs/php_codesniffer:"${PHPCS_BRANCH}" --no-update --no-suggest --no-scripts
     - |
       if [[ $TRAVIS_PHP_VERSION == "nightly" ]]; then
         # PHPUnit 7.x does not allow for installation on PHP 8, so ignore platform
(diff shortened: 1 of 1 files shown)


jrfnl:
Update: the initial Composer issue has been fixed in Composer 1.10.12 / Semver 3.1.0, but there is still a blocking issue, so until that's fixed, I'm leaving the second commit in place.

For more detailed information about the remaining issue, see: https://github.com/composer/semver/issues/109

dingo-d (approved):

jrfnl:
In the mean time Composer `1.10.13` / Semver `3.2.0` have been released which should fix the Composer related issue.

I've retriggered the builds for the two doc PRs from @NielsdeBlaauw . If both pass, I'll remove the second commit from this PR as it is no longer needed.

GaryJones (approved):

PR #1955, opened 2020-10-14
Warn when using slow PHP functions for type casting

kraftbj:
Following up on https://core.trac.wordpress.org/ticket/42918 and https://core.trac.wordpress.org/changeset/49108.

We can warn that type casting should be used instead.


WordPress/Sniffs/PHP/DiscouragedPHPFunctionsSniff.php
```diff
@@ -97,6 +97,16 @@ public function getGroups() {
...
 				),
 			),
+			'typecasting' => array(
+				'type'      => 'warning',
+				'message'   => '%s() can be replaced with type casting.',
+				'functions' => array(
+					'intval',
+					'strval',
+					'floatval',
+					'boolval',
```
jrfnl, line 107: What about `doubleval()` and `settype()` ?

WordPress/Sniffs/PHP/DiscouragedPHPFunctionsSniff.php
```diff
@@ -97,6 +97,16 @@ public function getGroups() {
 					'str_rot13',
 				),
 			),
+			'typecasting' => array(
+				'type'      => 'warning',
+				'message'   => '%s() can be replaced with type casting.',
```
jrfnl, line 102: Should this message be expanded a little to explain why ?

Maybe something along the lines of _"`%s()` should only be used in callbacks. In all other cases, using type casting will be more performant."_

jrfnl (changes requested):
Generally speaking, I like this addition.

Two questions I'd like to pose though (aside from the comments left in-line):
1. Does this belong in this sniff ? The other issues addressed in this sniff are more severe than this.
   This feels more like a "best practice" kind of thing, than something which we should discourage because it can cause problems.
2. Type casting can _not_ be seen as a straight replacement once another operator is in play as it is subject to operator precedence rules.
    This can lead to unexpected results if the developer doing the replacing is not that well-schooled in operator precedence.
    To illustrate: https://3v4l.org/sL8Y6

I'd like to hear your thoughts on this.

kraftbj:
For 1: Agree that it feels like a lesser issue than the current ones. I don't feel confident enough with WPCS (or PHPCS in general) to have a strong opinion on a meta question like this. No objection to breaking it out into a new sniff or move to a different one.

For 2: that's a really good point. What about changing the warning output to highlight your point or to make clearer that there are still valid uses for those functions depending on your need? "intval() is discouraged as slow compared to type casting. Please review this usage and determine if it can be replaced with type casting." or something like that?

WordPress/Sniffs/PHP/DiscouragedPHPFunctionsSniff.php
kraftbj: Added in 2aeb355

WordPress/Sniffs/PHP/DiscouragedPHPFunctionsSniff.php
kraftbj: Updated to this text in 2aeb355

jrfnl:
@kraftbj Eh.. you did give me a good laugh with that new sniff name, but we may need to reconsider it, code of conduct and all.... Try reading it out loud while looking at the spam folder in your email box... :innocent: 

Other than that, thinking iteratively based on the changed PR, I'm thinking that this sniff could be generically useful for more than just the WP community.
If others agree, I'd like to propose we move this sniff to [PHPCSExtra](https://github.com/PHPCSStandards/PHPCSExtra/), either now or in the near future.
_Note: It can't currently be pulled to PHPCSExtra as-is as the abstract sniff for this is not yet in place in PHPCSUtils. But once it is, I think it would be a good sniff for PHPCSExtra_.

Brandon: for context, a lot of the more generically reusable sniff writing utilities are moving to PHPCSUtils in WPCS 3.0.0 and sniffs which are not WP/WPCS specific are moving to PHPCSExtra to give them more visibility.

More notes:
* We're trying to add documentation to all sniffs, so I think it would be great for a new sniff to have the documentation in place straight away. For PHPCSExtra, documentation for each sniff is a requirement. (For WPCS not yet, but will be once all existing sniffs have docs)
* What about adding the sniff to the `CodeAnalysis` category ?
* I think this sniff could in select cases (no operators within the parenthesis) be auto-fixable. Should it be ?

kraftbj:
!!! That name was a completely innocent mistake! One track mind... on code. 😬 Code review is for more than just the code working, eh?

Immediate reaction aside, it is late here so will read the rest and respond after sleep.

kraftbj:
> * We're trying to add documentation to all sniffs, so I think it would be great for a new sniff to have the documentation in place straight away. 

Added in 46183c1 .

> * What about adding the sniff to the `CodeAnalysis` category ?

No objection from me! 83e4c42

> * I think this sniff could in select cases (no operators within the parenthesis) be auto-fixable. Should it be ?

I'm not confident if there are edge cases to that which would cause unexpected behavior or, like a case I saw yesterday `(bool) intval( $var )` where I'm not certain that was intentional in the first place.

WordPress/Docs/CodeAnalysis/PerfomantPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,22 @@
+<documentation title="Prefer Performant PHP">
```
jrfnl, line 1: The file name contains a typo (missing `r`).

WordPress/Sniffs/CodeAnalysis/PerformantPHPFunctionsSniff.php
```diff
@@ -0,0 +1,53 @@
...
+	 *  )
+	 * )
+	 *
+	 * @return array
+	 */
+	public function getGroups() {
+		return array(
+			'typecasting' => array(
+				'type'      => 'warning',
+				'message'   => '%s() should only be used in callbacks. All other cases, using type casting will be more performant.',
```
jrfnl, line 40:
```suggestion
				'message'   => '%s() should only be used in callbacks. In all other cases, using type casting will be more performant.',
```

WordPress/Sniffs/CodeAnalysis/PerformantPHPFunctionsSniff.php
```diff
@@ -0,0 +1,53 @@
...
+				'functions' => array(
+					'intval',
```
jrfnl, line 42: Maybe order this list alphabetically ?

WordPress/Tests/CodeAnalysis/PerformantPHPFunctionsUnitTest.inc
```diff
@@ -0,0 +1,10 @@
+<?php
+
+// These calls are slow and type casting is more performant. Warning.
+intval( '10' );
+strval( 100 );
+floatval( '100' );
+boolval( 1 );
+doubleval( '100' );
+$var = 100;
```
jrfnl, line 9: This line is redundant and adds no value to the tests.

Other than that, while it doesn't matter for the actual sniff, the test code does not reflect how these functions are commonly used (except for the last one). Casting a hard-coded value is redundant anyway, in that case, the hard-coded value should be declared with the correct type. Casting/use of these functions will generally only be done on variables or expressions.

WordPress/Sniffs/CodeAnalysis/PerformantPHPFunctionsSniff.php
```diff
@@ -0,0 +1,53 @@
...
+use WordPressCS\WordPress\AbstractFunctionRestrictionsSniff;
+
+/**
+ * Encourages the use of more performant PHP.
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since 3.0.0
+ */
+class PerformantPHPFunctionsSniff extends AbstractFunctionRestrictionsSniff {
```
jrfnl, line 21: This name seems to imply that more groups will be added at a later point. Do you have any other ones mind ?

Also, the suggestion is to *not* use these functions, but use more performant alternatives, while the name of the sniff gives the impression that the functions would be more performant.

Altogether, might be good to have another think about the name.

And yes, naming things is hard ;-)

WordPress-Extra/ruleset.xml
```diff
@@ -87,6 +87,7 @@
 
 	<rule ref="WordPress.PHP.DevelopmentFunctions"/>
 	<rule ref="WordPress.PHP.DiscouragedPHPFunctions"/>
+	<rule ref="WordPress.CodeAnalysis.PerformantPHPFunctions"/>
```
jrfnl, line 90: Move this up two lines for alphabetical order or have it in it's own section with documentation on what this does and the links like everywhere else.

WordPress/Docs/CodeAnalysis/PerfomantPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,22 @@
...
+    <standard>
+        <![CDATA[
+    Various PHP functions are legacy and less performant than other language constructs.
```
jrfnl, line 4: These functions aren't necessarily legacy, but should basically only be used in callbacks.

The "standard" description as you have it now is also more generic than what the sniff actually does.

WordPress/Docs/CodeAnalysis/PerfomantPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,22 @@
...
+    This sniff will help identify use of functions that may be better served using other options.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using native type casting.">
+            <![CDATA[
+$var = '100'; // string.
```
jrfnl, line 11: For a more realistic/more relatable example, you may want to use a function in the example which has multiple return types. (in both code samples)

```php
$var = fn_returning_int_or_false();
```

jrfnl (commented):
Hi Brandon, looking good!

The [second point of my original feedback](https://github.com/WordPress/WordPress-Coding-Standards/pull/1955#pullrequestreview-508850289), however, has not been addressed yet. And without addressing it, this will be a dangerous sniff as inexperienced developers will introduce bugs by blindly following the sniff advise.

Other than that, I've left a few more remarks, more dotting the i's and crossing the t's than anything structurally wrong.

jrfnl:
While it would be nice to have a sniff for this, the PR as-is is not mergable with too many open review comments and no action taken on them in over two years, so I'm going to close this PR.

If at some point a new attempt at a sniff for this is made, it should be pulled to PHPCSExtra.

Issue #1915, opened 2020-07-02
Remove problematic words from the code


dingo-d:
As discussed in #1908 we should scan the code for any problematic and non-inclusive words (`whitelist/blacklist` for instance).

The principles discussed in https://github.com/PHPCSStandards/PHPCSExtra/issues/59 should be taken into the account. I think that sniff will be useful if included in standards for this repo 🙂 

Also, a thing to look out for is pronouns in comments. I know I've done that many times (haven't checked this repo in details for this) - using `he` in comments. Like:

```php
/**
 * This code will force a user to do this.
 * He will then be redirected.
 */
```

Instead, a more inclusive version can be used

```php
/**
 * This code will force a user to do this.
 * They will then be redirected.
 */
```

A small change that doesn't hurt anybody, but makes it a bit more inclusive.

Note to anybody who wants to point out that this is blown out of proportion, please, don't even try. While the offense is taken, not given, using non-inclusive words means creating a non-inclusive environment, and inclusivity brings diversity and a happier space to work in. And this is what we should always strive for.

jrfnl:
I've just added this issue to the 3.0 milestone.

I'm pretty sure there are some public properties which people can set from a custom `[.]phpcs.xml[.dist]` which uses "whitelist"/"blacklist" terminology.

Changing the names of those properties will be a BC-break for end-users, so should be done in a major version (which we just happen to be working on right now).

jrfnl:
This was largely addressed by #2065, though a cursory word search for other potentially problematic words should still be done before the issue can be closed.

jrfnl:
Anyone still up for a cursory search ? If not, I suggest we close this issue as fixed. Follow-up issues can be opened if other problematic terminology would be encountered in the future.

GaryJones:
Using https://github.com/get-woke/woke, it detected:

<details>
<summary>Expand for raw results</summary>

```
WordPress/Helpers/IsUnitTestTrait.php:62:51-60: `whitelist` may be insensitive, use `allowlist`, `inclusion list` instead (warning)
	 *               Renamed from `$custom_test_class_whitelist` to `$custom_test_classes`.
	                                                  ^
WordPress/Helpers/IsUnitTestTrait.php:76:46-55: `whitelist` may be insensitive, use `allowlist`, `inclusion list` instead (warning)
	 *               - Renamed from `$test_class_whitelist` to `$known_test_classes`.
	                                             ^
[...]

Issue #1877, opened 2020-04-01
WordPress CS 3.0.0 planning & roadmap


jrfnl:
Work on WordPressCS 3.0.0 will start in the near future. @GaryJones, @dingo-d and me have recently had a call to discuss what we envision for WPCS 3.0.0.

This issue is intended to keep track of what needs to be done, as well as give openness about the changes envisioned for WPCS 3.0.0.

Milestone: https://github.com/WordPress/WordPress-Coding-Standards/milestone/25

Planning:
* Ready when it's ready.
* Most of the initial PRs are tentatively expected in the ~~second half of April/beginning of May (2020)~~ July 2020.
* There will be at least one release candidate before the actual release of WPCS 3.0.0.

Environment changes:
* The minimum PHPCS requirement will go up to - at least - PHPCS 3.5.0.
* WPCS will get two (three) new dependencies:
    - [PHPCSUtils](https://github.com/PHPCSStandards/PHPCSUtils) - a set of utility functions and other helpers for sniff writers
    - [PHPCSExtra](https://github.com/PHPCSStandards/PHPCSExtra) - a set of generic sniffs
    - The [DealerDirect Composer PHPCS plugin](https://github.com/Dealerdirect/phpcodesniffer-composer-installer) - this is a dependency of the above two packages, so we will automatically benefit from it.
* The installation instructions will change.
    - A Composer based project/global install will be the recommended and only supported manner of installation.
    - Stand-alone installation from a downloaded zip file, using a git clone or using `composer create-project` will still be possible, but no support will be given for those installations.

Note: PHPCSUtils and PHPCSExtra both are on 1.0.0-alphaX releases at the time of writing this.
The first stable `1.0.0` for both will be tagged before the first WPCS 3.0.0 RC will be (pre-)released.

Roadmap / what needs to be done:
- [x] Update the environment files (`composer.json`, `.travis.yml` etc) to allow the changes to be made. - PR #1905
- [x] Remove all previously deprecated code - #1583, #1819 - PRs #1908, #1909
- [x] Remove all sniffs which can be replaced by upstream sniffs now the minimum PHPCS version has been upped (if any) and add the PHPCS native sniff to the ruleset(s) instead. - PR #1910
- [x] Remove all sniffs which can be replaced by sniffs in the PHPCSExtra standard and add the PHPCSExtra sniff to the ruleset(s) instead. - PR #1911, #1912, #1919, #1920 (more to come)
[...]

PR #1960, opened 2020-11-20
3.0: Move "is unit test" utility functions to dedicated trait

jrfnl:
This PR will be easiest to review by looking through the individual commits.

Related to #1877

### :heavy_plus_sign:  Sniff: remove the `is_token_in_test_method()` method :new: 

This method is currently unused in WPCS and it is not 100% clear what the method _should_ do, so removing the method for now.

If needs be it can be brought back later if/when there is more clarity on what the method is supposed to solve.

### Move "is this test code" related utilities to dedicated `IsUnitTestTrait`

The "is this test code" related utilities are only used by a small set of sniffs, so are better placed in a dedicated trait.

This moves the (`public`) `$custom_test_class_whitelist` and the `$custom_test_class_whitelist` properties as well as ~~the `is_token_in_test_method()` and~~ the `is_test_class` method to a new `WordPressCS\WordPress\Helpers\IsUnitTestTrait` and starts using that trait in the relevant sniffs.

The trait has been made stand-alone, i.e. it does not presume that a sniff implementing it extends the `WordPressCS\WordPress\Sniff` class.
This makes it easier to (re-)use the traits, even when a sniff extends an abstract sniff class from PHPCS itself or from PHPCSUtils.

~~Note: the `is_token_in_test_method()` method is currently unused in WPCS, but seems useful enough to keep in place.~~

Related to #1465

### IsUnitTestTrait::is_test_class(): use PHPCSUtils

This implements PHPCSUtils alternatives for PHPCS native methods and other PHPCSUtils utilities in the `IsUnitTestTrait::is_test_class()` code.

The functionality is unchanged (just less buggy).

### IsUnitTestTrait: rename protected property

The protected property `$test_class_whitelist` has been renamed to `$known_test_classes` as a step towards removing whitelist/blacklist terminology.

### IsUnitTestTrait: rename public property

The public property `$custom_test_class_whitelist` has been renamed to `$custom_test_classes` as a step towards removing whitelist/blacklist terminology.

Includes adjusting the property name in all tests which included the property.

Related to #1915

#### To do before release:
- [ ] Annotate this change in the wiki custom properties list

### IsUnitTestTrait: rename local variable

The function local variable `$whitelist` has been renamed to `$known_test_classes` as a step towards removing whitelist/blacklist terminology.

### IsUnitTestTrait: add missing WP Core test case classes

Add two more test cases which were added to `/tests/phpunit/includes/`.

Note: WP Core contains more test cases, but those are in the `tests/phpunit/tests` directory, while the test cases in  `/tests/phpunit/includes/` are supposed to be the "public test API", which is why those are the only ones in this list.

### IsUnitTestTrait: improve documentation

### DIFF
--- a/WordPress/Helpers/IsUnitTestTrait.php
+++ b/WordPress/Helpers/IsUnitTestTrait.php
@@ -19,6 +19,13 @@ use WordPressCS\WordPress\Sniff as WPCS_Sniff;
  * Helper utilities for sniffs which need to take into account whether the
  * code under examination is unit test code or not.
  *
+ * Usage instructions:
+ * - Add appropriate `use` statement(s) to the file/class which intends to use this functionality.
+ * - Now the sniff will automatically support the public `custom_test_classes` property which
+ *   users can set in their custom ruleset. Do not add the property to the sniff!
+ * - The sniff can call the methods in this trait to verify if certain code was found within
+ *   a test method or is a test class and will take the custom property into account.
+ *
  * @package WPCS\WordPressCodingStandards
  * @since   3.0.0 The properties and method in this trait were previously contained in the
  *                `WordPressCS\WordPress\Sniff` class and have been moved here.
@@ -28,13 +35,15 @@ trait IsUnitTestTrait {
 	/**
 	 * Custom list of classes which test classes can extend.
 	 *
-	 * This property allows end-users to add to the $known_test_classes via their ruleset.
-	 * This property will need to be set for each sniff which uses the
-	 * `is_test_class()` method.
-	 * Currently the method is used by the `WordPress.WP.GlobalVariablesOverride`,
+	 * This property allows end-users to add to the build-in `$known_test_classes`
+	 * via their custom PHPCS ruleset.
+	 * This property will need to be set for each sniff which uses this trait.
+	 *
+	 * Currently this property is used by the `WordPress.WP.GlobalVariablesOverride`,
 	 * `WordPress.NamingConventions.PrefixAllGlobals` and the `WordPress.Files.Filename` sniffs.
 	 *
 	 * Example usage:
+	 * ```xml
 	 * <rule ref="WordPress.[Subset].[Sniffname]">
 	 *  <properties>
 	 *   <property name="custom_test_classes" type="array">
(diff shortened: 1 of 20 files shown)


dingo-d:
> For example: for a token in a function nested within a test method, the method would previously return true and strictly speaking this is correct when the question is "is the token in a test method ?".

> However, this answer is incorrect if the question would be "is this a token which is unit test code ?" as the nested function would be in the global scope and not in the scope of the test class.

It seems to me that, in both cases, the current status of what this helper does is not truthful. 

If we want to cover both cases, a new helper would need to be created which will account for the scope of the unit test, and a check should be added in the current helper that would exclude the case if it's in a unit case, but just cover the part where we check if the token is in a test method.

If I'm not mistaken, this could be useful if anonymous functions or classes are used in a method within a test class. Don't ask me how I know of this case, let's just say that integration tests in WordPress require some level of imagination 😂

jrfnl:
> It seems to me that, in both cases, the current status of what this helper does is not truthful.

Well, as the method is not used in WPCS at this time, this discussion is theoretical.

Maybe we should just remove the method completely until such time that WPCS needs it again.

More than anything, it would be better for custom rulesets to exclude the test directories completely for these sniffs. This can't be done from within WPCS as we don't know what the name of the test directory is for individual projects.
These methods are only a guestimation to try and bow out before reporting.

GaryJones:
As this is going to be for a major release, I'm not against the behaviour here being clarified in a particular way, even if there is no extra helper method to help answer the other question.

jrfnl:
@dingo-d @GaryJones What can I do to move this PR forward ? I'm happy to adjust it, but at this moment there doesn't seem to be consensus on if and how it should be adjusted. Would it be possible to continue the above discussion ?

dingo-d:
I'm fine with removing it, especially since it's not being used at the moment. We can easily return it back (it's in the commits).

GaryJones:
I'm in agreement here.

jrfnl:
Ok, in that case, the method is hereby gone. I've added a new commit at the start of the stack and removed the "improve the method" commit. The changes have been annotated in the above PR description.

dingo-d (approved):

jrfnl:
Rebased now the GH actions PR has been merged. Would appreciate a merge if the build passes,

PR #1983, opened 2021-04-07
GH Actions: report CS violations in the PR

jrfnl:
### GH Actions: report CS violations in the PR

The cs2pr tool will allow to display the results from an action run in checkstyle format in-line in the PR code view, which should improve usability of the workflow results.

Ref: https://github.com/staabm/annotate-pull-request-from-checkstyle

### GH Actions: don't use cs2pr in quicktest

... as the `quicktest` workflow is run on `push` events, not on PRs, so displaying results in a PR is not relevant (and is done via the `test` workflow which _is_ run on `pull` events anyhow).

### DIFF
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -55,7 +55,6 @@ jobs:
           php-version: ${{ matrix.php }}
           ini-values: ${{ steps.set_ini.outputs.PHP_INI }}
           coverage: none
-          tools: cs2pr
 
       - name: Set PHPCS version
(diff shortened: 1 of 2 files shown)


dingo-d (approved):

GaryJones:
I think this now needs rebasing to have the tests run with the PHP 8.1 fixes merged.

jrfnl:
Well, these changes didn't affect the test runs on PHP 8.1, but even so, it will be nice to have the green checkmark, so rebased ;-)

PR #1997, opened 2021-06-14
Bug report template: various improvements

jrfnl:
After the umpteenth time of getting a non-reproducible bug report (not just in this repo, mind you), I figured it may help to add some more guidance to the bug report template.

Most important changes:
* Ask for the actual CLI command people have used.
* Ask for the custom ruleset used (if used).
* Give some guidance on how to find out the version numbers of the tooling used.

PR #1998, opened 2021-06-14
CS/QA: get rid of (most) warnings

jrfnl:
Though it is by the looks of it only a "beta" feature for the moment, I noticed in PR #1997 that warnings are now being shown for files unrelated to the current PR. To me this is just plain annoying as it is noise which distracts from the actual changes which need reviewing and can be confusing for contributors (should I fix these ? will my changes be accepted even though these warnings are there?).

image

With that in mind, I figured I might as well try to reduce this noise some more as well as fix up some other small issues.

###  CS: get rid of "commented out code" warnings

... which show through in PRs, even when the code in those files has not been adjusted in the PR.

There's only one such warning remaining now, which is _actual_ commented out code.

### DIFF
--- a/WordPress/AbstractArrayAssignmentRestrictionsSniff.php
+++ b/WordPress/AbstractArrayAssignmentRestrictionsSniff.php
@@ -153,9 +153,9 @@ abstract class AbstractArrayAssignmentRestrictionsSniff extends Sniff {
 		$inst = array();
 
 		/*
-		 * Covers:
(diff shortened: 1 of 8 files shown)


dingo-d (approved):

dingo-d:
The GH Actions warnings are just warnings, so I think they can be fixed later on (I usually ignore those if they don't have anything to do with the PR I'm working on and fix those later on).

GaryJones (approved):

PR #1999, opened 2021-06-22
GH Actions: turn display_errors on

jrfnl:
Turns out the default setting for `error_reporting` used by the SetupPHP action is `error_reporting=E_ALL & ~E_DEPRECATED & ~E_STRICT` and `display_errors` is set to `Off`.

For the purposes of CI, I'd recommend running with `E_ALL` and `display_errors=On` to ensure **all** PHP notices are shown.

In this script, error_reporting was already enabled, but the error display was not yet fixed. Sorted now.

### DIFF
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -44,9 +44,9 @@ jobs:
         id: set_ini
         run: |
           if [ "${{ matrix.phpcs_version }}" != "dev-master" ]; then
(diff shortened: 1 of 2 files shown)


jrfnl:
Note: the current build failure against PHP 8.1 was expected - this is an upstream issue with the PHP Parallel Lint package and I've already pulled a fix there: https://github.com/php-parallel-lint/PHP-Parallel-Lint/pull/64

dingo-d (approved):

GaryJones (approved):

PR #2023, opened 2021-12-30
README.md: Update the minimum supported PHP version

slaFFik:
In the readme.md file it is stated that PHP 5.2 is the minimum supported PHP version, while it's already PHP 5.6 for quite some time and technically WP does not work on PHP 5.2-5.5.


jrfnl (approved):
@slaFFik Thanks for this PR.

We held off on doing this update previously as a lot of plugins/themes using WPCS support multiple WP versions and would therefore still need to support PHP 5.2.

Having said that. as WP 5.2 (which changed the minimum supported version) came out some twenty months ago and there have been five (nearly six) WP majors since, I agree, this change can now be made.

dingo-d (approved):

PR #2024, opened 2022-01-18
Issue/trim error

kpverint:
PHP_CodeSniffer returns null which results in errors.


jrfnl:
@kpverint Thank you for trying to make WPCS better.

I think I see what you are trying to do here. Let's first define the problem you are trying to solve:

>  When running WPCS on PHP 8.1, you are seeing deprecation notices for passing `null` as a parameter to a function where that parameter is not nullable.

Correct ?

If you look at the _current_ state of the `develop` branch, you will see that those issues have already been fixed in #1984 and #1985.

If looks like you didn't update your dev clone of WPCS before creating this PR, which also explains the file conflicts which GH shows at the bottom of this page.

Suggest: close.

kpverint:
That's right. Thanks! :) Closing.

WraithKenny:
Any news when you will actually publish the fixes?

jrfnl:
When the release is ready.

WraithKenny:
Cool, I'll just keep downgrading my computer's PHP version since 11 months wasn't enough time.

jrfnl:
@WraithKenny Or you could start contributing instead of bashing people who actually do the work

WraithKenny:
> @WraithKenny Or you could start contributing instead of bashing people who actually do the work

It's 2 days from being fixed, but unpublished, for exactly one year. 

"...people who actually do the work." Where? What work?

PR #2026, opened 2022-02-04
WP/CronInterval: bug fix for parentheses

### LINKED ISSUE #2025: WordPress.WP.CronInterval.ChangeDetected can't handle intervals with parentheses
## Bug Description

The WordPress.WP.CronInterval.ChangeDetected sniff will show a `Detected changing of cron_schedules, but could not detect the interval value.` warning when the interval is surrounded with parentheses.

## Minimal Code Snippet
```bash
phpcs ./myfile.php
```
``` php
// myfile.php
<?php

function add_cron_schedule( $schedules ) {
		$schedules['my-schedule'] = [
			'interval' => ( 15 * MINUTE_IN_SECONDS ), // <-- works without the "()"
			'display'  => esc_html__( 'Every fifteen minutes' ),
		];

		return $schedules;
}

add_filter( 'cron_schedules', 'add_cron_schedule' );
```

## Error Code

N/A

## Custom ruleset

I used the ruleset of this repository. 

## Environment

| PHP version             | 7.2.22
| WPCS version            | `develop` (2.3.0)
| WPCS install type       | Composer project local and git clone (haven't tested others)

While debugging this I found that the parser does look for parentheses, but only parses the opening and not the closing.
After [this loop](https://github.com/WordPress/WordPress-Coding-Standards/blob/73f3d69b3a200935b80f5fd265bfbee8d642d082/WordPress/Sniffs/WP/CronIntervalSniff.php#L180) the `$value` is `"(15*MINUTE_IN_SECONDS"`

jrfnl:
This allows for a time calculation to be wrapped in parentheses.

Fixes #2025

---

Note: just looking at the code, I can see multiple additional improvements which could/should be made:
* Only search for close parenthesis as a "valueEnd" if the `$valueStart` token is an open parenthesis.
* Allow for more assignment operators - the new schedule could be build up bit by bit and only assigned to the array later.

These additional things are not addressed here. This just addresses the immediate issue.

### DIFF
--- a/WordPress/Sniffs/WP/CronIntervalSniff.php
+++ b/WordPress/Sniffs/WP/CronIntervalSniff.php
@@ -177,10 +177,15 @@ class CronIntervalSniff extends Sniff {
 					$valueStart = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $operator + 1 ), null, true, null, true );
 					$valueEnd   = $this->phpcsFile->findNext( array( \T_COMMA, \T_CLOSE_PARENTHESIS ), ( $valueStart + 1 ) );
 					$value      = '';
-					for ( $j = $valueStart; $j < $valueEnd; $j++ ) {
+					for ( $j = $valueStart; $j <= $valueEnd; $j++ ) {
 						if ( isset( Tokens::$emptyTokens[ $this->tokens[ $j ]['code'] ] ) ) {
 							continue;
 						}
+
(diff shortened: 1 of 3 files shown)


diedexx (approved):
Code changes look good. I've also tested it and it fixes the issue 👍

GaryJones (approved):
✅

dingo-d (approved):

PR #2034, opened 2022-03-19
GH Actions/CS: fix build failure

jrfnl:
Fix failing installation of the xmllint tooling.

### DIFF
--- a/.github/workflows/ruleset-checks-sniffs.yml
+++ b/.github/workflows/ruleset-checks-sniffs.yml
@@ -54,7 +54,9 @@ jobs:
         uses: ramsey/composer-install@v2
 
       - name: Install xmllint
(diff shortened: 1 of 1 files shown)


dingo-d (approved):
I thought that was failing because GH was acting up so actions couldn't fetch the lib

jrfnl:
> I thought that was failing because GH was acting up so actions couldn't fetch the lib

It sort of is... the URL from which GH tries to download returns a 404, so the image GH uses contains a reference to an outdated URL. Adding the `apt-get update` fixes that.

GH will probably fix the Ubuntu image soon enough, but it's annoying when progress gets blocked by something like this, so hopefully this will sort it out for now and later

GaryJones (approved):

Issue #2048, opened 2022-05-06
Up the minimum version of PHPCS to 3.7.0 as soon as it is released


jrfnl:
## Introduction

The last three PHP releases, PHP 7.4, 8.0 and 8.1, introduced more syntax changes and new keywords in PHP than basically all PHP versions between PHP 5.0 to PHP 7.3 combined.

The work on getting all of these (tokenizer) changes accounted for and backfilled in PHPCS to allow for running PHPCS on older PHP versions, but still get the same token stream, is still on-going.

Along the same lines, the work of adding support for these new syntaxes and adding utility methods to PHPCSUtils to allow supporting these new syntaxes on older PHPCS versions as well is also still on-going.

### PHPCS Runtime compatibility

In this context, runtime compatibility means that PHPCS will _run_ on a PHP version without fatal errors or deprecation notices.

PHP | PHPCS
--- | ---
7.4 | 3.4.2 and higher
8.0 | 3.5.7 and higher
8.1 | 3.6.2 and higher

### Syntax support

In this context, syntax support means that the Tokenizer in PHPCS supports all new syntaxes of a new PHP version. It does not imply that all sniffs have yet been adjusted to allow for these new syntaxes correctly, nor that all utility functions included in PHPCS itself support the new syntaxes yet.

There may also still be minor bugs in the Tokenizer support for these new syntaxes. The PHPCS version listed is the first version in which all new syntaxes are supported and **_major_** known bugs have been resolved.

PHP | PHPCS
--- | ---
7.4 | 3.5.6 and higher
8.0 | 3.6.1 and higher
8.1 | 3.7.0 and higher, though this may still be partial support (expected)

## The problem

For sniffs to stay accurate, they need to be updated to take these new syntaxes into account.

For example, PHP 7.4 introduced arrow functions.
This means that all sniffs looking explicitly for function declarations - for instance to verify the formatting of these - will need to be updated to now also include arrow functions.

At the same time, sniffs which need to _skip over_ function declarations to prevent false positives and such, will now also need to _skip over_ arrow functions.

Arrow functions are supported in PHPCS as of PHPCS 3.5.3 (very buggy), but realistically PHPCS 3.5.6 is needed as by that time the worst bugs in the Tokenizer arrow function support were resolved.

While PHPCSUtils can - up to a point - provide utilities to backport support for these kind of new syntaxes, there are a couple of limitations:
[...]

PR #2049, opened 2022-05-15
Composer: up the minimum PHPCS version to 3.6.2

jrfnl:
While the minimum PHPCS version has already been upped to PHPCS 3.5.0 previously for the WPCS 3.0.0 branch, I'm proposing to up it yet further to the latest PHPCS version available.

PHP 8.0 and 8.1 introduced a lot of new syntaxes and, while a lot of this can be supported for older PHPCS versions via PHPCSUtils, this will always impact performance and precision, so it is strongly advisable to use the latest PHPCS version as the new minimum if we want to support the newly available PHP syntaxes.

For now, I'm upping the minimum version to PHPCS 3.6.2, which is the latest PHPCS release. This will unblock PRs to add support for PHP 8.0 features to WPCS.
Once PHPCS 3.7.0 has been released, the minimum PHPCS version should be upped again to unblock adding support for PHP 8.1 features to WPCS.

Also see #2048

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -36,7 +36,7 @@ When you introduce new `public` sniff properties, or your sniff extends a class 
 
 ## Pre-requisites
 * WordPress-Coding-Standards
-* PHP_CodeSniffer 3.3.1 or higher
+* PHP_CodeSniffer 3.6.2 or higher
 * PHPUnit 4.x, 5.x, 6.x or 7.x
 
 The WordPress Coding Standards use the `PHP_CodeSniffer` native unit test suite for unit testing the sniffs.
--- a/.github/ISSUE_TEMPLATE/dependency-change.md
+++ b/.github/ISSUE_TEMPLATE/dependency-change.md
@@ -4,7 +4,7 @@ about: A reminder to take action when a WPCS dependency changes
 
 ---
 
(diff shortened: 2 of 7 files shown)


dingo-d (approved):

jrfnl:
Note: the "required statuses" for the protected `develop` branch will need to be updated for this PR to become mergable.

GaryJones (approved):

dingo-d:
@jrfnl 

> Note: the "required statuses" for the protected develop branch will need to be updated for this PR to become mergable.

Do we need to remove some, is this just going to be temporary or?

jrfnl:
> @jrfnl
> 
> > Note: the "required statuses" for the protected develop branch will need to be updated for this PR to become mergable.
> 
> Do we need to remove some, is this just going to be temporary or?

We need to remove the `PHPCS 3.5.0` builds and add the `PHPCS 3.6.2` builds (and same again when we update to PHCS 3.7.0).

For now, let's just update the required statuses for the `develop` branch and leave the `master` branch for later once everything has settled.

jrfnl:
While updating the required statuses, I realized that the "ruleset check" also contains a PHPCS version reference, so I've updated that workflow now as well.

As a side-note: I kind of have a feeling the test files used for that are sorely out of date, so we may need to discuss what to do with that check and whether we should keep it at all.

PR #2050, opened 2022-05-15
IsUnitTestTrait: update test class list

jrfnl:
Update the test class list based on changes in WP Core:
* The `Block_Supported_Styles_Test` was never a `TestCase`, but was a concrete test placed in the wrong directory.
* The `PHPUnit_Adapter_TestCase` class was added to allow for PHPUnit cross-version support.

Other than that, the list is now alphabetized within each category and the categories are annotated with comments.

### DIFF
--- a/WordPress/Helpers/IsUnitTestTrait.php
+++ b/WordPress/Helpers/IsUnitTestTrait.php
@@ -79,16 +79,21 @@ trait IsUnitTestTrait {
 	 * @var string[]
 	 */
 	protected $known_test_classes = array(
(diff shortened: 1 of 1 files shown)


dingo-d (approved):

GaryJones (approved):

jrfnl:
@GaryJones Just checking - any particular reason why you didn't merge this after approving (as the second approver) ? Same for the other open PR.

jrfnl:
FYI: As I've now changed the required statuses, I will rebase this PR (and the other open one) without changes, so the required statuses pass again.

PR #2053, opened 2022-05-16
Add Code of Conduct

jrfnl:
Based on the [Contributor Covenant](https://www.contributor-covenant.org) version 2.0.

As there is no email address available for the project, I'm proposing to use the WP Core Slack channel #core-coding-standards as the official channel to report violations. Depending on the violation, details can be shared in private via DMs if needed to protect the privacy of the parties involved.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,128 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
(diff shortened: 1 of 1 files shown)


dingo-d (approved):

jeffpaul:
As noted in https://make.wordpress.org/project/2022/04/29/discussion-contrib-handbook-part-2/, it sounds like a project Code of Conduct is coming.  Perhaps it would be better to inherit from that (though I admit that 10up uses [Contributor Covenant](https://www.contributor-covenant.org/) for our projects as well)?

jeffpaul:
@jrfnl I can see the benefit of getting something in place and we can iterate once the larger project has formalized something to inherit to all downstream teams and components.

jeffpaul:
I'm not able to add myself as a review and note my approval of the PR, so leaving a comment to that effect: ✅ thanks for adding this, always good to allow people to know how we expect them to behave.

GaryJones (approved):

PR #2061, opened 2022-06-17
Helper classes: explicitly state there is no BC promise

jrfnl:
For the `Helper` classes which do not contain WordPress specific code and for which it is likely that PHPCSUtils may at some point replace the functionality, make it very clear that those classes are not covered by a promise of backward-compatibility.

By doing this, we buy ourselves the freedom to remove methods from these classes/the classes completely if/when PHPCSUtils would offer similar functionality, without the need to create a new major release of WPCS when we do.


dingo-d (approved):

GaryJones (approved):
✅

PR #2062, opened 2022-06-17
3.0: Move "variable" related utilities to dedicated VariableHelper + use PHPCSUtils + handle match

jrfnl:
### Move "array access keys" related utilities to dedicated VariableHelper

The "array access keys" related utilities are only used by a small set of sniffs, so are better placed in a dedicated class.

This commit moves the `get_array_access_keys()` method and the `get_array_access_key()` method to a new `WordPressCS\WordPress\Helpers\VariableHelper` and starts using that class in the relevant sniffs.

In contrast to some of the other "move methods out of the Sniff class" PRs, these methods have been moved to a class and made `static` - instead of moved to a `trait`.
The reason for this difference is that the methods in other "moves" are setting properties which the sniff classes would need access to, while these methods are 100% stand-alone.

**Note**:
It is expected for PHPCSUtils to have dedicated methods for the same at some point in the future. If/when those methods become available, it is recommended for the sniffs to start using the PHPCSUtils methods. With this in mind, this class has been marked as "internal" without BC promise.

Related to #1465

### VariableHelper::get_array_access_keys() use PHPCSUtils

Use the PHPCSUtils `GetTokensAsString::compact()` method to retrieve array access keys without extraneous whitespace or comments.

### Move "is comparison" related utility to dedicated VariableHelper

The "is comparison" related utility method is only used by a small set of sniffs, so is better placed in a dedicated class.

As this method also expects a `T_VARIABLE` token as input, the `VariableHelper` class seems appropriate.

This commit moves the `is_comparison()` method to the new `WordPressCS\WordPress\Helpers\VariableHelper` and starts using that class in the relevant sniffs.

**Note**:
It is expected for PHPCSUtils to have dedicated methods for the same at some point in the future. If/when those methods become available, it is recommended for the sniffs to start using the PHPCSUtils methods.

Related to #1465

### VariableHelper::is_comparison(): use PHPCSUtils

... to get the potential parentheses opener.

### PHP 8.0 | VariableHelper::is_comparison(): add support for variables being compared in a `match` expression

PHP 8.0 introduced `match` expressions, which do a (strict) comparison on the value in the condition against the options in the `match` body.
Ref: https://www.php.net/manual/en/control-structures.match.php

This commit updates the `is_comparison()` method to allow for variables in the "condition" part of a `match` expression to be considered part of a comparison.

Tested via the `ValidatedSanitizedInput` sniff.

---

N.B. While reviewing the "array access keys" methods, I noticed that the methods do not account for array assignments without a key, i.e. `$var[]`. This may need improved handling at some point, but is outside the scope of this PR.

### DIFF
--- a/WordPress/Helpers/VariableHelper.php
+++ b/WordPress/Helpers/VariableHelper.php
@@ -11,6 +11,7 @@ namespace WordPressCS\WordPress\Helpers;
 
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Utils\GetTokensAsString;
 
 /**
  * Helper utilities for working with variables representing arrays.
@@ -74,9 +75,11 @@ final class VariableHelper {
 				break;
 			}
 
-			$key = $phpcsFile->getTokensAsString(
+			$key = GetTokensAsString::compact(
+				$phpcsFile,
 				( $open_bracket + 1 ),
-				( $tokens[ $open_bracket ]['bracket_closer'] - $open_bracket - 1 )
+				( $tokens[ $open_bracket ]['bracket_closer'] - 1 ),
+				true
 			);
 
 			$keys[]  = trim( $key );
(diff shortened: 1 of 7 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2063, opened 2022-06-17
DB/DirectDatabaseQuery: remove custom ignore annotation support

jrfnl:
Follow up on #1583 / PR #1908 which removed the `Sniff::has_whitelist_comment()`.

The `WordPress.DB.DirectDatabaseQuery` had its own - even more custom - ignore annotation mechanism.

This commit removes that mechanism and updates (and cleans up) the unit tests.


WordPress/Sniffs/DB/DirectDatabaseQuerySniff.php
```diff
@@ -163,22 +159,15 @@ public function process_token( $stackPtr ) {
 			}
 		}
 
-		// Flag instance if not whitelisted.
-		if ( ! $whitelisted_db_call ) {
-			$this->phpcsFile->addWarning( 'Usage of a direct database call is discouraged.', $stackPtr, 'DirectQuery' );
-		}
+		$this->phpcsFile->addWarning( 'Usage of a direct database call is discouraged.', $stackPtr, 'DirectQuery' );
```
GaryJones, line 162:
```suggestion
		$this->phpcsFile->addWarning( 'Use of a direct database call is discouraged.', $stackPtr, 'DirectQuery' );
```

GaryJones (approved):
One minor suggestion for simplified language, but approving anyway as not a blocker.

WordPress/Sniffs/DB/DirectDatabaseQuerySniff.php
jrfnl: Out of scope of this PR, but I've updated it anyway.

Issue #1917, opened 2020-07-06
Update wiki for changes in 3.0.0


jrfnl:
As discussed in #1908, the wiki pages should be reviewed and references to "whitelisting" code should be replaced with other phrasing, such as "ignoring code"

The wiki should also be updated to reflect that support for the old-style WPCS native ignore comments has been removed in WPCS 3.0.0.

jrfnl:
List of things which will need updating in the wiki before the 3.0.0 release:
- [ ] [Running in Travis](https://github.com/WordPress/WordPress-Coding-Standards/wiki/Running-in-Travis) - the workflow will need to be changed to not clone the repos, but use a Composer install.
- [ ] [Customizable sniff properties](https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties) - references to removed sniffs should be moved to the bottom with a note that those sniffs are no longer available in WPCS 3.0+.
- [ ] [Customizable sniff properties](https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties) - the names of custom properties which have been renamed/changed/added will need to be updated.
- [ ] [whitelist code that is being flagged with errors](https://github.com/WordPress/WordPress-Coding-Standards/wiki/Whitelisting-code-which-flags-errors) - page probably should be removed (or rewritten)

Note: this list may not be complete....

dingo-d:
> [Running in Travis](https://github.com/WordPress/WordPress-Coding-Standards/wiki/Running-in-Travis) - the workflow will need to be changed to not clone the repos, but use a Composer install.

We will leave this for legacy reasons? Since we're not running checks in TravisCI any more 🤷🏼‍♂️

jrfnl:
> We will leave this for legacy reasons? Since we're not running checks in TravisCI any more 🤷🏼‍♂️

This is not about _our_ CI, but an example for projects _using_ WPCS. There may still be (mostly commercial plugins/themes) running _their_ CI on Travis, so for them, it's good to have an example for how to set it up.

jrfnl:
Wiki pages review list - those pages marked with a ✔️ have been updated for WordPressCS 3.0.0:

- [ ] [Home](https://github.com/WordPress/WordPress-Coding-Standards/wiki)
- [ ] [Customizable sniff properties](https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties)
[...]

PR #2065, opened 2022-06-17
Remove typical non-inclusive language from the repo

jrfnl:
Note: not claiming completeness - this PR is based on a code-base wide search for the terms `whitelist` and `blacklist`. Other potential non-inclusive terminology is not (yet) addressed, though I expect there won't be much else.

Related to #1915

---

### AbstractFunctionRestrictionsSniff: rename array key

... and update sniffs using that array key.

👉🏻 Note: this is a breaking change for any external sniffs extending this base class.

### PrefixAllGlobals: rename property [1]

As this is a `protected` property, this is an internal change only (unless an external standard would be extending the sniff).

### PrefixAllGlobals: rename property [2]
As this is a `protected` property, this is an internal change only (unless an external standard would be extending the sniff).

### PrefixAllGlobals: rename property [3]

As this is a `protected` property, this is an internal change only (unless an external standard would be extending the sniff).

### ValidVariableName: rename property [1]

As this is a `protected` property, this is an internal change only (unless an external standard would be extending the sniff).

### ValidVariableName: rename property [2]

👉🏻 As this is a `public` property, this is a BC-break and should be annotated as such in the upgrade guide/changelog.

### IniSet: rename property [1]

As this is a `protected` property, this is an internal change only (unless an external standard would be extending the sniff).

### IniSet: rename property [2]

As this is a `protected` property, this is an internal change only (unless an external standard would be extending the sniff).

### NoSilencedErrors: rename property [1]

👉🏻 As this is a `public` property, this is a BC-break and should be annotated as such in the upgrade guide/changelog.

Also note that the new name is in line with the name for the same property in an upcoming replacement sniff in PHPCSExtra. This should make the switch over to the replacement sniff more straight-forward (if and when).

### NoSilencedErrors: rename property [2]

👉🏻 As this is a `public` property, this is a BC-break and should be annotated as such in the upgrade guide/changelog.

Also note that the new name is in line with the name for the same property in an upcoming replacement sniff in PHPCSExtra. This should make the switch over to the replacement sniff more straight-forward (if and when).

### NoSilencedErrors: rename property [3]

As this is a `protected` property, this is an internal change only (unless an external standard would be extending the sniff).

Also note that the new name is in line with the name for the same property in an upcoming replacement sniff in PHPCSExtra. This should make the switch over to the replacement sniff more straight-forward (if and when).

### Various language usage updates in comments

Includes:
* Removing a reference to a custom ignore annotation which was already removed in PR 1908.
* Removing a (class-level) changelog entry referring to a custom ignore annotation which doesn't exist anymore.
* Use more inclusive language in various other comments.

### DIFF
--- a/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
+++ b/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
@@ -64,13 +64,14 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	public $prefixes = '';
 
 	/**
-	 * Prefix blacklist.
+	 * Prefix blocklist.
 	 *
 	 * @since 0.12.0
+	 * @since 3.0.0  Renamed from `$prefix_blacklist` to `$prefix_blocklist`.
 	 *
 	 * @var string[]
 	 */
-	protected $prefix_blacklist = array(
+	protected $prefix_blocklist = array(
 		'wordpress' => true,
 		'wp'        => true,
 		'_'         => true,
@@ -949,7 +950,7 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	 * Validate an array of prefixes as passed through a custom property or via the command line.
 	 *
 	 * Checks that the prefix:
-	 * - is not one of the blacklisted ones.
+	 * - is not one of the blocked ones.
 	 * - complies with the PHP rules for valid function, class, variable, constant names.
 	 *
 	 * @since 0.12.0
@@ -968,7 +969,7 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 		foreach ( $this->prefixes as $key => $prefix ) {
 			$prefixLC = strtolower( $prefix );
 
-			if ( isset( $this->prefix_blacklist[ $prefixLC ] ) ) {
+			if ( isset( $this->prefix_blocklist[ $prefixLC ] ) ) {
 				$this->phpcsFile->addError(
 					'The "%s" prefix is not allowed.',
 					0,
--- a/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
+++ b/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
@@ -148,6 +148,7 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	 * A list of core constants that are allowed to be defined by plugins and themes.
 	 *
 	 * @since 1.0.0
+	 * @since 3.0.0 Renamed from `$whitelisted_core_constants` to `$allowed_core_constants`.
 	 *
 	 * Source: {@link https://core.trac.wordpress.org/browser/trunk/src/wp-includes/default-constants.php#L0}
 	 * The constants are listed in the order they are found in the source file
@@ -157,7 +158,7 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	 *
 	 * @var array
 	 */
-	protected $whitelisted_core_constants = array(
+	protected $allowed_core_constants = array(
 		'WP_MEMORY_LIMIT'      => true,
 		'WP_MAX_MEMORY_LIMIT'  => true,
 		'WP_CONTENT_DIR'       => true,
@@ -468,7 +469,7 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 						return;
 					}
 
-					if ( isset( $this->whitelisted_core_constants[ $item_name ] ) ) {
+					if ( isset( $this->allowed_core_constants[ $item_name ] ) ) {
 						// Defining a WP Core constant intended for overruling.
 						return;
 					}
@@ -809,7 +810,7 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 		if ( ( 'define' !== $matched_content
 			&& isset( $this->allowed_core_hooks[ $raw_content ] ) )
 			|| ( 'define' === $matched_content
-			&& isset( $this->whitelisted_core_constants[ $raw_content ] ) )
+			&& isset( $this->allowed_core_constants[ $raw_content ] ) )
 		) {
 			return;
 		}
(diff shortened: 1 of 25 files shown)


WordPress/Sniffs/PHP/IniSetSniff.php
```diff
@@ -44,15 +44,16 @@ class IniSetSniff extends AbstractFunctionParameterSniff {
...
 	 * @var array Multidimensional array with parameter details.
-	 *     $whitelisted_options = array(
+	 *     $safe_options = array(
 	 *         (string) option name. = array(
 	 *             (string[]) 'valid_values' = array()
 	 *         )
 	 *     );
 	 */
-	protected $whitelisted_options = array(
+	protected $safe_options = array(
```
GaryJones, line 56: Why `safe` here, instead of `allowed` (which matches the Summary)?

WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
```diff
@@ -36,33 +36,35 @@ class NoSilencedErrorsSniff extends Sniff {
...
+	 * This property only affects whether the standard function list is
+	 * used. The custom functions list, if set, will always be respected.
 	 *
 	 * @since 1.1.0
+	 * @since 3.0.0 Renamed from `$use_default_whitelist` to `$usePHPFunctionsList`.
 	 *
 	 * @var bool
 	 */
-	public $use_default_whitelist = true;
+	public $usePHPFunctionsList = true;
```
GaryJones, line 51: Is anything going to flag this variable as not following snake_case formatting?

GaryJones (commented):
Noting that the wiki will also need updates as well once this is all released.

jrfnl:
> Noting that the wiki will also need updates as well once this is all released.

Correct, there is already an issue open for that as a reminder: #1917

WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
jrfnl: No, we have disabled the [variable naming sniff for this repo](https://github.com/WordPress/WordPress-Coding-Standards/blob/fd44ba4300f8cbe73ee1b4c02547acb0664b3e66/.phpcs.xml.dist#L19) as PHPCS itself uses camelCase variables and we sometimes extend upstream sniffs and need to be able to change the default value of (camelCase) properties, so this will not be flagged.

WordPress/Sniffs/PHP/IniSetSniff.php
jrfnl: Maybe I should update the summary ? The list is a minimal set of ini directives for which changing their value will not cause any problems for interoperability either between a plugin and WP or between plugins. In other words, these are not just _allowed_ to be changed, they are actually _safe_ to be changed.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -64,13 +64,14 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	public $prefixes = '';
 
 	/**
-	 * Prefix blacklist.
+	 * Prefix blocklist.
```
dingo-d, line 67: How come this is not `disallowed_list` (or `disallowList`)? Seeing how other replacements are _allowed_

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.php
```diff
@@ -32,7 +32,7 @@ public function getErrorList( $testFile = 'PrefixAllGlobalsUnitTest.1.inc' ) {
 		switch ( $testFile ) {
 			case 'PrefixAllGlobalsUnitTest.1.inc':
 				return array(
-					1   => 8, // 2 x error for blacklisted prefix passed. 4 x error for short prefixes. 2 x no prefix.
+					1   => 8, // 2 x error for blocked prefix passed. 4 x error for short prefixes. 2 x no prefix.
```
dingo-d, line 35: Here as well (I guess it's kinda related to my first comment). _Disallowed_ would work as well as _blocked_

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
jrfnl: I kind of felt this needed a stronger word. "Disallow", for me, has a connotation along the lines of "we choose to disallow", while "block" feels more restrictive/a stronger choice of words, which felt appropriate as these are prefixes which REALLY should be blocked (disallowed) as they WILL cause problems when used.

dingo-d (approved):

GaryJones (approved):

GaryJones:
Approved, but not merged in case https://github.com/WordPress/WordPress-Coding-Standards/pull/2065#discussion_r900815306 wants to be addressed (fix summary).

WordPress/Sniffs/PHP/IniSetSniff.php
jrfnl: Addressed in: https://github.com/WordPress/WordPress-Coding-Standards/pull/2065/commits/be76478f0d283e6349e2cbde737c9d29e8d044e6

jrfnl:
I've rebased the PR + added one commit to address the documentation for the `IniSet` sniff: https://github.com/WordPress/WordPress-Coding-Standards/pull/2065/commits/be76478f0d283e6349e2cbde737c9d29e8d044e6

While going through the commits again, there was one property I still wondered about whether I chose the right name.
I'll flag it inline and would love to hear your opinions.

WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
```diff
@@ -36,33 +36,35 @@ class NoSilencedErrorsSniff extends Sniff {
...
+	 * Allows users to pass a list of additional functions for which to allow
+	 * the use of the silence operator. This list can be set in a custom ruleset.
 	 *
 	 * @since 1.1.0
+	 * @since 3.0.0 Renamed from `$custom_whitelist` to `$customFunctionsList`.
 	 *
 	 * @var array
 	 */
-	public $custom_whitelist = array();
+	public $customFunctionsList = array();
```
jrfnl, line 64: ☝🏻 This one.

This is a `public` property, so manipulable from a custom ruleset.

I wonder if `$customFunctionsList` isn't too vague and whether this should be changed to `$customAllowedFunctionsList` ?

dingo-d: `$customAllowedFunctionsList` sounds clearer, so I'd go with that 👍🏼

jrfnl: I've fixed this up in the original commit.

jrfnl:
Anyone still wants to give this a look over after the update of the property description + the rename of the NoSilencedErrors property ?

If so, please do and merge if all is okay.

dingo-d (approved):

PR #2073, opened 2022-08-07
WP/CronInterval: bugfix for fully qualified time constants

jrfnl:
If the WP time constants would be used in their fully qualified form, as is often encountered in namespaced files, the sniff would no longer be able to correctly determine the interval being scheduled.

Note: To prevent constants from a different namespace accidentally matching the WP native constants, the namespace separator is replace by a space character rather than just ignored. The regex before running the `eval()` will prevent code with identifier names other than the WP time constants from being passed to the `eval()`, meaning those will still be reported as "interval undetermined".

This commit fixes that.
Includes unit tests.

### DIFF
--- a/WordPress/Tests/WP/CronIntervalUnitTest.inc
+++ b/WordPress/Tests/WP/CronIntervalUnitTest.inc
@@ -160,3 +160,45 @@ add_filter( 'cron_schedules', function ( $schedules ) {
 	];
 	return $schedules;
 } ); // Warning: 8 min.
+
+// Correctly handle fully qualified WP time constants.
+class FQNConstants {
+	public function add_schedules() {
+		add_filter( 'cron_schedules', array( $this, 'add_weekly_schedule' ) ); // Ok: > 15 min.
+		add_filter( 'cron_schedules', array( $this, 'add_eight_minute_schedule' ) ); // Warning: 8 min.
+		add_filter( 'cron_schedules', array( $this, 'add_hundred_minute_schedule' ) ); // Warning: time undetermined.
+		add_filter( 'cron_schedules', array( $this, 'sneaky_fake_wp_constant_schedule' ) ); // Warning: time undetermined.
+	}
+
+	public function add_weekly_schedule( $schedules ) {
+		$schedules['weekly'] = [
+			'interval' => \WEEK_IN_SECONDS,
+			'display'  => \__( 'Once Weekly', 'text-domain' ),
+		];
+		return $schedules;
+	}
+
+	public function add_eight_minute_schedule( $schedules ) {
+		$schedules['every_8_minutes'] = [
+			'interval' => (8 * \MINUTE_IN_SECONDS),
+			'display' => __( 'Once every 8 minutes' )
+		];
+		return $schedules;
+	}
+
+	public function add_hundred_minute_schedule( $schedules ) {
+		$schedules['every_100_minutes'] = [
+			'interval' => (100 * My\Name\MINUTE_IN_SECONDS),
+			'display' => __( 'Once every 100 minutes' )
+		];
+		return $schedules;
+	}
+
+	public function sneaky_fake_wp_constant_schedule( $schedules ) {
+		$schedules['every_100_seconds'] = [
+			'interval' => (100 * MINUTE_\IN_\SECONDS),
+			'display' => __( 'Once every 100 minutes' )
+		];
+		return $schedules;
+	}
+}
--- a/WordPress/Sniffs/WP/CronIntervalSniff.php
+++ b/WordPress/Sniffs/WP/CronIntervalSniff.php
@@ -182,6 +182,11 @@ class CronIntervalSniff extends Sniff {
 							continue;
 						}
 
+						if ( \T_NS_SEPARATOR === $this->tokens[ $j ]['code'] ) {
+							$value .= ' ';
+							continue;
+						}
+
 						if ( $j === $valueEnd && \T_COMMA === $this->tokens[ $j ]['code'] ) {
 							break;
 						}
--- a/WordPress/Tests/WP/CronIntervalUnitTest.php
+++ b/WordPress/Tests/WP/CronIntervalUnitTest.php
@@ -52,6 +52,9 @@ class CronIntervalUnitTest extends AbstractSniffUnitTest {
 			115 => 1,
 			133 => 1,
 			156 => 1,
+			168 => 1,
+			169 => 1,
+			170 => 1,
 		);
 	}
 }


WordPress/Tests/WP/CronIntervalUnitTest.inc
```diff
@@ -160,3 +160,45 @@ add_filter( 'cron_schedules', function ( $schedules ) {
...
 	return $schedules;
 } ); // Error: 8 min.
+
+// Correctly handle fully qualified WP time constants.
+class FQNConstants {
+	public function __construct() {
+		add_filter( 'cron_schedules', array( $this, 'add_weekly_schedule' ) );
+		add_filter( 'cron_schedules', array( $this, 'add_eight_minute_schedule' ) ); // Error: 8 min.
+		add_filter( 'cron_schedules', array( $this, 'add_hundred_minute_schedule' ) ); // // Warning: time undetermined.
+		add_filter( 'cron_schedules', array( $this, 'sneaky_fake_wp_constant_schedule' ) ); // // Warning: time undetermined.
```
GaryJones, line 170:
```suggestion
		add_filter( 'cron_schedules', array( $this, 'add_hundred_minute_schedule' ) ); // Warning: time undetermined.
		add_filter( 'cron_schedules', array( $this, 'sneaky_fake_wp_constant_schedule' ) ); // Warning: time undetermined.
```

WordPress/Tests/WP/CronIntervalUnitTest.inc
```diff
@@ -160,3 +160,45 @@ add_filter( 'cron_schedules', function ( $schedules ) {
 	];
 	return $schedules;
 } ); // Error: 8 min.
+
+// Correctly handle fully qualified WP time constants.
+class FQNConstants {
+	public function __construct() {
+		add_filter( 'cron_schedules', array( $this, 'add_weekly_schedule' ) );
+		add_filter( 'cron_schedules', array( $this, 'add_eight_minute_schedule' ) ); // Error: 8 min.
```
GaryJones, line 168: This comment suggests it would be an "Error", but there's nothing in the `getErrorList()` array?

Maybe this one was not meant to produce an error anymore?

WordPress/Tests/WP/CronIntervalUnitTest.inc
```diff
@@ -160,3 +160,45 @@ add_filter( 'cron_schedules', function ( $schedules ) {
 	];
 	return $schedules;
 } ); // Error: 8 min.
+
+// Correctly handle fully qualified WP time constants.
+class FQNConstants {
+	public function __construct() {
```
GaryJones, line 166:
```suggestion
	public function add_schedules() {
```

Change only because adding `add_*()` to class constructors is an anti-pattern that makes code less testable generally and I'd rather the fixtures didn't promote a bad practice.

GaryJones (changes requested):

WordPress/Tests/WP/CronIntervalUnitTest.inc
jrfnl: Throughout the test case file, it uses `// Error`. This is tied to the history of the sniff.

* This sniff was originally part of the set of `VIP` sniffs. At that time, the "interval too short" notice was an error.
* PR #1242, and in particular commit 52faccb08145e050e983cc123a97bc17f5a8a199, moved the sniff from the `VIP` category to the `WP` category (as one of the sniff we wanted to keep widely available when the `VIP` category would be removed). The preceding discussion on this can be found in #1157 and was also related to #1309.
* The [next commit](https://github.com/WordPress/WordPress-Coding-Standards/pull/1242/commits/77e908624be1ccd99057e9af70c3ece2f5e3c374) (part of the same discussion and PR) then changed the notice from an `error` to a `warning`, but didn't update the inline comments in the test case file.

I agree that it should say `Warning` in the test case file now, but that is outside the scope of this PR, so I have opened PR #2074 to address this for the pre-existing tests and will fix it up for the new tests in this PR.

WordPress/Tests/WP/CronIntervalUnitTest.inc
jrfnl: The constructor is used in some other places in this test case file as well but I'll update the new example.

jrfnl:
Added a second commit addressing the review remarks for easier verification, Please squash the commit on merge (or I can squash and force push ahead of merge).

WordPress/Tests/WP/CronIntervalUnitTest.inc
```diff
@@ -160,3 +160,45 @@ add_filter( 'cron_schedules', function ( $schedules ) {
 	];
 	return $schedules;
 } ); // Error: 8 min.
+
+// Correctly handle fully qualified WP time constants.
+class FQNConstants {
+	public function add_schedules() {
+		add_filter( 'cron_schedules', array( $this, 'add_weekly_schedule' ) );
+		add_filter( 'cron_schedules', array( $this, 'add_eight_minute_schedule' ) ); // Warning: 8 min.
```
GaryJones, line 168:
```suggestion
		add_filter( 'cron_schedules', array( $this, 'add_eight_minute_schedule' ) ); // Warning: 8 min, which is less than the 15 minutes default from `$this->min_interval`.
```

Doesn't have to be this exact wordy comment, but it took me a while to understand why this would be a warning, since the fix addressed this exact syntax parsing.

```suggestion
		add_filter( 'cron_schedules', array( $this, 'add_eight_minute_schedule' ) ); // Warning (CronSchedulesInterval): 8 min.
```

..would also work, for instance, since there are multiple warning violation codes for this sniff.

GaryJones (approved):
Approving, with one suggestion / self-note.

WordPress/Tests/WP/CronIntervalUnitTest.inc
jrfnl: The two types of warnings in the test case file are distinguished from each other by `Warning: time undetermined` vs `Warning: # min.` This is the same throughout the file, so changing that format should be done for the whole file, not for a new test only.

And yes, while the false positive I was addressing is basically what's in the first new test - `add_weekly_schedule()` -, I added the extra tests to make sure I would not be introducing _false negatives_ with the fix I was making.

Using "which is less than the 15 minutes default from `$this->min_interval`" would make things less clear as the `$min_interval` is a configurable `public` property and the test file contains tests with different settings for `min_interval`.

What about if I add ` // Ok: > 15 min.` to the first new test case ? That should help make the distinction clearer. Would that work for you ?

GaryJones: No change necessary at this point - I think I was just seeing the warning message out of context of understanding there _was_ a minimum interval.

jrfnl: I'd added the `// Ok: > 15 min.` in the previous update anyway, so, hopefully this will help next time someone looks at it.

jrfnl:
I've rebased  the PR (without changes) to get passed the imaginary merge conflict. I've squashed the second commit with the updates after review into the first one now as well, so if found okay, this PR can now be merged normally.

dingo-d (approved):

PR #2081, opened 2022-09-02
Add filter_var_array and filter_input_array to escapingFunctions

anubisthejackle:
When running [`filter_var_array`](https://www.php.net/manual/en/function.filter-var-array.php) I noticed that this sniff alerted the line wasn't properly sanitizing my input variable. 

This addresses that by adding `filter_var_array` and `filter_input_array` beside their corresponding functions in `$escapingFunctions`


jrfnl:
@anubisthejackle Thanks for the PR. Unfortunately it is unclear what you want to achieve with this and why you are making this change in the first place. There also no tests.

I'm more inclined to remove the `filter*()` function group completely from the `$escapingFunctions` list.
First off these are _not_ escaping functions, but input validation/sanitization functions.
Secondly, without specific filter constants being passed, those functions actually don't have much effect and will not even do that, so will leave an application wide open to security issues.

PR #2085, opened 2022-09-26
Added auto escaped function `get_block_wrapper_attributes()` to list of escaped functions.

lewis-elborn:
This PR is a minor change adding `get_block_wrapper_attributes()` to the list of already escaped functions. See (PR [#44473](https://github.com/WordPress/gutenberg/pull/44473) in Gutenburgs PR's).


jrfnl:
Please undo all the (incorrect) CS changes and only make the change you are actually proposing.

Also, I've had a quick look at the `get_block_wrapper_attributes()` function and the `$key` used is not escaped and as it is not a hard-coded value, this leaves the code wide-open to all sorts of injection, so until that function is actually made safe, I would not recommend adding it to the "safe" functions list. This should probably be flagged up for the @WordPress/security-team ...

https://developer.wordpress.org/reference/functions/get_block_wrapper_attributes/

lewis-elborn:
@jrfnl  Sorry, I'm finding my way around pull requests for the first time. I only wanted to edit one line, but the Basic QA checks failed.

jrfnl:
> @jrfnl Sorry, I'm finding my way around pull requests for the first time. I only wanted to edit one line, but the Basic QA checks failed.

That's fine. You can run `composer check-cs` on the command line to see what is being flagged and `composer fix-cs` should be able to auto-fix most issues.

For additional guidelines - like adding tests to cover your change - please have a look at the [CONTRIBUTING](https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/.github/CONTRIBUTING.md) docs.

lewis-elborn:
Thanks for your patience, @jrfnl. I got there in the end though. 😅

jrfnl:
@lewis-elborn No worries, but now (aside from the PR needing a test) the question remains whether the function should be added, so my remark about this above.

jrfnl:
P.S.: thanks for persisting in trying to get the CI to pass ;-)

peterwilsoncc:
It looks like the function only supports a fixed set of attributes, see [source code reference](https://github.com/WordPress/wordpress-develop/blob/2bb5679d666474d024352fa53f07344affef7e69/src/wp-includes/class-wp-block-supports.php#L176-L178).

However, I think it would be good to modify WordPress to use `esc_attr( $key )` for hardening purposes prior to adding the function to the safe list. Doing so will prevent issues in the event a bug is introduced at a latter date.

Does that work for you both?

jrfnl:
> It looks like the function only supports a fixed set of attributes, see [source code reference](https://github.com/WordPress/wordpress-develop/blob/2bb5679d666474d024352fa53f07344affef7e69/src/wp-includes/class-wp-block-supports.php#L176-L178).

@peterwilsoncc Except it doesn't - arbitrary attributes passed via the parameter which are not in that list are added without any validation: https://github.com/WordPress/wordpress-develop/blob/2bb5679d666474d024352fa53f07344affef7e69/src/wp-includes/class-wp-block-supports.php#L198-L202

I'll pass you a code sample in private (wouldn't want to give anyone any ideas).

> However, I think it would be good to modify WordPress to use `esc_attr( $key )` for hardening purposes prior to adding the function to the safe list. Doing so will prevent issues in the event a bug is introduced at a latter date.

I'd highly improving the security. The security bug already exists, this is not a hypothetical future situation.

I wonder if this shouldn't use `sanitize_key()` instead of `esc_attr()` ?

peterwilsoncc:
> arbitrary attributes passed via the parameter which are not in that list are added without any validation

Thanks @jrfnl, I misread that line of code.

As discussed elsewhere, I am not sure what is best to use for the escaping of attribute names. Reviewing the attribute spec, it looks like sanitize_title is closer than esc_attr. https://html.spec.whatwg.org/multipage/syntax.html#attributes-2

dingo-d:
Just chiming in that we have a PR opened that adds whole lot of other safe functions to the list (#2082).

IMO this PR should be closed, and an issue opened instead. 

There we can discuss the problem of whether the `get_block_wrapper_attributes()` is correctly escaped or not, and should be added to the list.

jrfnl:
Closing this PR now as the function should not be on the safe list as-is.

PR #2090, opened 2022-10-25
Update for PHPCSUtils 1.0.0-alpha4

jrfnl:
**_Note: this is a minimal update to get rid of some deprecations and such. Once this has been merged, a lot more PRs will follow which will actually implement the new functionality from PHPCSUtils 1.0.0-alpha4._**

Ref: https://github.com/PHPCSStandards/PHPCSUtils/releases/tag/1.0.0-alpha4

### Composer: up the minimum PHPCS version to 3.7.1

Follow up on #2058

The minimum PHPCS version for PHPCSUtils 1.0.0 will be 3.7.1, so the minimum PHPCS version for WPCS should be in line with that.

The difference between 3.7.0 and 3.7.1 is very small. They were released just days apart with 3.7.1 fixing a pertinent bug in the retokenization of reserved keywords, which was updated in PHPCS 3.7.0.

### EnqueuedResourcesSniff: use Collections method instead of property

All the `Collections` class properties have been deprecated in PHPCSUtils 1.0.0-alpha4 in favour of methods with the same name, due to too many new tokens being introduced in PHP itself and the properties therefor not being a viable path for the future.

### CronInterval: add the parameter name to allow for named parameters

PHPCSUtils 1.0.0-alpha4 supports named parameters in the `PassedParameters` class, but requires for the parameter name to be passed either as a string or an array (of multiple names in case of a renamed parameter).

The `CronInterval` sniff not passing the parameter name would cause the tests to fail - even when the tests do not contain a test for named parameters yet -, so let's fix the function call.

Note: this doesn't add proper support for named parameters to the sniff yet. That will be addressed later.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -36,7 +36,7 @@ When you introduce new `public` sniff properties, or your sniff extends a class 
 
 ## Pre-requisites
 * WordPress-Coding-Standards
-* PHP_CodeSniffer 3.7.0 or higher
+* PHP_CodeSniffer 3.7.1 or higher
 * PHPUnit 4.x, 5.x, 6.x or 7.x
 
 The WordPress Coding Standards use the `PHP_CodeSniffer` native unit test suite for unit testing the sniffs.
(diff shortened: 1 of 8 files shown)


dingo-d (approved):
🥳🥳🥳

jrfnl:
FYI: I've updated the branch protection rules for `develop` to match.

GaryJones (approved):
✅

PR #2091, opened 2022-10-26
Sniff: start using PHPCSUtils MessageHelper

jrfnl:
### Sniff::addMessage(): switch over to the PHPCSUtils version

PHPCSUtils 1.0.0-alpha4 has a new `MessageHelper` class containing methods for the same, so we can now switch out the WPCS native methods for the PHPCSUtils ones.

Includes changing some long lines to multi-line function calls.

### Sniff::string_to_errorcode(): switch over to the PHPCSUtils version

PHPCSUtils 1.0.0-alpha4 has a new `MessageHelper` class containing a method for the same, so we can now switch out the WPCS native method for the PHPCSUtils one.


dingo-d (approved):

GaryJones:
Am I right in thinking that:

 1. The PHPCSUtils MessageHelper doesn't provide any noticeable improvements over the code that was already in WPCS, but it just pulls it up to PHPCSUtils and out of the WPCS weeds?
 2. The `MessageHelper::addMessage()` is used when we don't know until runtime whether it should be an error or warning, and so avoids having the conditional `addError()` / `addWarning()` in the sniff code directly?

GaryJones (approved):
✅

jrfnl:
> Am I right in thinking that:
> 
> 1. The PHPCSUtils MessageHelper doesn't provide any noticeable improvements over the code that was already in WPCS, but it just pulls it up to PHPCSUtils and out of the WPCS weeds?

Mostly correct.
1. In contrast to utility methods in WPCS, all code in PHPCSUtils is actually 100% unit tested. In WPCS this was only tested via integration tests in the sniffs.
2. The `MessageHelper::stringToErrorcode()` method has an optional second parameter `$strtolower = false`, which when set to `true` will lowercase the text passed as well.
    This can be useful to ensure that arbitrary texts, like function names, will always be in the same case when used in an error code.

> 2. The `MessageHelper::addMessage()` is used when we don't know until runtime whether it should be an error or warning, and so avoids having the conditional `addError()` / `addWarning()` in the sniff code directly?

Correct.

PR #2092, opened 2022-10-26
TextStringHelper: remove the strip_interpolated_variables() method

jrfnl:
... in favour of the significantly more comprehensive PHPCSUtils `TextStrings::stripEmbeds()` method.

This improves the handling of embedded variables in double quoted strings (and consequently fixes some bugs).

Bug confirmed via the tests added to the `ValidPostTypeSlug` sniff.

Without this fix, the "text stripped of embeds" for these test cases would look like this:
```
string(24) "testing123-${(foo)}-test"
string(19) "testing123-"]}-test"
```

I.e. the embeds wouldn't be, or wouldn't be completely, stripped.

With the fix, the embeds in both texts are correctly stripped, resulting in:
```
string(16) "testing123--test"
```

As for the sniff, for the `ValidPostTypeSlug` sniff, the bug meant that the sniff would report three false positives:
```
 55 | ERROR   | register_post_type() called with invalid post type "testing123-${(foo)}-test". Post type contains invalid characters. Only lowercase
    |         | alphanumeric characters, dashes, and underscores are allowed.
 55 | ERROR   | A post type slug must not exceed 20 characters. Found: "testing123-${(foo)}-test" (24 characters).
 56 | ERROR   | register_post_type() called with invalid post type "testing123-${foo["${bar['baz']}"]}-test". Post type contains invalid characters.
    |         | Only lowercase alphanumeric characters, dashes, and underscores are allowed.
```

These are all fixed now.

### DIFF
--- a/WordPress/Helpers/TextStringHelper.php
+++ b/WordPress/Helpers/TextStringHelper.php
@@ -21,20 +21,10 @@ namespace WordPressCS\WordPress\Helpers;
  * the future by functions from PHPCSUtils.}
  *
  * @package WPCS\WordPressCodingStandards
- * @since   3.0.0 The constant and methods in this class were previously contained in the
- *                `WordPressCS\WordPress\Sniff` class and have been moved here.
+ * @since   3.0.0 The method in this class was previously contained in the
+ *                `WordPressCS\WordPress\Sniff` class and has been moved here.
  */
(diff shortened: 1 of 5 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2093, opened 2022-10-26
Sniff::is_foreach_as(): switch over to the PHPCSUtils version

jrfnl:
PHPCSUtils 1.0.0-alpha4 has a new `Context` class containing a more versatile method for the same, so we can now switch out the WPCS native method for the PHPCSUtils one.

Ref: https://phpcsutils.com/phpdoc/classes/PHPCSUtils-Utils-Context.html#method_inForeachCondition


dingo-d (approved):

GaryJones (approved):
✅

PR #2094, opened 2022-10-26
TextStringHelper: remove in favour of PHPCSUtils

jrfnl:
### WP/I18n: use PHPCSUtils for checking string embeds

Switch out the use of the WPCS native `TextStringHelper::get_interpolated_variables()` method in favour of the significantly more comprehensive PHPCSUtils `TextStrings::getEmbeds()` method.

This improves the handling of embedded variables in double quoted strings (and consequently fixes some bugs).

For some samples of embedded variables/expressions, the sniff would previously correctly report on the embed, but would contain an incomplete representation of the embed in the error message:
```
 223 | ERROR   | [ ] The $text arg must not contain interpolated variables. Found "$foo".
 224 | ERROR   | [ ] The $text arg must not contain interpolated variables. Found "$foo".
 225 | ERROR   | [ ] The $text arg must not contain interpolated variables. Found "$foo".
```

For other embedded variables/expressions, the sniff would previously not throw an error (false negative).
```php
// No error, while an error should be thrown.
__( "{$foo['bar']->baz()()}", 'my-slug' );
__( "${foo->bar}", 'my-slug' );
__( "${foo["${bar['baz']}"]}", 'my-slug' );
```

With the implemented fix, all embeds will now result in the expected error and the error messages will contain a correct representation of the embed:
```
 222 | ERROR   | [ ] The $text arg must not contain interpolated variables or expressions. Found "${foo}".
 223 | ERROR   | [ ] The $text arg must not contain interpolated variables or expressions. Found "{$foo['bar']}".
 224 | ERROR   | [ ] The $text arg must not contain interpolated variables or expressions. Found "{$foo->bar()}".
 225 | ERROR   | [ ] The $text arg must not contain interpolated variables or expressions. Found "{$foo['bar']->baz()()}".
 226 | ERROR   | [ ] The $text arg must not contain interpolated variables or expressions. Found "${foo->bar}".
 227 | ERROR   | [ ] The $text arg must not contain interpolated variables or expressions. Found "${foo["${bar['baz']}"]}".
```

Includes minor improvement to the error message.
To prevent an unnecessary BC-break, I'm electing not to change the error code.

### DB/PreparedSQL: use PHPCSUtils for checking string embeds

Switch out the use of the WPCS native `TextStringHelper::get_interpolated_variables()` method in favour of the significantly more comprehensive PHPCSUtils `TextStrings::getEmbeds()` method.

This improves the handling of embedded variables in double quoted strings (and consequently fixes some bugs).

For some samples of embedded variables/expressions, the sniff would previously correctly report on the embed, but would contain an incomplete/incorrect representation of the embed in the error message:
```php
$wpdb->query( "SELECT * FROM {$wpdb->bar()} WHERE post_title LIKE '{$title->bar()}';" );
//  106 | ERROR | Use placeholders and $wpdb->prepare(); found interpolated variable $title at "SELECT * FROM {$wpdb->bar()} WHERE post_title LIKE '{$title->bar()}';"

$wpdb->query( "SELECT * FROM ${wpdb->{$baz}} WHERE post_title LIKE '${title->{$sub}}';" );
// 108 | ERROR | Use placeholders and $wpdb->prepare(); found interpolated variable $sub at "SELECT * FROM ${wpdb->{$baz}} WHERE post_title LIKE '${title->{$sub}}';"
```

In other cases, the `$wpdb` variable was not correctly recognized as such (false positive):
```php
$wpdb->query( "SELECT * FROM ${wpdb->{$baz}} WHERE post_title LIKE '${title->{$sub}}';" );
//  108 | ERROR | Use placeholders and $wpdb->prepare(); found interpolated variable $baz at "SELECT * FROM ${wpdb->{$baz}} WHERE post_title LIKE '${title->{$baz}}';"
```

For other embedded variables/expressions, the sniff would previously not throw an error (false negative).
```php
// No error, while an error should be thrown.
$wpdb->query( "SELECT * FROM ${wpdb->bar} WHERE post_title LIKE '${title->sub}';" ); // Bad x 1.
$wpdb->query( "SELECT * FROM ${wpdb->{${'a'}}} WHERE post_title LIKE '${title->{${'sub'}}}';" ); // Bad x 1.
```

With the implemented fix, all embeds will now result in the expected errors and the error messages will contain a correct representation of the embed:
```
 106 | ERROR | Use placeholders and $wpdb->prepare(); found interpolated variable {$title->bar()} at "SELECT * FROM {$wpdb->bar()} WHERE post_title LIKE '{$title->bar()}';"
 107 | ERROR | Use placeholders and $wpdb->prepare(); found interpolated variable ${title->bar} at "SELECT * FROM ${wpdb->bar} WHERE post_title LIKE '${title->bar}';"
 108 | ERROR | Use placeholders and $wpdb->prepare(); found interpolated variable ${title->{$baz}} at "SELECT * FROM ${wpdb->{$baz}} WHERE post_title LIKE '${title->{$baz}}';"
 109 | ERROR | Use placeholders and $wpdb->prepare(); found interpolated variable ${title->{${'a'}}} at "SELECT * FROM ${wpdb->{${'a'}}} WHERE post_title LIKE '${title->{${'a'}}}';"
```

Note: the matching of `$wpdb` has been improved, but does not check whether a known table name is accessed or anything. This is no different than before, but may warrant further improvement in the future.

### DB/PreparedSQL: improve matching of $wpdb

While probably still not perfect, this improves how the `$wpdb` variable is matched in embeds.

* The regex will match all possible "start of" embeds - `$wpdb`, `{$wpdb` and `${wpdb`.
* The regex requires that the `$wpdb` variable is followed by an object operator - either `->` or the PHP 8.1 `?->` - to access a table name.
    Note: I'm not sure we should actually allow `?->` as in that case, if `$wpdb` would be `null`, the result would be `null`, which would convert to an empty string in a double quoted string.
    Then again, that would just make for an invalid query and that is not the concern of this sniff (which is why I included it anyway).
* `$wpdb` on its own will no longer match as it would be invalid anyway (the `WPDB` class does not have a `__toString()` method).

Includes one extra unit test, though the principle of it is already covered by most existing tests.

### DB/PreparedSQLPlaceholders: use PHPCSUtils for checking string embeds and improve `$wpdb` matching

This change is along the same lines as the change made for the `PreparedSQL` sniff, but as the `PreparedSQLPlaceholders` sniff only uses the embeds to prevent duplicate messages, extra unit tests which would actually fail would be pretty complex to create.

All the same, this should be considered safeguarded by the existing unit tests.

### Security/ValidatedSanitizedInput: use PHPCSUtils for checking string embeds

Switch out the use of the WPCS native `TextStringHelper::get_interpolated_variables()` method in favour of the significantly more comprehensive PHPCSUtils `TextStrings::getEmbeds()` method.

This improves the handling of embedded variables in double quoted strings (and consequently fixes some bugs).

For various embedded variables/expressions, the sniff would previously not throw an error (false negative), for example:
```php
// No error, while an error should be thrown.
echo "{$_GET['bar']}";
echo "${_REQUEST['bar']}";
echo "${_GET["${bar}"]}";
echo "${_FILES["${bar['baz']}"]}";
```

With the implemented fix, all embedded superglobals the sniff looks for, will now result in the expected error.

Note: the "bad" comment for the heredoc is _above_ the test instead of as a trailing comment as a trailing comment would be a parse error in PHP < 7.3.

### TextStringHelper: remove the class

After the implementation of the PHPCSUtils `TextString::getEmbeds()` function in all relevant sniffs, the `TextStringHelper::get_interpolated_variables()` method is no longer used in WPCS, so this Helper class has become redundant.


dingo-d (approved):

GaryJones (approved):
✅

PR #2095, opened 2022-10-27
:sparkles: New `WordPress.WhiteSpace.ObjectOperatorSpacing` sniff

jrfnl:
This sniff largely defers to the upstream `Squiz.WhiteSpace.ObjectOperatorSpacing` sniff, which was previously already included.

The only real difference between the two sniffs is that for class resolution using `ClassName::class`, new lines around the `::` object operator will not be ignored.

The WPCS sniff has the same default property settings as the upstream sniff and changes to the property values should be made in the ruleset.

Includes docs.
Includes unit tests covering the difference and safeguarding that the sniff doesn't overreach.

Related to: https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#magic-constants

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -526,7 +526,7 @@
 	<rule ref="Generic.Classes.OpeningBraceSameLine"/>
 
 	<!-- Object operators should not have whitespace around them unless they are multi-line. -->
-	<rule ref="Squiz.WhiteSpace.ObjectOperatorSpacing">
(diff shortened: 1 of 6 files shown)


jrfnl:
Note: there will be a follow-up PR to address the rest of that handbook section once PHPCSExtra 1.0.0-alpha4 has been tagged.

dingo-d (approved):

GaryJones (approved):
✅

PR #2096, opened 2022-10-27
WPCS-Extra: enable the `Generic.CodeAnalysis.UnusedFunctionParameter` sniff

### LINKED ISSUE #1510: PHPCS 3.4.0: Activate select codes from Generic.CodeAnalysis.UnusedFunctionParameter sniff
## Rationale
The upstream `Generic.CodeAnalysis.UnusedFunctionParameter` checks if all parameters declared in a function signature are used within the function.

This sniff was, so far, not added to WPCS as it contained the following two flaws making it unusable within a WP context:
* The sniff did not check properly if a class implemented an interface or extended another class.
    In those cases, the sniff would throw warnings for unused function parameters which couldn't be removed as the function signature of the method in the child class has to be the same as the one in the parent class/interface.
* The sniff would throw warnings for function parameters being unused *before* the last *used* parameter.
    The whole WordPress hook mechanism is based on call-backs. A function hooked into a filter/action, may not use all parameters the action/filter passes.
    Example: an action hook may pass four parameters, a hooked in function may only use parameter 1 and 3.
    In that case, a warning for the second parameter being unused would be useless (as it cannot be removed), while a warning for the fourth parameter not being used would be _useful_.

Upstream PR https://github.com/squizlabs/PHP_CodeSniffer/pull/1318 improves the sniff to take all those situations into account and introduces a large range of modular error codes covering them.
This PR has now been merged upstream and will be released as part of PHPCS 3.4.0.

Of the newly introduced error codes, I'd like to suggest adding the following codes to WPCS:
* Found
* FoundAfterLastUsed

For the following errorcodes, some more tests would need to be run to see if these should be added to the standard or not:
* FoundInExtentedClass
* FoundInExtentedClassAfterLastUsed
* FoundInImplementedInterface
* FoundInImplementedInterfaceAfterLastUsed

I'd like to suggest adding the error codes from this sniff to the `Extra` ruleset once the minimum PHPCS requirement has been upped to PHPCS 3.4.0.

## References
* Upstream (merged) PR: https://github.com/squizlabs/PHP_CodeSniffer/pull/1318 - Please see the whole thread for the different error codes now available.
* https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/382#discussion-diff-29981655 - earlier discussion about this sniff where it was decided to not add the sniff to WPCS at that time.

## Action Checklist
Once PHPCS 3.4.0 has been released and WPCS is ready/willing to up the minimum required PHPCS version:
- [x] Update the PHPCS version requirements in the `composer.json` file.
- [x] Update references to the WPCS minimum PHPCS version requirements in the Readme and wherever else needed.
- [x] Add select errorcodes from the `Generic.CodeAnalysis.UnusedFunctionParameter` sniff to the `WordPress-Extra` `ruleset.xml` file and remove the commented out reference to the sniff from the ruleset.

jrfnl:
The reason why the sniff was disabled up to now was that WP uses callbacks a lot and the hooked in functions may not use all parameters passed to them.

This concern was previously raised in https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/382#discussion_r29981655, also see https://github.com/WordPress/WordPress-Coding-Standards/pull/382/commits/522b35f3ebac915d6eebb92624fc703c193aee60

To this end, I've send in a [PR upstream](https://github.com/squizlabs/PHP_CodeSniffer/pull/1318) to address these concerns, which has subsequently been merged. The adjusted version of the sniff was released in PHPCS 3.4.0, which means the sniff can now be added.

Note: the sniff has since then also been adjusted to handle arrow functions correctly, as well as constructor property promotion.

Fixes #1510

### DIFF
--- a/WordPress-Extra/ruleset.xml
+++ b/WordPress-Extra/ruleset.xml
@@ -48,13 +48,23 @@
 	</rule>
 
 	<!-- Check correct spacing of language constructs. This also ensures that the
-	     above rule for not using brackets with require is fixed correctly.
+		 above rule for not using brackets with require is fixed correctly.
 		 https://github.com/WordPress/WordPress-Coding-Standards/issues/1153 -->
 	<rule ref="Generic.WhiteSpace.LanguageConstructSpacing"/>
 
-	<!-- Hook callbacks may not use all params -->
-	<!-- https://github.com/WordPress/WordPress-Coding-Standards/pull/382#discussion_r29981655 -->
-	<!--<rule ref="Generic.CodeAnalysis.UnusedFunctionParameter"/>-->
+	<!-- Check that functions use all parameters passed.
+		 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/xxx -->
+	<rule ref="Generic.CodeAnalysis.UnusedFunctionParameter">
+		<!-- Allow for callback functions which may not need all parameters passed. -->
+		<exclude name="Generic.CodeAnalysis.UnusedFunctionParameter.FoundBeforeLastUsed"/>
+		<!-- Allow for functions in extended classes/implemented interfaces. -->
+		<exclude name="Generic.CodeAnalysis.UnusedFunctionParameter.FoundInExtentedClass"/>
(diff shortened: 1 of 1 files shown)


GaryJones (approved):
✅ This is a nice change for WP :-)

PR #2097, opened 2022-10-27
Core: move rules related to include/require statements from `Extra` to `Core`

jrfnl:
> 1. No parentheses shall be used for `include[_once]` and `require[_once]` statements.
>    As `include[_once]` and `require[_once]` are language constructs, they do not need parentheses around the path.
> 2. There should be exactly one space between the `include[_once]` and `require[_once]` keyword and the start of the path.
> 3. It is strongly recommended to use `require[_once]` for unconditional includes.
>    When using `include[_once]` PHP will throw a warning when the file is not found, but will continue execution which will almost certainly lead to other errors/warnings/notices being thrown if your application depends on the file being available, potentially leading to security leaks. For that reason, `require[_once]` is generally the better choice as it will throw a Fatal Error if the file cannot be found.

Refs:
* https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/ - Include/Require section
* https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#writing-include-require-statements
* WordPress/wpcs-docs#111
* WordPress/WordPress-Coding-Standards#1007
* WordPress/WordPress-Coding-Standards#1143
* WordPress/WordPress-Coding-Standards#1145
* WordPress/WordPress-Coding-Standards#1153
* WordPress/WordPress-Coding-Standards#1163

Closes #1862

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -66,6 +66,31 @@
 
 	<!--
 	#############################################################################
+	Handbook: General - Writing include/require statements.
+	Ref: https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#single-and-double-quotes
+	#############################################################################
+	-->
+
+	<!-- Covers rule: include[_once] and require[_once] are language constructs, they do
+		 not need parentheses around the path, so those shouldn't be used. -->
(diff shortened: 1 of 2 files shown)


GaryJones (approved):
✅

dingo-d (approved):

Issue #1101, opened 2017-08-14
Function declaration verification


jrfnl:
Following up on my remark here: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1084#issuecomment-320933566

> The _T_FUNCTION and T_CLOSURE (and T_USE for that matter) tokens have been added to that sniff in WPCS, but are not control structures. They should be handled separately completely as the control structure sniff is not suitable for those.

### Inventarisation

Function declaration verification basically can be broken down into a number of separate issues/checks:

1. **Spacing around the `function` / `use` keywords.**
    Status: Problematic. Currently covered by the `WordPress.WhiteSpace.ControlStructureSpacing` sniff, but this is leading to issues.
    Viable upstream sniffs: `PEAR.Functions.FunctionDeclaration`, `Squiz.Functions.FunctionDeclaration` and `Squiz.Functions.MultiLineFunctionDeclaration`
1. **Formatting of multi-line function declarations, i.e. parameters declarations and/or `use` statement over multiple lines.**
    Status: Currently not covered.
    Viable upstream sniffs: `Squiz.Functions.MultiLineFunctionDeclaration`
1. **Spacing around the function parameters and the comma's separating these.**
    Status: Covered in `WordPress-Core`
    Sniff: `Squiz.Functions.FunctionDeclarationArgumentSpacing`
1. **Spacing around variables being imported via a closure `use` statement and the comma's separating these.**
    Status: Covered in `WordPress-Core`
    Sniff: `Squiz.Functions.FunctionDeclarationArgumentSpacing`
1. **(No) Spacing within the parentheses if there are no parameters.**
    Status: Covered in `WordPress-Core`
    Sniff: `Squiz.Functions.FunctionDeclarationArgumentSpacing`
1. **Spacing of a return type declaration.**
    Status: Open PR #1084
1. **Placement of the opening curly brace.**
    Status: Covered in `Wordpress-Core`
    Sniff: `Generic.Functions.OpeningFunctionBraceKernighanRitchie`
1. **Placement of the closing curly brace.**
    Status: Partially covered in `Wordpress-Core`. If the closing brace is the first non whitespace token of a new line, the indent is checked. That it should be on a new line, is, however, currently **NOT** checked.
    Sniff: `Generic.WhiteSpace.ScopeIndent`
[...]

PR #2101, opened 2022-10-28
Core: move rules related to explicit property/method visibility from `Extra` to `Core`

jrfnl:
> 1. For all constructs which allow it (properties, methods), visibility should be explicitly declared.
> 2. Using the var keyword for property declarations is not allowed.

The `PSR2.Classes.PropertyDeclaration` sniff contains a multitude of checks, amongst which two error codes which IMO should probably not (yet) go into the `Core` ruleset.
* `Underscore` - which warns against prefixing a property name with an underscore, advising to use visibility instead. While I'm 100% in favour of this, introducing this for WP Core would be problematic as renaming (`public`/`protected`) properties would be a BC-break.
* `Multiple` - which throws an error for multi-property declarations. I have excluded it as it is not something which is explicitly forbidden in the handbook, nor was it mentioned in the Make post. Having said that, I have scanned WP Core with this check in place and it would not yield any issues, so "silently" introducing it should not be a problem.

As these rules were previously already included in the `Extra` ruleset, those error code which are being silenced for `Core`, will be "unsilenced" again for `Extra`.

Refs:
* https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/ - Visibility should always be declared section
* https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#visibility-should-always-be-declared
* WordPress/wpcs-docs#107
* WordPress/WordPress-Coding-Standards#1101
* WordPress/WordPress-Coding-Standards#1103

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -348,6 +348,27 @@
 
 	<!--
 	#############################################################################
+	Handbook: Object-Oriented Programming - Visibility should always be declared.
+	Ref: https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#visibility-should-always-be-declared
+	#############################################################################
+	-->
+	<!-- Covers rule: For all constructs that allow it (properties, methods, class constants since PHP 7.1),
+		 visibility should be explicitly declared. -->
+	<rule ref="PSR2.Classes.PropertyDeclaration"/>
+	<rule ref="PSR2.Classes.PropertyDeclaration.Underscore">
+		<severity>0</severity>
+	</rule>
+	<rule ref="PSR2.Classes.PropertyDeclaration.Multiple">
+		<severity>0</severity>
+	</rule>
+	<rule ref="Squiz.Scope.MethodScope"/>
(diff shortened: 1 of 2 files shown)


dingo-d (approved):

GaryJones (approved):
✅

jrfnl:
> * `Multiple` - which throws an error for multi-property declarations. I have excluded it as it is not something which is explicitly forbidden in the handbook, nor was it mentioned in the Make post. Having said that, I have scanned WP Core with this check in place and it would not yield any issues, so "silently" introducing it should not be a problem.

@GaryJones @dingo-d Gentlemen, just checking - did either or you have an opinion on the above ?

GaryJones:
PSR-12 says `There MUST NOT be more than one property declared per statement.`, so I'd be happy to align with that as well.

dingo-d:
Definitely apply the no multiple properties per statement 👍🏼

jrfnl:
Thanks for your responses. Actioned in #2104.

PR #2105, opened 2022-11-04
Extra: fix typo in rule

jrfnl:
Oops... (follow up to #2096)

PR #2107, opened 2022-11-05
Extra: fix (more) typos in rules

jrfnl:
Argh... two more... (follow up to #2096 and #2105)

(Happens when I'm working on too many things at the same time...)


dingo-d (approved):

GaryJones (approved):
✅

PR #2108, opened 2022-11-12
Updated - WP.AlternativeFunctions: Add more functions to filesystem functions list

sandeshjangam:
Closes #1265

- [x] Check the current state of `WP_File_System` and remove any functions from the list which don't have a `WP_File_System` counterpart.
If there is no replacement function, the sniff would throw a non-actionable error/warning, which effectively is just noise.
- [x] For functions which do have a WP counter-functions (outside `WP_File_System`), set those up in separate groups with an informative warning/error message.
- [x] Remove `file_get_contents()` from the list as it is already checked in another group in the same sniff (with logic to prevent triggering on opening of local files) and we don't want to throw duplicate notices.


WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -139,6 +157,9 @@ public function getGroups() {
 					'file_put_contents',
 					'fsockopen',
 					'pfsockopen',
+					// 'fputcsv',
```
dingo-d, line 160: You can remove comments out 🙂

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -139,6 +157,9 @@ public function getGroups() {
...
+					// 'fputs',
+					// 'ftruncate',
```
sandeshjangam, line 162:
```suggestion
					'fputs',
```

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
sandeshjangam: Removed...

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -126,11 +126,29 @@ public function getGroups() {
 				),
 			),
 
+			'unlink' => array(
+				'type'      => 'warning',
+				'message'   => '%s() is discouraged. Use wp_delete_file() for remote URLs instead.',
+				'since'     => 'x.x.x',
```
jrfnl, line 132: Looking at the docs, the `wp_delete_file()`function was added in WP 4.2.0, so I believe this should be set to `4.2.0`.
Ref: https://developer.wordpress.org/reference/functions/wp_delete_file/

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -126,11 +126,29 @@ public function getGroups() {
 				),
 			),
 
+			'unlink' => array(
+				'type'      => 'warning',
+				'message'   => '%s() is discouraged. Use wp_delete_file() for remote URLs instead.',
```
jrfnl, line 131: Just wondering: why specifically for "remote URLs" ? And if so, no check is being done in the sniff that the warning will not show for local files. Should that be added ? Or should the `"for remote URLs" be removed ?

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -126,11 +126,29 @@ public function getGroups() {
...
 				'message'   => 'File operations should use WP_Filesystem methods instead of direct PHP filesystem calls. Found: %s()',
 				'since'     => '2.5.0',
 				'functions' => array(
+					'chgrp',
+					'chmod',
+					'chown',
+					'is_writable',
+					'is_writeable',
+					'mkdir',
+					'rename',
```
jrfnl, line 149: Based on https://github.com/WordPress/WordPress-Coding-Standards/pull/1265#issuecomment-353495333, the WP_FileSystem counterpart for `rename` would be `move`, so should this be in a a separate group so the replacement can be mentioned ?

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -126,11 +126,29 @@ public function getGroups() {
...
+			'unlink' => array(
+				'type'      => 'warning',
+				'message'   => '%s() is discouraged. Use wp_delete_file() for remote URLs instead.',
+				'since'     => 'x.x.x',
+				'functions' => array(
+					'unlink',
+				),
+			),
+
 			'file_system_read' => array(
```
jrfnl, line 138:
```suggestion
			'file_system_operations' => array(
```

The `file_system_read` name is inaccurate. It is like this for historic reasons and changing it would be a - albeit minor - breaking change. Having said that, we are now at `3.0.0`, so if we want to fix the name, now would be the time.

jrfnl (changes requested):
Thanks for making this update @sandeshjangam ! Looking good.

I've left some small remarks. Other than that, could you please add some tests for the new functions ?

Adding tests for this, is basically just a case of adding a line with a function call to each of these functions in the `WordPress/Tests/WP/AlternativeFunctionsUnitTest.inc` file and then adding the line numbers for the lines containing those function calls to the `WordPress/Tests/WP/AlternativeFunctionsUnitTest.php` file.

jrfnl:
@sandeshjangam Just checking - will you have a change to update the PR in the next few days ? I'd like to include this in WPCS 3.0.0, but if not, I'll just move it out of the milestone.

sandeshjangam:
@jrfnl Trying to finish it today itself. I will update you here.

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
sandeshjangam: Added

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
sandeshjangam: Removed. Dumb mistake

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
sandeshjangam: Added in separate group

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
sandeshjangam: Updated

dingo-d (approved):

sandeshjangam:
@jrfnl Changes done as per suggestions and added test cases also.

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -126,11 +126,37 @@ public function getGroups() {
...
+				'message'   => '%s() is discouraged. Use wp_delete_file() to delete a file.',
+				'since'     => '4.2.0',
+				'functions' => array(
+					'unlink',
+				),
+			),
+
+			'rename' => array(
+				'type'      => 'warning',
+				'message'   => '%s() is discouraged. Use WP_Filesystem::move method to rename a file.',
```
GaryJones, line 140:
```suggestion
				'message'   => '%s() is discouraged. Use WP_Filesystem::move() to rename a file.',
```

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -126,11 +126,37 @@ public function getGroups() {
...
+				'since'     => '2.5.0',
+				'functions' => array(
+					'rename',
+				),
+			),
+
+			'file_system_operations' => array(
 				'type'      => 'warning',
 				'message'   => 'File operations should use WP_Filesystem methods instead of direct PHP filesystem calls. Found: %s()',
```
GaryJones, line 149:
```suggestion
				'message'   => 'File operations should use WP_Filesystem methods instead of direct PHP filesystem calls. Found: %s().',
```

WordPress/Tests/WP/AlternativeFunctionsUnitTest.inc
```diff
@@ -78,3 +78,15 @@ function curl_version_ssl() {} // OK.
...
+rename(); // Warning.
+chgrp(); // Warning
+chmod(); // Warning
+chown(); // Warning
+is_writable(); // Warning
+is_writeable(); // Warning
+mkdir(); // Warning
+rmdir(); // Warning
+touch(); // Warning
+fputs(); // Warning
```
GaryJones, line 92:
```suggestion
chgrp(); // Warning.
chmod(); // Warning.
chown(); // Warning.
is_writable(); // Warning.
is_writeable(); // Warning.
mkdir(); // Warning.
rmdir(); // Warning.
touch(); // Warning.
fputs(); // Warning.

```

GaryJones (changes requested):

jrfnl (commented):
Other than those things pointed out by @GaryJones, this PR looks good to me. Thanks for making those updates @sandeshjangam !

sandeshjangam:
Thanks for the feedback. All changes are done now.

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -126,11 +126,37 @@ public function getGroups() {
...
 				'since'     => '2.5.0',
 				'functions' => array(
+					'chgrp',
+					'chmod',
+					'chown',
+					'is_writable',
+					'is_writeable',
+					'mkdir',
+					'rmdir',
+					'touch',
```
GaryJones, line 159: While these new function names are in alphabetical order, could you make the whole list alphabetical please? That'll make it easy to read in the future.

GaryJones (changes requested):

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
jrfnl: @GaryJones I'm fine with merging this as it is - I will sort out the alphabetization later. I have PR to update this sniff lined up as part of the PHPCSUtils/modern PHP PR series anyway, so I'll just do it in that PR.

jrfnl: (and yes, noticed it myself too and already put it on my to-do list)

GaryJones (approved):

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
GaryJones: 👍🏻

PR #2112, opened 2022-11-25
Add sniff that will check that capabilities are used correctly.

dingo-d:
The sniff will check if the functions that are accepting capabilities as the argument actually use capabilities ([as described in the support article](https://wordpress.org/support/article/roles-and-capabilities/)) and not roles.

The sniff was rewritten from the original one in #1364 to include helper functions from PHPCSUtils. The list of core functions that are using capabilities as a parameter was updated with changes up to WordPress 6.1.0. The list of capabilities was updated as well. The sniff is also compatible with PHP 8 (named arguments).

Original PRs:
https://github.com/WordPress/WordPress-Coding-Standards/pull/1364
https://github.com/WPTT/WPThemeReview/pull/36

Ref:
https://make.wordpress.org/themes/handbook/review/required/#10-classic-themes
https://developer.wordpress.org/plugins/users/roles-and-capabilities/
https://developer.wordpress.org/plugins/security/checking-user-capabilities/

Closes: #1364

### DIFF
--- a/WordPress/Sniffs/WP/CapabilitiesSniff.php
+++ b/WordPress/Sniffs/WP/CapabilitiesSniff.php
@@ -1,0 +1,478 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Sniffs\WP;
+
+use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Utils\MessageHelper;
+use PHPCSUtils\Utils\PassedParameters;
+use PHPCSUtils\Utils\TextStrings;
+use WordPressCS\WordPress\AbstractFunctionParameterSniff;
+use WordPressCS\WordPress\Helpers\MinimumWPVersionTrait;
+
+/**
+ * Check that capabilities are used correctly.
+ *
+ * User capabilities should be used, not roles or deprecated capabilities.
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since   3.0.0
+ */
+class CapabilitiesSniff extends AbstractFunctionParameterSniff {
+
+	use MinimumWPVersionTrait;
+
+	/**
+	 * List of custom capabilities.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var array
+	 */
+	public $custom_capabilities = array();
+
+	/**
+	 * The group name for this group of functions.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var string
+	 */
+	protected $group_name = 'caps_not_roles';
+
+	/**
+	 * List of functions that accept roles and capabilities as an argument.
+	 *
+	 * The functions are defined in `wp-admin/includes/plugin.php` and
+	 * `/wp-includes/capabilities.php`.
+	 * The list is sorted alphabetically.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var array<string, array> The key is the name of a function we're targetting,
+	 *                           the value is an array containing the 1-based parameter position
+	 *                           of the "capability" parameter within the function, as well as
+	 *                           the name of the parameter as declared in the function.
+	 *                           If the parameter name has been renamed since the release of PHP 8.0,
+	 *                           the parameter can be set as an array.
+	 */
+	protected $target_functions = array(
+		'add_comments_page'         => array(
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_dashboard_page'        => array(
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_links_page'            => array(
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_management_page'       => array(
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_media_page'            => array(
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_menu_page'             => array(
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_object_page'           => array( // Deprecated since WP 4.5.0.
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_options_page'          => array(
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_pages_page'            => array(
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_plugins_page'          => array(
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_posts_page'            => array(
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_submenu_page'          => array(
+			'position' => 4,
+			'name'     => 'capability',
+		),
+		'add_theme_page'            => array(
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_users_page'            => array(
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'add_utility_page'          => array( // Deprecated since WP 4.5.0.
+			'position' => 3,
+			'name'     => 'capability',
+		),
+		'author_can'                => array(
+			'position' => 2,
+			'name'     => 'capability',
+		),
+		'current_user_can'          => array(
+			'position' => 1,
+			'name'     => 'capability',
+		),
+		'current_user_can_for_blog' => array(
+			'position' => 2,
+			'name'     => 'capability',
+		),
+		'map_meta_cap'              => array(
+			'position' => 1,
+			'name'     => 'cap',
+		),
+		'user_can'                  => array(
+			'position' => 2,
+			'name'     => 'capability',
+		),
+	);
+
+	/**
+	 * List of core roles which should not to be used directly.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var array<string, bool> Role available in WP Core.
+	 */
+	private $core_roles = array(
+		'super_admin'   => true,
+		'administrator' => true,
+		'editor'        => true,
+		'author'        => true,
+		'contributor'   => true,
+		'subscriber'    => true,
+	);
+
+	/**
+	 * List of known primitive and meta core capabilities.
+	 *
+	 * Sources:
+	 * - {@link https://wordpress.org/support/article/roles-and-capabilities/ Roles and Capabilities handbook page}
+	 * - The `map_meta_cap()` function in the `src/wp-includes/capabilities.php` file.
+	 * - The tests in the `tests/phpunit/tests/user/capabilities.php` file.
+	 *
+	 * List is sorted alphabetically.
+	 *
+	 * {@internal To be updated after every major release. Last updated for WordPress 6.1.0.}
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var array<string, bool> All capabilities available in core.
+	 */
+	private $core_capabilities = array(
+		'activate_plugin'             => true,
+		'activate_plugins'            => true,
+		'add_comment_meta'            => true,
+		'add_post_meta'               => true,
+		'add_term_meta'               => true,
+		'add_user_meta'               => true,
+		'add_users'                   => true,
+		'assign_categories'           => true,
+		'assign_post_tags'            => true,
+		'assign_term'                 => true,
+		'create_app_password'         => true,
+		'create_sites'                => true,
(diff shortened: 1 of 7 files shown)


jrfnl:
One remark: as part of working with @dingo-d on this sniff, I updated the list of known capabilities with anything I could find in Core. I would very much like to invite someone to very critically review that list as there is a risk that I may have added non-officially supported capabilities.

WordPress/Tests/WP/CapabilitiesUnitTest.inc
```diff
@@ -0,0 +1,88 @@
...
+add_plugins_page( 'page_title', 'menu_title', 'super_admin', 'menu_slug', 'function' ); // Error.
+add_users_page( 'page_title', 'menu_title', 'administrator', 'menu_slug', 'function' ); // Error.
+add_management_page( 'page_title', 'menu_title', 'editor', 'menu_slug', 'function' ); // Error.
+add_options_page( 'page_title', 'menu_title', 'level_10', 'menu_slug', 'function' ); // Error.
+if ( current_user_can( 'super_admin' ) ) { } // Error.
+if( current_user_can_for_blog( '1', 'editor' ) ) { }  // Error.
+add_dashboard_page(
+	'page_title',
+	'menu_title',
+	'super_admin' /* Error */, // Error.
```
jrfnl, line 17:
```suggestion
	'super_admin' /* Comment */, // Error.
```

The `error` here is confusing as the `// Error` indicates the line should result in an error already. What is actually being checked is that a comment is ignored/disregarded, so let's just call it `/* Comment */`.

WordPress/Tests/WP/CapabilitiesUnitTest.inc
```diff
@@ -0,0 +1,88 @@
...
+	'function'
+);
+add_utility_page(
+	'page_title'
+	,'menu_title'
+	,'super_admin' // Error.
+	,'menu_slug'
+	,'function'
+	,'icon_url'
+);
```
jrfnl, line 28: Nice one! I always forget there are people who code like this! 👍🏻

WordPress/Tests/WP/CapabilitiesUnitTest.inc
```diff
@@ -0,0 +1,88 @@
...
+add_menu_page($p, $t, 'level_' . $level ); // Low severity warning.
+
+// Parse error, but just making sure we account for all possibilities.
+add_menu_page($p, $t, 'level_' 'level' ); // Low severity warning.
+
+// Another parse error, but the sniff should still handle this correctly (by bowing out).
+add_menu_page( $p, $t, /* deliberately empty */, $slug, );
+
+add_posts_page( 'page_title', 'menu_title', self /* comment */   ::   CAPABILITY, 'menu_slug', 'function' ); // Warning.
+add_posts_page( 'page_title', 'menu_title', 'admin' /* comment */   . 'istrator', 'menu_slug', 'function' ); // Warning.
```
jrfnl, line 88: Noticed some things we don't have tests for. It's arbitrary whether or not to add them as they don't necessarily test the functionality of the sniff itself, but rather the functionality of the abstract sniff, but then again, the abstract sniff doesn't have its own tests, so until then, it kind of makes sense to have some tests for these...:
* Parse error/live coding, i.e. `add_posts_page( 'page_title',` (and nothing else - should be last test in the test file).
* `$obj->current_user_can()`
* `My\NS\add_posts_page(...)`
* etc

WordPress/Tests/WP/CapabilitiesUnitTest.php
```diff
@@ -0,0 +1,90 @@
...
+
+	/**
+	 * Set warnings level to 3 to trigger suggestions as warnings.
+	 *
+	 * @param string                  $filename The name of the file being tested.
+	 * @param \PHP_CodeSniffer\Config $config   The config data for the run.
+	 *
+	 * @return void
+	 */
+	public function setCliValues( $filename, $config ) {
```
jrfnl, line 86: Suggestion: move this function to the top of the test file to be in line with other tests where we adjust these values.

jrfnl (changes requested):
Looking great! Just some final touches and then we can get this merged (after ? years... ) 🎉

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,464 @@
...
+use PHPCSUtils\Utils\MessageHelper;
+use PHPCSUtils\Utils\PassedParameters;
+use PHPCSUtils\Utils\TextStrings;
+use WordPressCS\WordPress\AbstractFunctionParameterSniff;
+use WordPressCS\WordPress\Helpers\MinimumWPVersionTrait;
+
+/**
+ * Check that capabilities are used correctly.
+ *
+ * User capabilities should be used not roles nor deprecated capabilities.
```
GaryJones, line 22:
```suggestion
 * User capabilities should be used, not roles or deprecated capabilities.
```

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,465 @@
...
+		'upload_plugins'              => true,
+		'upload_themes'               => true,
+		'view_site_health_checks'     => true,
+	);
+
+	/**
+	 * List of deprecated core capabilities.
+	 *
+	 * User Levels were deprecated in version 3.0.
+	 * To be updated after every major release.
```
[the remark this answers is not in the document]
GaryJones: Should that second sentence be marked as an internal comment?

dingo-d:
Blocked by: #2114. Once that PR is merged I will rebase and the tests should be passing.

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,470 @@
...
+				array(
+					$matched_parameter,
+					$matched_content,
+				)
+			);
+			return;
+		}
+
+		$this->phpcsFile->addWarning(
+			'"%s" is an unknown capability. Check the "%s()" function call, and that the capability is registered with WordPress. If this is a custom capability, make sure you have passed it to the custom_capabilities sniff property.',
```
jrfnl, line 460:
```suggestion
			'Found unknown capability "%s" in function call to %s(). Please check the spelling of the capability. If this is a custom capability, please verify the capability is registered with WordPress via a call to WP_Role(s)->add_cap().' . \PHP_EOL . 'Custom capabilities can be made known to this sniff by setting the "custom_capabilities" property in the PHPCS ruleset.',
```

Along the same lines as above. Trying to make the error message more actionable for end-users. Hoping this makes it clearer ?

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,470 @@
...
+	);
+
+	/**
+	 * List of core roles which should not to be used directly.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var array<string, bool> Role available in WP Core.
+	 */
+	protected $core_roles = array(
```
jrfnl, line 156: Any particular reason why this property is `protected` ? 

```suggestion
	private $core_roles = array(
```

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,470 @@
...
+	 * Sources:
+	 * - {@link https://wordpress.org/support/article/roles-and-capabilities/ Roles and Capabilities handbook page}
+	 * - The `map_meta_cap()` function in the `src/wp-includes/capabilities.php` file.
+	 * - The tests in the `tests/phpunit/tests/user/capabilities.php` file.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var array<string, bool> All capabilities available in core.
+	 */
+	protected $core_capabilities = array(
```
jrfnl, line 180: Any particular reason why this property is `protected` ? 

```suggestion
	private $core_capabilities = array(
```

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,470 @@
...
+	 *
+	 * {@internal To be updated after every major release. Last updated for WordPress 6.1.0.}
+	 *
+	 * @link https://github.com/WordPress/wordpress-develop/blob/master/tests/phpunit/tests/user/capabilities.php
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var array<string, string> All deprecated capabilities in core.
+	 */
+	protected $deprecated_capabilities = array(
```
jrfnl, line 330: Any particular reason why this property is `protected` ? 


```suggestion
	private $deprecated_capabilities = array(
```

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,470 @@
...
+
+			if ( \T_CONSTANT_ENCAPSED_STRING !== $this->tokens[ $i ]['code']
+				|| null !== $first_non_empty
+			) {
+				// Throw warning at low severity.
+				$this->phpcsFile->addWarning(
+					"Couldn't determine the value passed to the capability parameter. Please check if it matches a valid capability.",
+					$i,
+					'NeedsManualCheck',
+					array( $parameter['clean'] ),
```
jrfnl, line 385:
```suggestion
					array( $function_details['name'], $matched_content, $parameter['clean'] ),
```

Note: `$parameter['clean']` was unused without the `Found: ... ` addition.

jrfnl (changes requested):
Thanks @dingo-d for working on this and making all those changes!

Went through it one last time with a fine toothcomb:
* Ran PHPCS with `--generator=text` to see the CLI docs output.
* Ran PHPCS against the test files to verify the error messages and codes.
* Re-ran PHPCS against the test files with a custom ruleset with the `<config name="minimum_supported_wp_version" value="2.9"/>` directive set.
* Re-ran PHPCS against the test files using the CLI `--runtime-set minimum_supported_wp_version 2.9` arg.
* Ran the tests.

These final remarks are mostly just small peanuts, dotting the i's and crossing the t's. The logic of the sniff is sound and doesn't need any changes anymore, the tests are good and comprehensive, the docs are clear. We're nearly there.

WordPress/Sniffs/WP/CapabilitiesSniff.php
```diff
@@ -0,0 +1,470 @@
...
+		 */
+		$matched_parameter = TextStrings::stripQuotes( $this->tokens[ $first_non_empty ]['content'] );
+
+		if ( isset( $this->core_capabilities[ $matched_parameter ] ) ) {
+			return;
+		}
+
+		if ( empty( $matched_parameter ) ) {
+			$this->phpcsFile->addError(
+				'Found an empty capability parameter in the function call "%s()"',
```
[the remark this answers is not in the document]
dingo-d: Maybe 

```suggestion
				'An empty string is not a valid capability. Empty string found as a parameter in a function call to %s()',
```

?

WordPress/Sniffs/WP/CapabilitiesSniff.php
dingo-d: No reason, this is left over from the original PR 👍🏼

WordPress/Tests/WP/CapabilitiesUnitTest.php
```diff
@@ -0,0 +1,130 @@
...
+		$config->warningSeverity = 3;
```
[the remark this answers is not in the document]
dingo-d: Yes, only the first file has the examples with the low-severity warnings. I'll add it to the condition list 👍🏼

WordPress/Sniffs/WP/CapabilitiesSniff.php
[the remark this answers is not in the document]
jrfnl: I like where you're going with this. What about ?

```suggestion
				'An empty string is not a valid capability. Empty string found as the %s parameter in a function call to %s()"',
```
... and then adding `$function_details['name']` to  `$data` passed ?

jrfnl (approved):
> I would very much like to invite someone to very critically review that list as there is a risk that I may have added non-officially supported capabilities.

The review of the capabilities list is IMO not a blocker for merging, but would still be very welcome!

@sandeshjangam Would this ☝🏻 be something you'd be interested in helping us with ?

---

Other than the above remark from earlier on, everything looks to have been addressed.

Still saw three inconsistencies, so I've fixed those in a new commit (so you all can see easily what I changed). Feel free to squash on (or before) merge.

Approved (at long last). 🎉

GaryJones (approved):
✅

dingo-d:
I agree with the added changes 👍🏼

jrfnl:
Rebased without changes and squashed the extra commit into the original. Will merge once the build has passed.

PR #2114, opened 2022-11-25
Minimum version trait fix

dingo-d:
Discovered while working on #2112 that the command line argument wasn't being picked up by the `getCommandLineData` helper in the `get_wp_version_from_cli` method.

@jrfnl and I did all the manual tests to confirm that nothing will break with this change.

### DIFF
--- a/WordPress/Helpers/MinimumWPVersionTrait.php
+++ b/WordPress/Helpers/MinimumWPVersionTrait.php
@@ -9,7 +9,6 @@
 
 namespace WordPressCS\WordPress\Helpers;
 
-use PHP_CodeSniffer\Files\File;
 use PHPCSUtils\BackCompat\Helper;
 
 /**
@@ -81,12 +80,9 @@ trait MinimumWPVersionTrait {
 	 * @since 0.14.0
 	 * @since 3.0.0  - Moved from the Sniff class to this dedicated Trait.
 	 *               - Renamed from `get_wp_version_from_cl()` to `get_wp_version_from_cli()`.
-	 *               - Now requires the $phpcsFile object to be passed in.
-	 *
-	 * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
 	 */
-	protected function get_wp_version_from_cli( File $phpcsFile ) {
-		$cli_supported_version = Helper::getCommandLineData( $phpcsFile, 'minimum_supported_wp_version' );
+	protected function get_wp_version_from_cli() {
+		$cli_supported_version = Helper::getConfigData( 'minimum_supported_wp_version' );
 
 		if ( empty( $cli_supported_version ) ) {
 			return;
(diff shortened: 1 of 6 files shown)


jrfnl (approved):
LGTM. Very happy this bug was discovered before release!

While in a call with @dingo-d, I tested this change in two ways (both with the upcoming `WP.Capabilities` sniff - see PR #2112):
1. Manually passing `--runtime-set minimum_supported_wp_version 2.9` on the command line and running PHPCS against the test files for the `WP.Capabilities` sniff.
    Prior to this change, the sniff would still return an error, while now, with this change, it returns a warning.
    Also confirmed that changing the version to `6.1` (anything 3.0 or higher) will yield an error instead of a warning.
2. With a minimal ruleset containing the below:
    ```xml
<?xml version="1.0"?>
<ruleset name="Test">
	<config name="minimum_supported_wp_version" value="2.9"/>
	<rule ref="WordPress.WP.Capabilities"/>
</ruleset>
    ```
    Again, running PHPCS from the command line using the above ruleset correctly shows a `warning` instead of an `error` and changing the `2.9` to `3.5` brings it back to an error.
    

One sidenote - I know there are some changes in PHPCS 4.x regarding command line settings (which may have been the reason I used the other helper function originally).
As things are now, we'll figure out what is needed for PHPCS 4.x once we start running the tests against it.

GaryJones:
I'm trying to understand the change here.

It's switched from `PHPCSUtils\BackCompat\Helper::getCommandLineData()` to `PHPCSUtils\BackCompat\getConfigData()`. Both are defined [here](https://github.com/PHPCSStandards/PHPCSUtils/blob/develop/PHPCSUtils/BackCompat/Helper.php).

The `getConfigData()` implies to use `getCommandLineData()` if the `$phpcsFile` is available - judging by the changes here, it suggests it is/was available?

What's the difference between the two Helper methods, and why was one not working here? Why was the command line argument not picked up by the `getCommandLineData()` helper? Was `$phpcsFile` not available after all?

jrfnl:
> It's switched from `PHPCSUtils\BackCompat\Helper::getCommandLineData()` to `PHPCSUtils\BackCompat\getConfigData()`. Both are defined [here](https://github.com/PHPCSStandards/PHPCSUtils/blob/develop/PHPCSUtils/BackCompat/Helper.php).
> 
> The `getConfigData()` implies to use `getCommandLineData()` if the `$phpcsFile` is available - judging by the changes here, it suggests it is/was available?
> 
> What's the difference between the two Helper methods, and why was one not working here? Why was the command line argument not picked up by the `getCommandLineData()` helper? Was `$phpcsFile` not available after all?

PHPCS contains some subtle differences in how it handles various CLI arguments as well as some PHPCS cross-version differences.
I know I created both methods to overcome this, but I'd need to dig back in to be able to explain the fine details.

Looking at [the docs](https://phpcsutils.com/phpdoc/classes/PHPCSUtils-BackCompat-Helper.html), `Helper::getCommandLineData()` is more aimed at the PHPCS natively defined config settings, like `encoding` and `tab_width`, which can also be overruled via a `CodeSniffer.conf` file, while `Helper::getConfigData()` is more suited for `config` value passed using `runtime-set`.

GaryJones (approved):
✅

PR #2116, opened 2022-11-27
Group all the checks in one place and add script descriptions

### LINKED ISSUE #2115: Unify the composer testing scripts under one 'namespace'?

Should we maybe put all the scripts that are used for testing purposes (lint checks, cs checks, tests, check-complete, etc.) under one namespace (say `test`)? 

That way all the commands would be nicely grouped together when typing the `composer` command like this:

image

In addition to the above, we can also group them in one `test` command so that we can run just the `composer test` to run all of them which can be useful when working on sniffs.

The following scripts: `lint`, `lint-ci`, `check-cs`, `run-tests`, `check-complete` or `check-complete-strict` would be prefixed with `test:` prefix

And we'd add the following script to the scripts list as well:

```json
"test": [
  "@test:lint",
  "@test:lint-ci",
  "@test:check-cs",
  "@test:run-tests",
  "@test:check-complete-strict"
]
```

That would run all the 'tests' at once when the `composer test` is executed.

The impact is minimal, as only the workflows and Changelog would need to be updated (not sure if we have anything in the wiki, but that is also easy to update).

We could also add [script descriptions](https://getcomposer.org/doc/articles/scripts.md#custom-descriptions-) to explain what each command is doing.

Any thoughts?

dingo-d:
Closes #2115.

The PR groups the checks that are useful to pass before the sniff is pushed to the repo for a PR.

It also adds a description of all the custom commands defined in the `scripts` section.

I'm not too happy about the `check-complete` description, as it's a bit long so I need to split it into two lines, which doesn't look too nice.

But I couldn't omit all the necessary information about what this command is doing 🤷🏼‍♂️ 

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,17 +39,11 @@
 		"lint": [
 			"@php ./vendor/php-parallel-lint/php-parallel-lint/parallel-lint . -e php --show-deprecated --exclude vendor --exclude .git"
 		],
-		"lint-ci": [
-			"@php ./vendor/php-parallel-lint/php-parallel-lint/parallel-lint . -e php --show-deprecated --exclude vendor --exclude .git --checkstyle"
-		],
 		"check-cs": [
 			"@php ./vendor/squizlabs/php_codesniffer/bin/phpcs"
 		],
 		"fix-cs": [
 			"@php ./vendor/squizlabs/php_codesniffer/bin/phpcbf"
-		],
-		"install-codestandards": [
-			"Dealerdirect\\Composer\\Plugin\\Installers\\PHPCodeSniffer\\Plugin::run"
 		],
 		"run-tests": [
 			"@php ./vendor/phpunit/phpunit/phpunit --filter WordPress ./vendor/squizlabs/php_codesniffer/tests/AllTests.php"
@@ -59,7 +53,22 @@
 		],
 		"check-complete-strict": [
 			"@php ./vendor/phpcsstandards/phpcsdevtools/bin/phpcs-check-feature-completeness ./WordPress"
+		],
+		"check-all": [
+			"@lint",
+			"@check-cs",
+			"@run-tests",
+			"@check-complete-strict"
 		]
+	},
+	"scripts-descriptions": {
+		"lint": "Lint PHP files against parse errors.",
+		"check-cs": "Run the PHPCS script against the entire codebase.",
+		"fix-cs": "Run the PHPCBF script to fix all the autofixable violations on the codebase.",
+		"run-tests": "Run all the unit tests for the WordPress Coding Standards sniffs.",
+		"check-complete": "Check if all the sniffs have tests.",
+		"check-complete-strict": "Check if all the sniffs have unit tests and XML documentation.",
+		"check-all": "Run all checks (lint, phpcs, feature completeness) and tests."
 	},
 	"support": {
 		"issues": "https://github.com/WordPress/WordPress-Coding-Standards/issues",
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -73,7 +73,7 @@ jobs:
 
       - name: Lint PHP files against parse errors
         if: ${{ matrix.phpcs_version == 'dev-master' }}
-        run: composer lint-ci
+        run: composer lint -- --checkstyle
 
       - name: Run unit tests
         run: composer run-tests
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -95,7 +95,7 @@ jobs:
 
       - name: Lint PHP files against parse errors
         if: ${{ matrix.phpcs_version == 'dev-master' }}
-        run: composer lint-ci | cs2pr
+        run: composer lint -- --checkstyle | cs2pr
 
       - name: Run the unit tests - PHP 5.4 - 8.0
         if: ${{ matrix.php < '8.1' }}


composer.json
```diff
@@ -59,8 +59,25 @@
...
 		]
 	},
+	"scripts-descriptions": {
+		"lint": "Lint PHP files against parse errors.",
+		"lint-ci": "Lint PHP files against parse errors, but output the results as Checkstyle XML (used for CI purposes).",
+		"check-cs": "Run the PHPCS script against the entire codebase.",
+		"fix-cs": "Run the PHPCBF script to fix all the autofixable violations on the codebase.",
+		"install-codestandards": "Manually install the coding standards, so that they are discoverable by PHPCS script. Used mostly for CI purposes",
+		"run-tests": "Run all the unit tests for the WordPress Coding Standards sniffs.",
+		"check-complete": "Check if all the sniffs are feature complete (has unit tests and XML documentation), \n but don't warn if the documentation is missing for the sniffs, or if there are documentation and test files present that are not associated with a particular sniff.",
```
GaryJones, line 77: Why force the line split? Why not let the CLI's settings choose where to wrap the line?

jrfnl, line 77: What about simplifying it to describe just what it does ? (and not what it doesn't do)
```suggestion
		"check-complete": "Check if all the sniffs have tests.",
```

composer.json
```diff
@@ -59,8 +59,25 @@
...
+			"@run-tests",
+			"@check-complete-strict"
 		]
 	},
+	"scripts-descriptions": {
+		"lint": "Lint PHP files against parse errors.",
+		"lint-ci": "Lint PHP files against parse errors, but output the results as Checkstyle XML (used for CI purposes).",
+		"check-cs": "Run the PHPCS script against the entire codebase.",
+		"fix-cs": "Run the PHPCBF script to fix all the autofixable violations on the codebase.",
+		"install-codestandards": "Manually install the coding standards, so that they are discoverable by PHPCS script. Used mostly for CI purposes",
```
jrfnl, line 75: Do we still need the `install-codestandards` script ? If not, now would be the time to remove it.

composer.json
```diff
@@ -59,8 +59,25 @@
 		],
 		"check-complete-strict": [
 			"@php ./vendor/phpcsstandards/phpcsdevtools/bin/phpcs-check-feature-completeness ./WordPress"
+		],
+		"test": [
```
jrfnl, line 63:
```suggestion
		"check-all": [
```

Just a suggestion, but IMO `test` can easily be confused with running unit tests, while this script does much more.

composer.json
```diff
@@ -59,8 +59,25 @@
...
+			"@lint",
+			"@check-cs",
+			"@run-tests",
+			"@check-complete-strict"
 		]
 	},
+	"scripts-descriptions": {
+		"lint": "Lint PHP files against parse errors.",
+		"lint-ci": "Lint PHP files against parse errors, but output the results as Checkstyle XML (used for CI purposes).",
```
jrfnl, line 72: Do we actually need the separate `lint-ci` script ?


In CI, we could just do `composer lint -- --checkstyle` to achive the same, so is a separate script necessary ?

composer.json
```diff
@@ -59,8 +59,25 @@
...
+		"check-cs": "Run the PHPCS script against the entire codebase.",
+		"fix-cs": "Run the PHPCBF script to fix all the autofixable violations on the codebase.",
+		"install-codestandards": "Manually install the coding standards, so that they are discoverable by PHPCS script. Used mostly for CI purposes",
+		"run-tests": "Run all the unit tests for the WordPress Coding Standards sniffs.",
+		"check-complete": "Check if all the sniffs are feature complete (has unit tests and XML documentation), \n but don't warn if the documentation is missing for the sniffs, or if there are documentation and test files present that are not associated with a particular sniff.",
+		"check-complete-strict": "Check if all the sniffs are feature complete (has unit tests and XML documentation).",
```
jrfnl, line 78:
```suggestion
		"check-complete-strict": "Check if all the sniffs have unit tests and XML documentation.",
```

jrfnl (commented):
Just left some questions.

composer.json
dingo-d: I mean, I can leave it out, but it looks too big either way:

image

image

Ideally I'd like to shorten this, but not sure to what 🙈

dingo-d: Yup, this is definitely better 😄 👍🏼

dingo-d: Just saw the comment by Juliette, I'll apply the short explanation so this won't bother us 😄

composer.json
dingo-d: Initially, I added `check-all`, but since I've mentioned test in the issue I was thinking it's better to leave it as in the issue. But I'm glad to see our thinking matched 👍🏼

jrfnl: The name used for the description needs to be updated too. (is the reason CI isn't passing)

composer.json
GaryJones: Agreed on this change.

composer.json
GaryJones: I've found the plugin to work really well, and haven't had to run this script manually for longer than I can remember. At the very least, a `composer install` or `composer update` is enough to trigger it anyway.

Let's scrap it, to avoid any confusion.

jrfnl (approved):

composer.json
```diff
@@ -59,8 +53,23 @@
 		],
 		"check-complete-strict": [
 			"@php ./vendor/phpcsstandards/phpcsdevtools/bin/phpcs-check-feature-completeness ./WordPress"
+		],
+		"check-all": [
```
GaryJones, line 57: Should this be documented in https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/.github/CONTRIBUTING.md?

GaryJones (approved):
Approved with one suggestion.

composer.json
jrfnl: @dingo-d If you like, I can add that to the doc updates commits I have in my WIP-3.0.0 branch ?

jrfnl: Added!

Issue #1803, opened 2019-09-23
"Hidden" standard with helper tools for WPCS maintainers


jrfnl:
There are a number of lists in WPCS which should be updated on a regular basis - on every WP release - and gathering the information needed for this manually is a big maintenance burden.

Think:
* List of WP native global variables - also see #924 
* List of deprecated hooks
* List of deprecated functions
* List of deprecated function arguments
* ... etc

It would be relatively easy to create a few sniffs which could be run over WP Core to gather this information.
Those sniffs however would have no value for the end-users of WPCS, only for WPCS developers.

So, what I'm proposing is as follows:
* Add a new folder `AnalyseWPCore` with a `ruleset.xml` and sniff directory.
* Add this folder to `.gitattributes` to be `export-ignored`. This way it will not be available in released versions of WPCS, but will only be available in git clones.
* Over time add some custom sniffs to that directory which can gather the information we need to easily update the above mentioned lists.

Potentially future enhancements:
* Add some custom reports which would pre-format the output of certain sniffs in a way that the output could just be copy/pasted into the right lists.
* Add a bash script to run the reports in one go and potentially even update the lists in WPCS automatically.

Opinions ?

FYI: I've already got a variation of the `PrefixAllGlobals` sniff ready for instance to allows us to update the global variables list easily.

jrfnl:
FYI: I've made a very very preliminary start with this as the sniffs have to be checked against WP Core for the WPCS 3.0.0 release **_again_** and manually checking everything is time consuming, error prone and tedious.

jrfnl:
Prior art in this field: https://github.com/JDGrimes/wp-deprecated-code-scanner

Slightly updated version to allow it to still run, but needs a lot more work: https://github.com/jrfnl/wp-deprecated-code-scanner/tree/feature/update-for-current-php

PR #2117, opened 2022-11-28
:sparkles: New `WordPress.WP.ClassNameCase` sniff

jrfnl:
... to check that any class name references to WP native classes used the case of the class as per the class declaration.

As class names in PHP are case-insensitive, this is a `warning`, not an `error`.

**Open questions:**
* Should the same by checked for classes from external dependencies, like PHPMailer, SimplePie, Requests etc ? We could have separate arrays (groups) for each of these.
* Is the name of the sniff correct or should it be made more generic ? I mean, at this time, WP only contains classes, no interfaces, traits or enums, but that may change in the future and the logic for checking this is largely the same, so it would make sense to add those checks to this sniff.

I'm proposing to add this sniff to the WP Core ruleset, so both Core as well as plugins and themes will benefit from it.
FYI: I've ran the sniff over WP Core and it triggers no issues.

This sniff was inspired by a discussion I had with @aristath and @SergeyBiryukov earlier today.

----

### Future scope:

1. Expand the places where we check for the use of class/oo names.
    A new abstract class/oo-name sniff is expected to be added to PHPCSUtils in the foreseeable future which will also check for the use of class names in `catch` statements, type declarations etc.
    Once that class is available, we should consider switching.
2. As maintaining the list of classes will be a manual task which would need to be executed after each release, it would be good to create some tooling which can update the list automatically and/or generate the update which would need to be applied.
    Related #1803.

### DIFF
--- a/WordPress/Sniffs/WP/ClassNameCaseSniff.php
+++ b/WordPress/Sniffs/WP/ClassNameCaseSniff.php
@@ -1,0 +1,416 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Sniffs\WP;
+
+use WordPressCS\WordPress\AbstractClassRestrictionsSniff;
+
+/**
+ * Verify whether references to WP native classes use the proper casing for the class name.
+ *
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since 3.0.0
+ */
+class ClassNameCaseSniff extends AbstractClassRestrictionsSniff {
+
+	/**
+	 * List of all WP native classes.
+	 *
+	 * To be updated after every major release.
+	 *
+	 * List is sorted alphabetically and based on the WIP autoloading PR.
+	 * {@link https://github.com/WordPress/wordpress-develop/pull/3470}
+	 *
+	 * Note: this list will be enhanced in the class constructor.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var string[] The class names in their "proper" case.
+	 *               The constructor will add the lowercased class name as a key to each entry.
+	 */
+	private $wp_classes = array(
+		'_WP_Dependency',
+		'_WP_Editors',
+		'_WP_List_Table_Compat',
+		'Automatic_Upgrader_Skin',
+		'Bulk_Plugin_Upgrader_Skin',
+		'Bulk_Theme_Upgrader_Skin',
+		'Bulk_Upgrader_Skin',
+		'Core_Upgrader',
+		'Custom_Background',
+		'Custom_Image_Header',
+		'File_Upload_Upgrader',
+		'ftp_pure',
+		'ftp_sockets',
+		'Gettext_Translations',
+		'IXR_Base64',
+		'IXR_Client',
+		'IXR_ClientMulticall',
+		'IXR_Date',
+		'IXR_Error',
+		'IXR_IntrospectionServer',
+		'IXR_Message',
+		'IXR_Request',
+		'IXR_Server',
+		'IXR_Value',
+		'Language_Pack_Upgrader',
+		'Language_Pack_Upgrader_Skin',
+		'MO',
+		'NOOP_Translations',
+		'Plugin_Installer_Skin',
+		'Plugin_Upgrader',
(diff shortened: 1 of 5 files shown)


GaryJones:
What are the conditions for when using the wrong case would have an effect? What problem is this sniff aiming to solve?

jrfnl:
> What are the conditions for when using the wrong case would have an effect? What problem is this sniff aiming to solve?

At this point it doesn't solve a "problem" as such, but then most coding standards don't, as PHP doesn't care about whitespace one way or another.

The sniff is about consistency: referencing to the class name as declared when using the class name makes the names consistent with each other.

While it doesn't have any runtime implications at this moment in time (as PHP treats class names case-insensitively anyway), it could have implications for a potential future autoloading implementation for WP Core, at which time, consistency between class names as declared and used _could_ become relevant.

GaryJones (approved):
✅

jrfnl:
@GaryJones Any opinions on the open questions I posted ?

WordPress/Sniffs/WP/ClassNameCaseSniff.php
```diff
@@ -0,0 +1,416 @@
...
+ * @package WPCS\WordPressCodingStandards
+ *
+ * @since 3.0.0
+ */
+class ClassNameCaseSniff extends AbstractClassRestrictionsSniff {
+
+	/**
+	 * List of all WP native classes.
+	 *
+	 * To be updated after every major release.
```
dingo-d, line 26: We could use `@internal` tag here, like in the capabilities sniff, with the last updated WP version. 

Something like:

```php
	 * {@internal To be updated after every major release. Last updated for WordPress 6.1.0.}	
```

dingo-d (commented):
Left one suggestion.

---

> Should the same by checked for classes from external dependencies, like `PHPMailer`, `SimplePie`, `Requests` etc? We could have separate arrays (groups) for each of these.

Yes, as I think I've seen people extending the `PHPMailer` class and modifying it. We should add a separate group for this. This will be a good way to futureproof the code for any possible additions later on.

> Is the name of the sniff correct or should it be made more generic? I mean, at this time, WP only contains classes, no interfaces, traits or enums, but that may change in the future and the logic for checking this is largely the same, so it would make sense to add those checks to this sniff.

I agree. Especially with interfaces, when you may want to override the core implementation (somewhere in the future).

Maybe `ClassConstructNameCase`? Or we can leave it as `ClassNameCase` since PSR-12 refers to all these constructs as `class`

> The term "class" refers to all classes, interfaces, and traits.

Ref: https://www.php-fig.org/psr/psr-12/#4-classes-properties-and-methods

jrfnl:
> Maybe `ClassConstructNameCase`? Or we can leave it as `ClassNameCase` since PSR-12 refers to all these constructs as `class`

Maybe  `OONameCase` ?

jrfnl:
> Yes, as I think I've seen people extending the PHPMailer class and modifying it. We should add a separate group for this. This will be a good way to futureproof the code for any possible additions later on.

Done for all relevant external dependencies included with Core (see separate commit).

Only thing left to do AFAIC is decide on the final name for the sniff.

dingo-d (approved):
I'm fine to leave it as is (my PSR-12 note), we can add the additional object-oriented constructs later to be checked, and denote this in the documentation of the sniff, and the Standard.xml example.

`OONameCase` could be a bit confusing at first 🤷🏼‍♂️ 

@GaryJones what do you think?

GaryJones:
I'm fine with leaving the name as is as well - can always rip out the insides of it to a new abstract class that sniff classes for the other OO types could extend in the future as well.

jrfnl:
Okay, let's leave the name as it is. We can always use @dingo-d's PSR12 comment if people have questions about the name 👍🏻  

> can always rip out the insides of it to a new abstract class that sniff classes for the other OO types could extend in the future as well.

Well, the current abstract sniff we are using sort of checks for class and interface use already (also sniffs for `T_IMPLEMENTS` and `T_EXTENDS` can also be used with interfaces which extend other interfaces).

IIRC (but it's been a while since I worked on it), the abstract sniff in Utils will be set up to sniff for both classes as well as interfaces. (Enums and possibly traits still need to be added.)

Reason being that OO construct names can often be used in the same places, independently of the underlying type - think: type declarations, `instanceof` etc., so the logic and detection mechanism in the abstract would need to be mostly the same for all these constructs.

Individual sniffs using the PHPCSUtils abstract could, of course, choose to remove some target tokens and/or just not pass it interface names etc

Either way, that's for later. Let's get this released first.

PR #2120, opened 2022-12-03
WP/Capabilities: tests - prevent CLI value leaking

jrfnl:
In the `CapabilitiesUnitTest::setCliValues()` method, the config value for `'minimum_wp_version'` is set for two test case files, but as the sniff did not have a test case file which would be run _after_ that, it would never get _unset_ again, which means it could negatively influence other tests as the value will persist and the value from the CLI will overrule "ruleset" passed values (including the values passed using `phpcs:set`).

This fixes this issue by adding an extra test case file, which makes sure that the CLI value gets reset/deleted.

Notes:
* The "proper" way to do this would be to use the test `tear_down()`, but unfortunately, the `tear_down()` doesn't have access to the `$config`, so resetting the value won't work.
* This is a known bug in the test framework for which I have had a PR open in PHPCS since March 2020. squizlabs/PHP_CodeSniffer#2899
* There are a couple of other tests in WPCS in which the same should probably be applied, however, those (currently) don't negatively affect other tests.

### DIFF
--- a/WordPress/Tests/WP/CapabilitiesUnitTest.4.inc
+++ b/WordPress/Tests/WP/CapabilitiesUnitTest.4.inc
@@ -1,0 +1,12 @@
+<?php
+
+if ( author_can( $post, 'read' ) ) { } // OK.
+
+/*
+ * Deprecated capabilities - just making sure the CLI option is reset to the default.
+ */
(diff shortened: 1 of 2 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2121, opened 2022-12-03
Update the minimum_wp_version to WP 5.8

### LINKED ISSUE #2031: Remove deprecation Warning on sanitize_url()
## Bug Description

The use of sanitize_url() is still listed as deprecated, however as of WordPress 5.9 it has been **un** deprecated

## Minimal Code Snippet

Use sanitize_url() in a line of code.
ERROR   sanitize_url() has been deprecated since WordPress version 2.8.0. Use esc_url_raw() instead.
                  (WordPress.WP.DeprecatedFunctions.sanitize_urlFound)

This can be fixed by removing it from the deprecation list.

I have not tested against develop, but I did check here:

https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/WP/DeprecatedFunctionsSniff.php#L412

So it's still on the sniffs :)

jrfnl:
The minimum version should be three versions behind the latest WP release, so what with 6.1 being out, it should now be 5.8.

Includes adding a little more variation to the example code in the property docblock in the `MinimumWPVersionTrait`.

Includes updating the list of deprecated functions in the `WP.DeprecatedFunctions` sniff based on the WP 6.1.1 release. Input for this based on @JDGrimes's WP deprecated code scanner, though that is getting more and more difficult to run as it is no longer maintained and is throwing lots of errors.

Notes:
* The `sanitize_url()` function has been _un_deprecated in WP 5.9. See #2031.
* The replacement for the deprecated `wp_cache_reset()` function has been changed in WP Core.
    Ref: https://core.trac.wordpress.org/changeset/52705
* The `_filter_query_attachment_filenames()` function listed the wrong deprecation version in Core. WPCS now contains the correct version.
    A patch has been submitted to fix the listing in Core (and has since been committed and backported).
    Refs:
    - https://core.trac.wordpress.org/changeset/54524
    - https://core.trac.wordpress.org/ticket/56791#comment:10
* Other than the above, no verification has been executed on the newly added deprecated function information.

Includes minor tweak to make the generation of the test lines array a little more descriptive.

Previous: #1881

Closes #2031

---

### 🆕 WP/DeprecatedFunctions: temporarily exclude one test on PHP 8.2

The logic in the PHP 8.2 Tokenizer needs to be adjusted to allow for PHP 8.2 Disjunctive Normal Types in combination with `readonly` properties.

While PHP 8.2 still allows for the `readonly` keyword being used as a function call - see: https://3v4l.org/hL5da -, the PHPCS tokenizer has not been adjusted yet to allow for this in a way which is compatible with DNT.

Also see:
* The discussion about this in squizlabs/PHP_CodeSniffer 3667
* php/php-src 9512

### DIFF
--- a/WordPress/Helpers/MinimumWPVersionTrait.php
+++ b/WordPress/Helpers/MinimumWPVersionTrait.php
@@ -48,15 +48,15 @@ trait MinimumWPVersionTrait {
 	 * Example usage:
 	 * <rule ref="WordPress.WP.DeprecatedClasses">
 	 *  <properties>
-	 *   <property name="minimum_wp_version" value="4.3"/>
+	 *   <property name="minimum_wp_version" value="4.9"/>
 	 *  </properties>
 	 * </rule>
 	 *
 	 * Alternatively, the value can be passed in one go for all sniffs using it via
 	 * the command line or by setting a `<config>` value in a custom phpcs.xml ruleset.
 	 *
-	 * CL: `phpcs --runtime-set minimum_wp_version 4.5`
-	 * Ruleset: `<config name="minimum_wp_version" value="4.5"/>`
+	 * CL:      `phpcs --runtime-set minimum_wp_version 5.7`
+	 * Ruleset: `<config name="minimum_wp_version" value="6.0"/>`
 	 *
 	 * @since 0.14.0 Previously the individual sniffs each contained this property.
 	 * @since 3.0.0  - Moved from the Sniff class to this dedicated Trait.
@@ -68,7 +68,7 @@ trait MinimumWPVersionTrait {
 	 *
 	 * @var string WordPress version.
 	 */
-	public $minimum_wp_version = '5.1';
+	public $minimum_wp_version = '5.8';
 
 	/**
 	 * Overrule the minimum supported WordPress version with a command-line/config value.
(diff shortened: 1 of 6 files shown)


dingo-d (approved):
What is causing the PHP 8.2 failure?

GaryJones (approved):
✅

GaryJones:
> What is causing the PHP 8.2 test failure?

The [call](https://github.com/WordPress/WordPress-Coding-Standards/pull/2121/files#diff-fb3c621fe4dd88976d45274dcd0c5a3297b96131a6617772fb6b8d80d78f1901R364) to `readonly()`.

jrfnl:
@dingo-d Thanks for catching that - I didn't get notified of the run failure as we still have `continue-on-error` in place (let's remove that in a next PR).

And yes, @GaryJones is correct, this is about the `readonly()` test.

I've now added an extra commit to temporarily exclude that test on PHP 8.2+ as this is a issue with the PHPCS tokenizer. Function calls to `readonly()` are still allowed on PHP 8.2: https://3v4l.org/hL5da (i.e. the readonly keyword exception in PHP which was put in place at my request is still in play in PHP 8.2)

dingo-d (approved):

jrfnl:
FYI: I've opened issue #2123 as a reminder that we should remove the test exception as soon as possible.

PR #2122, opened 2022-12-04
CS: minor tweaks/don't allow warnings

jrfnl:
This commit gets rid of the last two PHPCS warnings in the code base, which allows us to run the CS check without allowing for warnings anymore.


GaryJones (approved):
✅

dingo-d (approved):

PR #2125, opened 2022-12-05
GH Actions: various tweaks

jrfnl:
### GH Actions: rename the basic QA workflow file

... to be more descriptive. The workflows in that file don't just check the rulesets, so the name as it was, was confusing.

Includes updating the links for the badge in the README.

### GH Actions: remove redundant documentation

All steps in the workflows have descriptive names, so re-iterating the steps in a "docblock" above each job is redundant and error-prone maintenance-wise.

### GH Actions: move the name key down

... to be below the matrix so it is more intuitive to see what the variables in the `name` refer to.

### GH Actions: order matrix range in logical order

... instead of in reverse order.

### GH Actions: use PHP latest

... whenever appropriate instead of PHP 7.4.

`latest` will always be the last released stable PHP version, which will be PHP 8.1 until this Thursday, when it will become PHP 8.2.

Includes a few tweaks in the `quicktest` workflow to allow for PHPUnit 7 in combination with PHP 8+.

### GH Actions/basic-qa: move a step up

Validate the `composer.json` file before making any changes to it.

### GH Actions/test: remove unused step

This step is for PHPCS 4.x, but we're not testing against PHPCS 4.x at this time and when we do enable it again, this step may well no longer be needed anymore anyway, so let's remove it for now and re-evaluate what is needed when we re-enable testing against PHPCS 4.x.

### GH Actions/test: no longer allow build failures against PHP 8.2

Includes removing the redundant `allowed_failure` matrix key, as we can use a comparison in `continue-on-error` if needed.


dingo-d (approved):

GaryJones (approved):
✅

GaryJones:
Two now-invalid required checks disabled in Settings. Not sure if it needs a rebase.

PR #2127, opened 2022-12-07
GH Actions: automate some label management

jrfnl:
This is a quite straight-forward workflow to just remove some labels which should only be on open issues/open PRs and which should be removed once an issue or PR has been closed/merged.

Just attempting to reduce yet some more manual labour.

**_Note: this is the first time I'm using this particular action runner, so consider this experimental and tweaks may be needed still._**

### DIFF
--- a/.github/workflows/manage-labels.yml
+++ b/.github/workflows/manage-labels.yml
@@ -1,0 +1,57 @@
+name: Remove outdated labels
+
+on:
+  # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
+  pull_request_target:
+    types:
+      - closed
+  issues:
+    types:
+      - closed
+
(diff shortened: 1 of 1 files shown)


jrfnl:
> Will this action only run when we close an issue/PR? Or do we want to manually trigger it as well?

Valid question, which I also considered when creating the PR.

As things are, it will only run when we close an issue/PR. AFAICS running manually would be useless as the action runner needs the context of an issue/PR to run (i.e. the issue/PR identifier).

GaryJones (approved):
✅

jrfnl:
Looks like the workflow _is_ running on the right trigger, but not doing what it is supposed to be doing. Will send in an update to see if I can get it working. (but updates will need to be merged before we can see if it works...)

PR #2128, opened 2022-12-07
Core: add sniffs to check class/constant modifier keyword order

jrfnl:
Follow up on #2102

This adds two new sniffs to the Core ruleset to cover the keyword modifier order rules for class declarations and OO constant declarations.

Refs:
* https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#visibility-and-modifier-order
* WordPress/wpcs-docs#108
* PHPCSStandards/PHPCSExtra#142
* PHPCSStandards/PHPCSExtra#143


GaryJones (approved):
✅

PR #2129, opened 2022-12-07
Replace the `WordPress.WhiteSpace.PrecisionAlignment` sniff

jrfnl:
... with a similar sniff as merged upstream in PHPCSExtra as `Universal.WhiteSpace.PrecisionAlignment`.

The upstream sniff has various improvements, including:
* Detecting precision alignment in the leading whitespace for PHP 7.3 flexible heredoc/nowdoc closing identifiers (was false negative).
* Improved handling of (trailing) whitespace on otherwise blank lines, both in code as well as in comments and HTML. This will now be ignored by default. (was false positive).
* An additional option to report on precision alignment for whitespace found on otherwise blank lines (not relevant for WP as the `Squiz.WhiteSpace.SuperfluousWhitespace` is used).
* And it also includes a fixer. Note: the fixer works based on "best guess" and may not always result in the desired indentation.

Ref:
* PHPCSStandards/PHPCSExtra#119
* PHPCSStandards/PHPCSExtra#122
* PHPCSStandards/PHPCSExtra#123
* PHPCSStandards/PHPCSExtra#124


dingo-d (approved):

GaryJones (approved):
✅  Approved, but I'm wondering how WPCS can be sure that the Universal sniff defaults match what is wanted for WP, both now and in the future (when the Universal sniff may change) (other than the fact you're the developer of the Universal sniff...)

jrfnl:
> I'm wondering how WPCS can be sure that the Universal sniff defaults match what is wanted for WP, both now and in the future (when the Universal sniff may change) (other than the fact you're the developer of the Universal sniff...)

Well, to start, the default `public` property values are "sensible" (minimal) values, so I see no reason why these would ever need to be changed for this sniff.

Secondly, any change to the default values of the `public` properties would be a BC break and can only be made in a major release and has to be clearly annotated in the changelog (semver).

PR #2131, opened 2022-12-07
GH Actions: tweak label management workflow

jrfnl:
Follow up on #2127

Let's try this without the quotes around the labels. The example in the action runners' readme didn't contain quotes, I added them as the labels we use have a `:` in it, which has special meaning in YAML workflows, but clearly that's not working.

If this doesn't work either, I may need to report it as a bug to the action runner (or find another action runner for the same).

### DIFF
--- a/.github/workflows/manage-labels.yml
+++ b/.github/workflows/manage-labels.yml
@@ -21,8 +21,8 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           labels: |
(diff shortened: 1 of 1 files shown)


GaryJones (approved):
✅

dingo-d (approved):

dingo-d:
I guess it's working now

image

😄

jrfnl:
🎉 with 🍰

PR #2134, opened 2022-12-07
Core: add sniff to discourage using reserved keywords as param names

jrfnl:
The sniff in PHPCSExtra throws a `warning`. This is in line with the wording now used in the WP Core CS handbook.

All the same, as warnings for the `src` directory are currently ignored in CI for WP Core, we could consider making this an `error` once all known instances have been fixed, which should be soon as the fixes are all lined up to be committed, and we _really_ don't want new ones being introduced (as renaming the parameter after introduction would be a BC-break). For `Extra` we could then choose to make this less opinionated and revert it to a `warning`.

Refs:
* WordPress/wpcs-docs#120
* PHPCSStandards/PHPCSExtra#80

Related to WP Core tickets:
* https://core.trac.wordpress.org/ticket/51553
* https://core.trac.wordpress.org/ticket/55327
* https://core.trac.wordpress.org/ticket/55650
* https://core.trac.wordpress.org/ticket/56788

### DIFF
--- a/WordPress/AbstractFunctionRestrictionsSniff.php
+++ b/WordPress/AbstractFunctionRestrictionsSniff.php
@@ -330,15 +330,15 @@ abstract class AbstractFunctionRestrictionsSniff extends Sniff {
 	 *
 	 * @since 0.10.0
 	 *
-	 * @param string $function Function name.
+	 * @param string $function_name Function name.
 	 * @return string Regex escaped function name.
 	 */
-	protected function prepare_name_for_regex( $function ) {
-		$function = str_replace( array( '.*', '*' ), '@@', $function ); // Replace wildcards with placeholder.
-		$function = preg_quote( $function, '`' );
-		$function = str_replace( '@@', '.*', $function ); // Replace placeholder with regex wildcard.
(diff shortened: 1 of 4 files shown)


WordPress/AbstractFunctionRestrictionsSniff.php
```diff
@@ -333,12 +333,12 @@ public function process_matched_token( $stackPtr, $group_name, $matched_content
 	 * @param string $function Function name.
```
GaryJones, line 333:
```suggestion
	 * @param string $function_name Function name.
```

GaryJones (changes requested):
Needs a quick DocBock fix.

WordPress/AbstractFunctionRestrictionsSniff.php
jrfnl: Done. Thanks.

(I only realized after I pulled this PR we hadn't fixed the names in this codebase before 🙄)

dingo-d (approved):

GaryJones (approved):
✅

PR #2139, opened 2022-12-08
PHP/PregQuoteDelimiter: add support for PHP 8.0+ named parameters

jrfnl:
2. Verified the parameter name used is in line with the name as per the PHP 8.0 release.

Includes minor tweak to the warning message to remove mention of the parameter position.


GaryJones (approved):
✅

PR #2140, opened 2022-12-08
PHP/StrictInArray: add support for PHP 8.0+ named parameters

jrfnl:
1. Changed the `$target_functions` property to not only contain information about "always needed", but also about the target parameter name and position.
2. Adjusted the logic in the sniff to allow for named parameters using the new PHPCSUtils 1.0.0-alpha4 `PassedParameters::getParameterFromStack()` method..
3. The parameter names used are in line with the name as per the PHP 8.0 release.
    Also see: https://github.com/php/doc-en/pull/2044

### DIFF
--- a/WordPress/Sniffs/PHP/StrictInArraySniff.php
+++ b/WordPress/Sniffs/PHP/StrictInArraySniff.php
@@ -9,6 +9,7 @@
 
 namespace WordPressCS\WordPress\Sniffs\PHP;
 
+use PHPCSUtils\Utils\PassedParameters;
 use WordPressCS\WordPress\AbstractFunctionParameterSniff;
 
 /**
@@ -40,10 +41,8 @@ class StrictInArraySniff extends AbstractFunctionParameterSniff {
 	/**
 	 * List of array functions to which a $strict parameter can be passed.
 	 *
-	 * The $strict parameter is the third and last parameter for each of these functions.
-	 *
 	 * The array_keys() function only requires the $strict parameter when the optional
-	 * second parameter $search has been set.
+	 * second parameter $filter_value has been set.
 	 *
 	 * @link http://php.net/in-array
 	 * @link http://php.net/array-search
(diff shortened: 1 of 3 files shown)


dingo-d (approved):

WordPress/Sniffs/PHP/StrictInArraySniff.php
```diff
@@ -51,13 +52,26 @@ class StrictInArraySniff extends AbstractFunctionParameterSniff {
 	 *
 	 * @since 0.10.0
 	 * @since 0.11.0 Renamed from $array_functions to $target_functions.
+	 * @since 3.0.0  The format of the array value has changed from boolean to array.
 	 *
-	 * @var array <string function_name> => <bool always needed ?>
+	 * @var array <string function_name> => <array>
```
GaryJones, line 57: Needs a separate issue / PR, but I wonder if we could document the array / support IDE auto-completion / static analysis a little more:

```suggestion
	 * @var array<string, array{param_position: int, param_name: string, always_needed: bool}>
```

GaryJones (approved):
✅

WordPress/Sniffs/PHP/StrictInArraySniff.php
dingo-d: Is this psalm style for docblock annotations? In phpstan that would be something along these lines

```php
 * @var []array<string, mixed>
```

But this is something that should be addressed here: https://core.trac.wordpress.org/ticket/56070 (there is a call to make a proposal for a consistent doc standard in WordPress).

jrfnl: I'm not aware that is a supported format for any form of static analysis.

I think this should be addressed separately when the docs standards are being improved (any PR for this would need to include making sure the docs for the sniffs comply).

I mean, we could use the WP array format here, but let's be honest, that's a gotspe and not recognized nor supported anywhere outside of WP anyway.

jrfnl: P.S.: I've fixed the space ;-)

jrfnl: @dingo-d That ticket is only about the tag order, not about the param type formatting.

GaryJones: When the keys are known, you can use Array Shapes:

- https://phpstan.org/writing-php-code/phpdoc-types#array-shapes
- https://blog.jetbrains.com/phpstorm/2021/07/phpstorm-2021-2-release/#Array_shapes_PHPDoc_syntax_support
- https://psalm.dev/docs/annotating_code/type_syntax/array_types/#array-shapes

It's a step further than the regular `array<string, array>` format, but relatively well supported amongst tooling.

PR #2144, opened 2022-12-09
QA: URL review across codebase

### LINKED ISSUE #2064: Review the core ruleset against the WP handbook
Before we release the WPCS 3.0 we should do a review of the [WordPress-Core ruleset](https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress-Core/ruleset.xml).

We should check the sniffs (if any need to be removed/added or reorganized) and take into account the development of the [restructuring handbook sections](https://github.com/WordPress/wpcs-docs/issues/93) issue - once we come to a consensus reorganize everything to match the handbook sections as well.

jrfnl:
### QA: ensure URLs mentioned use https

... whenever possible.

Includes replacing some URLs with alternatives if the original URL no longer contained the relevant information, though only the URLs prefixed with `http:` have been verified, not all URLs.

Includes removing/replacing a few URLs if the page referenced either doesn't exist anymore or no longer contains the relevant information.

Including making sure that links to the PHP manual use language agnostic links.

Also, URLs used in test case files have not been verified, though some have been changed to `https`. Some have been left as `http` though to make sure that if a sniff is checking the contents of text/docs and looking for URLs, both `http` as well as `https` URLs are recognized and tested.

### QA: update/fix a number of URLs

Based on a search for [`https://make.wordpress.org/`](https://make.wordpress.org/), verified all found URLs and fixed those which needed fixing.

Closes #2064


dingo-d (approved):
Looks good to me, I didn't check all the links mostly because Juliette did so on our workshop session 😄

GaryJones (approved):
✅

PR #2145, opened 2022-12-09
Sniff::$formattingFunctions: add `antispambot()` to the list

### LINKED ISSUE #1246: Add antispambot to list of functions that don't need escaping
The function `antispambot` takes an email address, and obfuscates it by encoding it using HTML entities. [Documentation](https://codex.wordpress.org/Function_Reference/antispambot)

For example, this code:

```php
echo antispambot( 'john.doe@mysite.com' );
```

This will output the email like this in the HTML:

```
&#106;&#111;h&#110;&#46;&#100;&#111;&#101;&#64;mysit&#101;.&#99;&#111;&#109;
```

PHPCS complains about that code, saying it should require escaping, when it doesn't.

jrfnl:
Fixes #1246

Includes unit tests.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -358,14 +358,15 @@ abstract class Sniff implements PHPCS_Sniff {
 	 * @var array
 	 */
 	protected $formattingFunctions = array(
-		'array_fill' => true,
-		'ent2ncr'    => true,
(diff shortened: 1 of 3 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2148, opened 2022-12-14
New `WPHookHelper` utility class, incl PHP 8.0+ named param helper

jrfnl:
### Sniff: remove the $hookFunctions property

... as it is not currently used anywhere in the WPCS codebase.

### Move "hook" related utilities to dedicated WPHookHelper

The "hook" related property is only used by a small set of sniffs, so are better placed in a dedicated class.

This commit moves the `$hookInvokeFunctions` property to a new `WordPressCS\WordPress\Helpers\WPHookHelper` class and starts using that class in the relevant sniffs.

Note:
* The property is now `private` (and `static`).
* Access to the property can be obtained via the new `WPHookHelper::get_functions()` method.
* This new method can also pre-filter the list to only return thos functions which are not specifically to invoke deprecated hooks.

Related to #1465

### WPHookHelper: add new get_hook_name_param() method

... to support retrieving the hook name parameters from a stack of parameters, while supporting PHP 8.0+ function calls using named arguments.

This new method will be tested via the implementations in the `ValidHookName` and the `PrefixAllGlobals` sniffs (upcoming).

### DIFF
--- a/WordPress/Helpers/WPHookHelper.php
+++ b/WordPress/Helpers/WPHookHelper.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Helpers;
+
+/**
+ * Helper utilities for recognizing functions related to the WP Hook mechanism.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @since   3.0.0 The property in this class was previously contained in the
+ *                `WordPressCS\WordPress\Sniff` class and has been moved here.
+ */
+final class WPHookHelper {
+
+	/**
+	 * A list of functions that invoke WP hooks (filters/actions).
+	 *
+	 * @since 0.10.0
+	 * @since 0.11.0 Changed from public static to protected non-static.
+	 * @since 3.0.0  - Moved from the Sniff class to this class.
+	 *               - The property visibility has changed from `protected` to `private static`.
+	 *                 Use the `get_function_names()` method for access.
+	 *
+	 * @var array<string, bool>
+	 */
+	private static $hookInvokeFunctions = array(
+		'do_action'                => true,
+		'do_action_ref_array'      => true,
+		'do_action_deprecated'     => true,
+		'apply_filters'            => true,
+		'apply_filters_ref_array'  => true,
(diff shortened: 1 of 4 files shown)


dingo-d (approved):

WordPress/Helpers/WPHookHelper.php
```diff
@@ -0,0 +1,114 @@
...
+	 *               - The format of the value has changed from a non-relevant boolean to
+	 *                 an array with the parameter position and name(s) for the hook name parameter.
+	 *
+	 * @var array<string, <string, int|string|array>> Function name as key, array with target
+	 *                                                parameter position and name(s) as value.
+	 */
+	private static $hookInvokeFunctions = array(
+		'do_action' => array(
+			'position' => 1,
+			'name'     => 'hook_name',
```
GaryJones, line 40: In what way would this key-value be used?

GaryJones (changes requested):

WordPress/Helpers/WPHookHelper.php
```diff
@@ -0,0 +1,114 @@
...
+	 *
+	 * @since 0.10.0
+	 * @since 0.11.0 Changed from public static to protected non-static.
+	 * @since 3.0.0  - Moved from the Sniff class to this class.
+	 *               - The property visibility has changed from `protected` to `private static`.
+	 *                 Use the `get_function_names()` method for access.
+	 *               - The format of the value has changed from a non-relevant boolean to
+	 *                 an array with the parameter position and name(s) for the hook name parameter.
+	 *
+	 * @var array<string, <string, int|string|array>> Function name as key, array with target
```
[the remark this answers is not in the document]
jrfnl: > The possibility of `name` being an array isn't represented in the rest of the DocBlock, the current values, or in its usage.

But it should still be mentioned to prevent confusion later. The `name` key can have either a string or an array as value to allow for parameters which have been renamed since the release of PHP 8.0,. And as WP still doesn't officially support named parameters, the chances of a parameter rename happening at some point in the future is realistic.

Also, on the next line , it says "parameter position and name(s) as value.", same as in the changelog entry, so the possibility of `name` being an array entry is loosely mentioned already in the DocBlock.

WordPress/Helpers/WPHookHelper.php
jrfnl: It is used in the `get_hook_name_param()` function, which is being added in this PR, to allow for supporting PHP 8.0+ function calls using named parameters. The implementation in the two sniffs using this helper class will follow in follow-up PRs.

GaryJones (approved):
✅

PR #2149, opened 2022-12-14
MinimumWPVersionTrait: add new `wp_version_compare()` method (bug fix)

jrfnl:
... and implement it in the relevant places.

The PHP native `version_compare()` function expects two "PHP-standardized" version number strings and has some quirks if the numbers are not "PHP-standardized", i.e. one of the two numbers is a `#.#`-style version number, while the other is a `#.#.#`-style version number.
See: https://3v4l.org/OUMWE

This helper method works round those quirks and will ensure that the version comparisons done with the `$this->minimum_wp_version` are correct.

This fixes some subtle (and unreported) bugs with the version comparisons.

### DIFF
--- a/WordPress/Helpers/MinimumWPVersionTrait.php
+++ b/WordPress/Helpers/MinimumWPVersionTrait.php
@@ -97,4 +97,27 @@ trait MinimumWPVersionTrait {
 		}
 	}
 
+	/**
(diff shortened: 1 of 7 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2150, opened 2022-12-15
Arrays/ArrayKeySpacingRestrictions: small bug fix

jrfnl:
If an auto-fix contains more than one instruction, it should be wrapped in `beginChangeset()`/`endChangeset()` function calls.

This auto-fix code could potentially touch two tokens, so should included the changeset wrapper function calls.

FYI: the wrapper calls are used to potentially revert changes if they would conflict with changes made by another sniff in the same fixer loop. In that case, the changes are attempted again in the next fixer loop, with the fixer trying up to 50 loops before giving up.


dingo-d (approved):

GaryJones (approved):
✅

PR #2151, opened 2022-12-15
Arrays/CommaAfterArrayItem: minor tweaks

jrfnl:
Use the PHPCSUtils `Collections::arrayOpenTokensBC()` to future-proof the sniff for issues with the array tokenization, which keep cropping up every other PHP release.


dingo-d (approved):

GaryJones (approved):
✅

PR #2152, opened 2022-12-15
PHP/YodaConditions: implement PHPCSUtils + more tweaks

jrfnl:
### PHP/YodaConditions: minor tweaks

I've only done a cursory review of this sniff, i.e. not a deep dive analysis as [the intention is for this sniff to be removed](https://make.wordpress.org/core/2022/06/14/upcoming-disallow-assignments-in-conditions-and-remove-the-yoda-condition-requirement-for-php) as soon as consensus is reached.

With that in mind, I can't justify spending much time on improving this seriously flawed sniff, which doesn't protect against anything.


dingo-d (approved):

GaryJones (approved):
✅

PR #2154, opened 2022-12-15
Files/FileName: implement PHPCSUtils, exclude test files and more tweaks

### LINKED ISSUE #1995: FileNameSniff: do not apply to test files
### Current situation

The WordPress Coding Standards currently enforce file names to:
1. Be all lowercase;
2. Use hyphens instead of underscores as word separators;
3. Have a `class-` prefix for files containing classes.
4. Have a `-template` suffix for files containing template tags.

The first two rules are always applied.
The third rule is selectively applied to files containing classes which are not test classes.
The fourth rule is selectively applied to the `wp-includes` folder only.

Like with any sniff/error code, this sniff (or any of the error codes) can be disabled from a custom ruleset, either completely or selectively for certain files/directories.

### What needs to change and why

At this time, the WP Core test suite supports PHPUnit 5-7, but to allow for testing on PHP 8.1, it is very likely that a wider range of (newer) PHPUnit versions will need to be supported soon.

Aside from that, the WPCS ruleset is not only used by WP Core, but also by a large number of WP related projects, such as plugins and themes, which may already be testing against more recent versions of PHPUnit.

As of PHPUnit 9.1.0, support for test file names which do not match the class name of the test class within, is deprecated. Support is expected to be removed completely in a future PHPUnit version.
Effectively, this means that PHPUnit expects PSR4 to be followed for test classes.

Note: The file name comparison with the class name in PHPUnit is done in a case-sensitive manner (as per PSR4).

**In effect, this means that the file name rules - as they are - should no longer be applied to test classes as it will make the test classes incompatible with PHPUnit >= 9.1.0.**

The `FileName` sniff will need to be adjusted to ignore test classes completely.
Additionally a new sniff should be enabled/created to *enforce* PSR4 compliant file names for all test class files.

Ref: https://github.com/sebastianbergmann/phpunit/pull/4109/

Also related: https://core.trac.wordpress.org/ticket/53010

jrfnl:
### Files/FileName: remove some redundant code

... which is no longer needed now support for PHPCS < 3.7.1 has been dropped.

(check the diff ignoring whitespace differences to see the actual change)

### Files/FileName: split the logic off into three methods

... to lower the cognitive complexity and nesting levels of the code.

Includes two additional tests to ensure all code is properly covered.

### Files/FileName: minor efficiency tweak

As both arrays have unique string keys, this will do fine.

### Files/FileName: exclude test files altogether

As per #1995, test classes should be excluded completely from this rule to prevent issues with PHPUnit > 9.1.0.

Includes test.

Fixes #1995

Refs:
* WordPress/wpcs-docs#123

### Files/FileName: minor docs tweak


dingo-d (approved):

GaryJones (changes requested):

GaryJones:
I can see there are some test files with extra `.` in (and not just underscores), but I was wondering if an incorrect file name like `class.my-class.php` might be detected, as the checking for hyphenation only replaces underscores.

I also see `substr( $file_name, 0, -4 )`, which is assuming that a PHP-rendered file will always have a three letter file extension, but I wonder if someone could break that assumption with `--extensions=php3,myphp` or similar.

jrfnl:
> I can see there are some test files with extra `.` in (and not just underscores), but I was wondering if an incorrect file name like `class.my-class.php` might be detected, as the checking for hyphenation only replaces underscores.

I don't think it is, but did you test it ? If a test would proof it not to be handled, it might be good to open an issue about that (out of scope for this PR).

> I also see `substr( $file_name, 0, -4 )`, which is assuming that a PHP-rendered file will always have a three letter file extension, but I wonder if someone could break that assumption with `--extensions=php3,myphp` or similar.

They most definitely could, but is that likely for a WP file or plugin ? I mean, for that to work on a server, either additional `.htaccess` rules or nginx rules (or similar) would need to be set up to support that.

All the same, I agree it may be good to clarify the file name rule in the handbook to be explicit about file extensions. Once that is done, this sniff would need to be updated. (again, out of scope for this PR)

GaryJones:
I've not tested, but seems that any fixes are felt to be outside of this PR anyway.

GaryJones (approved):

PR #2155, opened 2022-12-17
Sniff: move the `has_nonce_check()` method to the `Security/NonceVerification` Sniff

jrfnl:
The `Sniff::has_nonce_check()` method and the associated property `$nonceVerificationFunctions` are _only_ used by the `WordPress.Security.NonceVerification` sniff, so these don't belong in the generic base `Sniff` class.

A [code search](https://sourcegraph.com/search?q=context:global+has_nonce_check%28+-file:%5E%28.*/%29%3Fvendor/wp-coding-standards/.*%24&patternType=standard&case=yes&sm=0) also shows these are not used by sniffs from external standards, so I'm also making both `private`.


GaryJones (approved):
✅

dingo-d (approved):

PR #2156, opened 2022-12-17
Sniff: move the `$cache*Functions` properties to the `DB/DirectDatabaseQuery` Sniff

jrfnl:
The `Sniff::$cacheGetFunctions`, `Sniff::$cacheSetFunctions` and `Sniff::$cacheDeleteFunctions` properties are _only_ used by the `WordPress.DB.DirectDatabaseQuerySniff` sniff, so these don't belong in the generic base `Sniff` class.

A [code search](https://sourcegraph.com/search?q=context:global+-%3Ecache%28Get%7CSet%7CDelete%29Functions+-file:%5E%28.*/%29%3Fvendor/wp-coding-standards/.*%24&patternType=regexp&case=yes&sm=0) also shows these properties are not used by sniffs from external standards, so this change should have no noticeable impact.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -423,54 +423,6 @@ abstract class Sniff implements PHPCS_Sniff {
 	);
 
 	/**
-	 * A list of functions that get data from the cache.
-	 *
-	 * @since 0.6.0
-	 * @since 0.11.0 Changed from public static to protected non-static.
-	 *
-	 * @var array
-	 */
-	protected $cacheGetFunctions = array(
-		'wp_cache_get' => true,
-	);
-
-	/**
-	 * A list of functions that set data in the cache.
-	 *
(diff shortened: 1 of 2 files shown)


GaryJones (approved):
Why were these put in the Sniff class originally?

jrfnl:
> Why were these put in the Sniff class originally?

I wondered the exact same thing. This was before my time. I suspect they were put there with the _idea_ that more sniffs _may_ at some point need access and that never happened, but who knows ... 🤷🏻‍♀️

jrfnl:
Note: if at some hypothetical point in the future, there would be a need for these lists to be more widely available, they can always still be moved to a new `Helper` class or trait, but at this point in time, I don't see the need for that.

dingo-d (approved):

PR #2157, opened 2022-12-17
Move "property handling" related utilities to dedicated `RulesetPropertyHelper`

jrfnl:
The "property handling" related utilities are only used by a limited set of sniffs, so are better placed in a dedicated class.

This commit moves the `merge_custom_array()` method to a new `WordPressCS\WordPress\Helpers\RulesetPropertyHelper` and starts using that class in the relevant sniffs.

Note: the function has been moved without changes.

Related to #1465


GaryJones (approved):
✅

dingo-d (approved):

PR #2159, opened 2022-12-19
Move "global variable" related utilities to dedicated `WPGlobalVariablesHelper`

jrfnl:
The "global variable" related utilities are only used by a small number of sniffs, so are better placed in a dedicated class.

This commit moves the `$wp_globals` property to a new `WordPressCS\WordPress\Helpers\WPGlobalVariablesHelper` and starts using that class in the relevant sniffs.

Note:
* The property is now `private` (and `static`). This protects the property against inadvertent changes.
* Access to the property can be obtained via the new `WPGlobalVariablesHelper::get_names()` method.
* Alternatively, the new `WPGlobalVariablesHelper::is_wp_global()` method can be used to check if a given variable name is the name of a WP global variable.

Related to #1465


GaryJones (approved):
✅

WordPress/Helpers/WPGlobalVariablesHelper.php
```diff
@@ -0,0 +1,313 @@
...
+	 * List of global WP variables.
+	 *
+	 * @since 0.3.0
+	 * @since 0.11.0 Changed visibility from public to protected.
+	 * @since 0.12.0 Renamed from `$globals` to `$wp_globals` to be more descriptive.
+	 * @since 0.12.0 Moved to the `Sniff` class from the WordPress.Variables.GlobalVariables sniff.
+	 * @since 3.0.0  - Moved from the Sniff class to this class.
+	 *               - The property visibility has changed from `protected` to `private static`.
+	 *                 Use the `get_names()` method for access or use the `is_wp_global()` method
+	 *                 to check if a new is on the list.
```
jrfnl, line 31:
```suggestion
	 *                 to check if a name is on the list.
```

Oops - just noticed this typo. Fixed now.

dingo-d (approved):

PR #2160, opened 2022-12-19
DateTime/CurrentTimeTimestamp: add support for PHP 8.0+ named parameters + minor tweak

jrfnl:
### DateTime/CurrentTimeTimestamp: add support for PHP 8.0+ named parameters

### DateTime/CurrentTimeTimestamp: minor code simplification

By switching to using that key, some additional logic which was in place to prevent false negatives when the parameter contained comments can be removed.

Safeguarded by pre-existing unit test on line 17-20.


GaryJones (approved):
✅

dingo-d (approved):

PR #2161, opened 2022-12-19
PHP/IniSet: add support for PHP 8.0+ named parameters + bug fix

jrfnl:
### PHP/IniSet: add support for PHP 8.0+ named parameters

    Ref: https://www.php.net/manual/en/function.ini-set.php

### PHP/IniSet: prevent some false positives

By switching to using that key, some false positives/incorrect error message being thrown situations get prevented.

Includes ensuring that the information displayed in the error message will also no longer contain comments.

### PHP/IniSet: minor code simplification

Check using `empty()` instead `! isset()` for efficiency as it will prevent a potentially unncessary call to `in_array()`.

### PHP/IniSet: minor message tweak for clarity


GaryJones (approved):
✅

dingo-d (approved):

PR #2162, opened 2022-12-19
PHP/NoSilencedErrors: implement PHPCSUtils methods

jrfnl:
### PHP/NoSilencedErrors: add extra test

... to cover a previously uncovered condition (and ensure that the configurable `$context_length` parameter is handled correctly).

### PHP/NoSilencedErrors: implement PHPCSUtils methods

As the methods from `GetTokensAsString` work slightly differently from the PHPCS native `File::getTokensAsString()` method, this would result in a "one token longer" code snippet. The `-1` in the function call in line 225 has been added to prevent this change in behaviour without the need for people to update the `context_length` property value in their custom ruleset (which would be an unnecessary BC-break).

Functionally, using the `GetTokensAsString::compact()` method will result in the code snippet in the error message being "cleaner" as those will no longer contain comments found in the code. The PHPCSUtils method will also "compact" the whitespace, so no need for the `str_replace()` call anymore.

No additional tests added, the existing tests (line 58) already cover this.

Note: it is still the intention for a similar sniff to become available via PHPCSExtra, but that will be in a later release.

### DIFF
--- a/WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
+++ b/WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
@@ -10,6 +10,8 @@
 namespace WordPressCS\WordPress\Sniffs\PHP;
 
 use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\BackCompat\BCFile;
+use PHPCSUtils\Utils\GetTokensAsString;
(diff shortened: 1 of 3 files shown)


GaryJones (approved):
✅

dingo-d (approved):

PR #2163, opened 2022-12-19
PHP/StrictInArray: prevent some false positives

jrfnl:

By switching to using that key, some false positives get prevented.


GaryJones (approved):
✅

dingo-d (approved):

PR #2165, opened 2022-12-20
Move "is assignment" related utility to dedicated `VariableHelper`

jrfnl:
The "is assignment" related utility method is only used by a small set of sniffs, so is better placed in a dedicated class.

As this method also primarily expects a `T_VARIABLE` token as input, the `VariableHelper` class seems appropriate.

This commit moves the `is_assignment()` method to the new `WordPressCS\WordPress\Helpers\VariableHelper` and starts using that class in the relevant sniffs.

Related to #1465

### DIFF
--- a/WordPress/Helpers/VariableHelper.php
+++ b/WordPress/Helpers/VariableHelper.php
@@ -179,4 +179,60 @@ final class VariableHelper {
 
 		return false;
 	}
+
+	/**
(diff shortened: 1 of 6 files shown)


GaryJones (approved):
✅

dingo-d (approved):

PR #2166, opened 2022-12-20
NamingConventions/ValidHookName: implement PHPCSUtils and support modern PHP

jrfnl:
### NamingConventions/ValidHookName::transform_complex_string(): simplify with PHPCSUtils

Simplify the `transform_complex_string()` method by using the PHPCSUtils utilities to split a text string containing interpolated variables or expressions, instead of using custom logic to do this in the sniff.

Functionally this has no impact. The "old" code already handled complex expressions correctly.
Performance-wise, this could (should) have a small, positive impact as PHPCSUtils caches the results of the splitting of the text string and using the Utils functions allows for simplifying the logic in the function itself, which is called three times for each double quoted teẋt string.

Includes a few extra tests with some more complex variables/expressions.

### NamingConventions/ValidHookName: bug fix - disregard variable function calls

Iterates on previously pulled PR #2056.

As function calls via closure callbacks are getting more common, probably a good idea to make the sniff ignore parameters passed to variable function calls as well.

Includes a test which would previously result in false positives from the sniff.

### NamingConventions/ValidHookName: add support for PHP 8.0+ named parameters

1. Adjusted the way the correct parameter is retrieved to use the new `WPHookHelper::get_hook_name_param()` method, which uses the PHPCSUtils 1.0.0-alpha4 `PassedParameters::getParameterFromStack()` method under the hood.

Note: function calls with named arguments are not supported for the `do_action()` or `apply_filters()` functions as those have a parameter with a spread operator (but that's irrelevant for the sniff).

### DIFF
--- a/WordPress/Sniffs/NamingConventions/ValidHookNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidHookNameSniff.php
@@ -94,6 +94,11 @@ class ValidHookNameSniff extends AbstractFunctionParameterSniff {
 	 */
 	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
 
+		$hook_name_param = WPHookHelper::get_hook_name_param( $matched_content, $parameters );
+		if ( false === $hook_name_param ) {
+			return;
+		}
+
 		$regex = $this->prepare_regex();
 
 		$case_errors    = 0;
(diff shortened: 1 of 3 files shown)


GaryJones (approved):
✅

dingo-d (approved):

PR #2167, opened 2022-12-20
NamingConventions/ValidPostTypeSlug: implement PHPCSUtils and support modern PHP

jrfnl:
### NamingConventions/ValidPostTypeSlug: minor error message tweak

Always put quotes around the found value, even when the value was in a heredoc/nowdoc (which doesn't natively have quotes).

### NamingConventions/ValidPostTypeSlug: add support for PHP 7.3+ flexible heredoc/nowdocs

Includes tests to safeguard these are handled correctly.

Note: the tests are in a separate file for maintainability as any code which would be placed _after_ the flexible heredoc/nowdoc would not be tokenized correctly on PHP < 7.3, making the tests maintaining error prone if the tests were in the same file.

### NamingConventions/ValidPostTypeSlug: add support for PHP 8.0+ named parameters

2. Verified the parameter name used is in line with the name as per the WP 6.1 release.

Includes moving the `Empty` error up as the sniff contained a presumption of a correct function call passing the required parameter, while that presumption may be incorrect.

### NamingConventions/ValidPostTypeSlug: use "clean" info in error message

By switching to using that key, the error message displayed will be _cleaner_.

Includes unit tests demonstrating the issue.


GaryJones (approved):
✅

dingo-d (approved):

PR #2168, opened 2022-12-20
Security/PluginMenuSlug: implement PHPCSUtils and support modern PHP

jrfnl:
### Security/PluginMenuSlug: add support for PHP 8.0+ named parameters

### Security/PluginMenuSlug: minor test tweak

Adjust one test to use lowercase `__file__` to safeguard that the sniff correctly recognizes that these magic constants are case insensitive.

### Security/PluginMenuSlug: improve error position precision

Throw the error on the token containing the `__FILE__` magic constant instead of on the function call token.


GaryJones (approved):
✅

dingo-d (approved):

PR #2169, opened 2022-12-20
WP/AlternativeFunctions: implement PHPCSUtils and support modern PHP

jrfnl:
### WP/AlternativeFunctions: order lists alphabetically

Follow up on #2108

Orders function name lists alphabetically. The group ordering, while not in alphabetic order, is left as-is for now.

This commit does not add or delete any functions from the lists.

### WP/AlternativeFunctions: add missing `@since` tags

... for properties which were added to the sniff after the initial conception.

### WP/AlternativeFunctions: change property visibility

... from `protected` to `private` for properties which are only intended to be used by the sniff itself anyway.

### WP/AlternativeFunctions: bug fix - special casing for parse_url was not applied

The `wp_parse_url()` alternative to `parse_url()` was introduced in WP 4.4.0, but didn't have support for the `$component` parameter until WP 4.7.0.

For that reason, special casing was put in place in PR #1369 (WPCS 1.0.0) to examine whether the `$component` parameter was passed to `parse_url()` or not and to only advise the switch to `wp_parse_url()`:
1. If the `$component` parameter was _not_ used and the `minimum_wp_version` was `4.4` or higher.
2. If the `$component` parameter **_was_** used and the `minimum_wp_version` was `4.7` or higher.

Unfortunately, the special casing was looking for the wrong function name match, so has never worked.

The additional tests now added demonstrate the bug.

This commit, combined with the previously pulled `version_compare()` related fix from #2149, fixes the bug.

---

Includes a follow up to #2121

The "final" `// phpcs:set` annotations in the test case file should always reset to the default value for the property. The default value which was being used, however, was out of date.

Fixed now.

### WP/AlternativeFunctions: add support for PHP 8.0+ named parameters [1] / strip_tags()

1. Adjusted the way it is determined whether the PHP native function can be replaced with the WP version of the same using the updated PHPCSUtils 1.0.0-alpha4 `PassedParameters` functionality.
2. Verified the parameter name used is in line with the name as per the PHP 8.0 release.
    Ref: https://www.php.net/manual/en/function.strip-tags

### WP/AlternativeFunctions: add support for PHP 8.0+ named parameters [2] / parse_url()

1. Adjusted the way it is determined whether the PHP native function can be replaced with the WP version of the same using the updated PHPCSUtils 1.0.0-alpha4 `PassedParameters` functionality.
2. Verified the parameter name used is in line with the name as per the PHP 8.0 release.
    Ref: https://www.php.net/manual/en/function.parse-url

### WP/AlternativeFunctions: add support for PHP 8.0+ named parameters [3] / file_get_contents()

1. Adjusted the way it is determined whether the PHP native function can be replaced with the WP version of the same using the updated PHPCSUtils 1.0.0-alpha4 `PassedParameters` functionality.
    Ref: https://www.php.net/manual/en/function.file-get-contents

### WP/AlternativeFunctions: add support for PHP 8.0+ named parameters [4] / readfile/fopen/file_put_contents()

1. Adjusted the way it is determined whether the PHP native function can be replaced with the WP version of the same using the updated PHPCSUtils 1.0.0-alpha4 `PassedParameters` functionality.
    Refs:
    * https://www.php.net/manual/en/function.readfile
    * https://www.php.net/manual/en/function.fopen
    * https://www.php.net/manual/en/function.file-put-contents

### WP/AlternativeFunctions: prevent some false positives / file_get_contents()

By switching to using that key, some false negatives and false positives get fixed.

### WP/AlternativeFunctions: minor code tweak / readfile/fopen/file_put_contents()

While this doesn't make a difference functionally (no false positives/negatives), it may improve performance a tiny bit as there is less text for the `is_local_data_stream()` method to handle.


dingo-d (approved):
👍🏼

GaryJones (approved):
✅

PR #2170, opened 2022-12-20
WP/CapitalPDangit: implement PHPCSUtils and support modern PHP

jrfnl:
### WP/CapitalPDangit: use PHPCSUtils

Use the `Collections::arrayOpenTokensBC()` method to retrieve the tokens which are used for arrays.

Also changes the order in which array tokens are examined to first examine long array tokens as those are the more likely token to encounter.

### WP/CapitalPDangit: prevent false positives for lists

Along the same lines as keys used in arrays, keys used in lists (PHP 7.2+) would be a false positive in the majority of cases, so skip over both long as well as short lists.

Includes tests.

### WP/CapitalPDangit: always skip over array access keys

There was already code in place to ignore text strings used in array access keys.

This change makes the "skipping over" more efficient by doing so on the open bracket and doing so earlier in the sniff.
This also prevents false positives if the array key is not just a plain text string.

Covered by pre-existing tests + an additional test.

### WP/CapitalPDangit: add tests covering PHP 8.1+ enums

The `Tokens::$ooScopeTokens` array contains the `T_ENUM` token since PHPCS 3.7.0, so enums are automatically supported now, but let's safeguard this with a test anyway and update the warning message as well.

### WP/CapitalPDangit: add tests with namespaced names containing reserved keyword

... where the keyword is a token we are listening for.

### WP/CapitalPDangit: add tests with PHP 7.3 flexible heredoc/nowdoc

Note: these are in a separate file for maintainability as any code which would be placed _after_ the flexible heredoc/nowdoc would not be tokenized correctly on PHP < 7.3, making the tests maintainer error prone if the tests were in the same file.


GaryJones (approved):
✅

dingo-d (approved):

PR #2172, opened 2022-12-21
WP/CronInterval: implement PHPCSUtils and support modern PHP 

jrfnl:
### WP/CronInterval: implement PHPCSUtils and other minor tweaks

This commit contains no functional changes.

### WP/CronInterval: add some tests for edge cases

... to add code coverage for some edge cases already accounted for in the code.

### WP/CronInterval: bug fix / prevent parse errors in eval-ed code

Follow up on #2026 (not yet released) which added handling of time calculations wrapped in parentheses to the sniff.

This improves on the initial fix by making the code more stable and preventing parse errors in the code (which will eventually be eval-ed) due to unmatched parentheses.

### WP/CronInterval: add support for PHP 7.4+ arrow function callbacks

While these may not be the first choice for returning an adjusted array value, like is required for adjusting the `cron_schedules`, it is still perfectly possible to do so, so the sniff should handle callbacks declared via arrow functions correctly.

Includes unit tests.

### WP/CronInterval: add support for PHP 7.4/8.1 numbers

PHP 7.4 introduced numeric literals with underscores.
PHP 8.1 introduced explicit octal notation.

While PHPCS backfills the tokens for this, this sniff uses an `eval()` call and if either of these PHP 7.4/8.1 syntaxes would be passed to this `eval()` while PHPCS is being run on PHP < 7.4, this will result in a parse error on the code passed to `eval()`.

Luckily the sniff already contains protection against that, but that means, the sniff would throw the "Detected changing of cron_schedules, but could not detect the interval value." warning, instead of okay-ing correct values/throwing an error for incorrect values when either of those syntaxes is used.

By using the PHPCSUtils `Numbers::getCompleteNumber()` method, we can get access to the decimal value of the encountered number, which allows us to bypass the issue by using that in the code string passed to `eval()`.

Includes unit tests.

### WP/CronInterval: (properly) add support for PHP 8.0+ named parameters

Follow up to PR #2090 which already added preliminary support for named parameters to the sniff.

This commit now adds tests with named arguments in the `add_filter()` function calls and safeguards the prevention for false positives on callbacks containing a "cron_schedules" text string.

### WP/CronInterval: add support for PHP 8.1 first class callables

PHP 8.1 introduced a first class callables syntax, which can be used to express callbacks.

This commit adds support for callbacks passed to the `add_filter()` function expressed using this new syntax.

This prevents some false positives as well as prevents the "interval undetermined" warning being displayed for first class callables, where the interval _is_ actually determinable.

Note: there are more flaws in this sniff, like the scoping of the function search not being limited to the class/global scope depending on how the callback is defined, but that is outside the scope of this commit.

Includes unit tests.

Ref: https://wiki.php.net/rfc/first_class_callable_syntax

### DIFF
--- a/WordPress/Sniffs/WP/CronIntervalSniff.php
+++ b/WordPress/Sniffs/WP/CronIntervalSniff.php
@@ -14,6 +14,7 @@ use PHP_CodeSniffer\Util\Tokens;
 use PHPCSUtils\Tokens\Collections;
 use PHPCSUtils\Utils\Arrays;
 use PHPCSUtils\Utils\FunctionDeclarations;
+use PHPCSUtils\Utils\Numbers;
 use PHPCSUtils\Utils\PassedParameters;
 use PHPCSUtils\Utils\TextStrings;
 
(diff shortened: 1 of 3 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2173, opened 2022-12-21
WP/DeprecatedParameters: verify against WP Core

jrfnl:
Updated the array of parameters to search for based on a review of all calls to the `_deprecated_argument()` function in WP Core.

Findings:
* `get_user_option()` was listing the wrong WP version for its deprecation.
    Ref: https://developer.wordpress.org/reference/functions/get_user_option/
* `unregister_setting()` was looking for the wrong (positional) parameter. Includes updating the unit test.
    Ref: https://developer.wordpress.org/reference/functions/unregister_setting/
* The `_future_post_hook()` function contains a _soft_ deprecated parameter. Should still be detected by WPCS. As per the function docblock, I've set the expected value to `null`.
    Ref: https://developer.wordpress.org/reference/functions/_future_post_hook/
* The `_load_remote_block_patterns()` is a newly deprecated function (per WP 5.9.0).
    Ref: https://developer.wordpress.org/reference/functions/_load_remote_block_patterns/

Includes additional unit tests for the two newly listed functions.

### DIFF
--- a/WordPress/Sniffs/WP/DeprecatedParametersSniff.php
+++ b/WordPress/Sniffs/WP/DeprecatedParametersSniff.php
@@ -73,7 +73,20 @@ class DeprecatedParametersSniff extends AbstractFunctionParameterSniff {
 	 *    );
 	 */
 	protected $target_functions = array(
-
+		'_future_post_hook' => array(
+			1 => array(
+				'name'    => 'deprecated',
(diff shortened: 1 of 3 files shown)


dingo-d (approved):
👍🏼

GaryJones (approved):
✅

PR #2174, opened 2022-12-21
WP/DeprecatedParameterValues: support modern PHP + update for WP 6.1

jrfnl:
### WP/DeprecatedParameterValues: add support for PHP 8.0+ named parameters

Includes a change to the array format to allow for recording the parameter name.

### WP/DeprecatedParameterValues: verify against WP Core

Updated the array of parameters and values to search for based on a review of all calls to the `_deprecated_argument()` function in WP Core.

Includes tests.

Refs:
* https://developer.wordpress.org/reference/functions/add_option/
* https://developer.wordpress.org/reference/functions/get_option/
* https://developer.wordpress.org/reference/functions/update_option/


dingo-d (approved):

GaryJones (approved):
✅

PR #2175, opened 2022-12-21
WP/DiscouragedConstants: implement PHPCSUtils and support modern PHP 

jrfnl:
### WP/DiscouragedConstants: allow for the PHP 8.0+ nullsafe object operator

Ignore tokens preceded by the nullsafe object operator as they won't be the global constant.

Includes minor maintenance tweak to use a predefined array for all object operators.

Includes unit test.

### WP/DiscouragedConstants: allow for PHP 8.1+ enums

Ignore tokens preceded by the `enum` keyword as they won't be the global constant.

Includes minor maintenance tweak to use a predefined array for all OO keywords.

Includes unit test.

### WP/DiscouragedConstants: add support for PHP 8.0+ named parameters

    Ref: https://www.php.net/manual/en/function.define

### WP/DiscouragedConstants: prevent some false positives

By switching to using that key, some false negatives get fixed.

Includes ensuring that the information displayed in the error message will also not contain comments.

### WP/DiscouragedConstants: improve error position precision

... for multi-line function calls.

### WP/DiscouragedConstants: add tests with trait and enum constants

... to ensure those aren't recognized as if they were the global constant.

Enums are a PHP 8.1+ features.
Constants in traits are allowed since PHP 8.2.

### WP/DiscouragedConstants: remove redundant code

... which is no longer needed now support for PHPCS < 3.7.1 has been dropped.

### DIFF
--- a/WordPress/Sniffs/WP/DiscouragedConstantsSniff.php
+++ b/WordPress/Sniffs/WP/DiscouragedConstantsSniff.php
@@ -219,9 +219,16 @@ class DiscouragedConstantsSniff extends AbstractFunctionParameterSniff {
 		$clean_content = TextStrings::stripQuotes( $found_param['clean'] );
 
 		if ( isset( $this->discouraged_constants[ $clean_content ] ) ) {
+			$first_non_empty = $this->phpcsFile->findNext(
+				Tokens::$emptyTokens,
+				$found_param['start'],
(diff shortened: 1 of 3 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2176, opened 2022-12-21
WP/EnqueuedResources: various tweaks

jrfnl:
### WP/EnqueuedResources: add tests with PHP 7.3+ flexible heredoc/nowdocs

.. to safeguard these are handled correctly.

Note: these are in a separate file for maintainability as any code which would be placed _after_ the flexible heredoc/nowdoc would not be tokenized correctly on PHP < 7.3, making the tests maintainance error prone if the tests were in the same file.

### WP/EnqueuedResources: minor code performance tweak

Always retrieve the stack pointer to the last token of a potentially multi-line text string and skip to the end after analyzing the content.

While the `TextStrings::getEndOfCompleteTextString()` and the `TextStrings::getCompleteTextString()` have some overlap, the results of the overlapping code is cached, so this should barely affect performance, while skipping to the end of the text string will greatly improve performance for files containing long multi-line text strings.

### WP/EnqueuedResources: minor message tweak for clarity


dingo-d (approved):

GaryJones (approved):
✅

PR #2177, opened 2022-12-21
WP/EnqueuedResourceParameters: implement PHPCSUtils and support modern PHP

jrfnl:
### WP/EnqueuedResourceParameters: add support for PHP 8.0+ named parameters

### WP/EnqueuedResourceParameters: prevent false positives

By switching to using that key, a `null` value accompanied by a comment will no longer throw the wrong error.

### WP/EnqueuedResourceParameters: improve error position precision

... for multi-line function calls.

### WP/EnqueuedResourceParameters: add support for PHP 7.4/8.1 numbers

PHP 7.4 introduced numeric literals with underscores.
PHP 8.1 introduced explicit octal notation.

While PHPCS backfills the tokens for this, this sniff uses an `eval()` call and if either of these PHP 7.4/8.1 syntaxes would be passed to this `eval()` while PHPCS is being run on PHP < 7.4, this will result in a parse error on the code passed to `eval()`.

By using the PHPCSUtils `Numbers::getCompleteNumber()` method, we can get access to the decimal value of the encountered number, which allows us to bypass the issue by using that in the code string passed to `eval()`.

Includes unit tests.

### DIFF
--- a/WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
+++ b/WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
@@ -130,7 +130,7 @@ class EnqueuedResourceParametersSniff extends AbstractFunctionParameterSniff {
 		 */
 
 		$version_param = PassedParameters::getParameterFromStack( $parameters, 4, 'ver' );
-		if ( false === $version_param || 'null' === $version_param['raw'] ) {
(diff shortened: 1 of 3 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2178, opened 2022-12-21
Various sniffs: improve error messages

jrfnl:
### CodeAnalysis/EscapedNotTranslated: improvement to the error message

... to remove the presumption that positional parameters are being used.

### PHP/TypeCasts: minor message tweak for clarity

... and remove the advise to assign `null` as this could cause yet other problems due to the PHP 8.1 "passing null to non-nullable" deprecation (though passing an unassigned variable is in a way just as bad, but that's a whole other story).

### Security/SafeRedirect: minor message tweak for clarity

### DIFF
--- a/WordPress/Sniffs/PHP/RestrictedPHPFunctionsSniff.php
+++ b/WordPress/Sniffs/PHP/RestrictedPHPFunctionsSniff.php
@@ -37,7 +37,7 @@ class RestrictedPHPFunctionsSniff extends AbstractFunctionRestrictionsSniff {
 		return array(
 			'create_function' => array(
 				'type'      => 'error',
-				'message'   => '%s() is deprecated as of PHP 7.2 and removed in PHP 8.0, please use full fledged functions or anonymous functions instead.',
+				'message'   => '%s() is deprecated as of PHP 7.2 and removed in PHP 8.0. Please use declared named or anonymous functions instead.',
 				'functions' => array(
 					'create_function',
 				),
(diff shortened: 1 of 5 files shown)


dingo-d (approved):
Looks good to me, but I'm not a native speaker so I trust @GaryJones will be more thorough 😄

WordPress/Sniffs/PHP/RestrictedPHPFunctionsSniff.php
```diff
@@ -37,7 +37,7 @@ public function getGroups() {
 		return array(
 			'create_function' => array(
 				'type'      => 'error',
-				'message'   => '%s() is deprecated as of PHP 7.2 and removed in PHP 8.0, please use full fledged functions or anonymous functions instead.',
+				'message'   => '%s() is deprecated as of PHP 7.2 and removed in PHP 8.0, please use declared named or anonymous functions instead.',
```
GaryJones, line 40:
```suggestion
				'message'   => '%s() is deprecated as of PHP 7.2 and removed in PHP 8.0. Please use declared named or anonymous functions instead.',
```

GaryJones (changes requested):
Of the lines that have changed, I would suggest one extra change. The rest are 👍🏻 .

WordPress/Sniffs/PHP/RestrictedPHPFunctionsSniff.php
jrfnl: Done!

GaryJones (approved):
✅

PR #2179, opened 2022-12-22
PHPCS: ruleset tweak

jrfnl:
For this repo, we use a separate package for linting code, so there is no need to have the `Generic.PHP.Syntax` sniff running when doing a PHPCS scan (the sniff is included in the `WordPress-Extra` ruleset).

An analysis of the performance of a sniff run on this repo revealed that this was the slowest sniff by far. As excluding it will not have any QA impact, let's make CS runs on the code in this repo faster.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -47,6 +47,9 @@
 		<exclude name="WordPress.DB"/>
 		<exclude name="WordPress.Security"/>
 		<exclude name="WordPress.WP"/>
+
+		<!-- Linting is done in a separate CI job, no need to duplicate it. -->
+		<exclude name="Generic.PHP.Syntax"/>
 	</rule>
 
 	<!-- Check code for cross-version PHP compatibility. -->


jrfnl:
While probably not fully representative, on my machine on `develop`, a CS run on this repo takes nearly 5 minutes, while with this change, it takes about 8 seconds... (and that's without the cache active as that would influence the results otherwise)
I used the following command to test: `phpcs --report-source --no-cache`

dingo-d (approved):
5 minutes on all different PHP versions or? Because when I run it locally it takes about 2 seconds (develop is 1.99, this branch 1.84 sec).

GaryJones (approved):
✅ 

`develop` takes 2.6-2.8 seconds here to run `composer check-cs`, even without this change. 1.37 seconds with this change, so agreed that it is a slow sniff.

jrfnl:
Weird that the results om my machine are so different, but thanks for merging this anyway. As I said - we already lint in a different way, so it's not as if this changes any safeguards for the repo.

PR #2181, opened 2022-12-27
WP/GlobalVariablesOverride: implement PHPCSUtils and support modern PHP

jrfnl:
### WP/GlobalVariablesOverride: minor efficiency tweak

Change the format of the `$search` array to have the "value" as the key instead of as the value, which allows for using the more performance friendly `isset()` instead of `in_array()`.

### WP/GlobalVariablesOverride: add extra test

... to cover an edge case which was already handled in the code, but not covered by a test.

### WP/GlobalVariablesOverride: bug fix - allow for global statement closed via close tag

The parsing of a `global` variable import statement would walk until the next semi-colon, while the statement can also be ended by a PHP close tag.

This could lead to false positives as well as false negatives.

Fixed now.

Includes unit tests.

Includes removing a redundant variable assignment.

### WP/GlobalVariablesOverride: add support for PHP 7.4+ arrow functions

For the purposes of this sniff, default value assignments to parameters declared in an arrow function should be ignored to prevent false positives.

I've also reviewed the other places where _special handling_ is in place for functions/closures and have concluded that this does not apply to arrow functions.

I have updated line 333 to use the PHPCSUtils `Collections::functionDeclarationTokens()` token array for convenience, but it won't have any effect on arrow functions as `global` statements are not supported in arrow functions (well, that is, I haven't been able to set up any code which didn't result in a parse error for that. May need revisiting if someone more creative than me comes up with a way to do it...)

Includes unit test for the parameter default value part.

### WP/GlobalVariablesOverride: add test with PHP 7.4+ null coalesce assignment

PHP 7.4 introduced the null coalesce equal operator.

As the `VariableHelper::is_assignment()` uses the `Tokens::$assignmentTokens` array to determine whether something is an assignment and the `T_COALESCE_EQUAL` token is included in that array, no functional changes need to be made to support this.

Just adding a test to safeguard the support.

### WP/GlobalVariablesOverride: add test with PHP 8.0+ constructor property promotion

PHP 8.0 introduced constructor property promotion.

As default value assignments in function declaration statements are excluded already, no functional changes need to be made to support this.

Just adding a test to safeguard that this is handled correctly.

### WP/GlobalVariablesOverride: add test with PHP 8.1+ enums

PHP 8.1 introduced enums.

As enums cannot be used as test classes, there isn't really anything special about them, so no functional changes need to be made to support this.

Just adding a test to safeguard that they are handled correctly.

### WP/GlobalVariablesOverride: minor comment fix

### DIFF
--- a/WordPress/Sniffs/WP/GlobalVariablesOverrideSniff.php
+++ b/WordPress/Sniffs/WP/GlobalVariablesOverrideSniff.php
@@ -131,6 +131,13 @@ class GlobalVariablesOverrideSniff extends Sniff {
 		 * on that.
 		 * Examine variables not within a function scope and access to the `$GLOBALS`
 		 * variable based on the variable token.
+		 *
+		 * Note: No special handling here for code found within PHP 7.4+ arrow functions.
+		 * Arrow functions are "open", i.e. they have by value access to variables in the
+		 * surrounding scope, but they cannot modify the value.
+		 * Additionally, as they can only have one statement, a `global` statement _within_
+		 * an arrow function declaration will lead to a parse error as the result is
+		 * not a returnable value.
 		 */
 		$in_function_scope = Conditions::hasCondition( $this->phpcsFile, $stackPtr, array( \T_FUNCTION, \T_CLOSURE ) );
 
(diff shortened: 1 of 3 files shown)


dingo-d (approved):

GaryJones (approved):
✅️

PR #2185, opened 2023-01-07
NamingConventions/PrefixAllGlobals: implement PHPCSUtils and support modern PHP 

jrfnl:
### NamingConventions/PrefixAllGlobals: minor code tweak

Using `type` instead of `code` is not currently needed as all relevant tokens will exist in PHPCS 3.7.1.
Using `code` has a minor advantage as it will allow for IDE auto-completion.

### NamingConventions/PrefixAllGlobals: remove some redundant code

... which could never be reached anyway.

### NamingConventions/PrefixAllGlobals: add extra tests

... to cover some previously uncovered code.

None of these should lead to errors/warnings.

### NamingConventions/PrefixAllGlobals: implement PHPCSUtils

No functional changes.

### NamingConventions/PrefixAllGlobals: minor bug fix

The regex as-was allowed for the `x7f` (= DEL) character to be part of a name.

The origin of this bug was the PHP manual, where the regexes were originally incorrect. This was fixed in php/doc-en@6187696

No unit tests included as, to be honest, I haven't got a clue how to add a test with the DEL character ;-)

Includes adding a note about why this condition is not using PHPCSUtils.

### NamingConventions/PrefixAllGlobals: (mostly) ignore PHP 7.4+ arrow functions

For the purposes of this sniff:
* Parameters declared by arrow functions (with a default value assignment) do not need to be prefixed.
* New variables declared within an arrow function are local to the arrow function, so can be ignored.
* A `global` statement is not allowed within an arrow function.

And as an arrow function can only contain one statement, the chances of anything else within the arrow function needing to be flagged are slim, except when an arrow function declares a closure.
Note: nesting a named function declaration in an arrow function is not allowed (parse error).

PHPCS adds the `parenthesis_*` keys correctly and adds the arrow function parentheses to the `nested_parenthesis` array, so the function parameters can be handled same like parameters for closures/named functions.

However, while PHPCS does add the `scope_*` keys to the `T_FN` function, it does **not** add the `T_FN` to the `conditions` array for tokens between the `scope_opener` and the `scope_closer`.
The reason for this, is that the `scope_closer` of arrow functions can overlap with other `scope_closer`s, like in nested arrow function, which would make the `conditions` array unusable.

With the above in mind, I'm proposing to skip over the contents of arrow functions and ignore everything in them completely, with the one exception, that if an arrow function declares a closure, the closure will still be examined.

While this could potentially lead to some false negatives, I do believe this is the best we can do for now.

Includes unit tests.

### NamingConventions/PrefixAllGlobals: add test with PHP 7.4 null coalesce equals

... to safeguard assignments using that operator are recognized correctly.

### NamingConventions/PrefixAllGlobals: add test with PHP 8.0 nullsafe object operator

... to safeguard assignments to properties are not flagged when using that operator.

### NamingConventions/PrefixAllGlobals: add support for PHP 8.0+ named parameters

1. Adjusted the logic in the sniff to allow for named parameters using the new PHPCSUtils 1.0.0-alpha4 `PassedParameters::getParameterFromStack()` method and the new `WPHookHelper::get_hook_name_param()` method.
2. The parameter name used for `define()` is in line with the name as per the PHP 8.0 release.

Note: function calls with named arguments are not supported for the `do_action()` or `apply_filters()` functions as those have a parameter with a spread operator (but that's irrelevant for the sniff).

### NamingConventions/PrefixAllGlobals: prevent some false positives

By switching to using that key, some false positives get fixed.

Includes renaming the variable name to which the value was being assigned to allow it to continue to be descriptive.

The switch to using the `clean` key also makes sure that the information displayed in the error message will not contain comments, making those message cleaner too.

### NamingConventions/PrefixAllGlobals: add test with PHP 8.0+ constructor property promotion

PHP 8.0 introduced constructor property promotion.

As default value assignments in function declaration statements are excluded already, no functional changes need to be made to support this.

Just adding a test to safeguard that this is handled correctly.

### NamingConventions/PrefixAllGlobals: add support for PHP 8.1+ enums

PHP 8.1 introduced enums as a new OO structure.

Unless in a (prefixed) namespace, enum names should also be prefixed.

The `Tokens::$ooScopeTokens` array used in the `register()` method already contains the `T_ENUM` token, so the sniff is already listening for it.

This commit now adjusts the sniff to actually handle enums as well.

Note: enums cannot `extend`, so cannot ever be regarded as "test classes" for the purposes of this sniff.

### NamingConventions/PrefixAllGlobals: add tests with PHP 8.2+ constants in traits

... to ensure those aren't recognized as if they were the global constant.

Constants in traits are allowed since PHP 8.2.
The `final` keyword can be used with OO constants since PHP 8.1.

### NamingConventions/PrefixAllGlobals: bug fix - finding end of global statement

Prevent false positives when a global statement is ended by a PHP close tag.

Includes unit tests.

### NamingConventions/PrefixAllGlobals: bug fix - only check global statement in current scope

Prevent false positives when a nested function/closure declaration also contains a `global` statement.

Includes unit test.

### DIFF
--- a/WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.9.inc
+++ b/WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.9.inc
@@ -1,0 +1,18 @@
+<?php
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] acronym,tgmpa
+
+/*
+ * This test safeguards yet another edge case to make sure the sniff bows out correctly
+ * for unfinished global statements within functions (live coding/parse error).
+ *
+ * This test file should generate no errors or warnings.
+ */
+
+function acronym_unfinished_global_statement($param) {
+	global $var1, $var2 // Deliberately missing semi-colon.
+	
+	foreach ($param as $var2) {}
+}
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[]
--- a/WordPress/Docs/NamingConventions/PrefixAllGlobalsStandard.xml
+++ b/WordPress/Docs/NamingConventions/PrefixAllGlobalsStandard.xml
@@ -5,7 +5,7 @@
     >
     <standard>
     <![CDATA[
-    All globals terms must be prefixed with a theme/plugin specific term. Global terms include Namespace names, Class/Interface/Trait names (when not namespaced), Functions (when not namespaced or in an OO structure), Constants/Variable names declared in the global namespace, and Hook names.
+    All globals terms must be prefixed with a theme/plugin specific term. Global terms include Namespace names, Class/Interface/Trait/Enum names (when not namespaced), Functions (when not namespaced or in an OO structure), Constants/Variable names declared in the global namespace, and Hook names.
 
     A prefix must be distinct and unique to the plugin/theme, in order to prevent potential conflicts with other plugins/themes and with WordPress itself.
 
--- a/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
+++ b/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
@@ -473,7 +473,7 @@ class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 					break;
 
 				case \T_CONST:
-					// Constants in a class do not need to be prefixed.
+					// Constants in an OO construct do not need to be prefixed.
 					if ( true === Scopes::isOOConstant( $this->phpcsFile, $stackPtr ) ) {
 						return;
 					}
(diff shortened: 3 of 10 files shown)


dingo-d (approved):

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.9.inc
```diff
@@ -0,0 +1,18 @@
+<?php
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] acronym,tgmpa
+
+/*
+ * This test safeguards yet another edge cases and makes sure the sniff bows out correctly
```
GaryJones, line 6:
```suggestion
 * This test safeguards yet another edge case to make sure the sniff bows out correctly
```

GaryJones (approved):
Minor doc comment grammar improvement, but ✅  otherwise. Great work bringing this sniff up to scratch!

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.9.inc
jrfnl: Fixed in the appropriate commit.

Issue #864, opened 2017-03-03
Why is $wpdb->delete cacheable?


barryceelen:
**Noob question** (and probably not directly related):

The `delete` method is marked as cacheable in [DirectDatabaseQuerySniff.php](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/543bd68ede9558ac4c8c6910ec130535af11e465/WordPress/Sniffs/VIP/DirectDatabaseQuerySniff.php#L74).  Why would I want to cache a DELETE query?

JDGrimes:
Not entirely noob, there was some confusion about this back in #180 (from me 😊). The fact is though, that if you added something to the cache at some point, then you will probably want to remove it from the cache when it is deleted, to free up the cache. So that's why `$wpdb->delete()` is still flagged—not to add to the cache, but to delete from it.

JDGrimes:
Maybe there is a way that we can make this more obvious though?

barryceelen:
…which reminds me to check existing issues before opening a new one 🙈.
Thanks @JDGrimes!

PR #2186, opened 2023-01-07
DB/DirectDatabaseQuery: implement PHPCSUtils and support modern PHP 

### LINKED ISSUE #1947: DirectDatabaseQuery: don't demand caching or forbid direct DB calls for `TRUNCATE`
## Bug Description
Given the following code snippet:
```php
$wpdb->query(
	$wpdb->prepare(
		'TRUNCATE TABLE `%1$s`',
		plugin_get_table_name( 'Name' )
	)
);
```

WPCS will currently throw the following warnings:
```
 # | WARNING | Usage of a direct database call is discouraged. (WordPress.DB.DirectDatabaseQuery.DirectQuery)
 # | WARNING | Direct database call without caching detected. Consider using wp_cache_get() / wp_cache_set() or wp_cache_delete().
     |         | (WordPress.DB.DirectDatabaseQuery.NoCaching)
```

IMO a query to truncate a table may be dangerous and could warrant a warning be thrown for that reason, however, it *is* a valid query to use a direct database call for and it is also a query for which caching of the query is irrelevant.

The "cache" related issue here is that any existing "old" cache should probably be thrown away, but that can not be done based on the `truncate` query.

Possibly a recommendation to query the truncated table afterwards with caching to clear any caches could be given, but that still wouldn't catch all potentially set caches.

## Environment

| PHP version             | 7.4.8
| PHP_CodeSniffer version | `master`
| WPCS install type       | git clones
| IDE (if relevant)       | N/A

Loosely related to #1903

jrfnl:
### DB/DirectDatabaseQuery: remove redundant code

This code is a remnant of the old-style custom ignore comment mechanism which was removed in #2063.

### DB/DirectDatabaseQuery: add missing test case

... to cover the one uncovered line.

### DB/DirectDatabaseQuery: implement PHPCSUtils

👉 This change will be easier to review while ignoring whitespace changes.

### DB/DirectDatabaseQuery: add support for PHP 8.0+ nullsafe object operator

Prevent false negatives in case someone would use the PHP 8.0 nullsafe object operator on the `$wpdb` variable.

Also makes the token walking more stable by preventing an unrelated object operator from being recognized as if it were an operator operating on the `$wpdb` object (false positive).

Includes tests.

### DB/DirectDatabaseQuery: bug fix - code style independent token walking

Don't allow the sniff to get confused by comments.

Related to #763

Includes tests.

### DB/DirectDatabaseQuery: bug fix - finding end of statement

Prevent false positives from walking too far when a database query statement is ended by a PHP close tag.

Includes unit tests.

### DB/DirectDatabaseQuery: bug fix - ignore TRUNCATE queries

Prevent unsolvable false positives for `TRUNCATE` queries. Those cannot be cached and need a direct DB query. **(second opinion appreciated!)**

Includes unit tests.

Fixes #1947

### DIFF
--- a/WordPress/Sniffs/DB/DirectDatabaseQuerySniff.php
+++ b/WordPress/Sniffs/DB/DirectDatabaseQuerySniff.php
@@ -192,11 +192,11 @@ class DirectDatabaseQuerySniff extends Sniff {
 			return;
 		}
 
-		$endOfStatement = $this->phpcsFile->findNext( \T_SEMICOLON, ( $stackPtr + 1 ), null, false, null, true );
+		$endOfStatement = $this->phpcsFile->findNext( array( \T_SEMICOLON, \T_CLOSE_TAG ), ( $stackPtr + 1 ) );
 
 		// Check for Database Schema Changes.
 		for ( $_pos = ( $stackPtr + 1 ); $_pos < $endOfStatement; $_pos++ ) {
-			$_pos = $this->phpcsFile->findNext( Tokens::$textStringTokens, $_pos, $endOfStatement, false, null, true );
+			$_pos = $this->phpcsFile->findNext( Tokens::$textStringTokens, $_pos, $endOfStatement );
 			if ( false === $_pos ) {
 				break;
 			}
--- a/WordPress/Sniffs/DB/DirectDatabaseQuerySniff.php
+++ b/WordPress/Sniffs/DB/DirectDatabaseQuerySniff.php
@@ -183,7 +183,7 @@ class DirectDatabaseQuerySniff extends Sniff {
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


jrfnl:
Note: this sniff is not that stable as the token walking is not very strict. Realistically, it could do with a complete rewrite, but that is outside the scope of what I was trying to achieve now.

Also just saw #864, which makes me question the `TRUNCATE` change. As I said above, second opinion very much appreciated.

dingo-d (approved):
I'm fine with `TRUNCATE` not being flagged.

In most cases people using raw SQL know why they are doing it and if they know how to configure PHPCS they'll just exclude the error for where it's being used anyhow so 🤷🏼‍♂️

GaryJones (approved):
✅

PR #2187, opened 2023-01-08
Helpers/MinimumWPVersionTrait: reduce use of magic number + other minor improvement

### LINKED ISSUE #2171: Support "default" as value for "// phpcs:set ... minimum_wp_version"
Looking at the repeated `// phpcs:set WordPress.WP.AlternativeFunctions minimum_wp_version 5.8` reset comments in a [test file](https://github.com/WordPress/WordPress-Coding-Standards/blob/feature/alternativefunctions-phpcsutils-and-modern-php/WordPress/Tests/WP/AlternativeFunctionsUnitTest.inc), then it seems like it's an ever-increasing list of places in code where future updates would need to be made when the default value changes, which could lead to some being missed.

I wondered if something like `// phpcs:set WordPress.WP.AlternativeFunctions minimum_wp_version default` could be supported (and perhaps similar for any other numerical/version/often-changing values), such that the consuming sniff(s) could look up the `default` value in a single place and use the actual value. It centralises the default value for that WPCS release.

Very much an internally-facing potential improvement, which may not warrant the extra complexity, even if it was possible, which I'm not sure it is, as it may just be assigning a value to a class property - maybe a magic `__set()` in the [new trait](https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Helpers/MinimumWPVersionTrait.php) could be used?

```php
public function __set( $name, $value ) {
	if ( 'minimum_wp_version' === $name && 'default' === $value ) {
		$this->minimum_wp_version = '5.8'; // Update this for each WPCS release - no need to touch test files.
	}
}
```

jrfnl:
### Helpers/MinimumWPVersionTrait: split wp_version_compare() method

.. into two methods:
* `wp_version_compare()` which does the actual compare.
* `normalize_version_number()` to handle the version number normalization.

This is just a code simplification/removal of duplicate code. No functional changes.

### Helpers/MinimumWPVersionTrait: reduce use of magic number

The default value for the `minimum_wp_version` property is used in a number of places, mostly in the test files when resetting a changed value for the property to its default, making it a "magic number".

Preferably, a magic number should be assigned to a constant and the constant should be used everywhere.
Unfortunately, that's not possible as a) constants in traits are not allowed prior to PHP 8.2 and b) code in test case files does not get executed, so using the constant is not an option anyway.

This commit now makes couple of changes to improve/stabilize working with this "magic" number.
* The magic number is now only set in one place  - in the new `$default_minimum_wp_version` property.
* A user-provided `minimum_wp_version` property value from a ruleset will now be validated in the same way as a CLI value for the same.
    If no value is found or an invalid value is found, the default value for the `minimum_wp_version` property will be used.
    Includes making it explicit in the `composer.json` file that the `Filter` extension is a prerequisite for this package.
* In test case files, a "reset" of the `minimum_wp_version` property can now unset the value, lowering the maintenance burden when the default value for the `minimum_wp_version` property changes.
    Note: test files for the `Deprecated*` sniffs will still need updating whenever the value changes.

---

**Open question**

I wonder if we should rename the `get_wp_version_from_cli()` method to a better name. The method was already renamed in WPCS 3.0.0, so renaming it again does not add a breaking change which wasn't already there in the first place.

The reason is that, on the one hand, the `get_` prefix suggests that the method returns a value, which it doesn't. And on the other hand, the `_from_cli` isn't fully covering the functionality anymore.

Name suggestions welcome.

Some (rough) ideas:
* `validate_minimum_wp_version()`
* `set_minimum_wp_version()`

Opinions ?

---

Fixes #2171

### :new: Helpers/MinimumWPVersionTrait: rename method

... from `get_wp_version_from_cli()` to `set_minimum_wp_version()`.

### DIFF
--- a/WordPress/Helpers/MinimumWPVersionTrait.php
+++ b/WordPress/Helpers/MinimumWPVersionTrait.php
@@ -68,7 +68,22 @@ trait MinimumWPVersionTrait {
 	 *
 	 * @var string WordPress version.
 	 */
-	public $minimum_wp_version = '5.8';
+	public $minimum_wp_version;
+
+	/**
+	 * Default minimum supported WordPress version.
+	 *
+	 * By default, the minimum_wp_version presumes that a project will support the current
+	 * WP version and up to three releases before.
+	 *
+	 * {@internal This should be a constant, but constants in traits are not supported
+	 *            until PHP 8.2.}}
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var string WordPress version.
+	 */
+	private $default_minimum_wp_version = '5.8';
 
 	/**
 	 * Overrule the minimum supported WordPress version with a command-line/config value.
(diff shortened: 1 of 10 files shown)


dingo-d (approved):
I'm for the `set_minimum_wp_version`. Even though some validation does happen, in the end, what happens in the method is the setting of the `minimum_wp_version` property.

GaryJones (approved):
✅  I agree with the `set_*` method name suggestion.

jrfnl:
Rebased & added a commit to rename the method.

PR #2188, opened 2023-01-08
DB/PreparedSQL: implement PHPCSUtils and support modern PHP 

jrfnl:
### Sniff: move the $SQL*Functions properties to the DB/PreparedSQL Sniff

The `Sniff::$SQLEscapingFunctions` and `Sniff::$SQLAutoEscapedFunctions` properties are _only_ used by the `WordPress.DB.PreparedSQL` sniff, so these don't belong in the generic base `Sniff` class.

A [code search](https://sourcegraph.com/search?q=context:global+-%3ESQL%28AutoEscaped%7CEscaping%29Functions+-file:%5E%28.*/%29%3Fvendor/wp-coding-standards/.*%24&patternType=regexp&case=yes&sm=0) also shows these properties are not used by sniffs from external standards, so this change should have no noticeable impact.

Includes changing the visibility of the properties from `protected` to `private`.

### DB/PreparedSQL: change visibility of a property

... which is only intended for use by this sniff. Was `protected`, now `private`.

### DB/PreparedSQL: use predefined token collections [1]

Leverage predefined token collections in PHPCS and PHPCSUtils to fill the list of `$ignored_tokens`.

This commit start using the PHPCS native `Tokens::$bracketTokens` collection.
In addition to the parentheses and square bracket tokens, that collection includes the curly brace tokens. As those, in and of itself, are not problematic, I believe it is safe to include these.

Includes a test demonstrating the issue.

### DB/PreparedSQL: use predefined token collections [2]

Leverage predefined token collections in PHPCS and PHPCSUtils to fill the list of `$ignored_tokens`.

This commit start using the PHPCS native `Tokens::$castTokens` collection.

The `Tokens::$castTokens` collection contains _all_ cast tokens, not just the int/float/boolean cast tokens. However, as a cast token, in and of itself, is not problematic, I believe it is safe to include these.

What I mean by this, is that the sniff shouldn't trigger on seeing `(object)` in an SQL query. The sniff should trigger on the `$var`, which is being cast to an object, as an object cast is not a "safe" cast for use in SQL queries.

Having those extra cast tokens in the "ignore" list, therefore doesn't make the sniff any less effective as the sniff has a separate check for whether or not a type cast is considered "safe".

Includes a test demonstrating the issue.

### DB/PreparedSQL: use predefined token collections [3]

Leverage predefined token collections in PHPCS and PHPCSUtils to fill the list of `$ignored_tokens`.

This commit start using the PHPCS native `Tokens::$heredocTokens` collection.

The `Tokens::$heredocTokens` collection contains the same tokens as were already in the array + `T_HEREDOC`.
As the contents of `T_HEREDOC` tokens needs to be examined, that token is subsequently removed again from the array.

### DB/PreparedSQL: use predefined token collections + allow floats [4]

Leverage predefined token collections in PHPCS and PHPCSUtils to fill the list of `$ignored_tokens`.

This commit start using the PHPCS native `Tokens::$arithmeticTokens` collection.

Arithmetic operator tokens, including the `+` and `-` signs should not be flagged by the sniff.
Same goes for floating point numbers. Those should be treated the same as integers as SQL supports those without issue.

Includes a few tests demonstrating the issue.

### DB/PreparedSQL: use predefined token collections [5]

Leverage predefined token collections in PHPCS and PHPCSUtils to fill the list of `$ignored_tokens`.

This commit start using the PHPCSUtils `incrementDecrementOperators()` collection.

This effectively adds two additional tokens to the "safe" list, `++` and `--`. These tokens, in and of itself, are not problematic, so I believe it is safe to include these.

It could even be argued that increment/decrement tokens should be treated the same as "safe casted" tokens, as incrementing or decrementing a variable effectively turns it into a benign integer.

Includes a test demonstrating the issue.

### DB/PreparedSQL: allow for the PHP 8.0+ nullsafe object operator

Like for the "normal" object operator: don't throw a "not prepared" error for the nullsafe object operator itself.
Same goes for the `::` object operator.

Includes unit test.

### DB/PreparedSQL: add test with PHP 7.3+ flexible heredoc

... to safeguard the sniff handles this correctly, in the same way as "normal" heredocs.

Note: the tests are in a separate file for maintainability as any code which would be placed _after_ the flexible heredoc/nowdoc would not be tokenized correctly on PHP < 7.3, making the tests maintainer error prone if the tests were in the same file.

### DB/PreparedSQL: add test with PHP 7.4+ numeric literals with underscores and PHP 8.1+ explicit octals

... to safeguard the sniff handles this correctly, in the same way as "normal" integers.

### DB/PreparedSQL: minor documentation tweaks


dingo-d (approved):

GaryJones (approved):
✅

PR #2190, opened 2023-01-10
Move "snake case" related utility to dedicated SnakeCaseHelper + bug fix in method

jrfnl:
### Move "snake case" related utility to dedicated SnakeCaseHelper

The "snake case" related utility method is only used by a small set of sniffs, so is better placed in a dedicated class.

This commit moves the `get_snake_case_name_suggestion()` method to the new `WordPressCS\WordPress\Helpers\SnakeCaseHelper` and starts using that class in the relevant sniffs.

**Note**:
It is expected for PHPCSUtils to have dedicated methods for the same at some point in the future. If/when those methods become available, it is recommended for the sniffs to start using the PHPCSUtils methods.
With that in mind, this class is marked as not part of the public API.

Related to #1465

### SnakeCaseHelper::get_suggestion(): bug fix - underscore handling

The `SnakeCaseHelper::get_suggestion()` method should have no opinion on the use of leading/trailing/double underscores in a name.

In practice, however, it did and the suggested name would be incorrect.

Fixed now.

Includes tests to safeguard this fix in the `ValidFunctionName` test case file, though the current methodology for testing would not catch a regression.

To illustrate the issue - originally, the error messages for these new tests would read:
```
216 | ERROR | Method name "___Leading_Underscores" in class UnderScoreHandling is not in snake case format, try "leading_underscores"
 217 | ERROR | Method name "Multiple_______Underscores" in class UnderScoreHandling is not in snake case format, try "multiple____underscores"
 218 | ERROR | Method name "Trailing_Underscores___" in class UnderScoreHandling is not in snake case format, try "trailing_underscores"
```

With the fix, the error messages for the new tests read:
```
 216 | ERROR | Method name "___Leading_Underscores" in class UnderScoreHandling is not in snake case format, try "___leading_underscores"
 217 | ERROR | Method name "Multiple_______Underscores" in class UnderScoreHandling is not in snake case format, try "multiple_______underscores"
 218 | ERROR | Method name "Trailing_Underscores___" in class UnderScoreHandling is not in snake case format, try "trailing_underscores___"
```

### DIFF
--- a/WordPress/Helpers/SnakeCaseHelper.php
+++ b/WordPress/Helpers/SnakeCaseHelper.php
@@ -39,10 +39,8 @@ final class SnakeCaseHelper {
 	 * @return string
 	 */
 	public static function get_suggestion( $name ) {
-		$suggested = preg_replace( '`([A-Z])`', '_$1', $name );
+		$suggested = preg_replace( '`(?<!_|^)([A-Z])`', '_$1', $name );
 		$suggested = strtolower( $suggested );
-		$suggested = str_replace( '__', '_', $suggested );
-		$suggested = trim( $suggested, '_' );
 		return $suggested;
 	}
 }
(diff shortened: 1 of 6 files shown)


GaryJones (approved):
✅

dingo-d (approved):

PR #2191, opened 2023-01-10
NamingConventions/ValidFunctionName: implement PHPCSUtils and support modern PHP

jrfnl:
### NamingConventions/ValidFunctionName: add extra tests

The handling of function names only consisting of underscore was already special cased in the sniff, but not covered by tests.

Similarly, live coding was not covered.

### NamingConventions/ValidFunctionName: implement PHPCSUtils methods

PHP is largely case-insensitive for class and method name, but not completely: the case-insensitivety does not apply to the non-ascii range.

The PHPCSUtils methods recognize this correctly.

Includes minor efficiency tweak by only doing the PHP4 constructor/destructor comparison when it's a real class as a name match can never be made for anonymous classes anyway.

Includes unit tests.

Note: the test would previously pass, but only due to the `strtolower()` mangling both the class name as well as the method name equally.

### NamingConventions/ValidFunctionName: add tests covering PHP 8.1+ enums

The various PHPCSUtils `Scopes` functions support enums since PHPCSUtils 1.0.0-alpha4, so enums are automatically supported now, but let's safeguard this with a test anyway.

### NamingConventions/ValidFunctionName: minor comment fix

### DIFF
--- a/WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
@@ -128,7 +128,7 @@ class ValidFunctionNameSniff extends Sniff {
 	 *
 	 * @since 0.1.0
 	 * @since 3.0.0 Renamed from `processTokenWithinScope()` to `process_method_declaration()`.
-	 *              Method signature has been changed as well as this method no longer overloads
+	 *              Method signature has been changed as well, as this method no longer overloads
 	 *              a method from the PEAR sniff which was previously the sniff parent.
 	 *
 	 * @param int    $stackPtr   The position where this token was found.
--- a/WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
+++ b/WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
@@ -11,6 +11,7 @@ namespace WordPressCS\WordPress\Sniffs\NamingConventions;
 
 use PHPCSUtils\BackCompat\BCTokens;
 use PHPCSUtils\Utils\FunctionDeclarations;
+use PHPCSUtils\Utils\NamingConventions;
 use PHPCSUtils\Utils\ObjectDeclarations;
 use PHPCSUtils\Utils\Scopes;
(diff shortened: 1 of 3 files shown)


WordPress/Sniffs/NamingConventions/ValidFunctionNameSniff.php
```diff
@@ -142,19 +143,16 @@ protected function process_method_declaration( $stackPtr, $methodName, $currScop
...
-			return;
-		}
-
-		// PHP4 destructors are allowed to break our rules.
-		if ( '_' . $classNameLc === $methodNameLc ) {
-			return;
+			// PHP4 destructors are allowed to break our rules.
+			if ( NamingConventions::isEqual( $methodName, '_' . $className ) === true ) {
+				return;
+			}
```
GaryJones, line 155: I thought destructors didn't exist until PHP 5?

I've not seen them suggested with a leading underscore in PHP 4 as:

```php
class Foo {
	function Foo() {} // Constructor.
	function _Foo() {} // Destructor.
}
```
Was that the _de facto_ approach back then? How did they get triggered?

jrfnl: All good questions. To be honest, it's so far back, that this should be redundant and I hope to never come across that type of code again.

For the purposes of this PR, I've just kept the existing functionality, which already handled things in this way (whether correct or not).
Also note that the PHPCS native `PEAR.NamingConventions.ValidFunctionName` sniff, which this sniff was originally based on, treats `_ClassName()` methods the same, i.e. as PHP 4-destructors.

I realize this is not an answer to your question, but there's only so much time (research) I'm willing to spend on work around for code which shouldn't exist anymore in the first place.

dingo-d (approved):

jrfnl:
> ✅ Approving as is, but as WP requires PHP 5.6 minimum, we should reconsider the benefit of WPCS supporting coding practises from the PHP 4 era.

What about we open an issue to remove those exceptions in WPCS 4.0 ?

jrfnl:
I've opened issue #2279 as a reminder to remove the PHP 4 related exceptions from sniffs for WPCS 4.0.

PR #2193, opened 2023-01-10
CodeAnalysis/AssignmentInCondition: remove parse error related warning

jrfnl:
The `WordPress.CodeAnalysis.AssignmentInCondition` contained a `NonVariableAssignmentFound` error code, which would flag assignments in condition not involving a variable.

Generally speaking, sniffs stay silent on parse errors and a non-variable assignment in condition will in most cases be a parse error and in all other cases a runtime Fatal Error. This can be seen here: https://3v4l.org/oQgjR Uncomment any of the commented out lines to see the parse error/fatal error.

With that in mind, I propose to remove the `NonVariableAssignmentFound` error code from the sniff.


dingo-d (approved):

GaryJones (approved):
✅

PR #2195, opened 2023-01-11
Utils/I18nTextDomainFixer: implement PHPCSUtils and support modern PHP 

jrfnl:
### Utils/I18nTextDomainFixer: add load_script_textdomain() function

The `load_script_textdomain()` function was added to WP in WP 5.0.

Ref: https://developer.wordpress.org/reference/functions/load_script_textdomain/

### Utils/I18nTextDomainFixer: update lists of plugin/theme headers

... as per the current lists of supported headers in the linked theme/plugin handbook pages.

### Utils/I18nTextDomainFixer: add some extra tests

... for a few situations which were handled by the sniff, but so far not covered by tests.

### Utils/I18nTextDomainFixer: add support for PHP 8.0+ named parameters

Includes a change to the array format to allow for recording the parameter name.

Note: for now, I've not (yet) made the `MissingArgDomain` error auto-fixable in function calls with named arguments. I do think it's possible to do so, but also need to prioritize and I deem this sniff low priority (as it is "off" by default unless certain required properties are passed).

### DIFF
--- a/WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
+++ b/WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
@@ -10,6 +10,7 @@
 namespace WordPressCS\WordPress\Sniffs\Utils;
 
 use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\BackCompat\Helper;
 use PHPCSUtils\Utils\TextStrings;
 use WordPressCS\WordPress\AbstractFunctionParameterSniff;
 use WordPressCS\WordPress\Helpers\RulesetPropertyHelper;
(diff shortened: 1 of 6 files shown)


dingo-d (approved):

WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
```diff
@@ -72,50 +74,159 @@ class I18nTextDomainFixerSniff extends AbstractFunctionParameterSniff {
 	 * The WP Internationalization related functions to target for the replacements.
 	 *
 	 * @since 1.2.0
+	 * @since 3.0.0 The format of the value has changed from a numerically indexed
+	 *              array containing parameter positions to an array with the parameter
+	 *              position as the index and the parameter name as value.
 	 *
-	 * @var array <string function name> => <int parameter position>
+	 * @var array<string, <string, in|string>> Function name as key, array with target
```
GaryJones, line 81:
```suggestion
	 * @var array<string, array<string, in|string>> Function name as key, array with target
```

GaryJones (changes requested):

WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php
jrfnl: Thanks for catching that. Fixed now.

GaryJones (approved):

PR #2196, opened 2023-01-11
Arrays/ArrayIndentation: implement PHPCSUtils

jrfnl:
### Arrays/ArrayIndentation: add extra test

... for previously uncovered code.

###Arrays/ArrayIndentation: implement PHPCSUtils

---

Note: I've not spend much time on this sniff as it is slated to be replaced with a sniff from PHPCSExtra in the future anyway. If anything still crops up, a quick fix can be made if needs be.


dingo-d (approved):

GaryJones (approved):
✅

PR #2197, opened 2023-01-12
NamingConventions/Valid(Function|Variable)Name: bug fix for error determination

### LINKED ISSUE #1891: snake_case checking doesn't work with characters outside a-z0-9 range
## Bug Description

I get `Variable "$lähtöaika" is not in valid snake_case format, try "$lähtöaika"` error, when clearly the variable name is in valid snake_case format.

## Minimal Code Snippet
```php
$lähtöaika = true;
```

This is because the `isSnakeCase()` function only allows characters within `[a-z0-9_]`. However, this is not the whole set of allowed characters in PHP variable names. Quoting from the [PHP manual](https://www.php.net/manual/en/language.variables.basics.php):

> Variable names follow the same rules as other labels in PHP. A valid variable name starts with a letter or underscore, followed by any number of letters, numbers, or underscores. As a regular expression, it would be expressed thus: `^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$`

If the `isSnakeCase()` function is changed to use this regular expression (minus the uppercase A-Z), the problem is gone:

```php
public static function isSnakeCase( $var_name ) {
	return (bool) preg_match( '/^[a-z_\x80-\xff][a-z0-9_\x80-\xff]*$/', $var_name );
}
```

This is not perfect, as the 128-255 character range contains uppercase letters, so some non-snake variable names would still be accepted: `$lähtÖaika` would not trigger an error. I'm not sure if checking for these cases would be worth making the regex a lot more complicated, though.

## Error Code
`WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase`

## Environment

| PHP version             | 7.3.11
| PHP_CodeSniffer version | 3.4.2
| WPCS version            | 2.2.1
| WPCS install type       | global, VS Code plugin
| IDE (if relevant)       | VS Code 1.44.2

- [X] I have verified the issue still exists in the `develop` branch of WPCS.

jrfnl:
This commit fixes two, closely related, issues:
1. The "is a name in snake case ?" determination was done in three slightly different ways between the `ValidFunctionName` sniff and the `ValidVariableName` sniff and the `SnakeCaseHelper::get_suggestion()` method, which meant that a name which would be regarded valid snakecase for the one, may not be regarded as valid snakecase for the other.
    It could also result in the suggested name being the same as the existing name.
2. The "is a name in snake case ?" as well as the suggestion determination didn't take non-ASCII characters into account consistently.

While PHPCSUtils will at some point get better methods for handling these things, for now, we can still make a quick fix to improve this situation.

To that end, I've made the following changes:
* Instead of having each sniff determine whether or not something is valid snakecase itself, let the sniffs retrieve the suggested alternative name and compare the original name against that and only throw an error if the suggestion is not the same as the original name. This also allows for removing some code related to potential leading underscores (which looks to be redundant anyway).
* Improved the creation of the suggestion to handle non-ASCII names better (in as far as possible).

Includes:
* The `Mbstring` extension has been added as a suggested extension to the `composer.json` file.
* Adding one extra test run to the CI workflow which runs the tests without the Mbstring extension to safeguard the fallback code.

Includes unit tests.

Fixes #1891

---

Note: as of PHP 8.2, `strtolower()` no longer acts on non-ASCII characters, which largely got rid of the issue already and made for some interesting debugging. Unfortunately, we cannot rely on everyone being on PHP 8.2...

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -23,7 +23,13 @@ jobs:
       matrix:
         php: [ '5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3' ]
         phpcs_version: [ '3.7.1', 'dev-master' ]
-        #include:
+        extensions: [ '' ]
+
+        include:
+          - php: '7.4'
+            phpcs_version: 'dev-master'
+            extensions: ':mbstring' # = Disable Mbstring.
+
           # Add extra build to test against PHPCS 4.
           #- php: '7.4'
           #  phpcs_version: '4.0.x-dev as 3.9.99'
--- a/WordPress/Helpers/SnakeCaseHelper.php
+++ b/WordPress/Helpers/SnakeCaseHelper.php
@@ -9,6 +9,8 @@
 
 namespace WordPressCS\WordPress\Helpers;
 
+use PHPCSUtils\BackCompat\Helper;
+
 /**
  * Helper utilities for checking if a name is in snake_case.
  *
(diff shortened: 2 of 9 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2200, opened 2023-01-12
CS/QA: make (nearly) all classes `final`

jrfnl:
### QA: make all non-abstract sniff classes final

The PHPCS autoloader does not always play nice with sniffs which extend other sniffs, especially if both sniffs (parent and child) are included in the same standard.

To discourage sniffs extending each other, I propose making all non-abstract sniff classes `final`with the exception of four classes of which it is known that those are being extended based on a public code search.

The four exceptions are:
* `NamingConventions.ValidHookName`
* `Security.EscapeOutput`
* `Security.NonceVerification`
* `Security.ValidatedSanitizedInput`

If needs be, a sniff in an external standard can still _use_ logic from the WPCS sniffs. On the one hand, they can use the new `Helper` classes and traits. On the other hand, they can do a preliminary check on certain data and defer to a new or preinstantiated instance of a WPCS sniff for everything else.

Note: this list of sniffs being extended is likely incomplete.
If reports come in from external standards that they are running into trouble using alternative logic, we can still remove the `final` keyword from select sniffs in a patch release. However, adding the `final` keyword needs to be done in a major.

Ref:
* https://sourcegraph.com/search?q=context:global+use+WordPressCS+-repo:%5Egithub%5C.com/WordPress/WordPress-Coding-Standards%24+-repo:%5Egithub%5C.com/WPTT/WPThemeReview%24+-file:%5E%28.*/%29%3Fvendor/wp-coding-standards/.*%24&patternType=standard&case=yes&sm=0

### PHPCS: enforce classes to be abstract or final

... for the files in this package.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -70,6 +70,13 @@
 	<!-- Enforce PSR1 compatible namespaces. -->
 	<rule ref="PSR1.Classes.ClassDeclaration"/>
 
+	<!-- Enforce that classes are abstract or final. -->
+	<rule ref="Universal.Classes.RequireFinalClass">
+		<!-- ... with the exception of four sniffs which are known to be extended by external standards. -->
+		<exclude-pattern>/WordPress/Sniffs/NamingConventions/ValidHookNameSniff\.php$</exclude-pattern>
+		<exclude-pattern>/WordPress/Sniffs/Security/(EscapeOutput|NonceVerification|ValidatedSanitizedInput)Sniff\.php$</exclude-pattern>
+	</rule>
+
 
 	<!--
 	#############################################################################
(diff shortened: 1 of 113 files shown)


GaryJones (approved):
✅  I checked VIPCS, and it only extends abstract classes from WPCS.

GaryJones:
> On the one hand, they can use the new Helper classes and traits.
When I was tinkering before, I think I had trouble loading these from WPCS, so we might need some documentation here on what, if anything, needs to be done to make these available to external standards.

jrfnl:
> > On the one hand, they can use the new Helper classes and traits.
> 
> When I was tinkering before, I think I had trouble loading these from WPCS, so we might need some documentation here on what, if anything, needs to be done to make these available to external standards.

If both standards are registered with PHPCS (which the Composer plugin takes care of) and if all files basically follow the expected namespace scheme which the PHPCS autoloader expects, everything should work....
It's the same principle as how we can use PHPCSUtils, with PHPCSUtils registered with PHPCS as a standard, even though it doesn't contain any sniffs, the PHPCS autoloader handles the file retrieval without issue.

I'd need to see the errors you were getting in VIPCS to advise on how to fix that, but that'll have to wait for now.

dingo-d (approved):

PR #2201, opened 2023-01-12
CS/QA: various minor fixes, incl docs fixes

jrfnl:
### Ruleset: various minor doc updates

Rules which are "covered" by a sniff normally are prefixed with "Covers rule:", while rules not (yet) covered by a sniff are prefixed with "Rule: ".

This updates the ruleset inline docs to be consistent with that again.

Includes fixing an incorrect link and some minor text tweaks to be in line with the handbook texts.

### Minor CS/QA fixes


GaryJones (approved):
✅

dingo-d (approved):

PR #2206, opened 2023-02-15
Fix misspelled class in warning text

BrianHenryIE:


jrfnl:
@BrianHenryIE Thanks for your willingness to contribute, however, this is not a misspelling. See: https://developer.wordpress.org/reference/classes/wp_role/

BrianHenryIE:
Ah, both `WP_Role::add_cap()` and `WP_Roles->add_cap()` exist. I thought the `s` was misplaced.

PR #2215, opened 2023-03-24
GH Actions: various changes

jrfnl:
**_This is a prelim PR before updating the PHPCS version to allow for simplifying the management of the protected branches_**

**Note**:
**The protected branch settings will need an update before this PR becomes mergable, but after this, it shouldn't need updating that often anymore.
Once this PR has received approvals, I will update the required statuses and merge the PR myself.**

### GH Actions: minor tweak to composer install

Since Composer 2.2, we can be more specific about which platform requirements should be ignored.

This change ensures that only the "high" end of a PHP requirement will be ignored and no other platform requirements are ignored.

Ref: https://blog.packagist.com/composer-2-2/#-ignore-platform-req-improvements

### GH Actions/quicktest: simplify the matrix

The `include` section was set up in Travis (and transferred to GH Actions) as the _lowest_ supported PHPCS version at that time, wasn't compatible with the highest supported PHP version (yet).

As that is no longer the case, we can now bring this back to a simplified matrix for the Quick Test and just test against high/low PHP with high/low PHPCS.

### GH Actions/ruleset-test: minimize the install

Given that:
* The ruleset test doesn't need PHPUnit to run.
* The ruleset test is run on PHP `latest`, which conflicts with our highest supported PHPUnit version.
* And the Composer install is run without `--ignore-platform-*` resulting in PHPUnit 5.2.7 being installed (last PHPUnit version which didn't have a PHP requirement)

I propose to install with `--no-dev` for the ruleset test.

While this doesn't necessarily remove the "conflict", it should still make the Composer install faster and at least won't create a cache with a completely incorrect PHPUnit version.

### GH Actions: tweak the way the PHPCS versions are set

As things were, whenever the minimum PHPCS version would be changed, the branch protection settings for both the `master` and the `develop` branch would need to be updated and all "required builds" referencing the old PHPCS version would need to be removed, while new "required builds" would need to be added referencing the new minimum PHPCS version.

This was a fiddly process and time-consuming.

The change proposed in this commit takes advantage of the Composer `--prefer-lowest` setting to achieve the same without a hard-coded PHPCS version in the build name, which means that once the branch protection settings have been updated for this PR, they shouldn't need updating anymore for future PHPCS version bumps.

### DIFF
--- a/.github/workflows/basic-qa.yml
+++ b/.github/workflows/basic-qa.yml
@@ -120,6 +120,7 @@ jobs:
       - name: Install Composer dependencies
         uses: ramsey/composer-install@v2
         with:
+          composer-options: --no-dev
           # Bust the cache at least once a month - output format: YYYY-MM.
           custom-cache-suffix: $(date -u "+%Y-%m")
 
--- a/.github/workflows/basic-qa.yml
+++ b/.github/workflows/basic-qa.yml
@@ -98,7 +98,7 @@ jobs:
     strategy:
       matrix:
         php: [ 'latest' ]
-        phpcs_version: [ '3.7.1', 'dev-master' ]
+        phpcs_version: [ 'lowest', 'dev-master' ]
 
     name: "Ruleset test: PHP ${{ matrix.php }} on PHPCS ${{ matrix.phpcs_version }}"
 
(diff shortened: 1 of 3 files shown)


GaryJones (approved):
✅

dingo-d (approved):

jrfnl:
I've updated the branch protection settings. Merging now.

Issue #1967, opened 2020-12-18
PHP 8.0  vsprintf


str:
## Bug Description
When testing with PHP 8 and wpcs 2.3.0 I get the following error message:

```
PHP Fatal error:  Uncaught TypeError: vsprintf(): Argument #2 ($values) must be of type array, string given in /home/str/.config/composer/vendor/squizlabs/php_codesniffer/src/Files/File.php:1051
Stack trace:
#0 /home/str/.config/composer/vendor/squizlabs/php_codesniffer/src/Files/File.php(1051): vsprintf()
#1 /home/str/.config/composer/vendor/squizlabs/php_codesniffer/src/Files/File.php(670): PHP_CodeSniffer\Files\File->addMessage()
#2 /home/str/.config/composer/vendor/squizlabs/php_codesniffer/src/Files/File.php(778): PHP_CodeSniffer\Files\File->addError()
#3 /home/str/.config/composer/vendor/wp-coding-standards/wpcs/WordPress/Sniffs/WhiteSpace/ControlStructureSpacingSniff.php(395): PHP_CodeSniffer\Files\File->addFixableError()
#4 /home/str/.config/composer/vendor/wp-coding-standards/wpcs/WordPress/Sniff.php(910): WordPressCS\WordPress\Sniffs\WhiteSpace\ControlStructureSpacingSniff->process_token()
#5 /home/str/.config/composer/vendor/squizlabs/php_codesniffer/src/Files/File.php(496): WordPressCS\WordPress\Sniff->process()
#6 /home/str/.config/composer/vendor/squizlabs/php_codesniffer/src/Files/LocalFile.php(91): PHP_CodeSniffer\Files\File->process()
#7 /home/str/.config/composer/vendor/squizlabs/php_codesniffer/src/Runner.php(630): PHP_CodeSniffer\Files\LocalFile->process()
#8 /home/str/.config/composer/vendor/squizlabs/php_codesniffer/src/Runner.php(434): PHP_CodeSniffer\Runner->processFile()
#9 /home/str/.config/composer/vendor/squizlabs/php_codesniffer/src/Runner.php(114): PHP_CodeSniffer\Runner->run()
#10 /home/str/.config/composer/vendor/squizlabs/php_codesniffer/bin/phpcs(18): PHP_CodeSniffer\Runner->runPHPCS()
#11 {main}
[...]

PR #2219, opened 2023-04-11
README: make it clear that the current stable version (v2.3.0) does not work on PHP8

bluefuton:
I attempted to install PHPCS and WPCS on a new machine running PHP 8.0.28 today, and spent quite a bit of time researching why it wasn't working and what the appropriate remedy was.

Specifically I ran into this bug: https://github.com/WordPress/WordPress-Coding-Standards/issues/1967 and it doesn't appear that the fix for this will be released until v3.0.0.

In the meantime, it would be kind to developers if the project makes it clear upfront that PHP8 is not supported, and recommends PHP 7.4 instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ This project is a collection of [PHP_CodeSniffer](https://github.com/squizlabs/P
 
 ### Requirements
 
-The WordPress Coding Standards require PHP 5.4 or higher and [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) version **3.7.2** or higher.
+The WordPress Coding Standards require PHP 5.4 or higher and [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) version **3.7.2** or higher. The current stable release does not work on PHP 8, and for the moment we recommend that you use PHP 7.4.
 
 ### Composer
 


jrfnl:
@bluefuton Thank you for your willingness to contribute to WordPressCS.

While I appreciate what you are trying to do, making this change in the `develop` branch is not the right place as the `README` for `develop` applies to the upcoming 3.0.0 version and that version **_does_** work with PHP 8.0+.

jrfnl:
As there have been several issues about this, including issues about how to install WPCS `develop`, it might be an idea to "pin" one of those issues to top of the issue page. What do you think ?

If you think that would help, suggestions on which issue would be the best one to pin would be welcome.

bluefuton:
> While I appreciate what you are trying to do, making this change in the develop branch is not the right place as the README for develop applies to the upcoming 3.0.0 version and that version does work with PHP 8.0+.

Thanks @jrfnl! I did consider this, but the README for `develop` is the first thing users will see when they visit the project on Github. Could I try adjusting the text so it remains appropriate for v3.0.0?

> As there have been several issues about this, including issues about how to install WPCS develop, it might be an idea to "pin" one of those issues to top of the issue page. What do you think ?

Possibly https://github.com/WordPress/WordPress-Coding-Standards/issues/1967, as that's where I ended up - or creating a new issue called "PHP8 compatibility" summarizing the current situation?

My problem solving journey was:

1. Search error message: `phpcs vsprintf(): Argument #2 ($values) must be of type array, string given`
2. Landed on this helpful issue in Human Made's repo: https://github.com/humanmade/coding-standards/issues/278
3. Found the PR fixing the issue https://github.com/WordPress/WordPress-Coding-Standards/pull/1935
4. Discovered that the PR hadn't been released yet
5. Went looking for related issues in the repo, found https://github.com/WordPress/WordPress-Coding-Standards/issues/1967
6. Tried to downgrade to PHP 7.4 but Homebrew wouldn't allow it because it's unsupported. Found that there were ways around that, but decided not to continue faffing with Homebrew.
7. Manually replaced `WordPress/Sniffs/WhiteSpace/ControlStructureSpacingSniff.php` with the latest version from `develop`, and everything works for the moment :)

Thanks for all your hard work on v3.0.0! Sounds like it has been a big effort to get to PHP 8 compatibility.

README.md
```diff
@@ -58,7 +58,7 @@ This project is a collection of [PHP_CodeSniffer](https://github.com/squizlabs/P
 
 ### Requirements
 
-The WordPress Coding Standards require PHP 5.4 or higher and [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) version **3.7.2** or higher.
+The WordPress Coding Standards require PHP 5.4 or higher and [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) version **3.7.2** or higher. The current stable release does not work on PHP 8, and for the moment we recommend that you use PHP 7.4.
```
bluefuton, line 61:
```suggestion
The WordPress Coding Standards require PHP 5.4 or higher and [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) version **3.7.2** or higher. PHP 8 is only supported in WordPress Coding Standards v3.0.0 and above.
```

BWBama85:
> As there have been several issues about this, including issues about how to install WPCS `develop`, it might be an idea to "pin" one of those issues to top of the issue page. What do you think ?
> 
> If you think that would help, suggestions on which issue would be the best one to pin would be welcome.

Where are the instructions on how to install WPCS develop? If you follow the instructions in the readme under develop it installs 2.3.0.

dingo-d:
In your project's `composer.json` file you can add

```json
{
  "repositories": [
    {
      "type": "git",
      "url": "https://github.com/WordPress/WordPress-Coding-Standards.git"
    }
  ],
  "require-dev": {
    "wp-coding-standards/wpcs": "dev-develop"
  }
}
```

Then, running `composer install` will install the develop branch of WPCS.

Hope this helps.

jrfnl:
AFAIK you shouldn't even need to add the `repositories` key in `composer.json` ;-)

U1F (approved):
I think this makes perfect sense.

GaryJones:
Since WP 3.0 is close to release, I'm closing this one out.

PR #2223, opened 2023-04-16
Tests: improve code coverage

jrfnl:
### WPDBTrait: add some extra tests

... to cover some previously uncovered code paths/code patterns.

### CodeAnalysis/AssignmentInTernaryCondition: add extra test

... for parse error edge case / to cover some previously uncovered code paths/code patterns.

### Files/FileName: add some extra tests

... to cover some previously uncovered code paths/code patterns.

### WP/CapitalPDangit: add extra test

... for an edge case / to cover some previously uncovered code paths/code patterns.

### WP/DeprecatedParameterValues: add extra test

... for an edge case / to cover some previously uncovered code paths/code patterns.


GaryJones (approved):
✅

PR #2226, opened 2023-04-18
Move is_class_object_call() utility method to dedicated `ContextHelper` + support nullsafe object operator

jrfnl:
_Note: this PR is part of a series moving various method to this new `ContextHelper` class. Pulling each bit separately to allow for easier reviewing._

### Move is_class_object_call() utility method to dedicated `ContextHelper`

The `is_class_object_call()` utility method is only used by a small set of sniffs, so is better placed in a dedicated class.

This commit moves the `is_class_object_call()` method to the new `WordPressCS\WordPress\Helpers\ContextHelper` class and starts using that class in the relevant sniffs.

The method has also been renamed to `has_object_operator_before()` as it doesn't actually check if something is a (method) _call_, just that something OO (method, constant, property) is being accessed.

Related to #1465

This method is tested via the `WordPress.WP.DiscouragedFunctions` sniff.

### ContextHelper::has_object_operator_before(): simplify/use PHPCSUtils/support PHP 8.0 nullsafe object operator

* A non-inline HTML token will always have another non-empty token before it, if nothing else, the PHP open tag, so checking if `$before` is `false` is redundant.
* Implement use of the `Collections::objectOperators()` token group, which includes the PHP 8.0+ nullsafe object operator.

_Note: this automatically adds support for the nullsafe object operator to all sniffs using this method, either directly or indirectly._

Includes test via the `WordPress.WP.DiscouragedFunctions` sniff.

### DIFF
--- a/WordPress/AbstractFunctionRestrictionsSniff.php
+++ b/WordPress/AbstractFunctionRestrictionsSniff.php
@@ -11,6 +11,7 @@ namespace WordPressCS\WordPress;
 
 use PHP_CodeSniffer\Util\Tokens;
 use PHPCSUtils\Utils\MessageHelper;
+use WordPressCS\WordPress\Helpers\ContextHelper;
 use WordPressCS\WordPress\Helpers\RulesetPropertyHelper;
 use WordPressCS\WordPress\Sniff;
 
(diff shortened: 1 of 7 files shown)


GaryJones (approved):
✅

dingo-d (approved):

PR #2227, opened 2023-04-19
Move is_token_namespaced() utility method to dedicated ContextHelper + minor simplifications

jrfnl:
_Next one in the series_

### Move is_token_namespaced() utility method to dedicated ContextHelper

The `is_token_namespaced()` utility method is only used by a small set of sniffs, so is better placed in a dedicated class.

This commit moves the `is_token_namespaced()` method to the new `WordPressCS\WordPress\Helpers\ContextHelper` class and starts using that class in the relevant sniffs.

Related to #1465

This method is tested via the `WordPress.WP.DiscouragedFunctions` sniff.

### ContextHelper::is_token_namespaced(): minor simplifications

* A non-inline HTML token will always have another non-empty token before it, if nothing else, the PHP open tag, so checking if `$prev` is `false` is redundant.
* Same for the second time this is done, as the same applies for the `T_NS_SEPARATOR` token.

### DIFF
--- a/WordPress/AbstractFunctionRestrictionsSniff.php
+++ b/WordPress/AbstractFunctionRestrictionsSniff.php
@@ -224,7 +224,7 @@ abstract class AbstractFunctionRestrictionsSniff extends Sniff {
 			return false;
 		}
 
-		if ( $this->is_token_namespaced( $stackPtr ) === true ) {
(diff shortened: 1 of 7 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2228, opened 2023-04-19
NamingConventions/ValidVariableName: various minor changes

jrfnl:
### NamingConventions/ValidVariableName: use ContextHelper::has_object_operator_before()

This was previously not possible as the sniff extends an upstream PHPCS sniff and didn't have access to the methods in the `Sniff` class.

### NamingConventions/ValidVariableName: normalize parameter names

This is the only file in which snake_case parameter names are being used for the typical `$phpcsFile` and `$stackPtr` parameters.

This makes synching the file with upstream and otherwise editing the code counter-intuitive.

This commit normalizes the parameter names to use the expected case.


GaryJones (approved):
✅

dingo-d (approved):

PR #2229, opened 2023-04-19
Move `is_in_function_call()` utility method to dedicated `ContextHelper`

jrfnl:
The `is_in_function_call()` utility method is only used by a small set of sniffs, so is better placed in a dedicated class.

This commit moves the `is_in_function_call()` method to the new `WordPressCS\WordPress\Helpers\ContextHelper` class and starts using that class in the relevant sniffs.

Includes adding a type declaration for the `$valid_functions` parameter.

Related to #1465

This method will be tested via the `WordPress.Security.NonceVerification` sniff (via pre-existing and a few new tests).

### DIFF
--- a/WordPress/Helpers/ContextHelper.php
+++ b/WordPress/Helpers/ContextHelper.php
@@ -81,4 +81,86 @@ final class ContextHelper {
 
 		return true;
 	}
+
+	/**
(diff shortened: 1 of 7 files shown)


GaryJones (approved):
✅

dingo-d (approved):

PR #2230, opened 2023-04-20
Move `is_in_type_test()` utility method to dedicated `ContextHelper`

jrfnl:
The `is_in_type_test()` utility method is only used by a small set of sniffs, so is better placed in a dedicated class.

This commit moves the `is_in_type_test()` method and the associated `$typeTestFunctions` property to the new `WordPressCS\WordPress\Helpers\ContextHelper` class and starts using that class in the relevant sniffs.

Related to #1465

This method will be tested via the `WordPress.Security.NonceVerification` sniff (via pre-existing tests).

### DIFF
--- a/WordPress/Helpers/ContextHelper.php
+++ b/WordPress/Helpers/ContextHelper.php
@@ -28,6 +28,40 @@ use PHPCSUtils\Tokens\Collections;
 final class ContextHelper {
 
 	/**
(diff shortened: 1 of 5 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2231, opened 2023-04-21
Move is_in_isset_or_empty() utility method to dedicated `ContextHelper` + improve

jrfnl:
### Move is_in_isset_or_empty() utility method to dedicated `ContextHelper`

The `is_in_isset_or_empty()` utility method is only used by a small set of sniffs, so is better placed in a dedicated class.

This commit moves the `is_in_isset_or_empty()` method to the new `WordPressCS\WordPress\Helpers\ContextHelper` class and starts using that class in the relevant sniffs.

Related to #1465

This method will be tested via the `WordPress.Security.NonceVerification` sniff (via pre-existing and new tests).

### ContextHelper::is_in_isset_or_empty(): move functions array to property

... instead of re-defining it each time the function is called.

### ContextHelper::is_in_isset_or_empty(): use PHPCSUtils

... to simplify the determination of whether a token is within a call to `isset()` or `empty()`.

### ContextHelper::is_in_isset_or_empty(): add support for PHP 8.0 named parameters

Includes tests in the `WordPress.Security.NonceVerification` test file.

### DIFF
--- a/WordPress/Helpers/ContextHelper.php
+++ b/WordPress/Helpers/ContextHelper.php
@@ -254,8 +254,14 @@ final class ContextHelper {
 
 		$functionPtr = self::is_in_function_call( $phpcsFile, $stackPtr, self::$key_exists_functions );
 		if ( false !== $functionPtr ) {
-			$second_param = PassedParameters::getParameter( $phpcsFile, $functionPtr, 2 );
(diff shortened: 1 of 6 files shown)


GaryJones (approved):
✅

PR #2232, opened 2023-04-21
Move is_safe_casted() utility method to dedicated ContextHelper

jrfnl:
### Move is_safe_casted() utility method to dedicated ContextHelper

The `is_safe_casted()` utility method is only used by a small set of sniffs, so is better placed in a dedicated class.

This commit moves the `is_safe_casted()` method and the associated `$safe_casts` property to the new `WordPressCS\WordPress\Helpers\ContextHelper` class and starts using that class in the relevant sniffs.

Related to #1465

This method will be tested via the `WordPress.DB.PreparedSQL` sniff (via pre-existing tests).

### ContextHelper::is_safe_casted(): minor simplification

A non-inline HTML token will always have another non-empty token before it, if nothing else, the PHP open tag, so checking if `$prev` is `false` is redundant.

### DIFF
--- a/WordPress/Helpers/ContextHelper.php
+++ b/WordPress/Helpers/ContextHelper.php
@@ -30,6 +30,22 @@ use PHPCSUtils\Utils\PassedParameters;
 final class ContextHelper {
 
 	/**
+	 * Tokens which when they preceed code indicate the value is safely casted.
+	 *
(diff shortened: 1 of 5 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2233, opened 2023-04-21
WP/I18n: allow for new PHP 8.0+ `[s]printf()` placeholders

jrfnl:
PHP 8.0 silently (not mentioned in migration guide, changelog etc - I only discovered it by accident via a small note in the manual) introduced two new placeholder characters for the `*printf()` range of functions.

This commit updates the regexes used in the `WordPress.WP.I18n` sniff to allow them to recognize those characters as placeholders which should be examined.

Includes unit tests.

Ref: https://www.php.net/manual/en/function.sprintf.php


dingo-d (approved):

GaryJones (approved):
✅

PR #2235, opened 2023-04-24
Move get_list_variables() utility method to dedicated ListHelper + bug fix

jrfnl:
### Move get_list_variables() utility method to dedicated ListHelper

The `get_list_variables()` utility method is only used by a small set of sniffs, so is better placed in a dedicated class.

This commit moves the `get_list_variables()` method to the new `WordPressCS\WordPress\Helpers\ListHelper` class and starts using that class in the relevant sniff.

The functionality in this new class will likely be replaced at some point in the future by functions from PHPCSUtils.

Related to #1465

This method will be tested via the `WordPress.WP.GlobalVariablesOverride` sniff (via pre-existing tests).

### ListHelper::get_list_variables(): use PHPCSUtils / prevent false positives on complex list item keys

Use the PHPCSUtils `Lists::getAssignments()` method to parse a list to its individual items.

Notes:
* As `Lists::getAssignments()` method only expects the list stack pointer, the `$list_open_close` parameter is no longer needed and has been removed.
* With the method now being stand-alone, it would still be good to validate the received token is usable before passing it off to `Lists::getAssignments()`, so some defensive coding has been put in place to do so.
* Even so, the call to `Lists::getAssignments()` still has to be wrapped in a `try`/`catch` as it will throw an exception for live coding/unfinished list structures.

While working on this, I realized there was still a bug in PHPCSUtils. I've fixed that now and the fix is included in the PHPCSUtils `1.0.5` release.

The fix is safeguarded by the extra unit tests.


dingo-d (approved):

GaryJones (approved):
✅

PR #2236, opened 2023-04-25
AbstractFunctionRestrictionsSniff: add extra tests and various bug fixes

jrfnl:
### AbstractFunctionRestrictionsSniff: add test for non-listed function

... to ensure the abstract sniff doesn't act on functions which are not our target.

Tested via tests in the test file for the `WordPress.WP.DiscouragedFunctions` sniff.
_These tests are deliberately placed in this sniff as a) it doesn't have a WP version component, so the test file wouldn't need to be reorganized and b) the sniff is not likely to move to PHPCSExtra as it is WP specific._

### AbstractFunctionRestrictionsSniff::is_targetted_token(): remove some redundant code

1. The sniff only listens to `T_STRING`, so the check for a token code mismatch would never be `true`.
    Note: there are other abstracts which extend from this one, which do listen to other tokens, but those have their own `is_targetted_token()` method.
2. A `T_STRING` token will always have another non-empty token before it, if nothing else, the PHP open tag, so checking if `$prev` is `false` is redundant.

### AbstractFunctionRestrictionsSniff::is_targetted_token(): add tests with OO declarations

... to safeguard that the sniff skips out early for OO structure declarations and make sure that all OO structures are taking into account.

This shouldn't be a problem anymore since PHPCS 3.7.0, but just making sure.

Includes renaming the variable containing the array of tokens to skip out on.

Tested via tests in the test file for the `WordPress.WP.DiscouragedFunctions` sniff.

### AbstractFunctionRestrictionsSniff::is_targetted_token(): add tests with import use declarations

... to safeguard that the sniff does not act on class imports, does act on function imports, but does not act on aliases for imports.

Tested via tests in the test file for the `WordPress.WP.DiscouragedFunctions` sniff.

### AbstractFunctionRestrictionsSniff::is_targetted_token(): add test with live coding/parse error

Tested via tests in the test file for the `WordPress.WP.DiscouragedFunctions` sniff.

### AbstractFunctionRestrictionsSniff::is_targetted_token(): bug fix - false positive on function declaration returning by reference

Includes test in the test file for the `WordPress.WP.DiscouragedFunctions` sniff.

### AbstractFunctionRestrictionsSniff::is_targetted_token(): bug fix - false positive for class instantiation

... when the class has the same name as a targetted function.

Includes test in the test file for the `WordPress.WP.DiscouragedFunctions` sniff.

### AbstractFunctionRestrictionsSniff::is_targetted_token(): prevent false positives on PHP 8.0+ attributes

`T_STRING` tokens in PHP 8.0+ attributes are either class names or possibly constant names (as a parameter for the class instantiation).
They are never function calls (as things are).

This commit ensures that `T_STRING` tokens in attributes are not confused with function calls.

Tested via tests in the test file for the `WordPress.WP.DiscouragedFunctions` sniff.

### AbstractFunctionRestrictionsSniff::process_token(): add tests for the `exclude` property

Tests have been added to the test file for the `WordPress.WP.DiscouragedFunctions` sniff.

### AbstractFunctionRestrictionsSniff: add test with empty functions group

This test is added to the `WordPress.DB.RestrictedFunctions` test class as we need to test the sniff with both the `$unittest_groups` property set and not set, so we need to use a second class to test the abstract properly.

The `WordPress.DB.RestrictedFunctions` sniff was chosen deliberately as it is one of the few function related sniff which use the `'allow'` key, which will allow us to test the correct handling of that key as well.

### AbstractFunctionRestrictionsSniff: bug fix - function names should be compared case-insensitively

While WP natively doesn't use non-snakecase function names, the abstract may be used by external sniffs, so should handle both snake_case as well as camelCase/camelCaps methos correctly.

This commit ensures that the names of the functions in each group, as well as the names of the function in a potential `'allow'` key are handled case-insensitively.

Includes tests via the `WordPress.DB.RestrictedFunctions` sniff.

### AbstractFunctionRestrictionsSniff: add test with PHP 8.1+ first class callables

... to ensure the abstract sniff will cause those to be flagged as well.

Tested via tests in the test file for the `WordPress.WP.DiscouragedFunctions` sniff.

### DIFF
--- a/WordPress/AbstractFunctionRestrictionsSniff.php
+++ b/WordPress/AbstractFunctionRestrictionsSniff.php
@@ -223,7 +223,10 @@ abstract class AbstractFunctionRestrictionsSniff extends Sniff {
 			return false;
 		}
 
-		$prev = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, ( $stackPtr - 1 ), null, true );
+		$search                   = Tokens::$emptyTokens;
+		$search[ \T_BITWISE_AND ] = \T_BITWISE_AND;
+
+		$prev = $this->phpcsFile->findPrevious( $search, ( $stackPtr - 1 ), null, true );
 
 		// Skip sniffing on function, OO definitions or for function aliases in use statements.
 		$invalid_tokens  = Tokens::$ooScopeTokens;
--- a/WordPress/AbstractFunctionRestrictionsSniff.php
+++ b/WordPress/AbstractFunctionRestrictionsSniff.php
@@ -214,11 +214,6 @@ abstract class AbstractFunctionRestrictionsSniff extends Sniff {
 	 * @return bool
 	 */
 	public function is_targetted_token( $stackPtr ) {
-
-		if ( \T_STRING !== $this->tokens[ $stackPtr ]['code'] ) {
-			return false;
-		}
-
 		// Exclude function definitions, class methods, and namespaced calls.
 		if ( ContextHelper::has_object_operator_before( $this->phpcsFile, $stackPtr ) === true ) {
 			return false;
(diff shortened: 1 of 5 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2237, opened 2023-04-29
AbstractClassRestrictionsSniff: implement PHPCSUtils and various bug fixes

### LINKED ISSUE #2184: AbstractClassRestrictionsSniff fails to consider T_SELF, T_STATIC, maybe others before double-colon
## Bug Description

This code
https://github.com/WordPress/WordPress-Coding-Standards/blob/dcbb3ec50911a554c9a4a7876dccb11419dda4e9/WordPress/AbstractClassRestrictionsSniff.php#L138-L142
is attempting to extract the class name from a static call like `Class::method()`, by looking for preceding string tokens. However, when the call is like `self::method()`, the "self" is a T_SELF token rather than a string token. Same for `static::method()`, the "static" is a T_STATIC.

In these cases, the logic skips over the T_SELF or T_STATIC (and various other tokens) until it finds a string, and uses that instead.

I don't know whether there are other tokens that might also appear before a double-colon token and cause similar problems.

## Minimal Code Snippet

I'm not sure if this is the best reproduction, but it's where I started in tracking this down.

```bash
vendor/bin/phpcs -s --standard=WordPress --sniffs=WordPress.WP.ClassNameCase test.php
```

```php
<?php

class Foo {
    public function bar() {
        wp_styles();
        self::foo();

        wp_styles();
        static::foo();
    }
}
```

## Error Code

```
FILE: /tmp/test.php
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 6 | WARNING | It is strongly recommended to refer to classes by their properly cased name. Expected: WP_Styles Found: wp_styles
   |         | (WordPress.WP.ClassNameCase.Incorrect)
 9 | WARNING | It is strongly recommended to refer to classes by their properly cased name. Expected: WP_Styles Found: wp_styles
   |         | (WordPress.WP.ClassNameCase.Incorrect)
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

## Custom ruleset

n/a

## Environment

| PHP version             | 8.0.26
| PHP_CodeSniffer version | 3.7.1
| WPCS version            | dev-develop dcbb3ec
| IDE (if relevant)       | None

I don't think the `$local` parameter to `findPrevious()` on line 139 will be sufficient here, as it would still FP on something like `func( wp_styles(), self::method() )`.

jrfnl:
Note: there is a lot more that needs improving in this abstract, but as we intend to switch to an abstract from PHPCSUtils in the near future and as WP itself doesn't currently use these features very often, if at all, I'm leaving the other shortcoming for the switch to the PHPCSUtils abstract in a future 3.x release.

For the record, the sniff currently has false negatives on:
* `instanceof ClassName`
* Use of exception classes in a `catch()`.
* Use of class (or interface) names in type declarations in any form (plain, union, intersect, DNF / parameter, return or property types / constructor property promotion).
* Use of attribute classes in PHP 8.0+ attributes.

The sniff also does not distinguish correctly between classes and interfaces, i.e. it will examine `implements...` statements as if the names found are classes (while they are interfaces).

For `implements` statements, it does not handle multi-implements - `implements InterfaceA, InterfaceB` - correctly, nor does it handle interface declarations which potentially `extend` multiple interfaces correctly.

For now, I've just implemented PHPCSUtils and fixed the most pertinent bugs I could find in the functionality of the abstract as it is.

----

### AbstractClassRestrictionsSniff::get_namespaced_classname(): efficiency tweak

Instead of doing a preliminary search for the `T_NAMESPACE` token, just let the PHPCSUtils `Namespaces::determineNamespace()` method handle the namespace determination.

Finding the previous `T_NAMESPACE` token can be quite token walking intensive as if the token isn't used, it would mean walking all the way back to the start of the file before `false` is returned.

While the PHPCSUtils method will also do this, it will do this in a much more efficient and precise manner. Additionally, it will cache the results, meaning that it will only do this token walking once per file per PHPCS run (at most).

### AbstractClassRestrictionsSniff::get_namespaced_classname(): bug fix - namespace operator is case-insensitive

The `namespace` keyword as an operator was not being recognized correctly when not in lowercase, while PHP treats the keyword insensitively, so so should the sniff.

Fixed now.

Includes unit test.

### AbstractClassRestrictionsSniff::is_targetted_token(): bug fix - false positives icw hierarchy keywords

When a hierarchy keyword or a variable would be before the double colon, the sniff would potential walk too far back and identify an unrelated `T_STRING` token as the class name.

Includes unit tests (which would throw two errors on each line without the fix).

Fixes #2184

### AbstractClassRestrictionsSniff::is_targetted_token(): use PHPCSUtils [1]

... for the name determination based on a `::`.

Note: the condition checking for `T_NS_SEPARATOR` is not necessary as the `get_namespaced_classname()` method checks this as well and will return straight away in that case.

### AbstractClassRestrictionsSniff::is_targetted_token(): bug fix - false negative on class instantiation ending on close tag

Includes unit test.

### AbstractClassRestrictionsSniff::is_targetted_token(): bug fix - false negative on class instantiation with method chaining

Note: the `->` or `?->` object operators can only be used in this context if the `new ...` is wrapped in parentheses, so the `T_OBJECT_OPERATOR` was redundant in the search.

Includes unit tests.

### AbstractClassRestrictionsSniff::is_targetted_token(): bug fix - sniff does not bow out early on hierarchy keywords

The abstract class turns every found name into a fully qualified name, so, as things are, the hierarchy keywords would be prefixed with a `\` too, so the `in_array()` comparison would never match.

While this doesn't lead to false positives, it should still be fixed for performance reasons.

Fixed now.

Includes unit test.

### AbstractClassRestrictionsSniff::is_targetted_token(): bug fix - hierarchy keywords are case-insensitive

... so should be matched in a case-insensitive manner.

While this doesn't lead to false positives, it should still be fixed for performance reasons.

Fixed now.

Includes (adjusted) unit tests.

### AbstractClassRestrictionsSniff::is_targetted_token(): use PHPCSUtils [2]

... for the name determination based on a `new ...` etc.

Note: the condition checking for `T_NS_SEPARATOR` is not necessary as the `get_namespaced_classname()` method checks this as well and will return straight away in that case.

### AbstractClassRestrictionsSniff::is_targetted_token(): add tests with PHP 8.1+ enums

... which can also `implement`.


GaryJones (approved):
✅

dingo-d (approved):

PR #2243, opened 2023-05-22
GH Actions: fix fixer conflict check

jrfnl:
While checking something else, I came across a notice in the GH Actions logs showing that the fixer conflict check was failing.

Some backtracing shows that it has been failing for the past five months since PR #2132 was merged - see: https://github.com/WordPress/WordPress-Coding-Standards/actions/runs/3643130487

Unfortunately, the way the script was set up, this failure was not showing and not failing the builds. /cc @desrosj 

The changes I'm proposing now will fix that.

I've also made sure that once the current fixer conflicts are fixed, the build will properly succeed again: https://github.com/jrfnl/WordPress-Coding-Standards/actions/runs/5048089688/jobs/9055836148

Next up: fix the fixer conflicts and whatever other errors have crept in since the build originally started failing....

### GH Actions: temporarily ignore one test case file for the fixer conflict check

... until upstream PR squizlabs/PHP_CodeSniffer#3833 has been merged.

This will allow for the fixer conflict check to be enabled again already and to prevent new issues from being introduced.

PR #2244, opened 2023-05-27
Tests: fix unintentional (parse/compile) errors in test case files

jrfnl:
### Tests: use unique function names

.. to make it easier to do a parse/compile error check on test case files.

Related to #2243


dingo-d (approved):

GaryJones (approved):
✅

PR #2245, opened 2023-05-27
WP/EnqueuedResources: bring back try/catch

jrfnl:
Commit dec9d40b33939678ed57ae88e977b5fe7198a55b, which was part of PR #2176 removed this `try/catch` and while it shouldn't be needed anymore for valid code, it _could_ still be needed for code containing parse errors.

This commit doesn't completely revert the previous commit, but does bring the try/catch back to prevent this situation.

Includes unit test safeguarding the fix.

Related to #2243


dingo-d (approved):

GaryJones (approved):
✅

PR #2247, opened 2023-06-13
Core: add section about namespace declaration rules with select new sniffs

### LINKED ISSUE #1763: Proposal: Add rules about (formatting of) namespace declarations
PHP 5.3 introduced namespaces into PHP.

Refs:
* https://www.php.net/manual/en/migration53.new-features.php
* https://www.php.net/manual/en/language.namespaces.php

At this moment WPCS has no formatting rules for namespace declarations or the usage of namespaced constructs.

This issue, however, is purely about **namespace declarations**.

I would like to propose adding the following rules regarding namespace declarations to the WP Coding Standards handbook:
1. There should be exactly one space between the `namespace` keyword and the start of the namespace name in a namespace declaration.
2. Each "leaf" of a namespace name should be in `Camel_Caps`, i.e. each word should start with a capital and words should be separated by an underscore.
    Consecutive caps, like for acronyms, will be allowed.
    This is in line with other WP naming conventions.
3. There should be no whitespace or comments within the name part of the namespace declaration.
4. There should be exactly one blank line before a namespace declaration.
5. There should be at least one blank line after a namespace declaration.
    The "at least" bit prevent potential fixer conflicts with sniffs which check for a fixed amount of blank lines above function/class declarations.
6. There should be only one namespace declaration per file.
7. Scoped namespace declarations are not allowed.
8. Namespace declarations without name (= explicit global namespace) are not allowed.

### Valid code example
```php
<?php
/**
 * File docblock.
 */

namespace Prefix\Admin\Domain_URL\Sub_Domain\Event;

use ...
```

### Invalid code examples
```php
<?php
/**
 * File docblock.
 */
namespace Prefix \
	/* Here comes the real namespace */ Admin\DomainURL\SubDomain\Event;
use ...
```

```php
<?php
/**
 * File docblock.
 */
namespace Foo {
    // Code.
}

namespace {
    // Code.
}
```

If this proposal is accepted, we can then add sniff(s) to WPCS to check these rules automatically and - except for issues with comments - auto-fix them.

### Already covered
* The `namespace` keyword should be lowercase.
    This is already covered by the `Generic.PHP.LowerCaseKeyword.Found` sniff which is included in the `WordPress-Core` ruleset.

### Open question:
* Should we - at this time - define rules about the allowed compound depth of namespace names ? Or should we leave this open for now ?

### Explicitly not covered by this proposal:
* The name of a namespace should not start with a (superfluous) namespace separator.
    This would be a parse error in nearly all cases as PHP will in that case interpret the namespace name as the name of a global (namespaced) constant.
    Parse errors should be detected by linters and/or by the `Generic.PHP.Syntax` sniff.
* The order of various statements which can be at the start of a document.
    This will be covered in a separate proposal.
* Any relation between namespace declarations and autoloading / the path to the file.

### Comparison with PSR-12:
The only rule in the above proposal which is included in PSR-12 is rule 4 (one blank line before) and in a way rule 5 (at least one blank line after).
PSR-12 has no opinion on any of the other rules proposed here.

PSR-12 does have an opinion on the allowed compound namespace depth, but what with plugins and themes needing to use prefixing of namespaces, the PSR-12 rule for this may ultimately not be suitable for use in the context of WordPress.

### Comparison with PSR-1:
While PSR-1 (included in PSR-12) demands `StudlyCaps`/`PascalCase` for class names, it does not do so explicitly for namespace names.

PSR-1 does demand implementing PSR-0 or PSR-4 for autoloading, but as autoloading is explicitly not covered in this proposal, this is not something we currently need to take into account.

## Notes for the WPCS Implementation:
* Rule 1 can be covered by the upstream `Generic.WhiteSpace.LanguageConstructSpacing.IncorrectSingle` sniff. Also see issue #1616
* Rule 2 would need a dedicated WP sniff in the `NamingConventions` category.
* Rule 3 can be covered by an upstream sniff once it is pulled & merged. See: https://github.com/squizlabs/PHP_CodeSniffer/issues/2150
* Rule 4 can possibly be covered by the upstream PSR-12 sniff for the same, once it is created, pulled & merged.
* Rule 5 is already covered in `WordPress-Extra` by the upstream `PSR2.Namespaces.NamespaceDeclaration.BlankLineAfter` sniff. This sniff could be moved to `WordPress-Core`.
* Rule 6, 7 and 8 can be covered by upstream sniffs once pulled & merged. See: https://github.com/squizlabs/PHP_CodeSniffer/issues/2324

/cc @pento

jrfnl:
> ### Namespace declarations
>
> Each part of a namespace name should consist of capitalized words separated by underscores.
>
> Namespace declarations should have exactly one blank line before the declaration and at least one blank line after.
>
> There should be only one namespace declaration per file, and it should be at the top of the file. Namespace declarations using curly brace syntax are not allowed. Explicit global namespace declaration (namespace declaration without name) are also not allowed.

Refs:
* https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/ - Namespace declarations section
* https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#namespace-declarations
* WordPress/wpcs-docs#97
* WordPress/WordPress-Coding-Standards#1763
* PHPCSStandards/PHPCSExtra#4
* PHPCSStandards/PHPCSExtra#6
* PHPCSStandards/PHPCSExtra#50

Notes:
* The "_Each part of a namespace name should consist of capitalized words separated by underscores_" rule needs a new dedicated sniff.
    An issue will need to be opened about this in WPCS. The corresponding issue in PHPCSExtra is PHPCSStandards/PHPCSExtra#231
* The "_Namespace declarations should have exactly one blank line before the declaration and at least one blank line after_" will probably also need a new sniff.
    My research shows the following:
    - The `PSR2.Namespaces.NamespaceDeclaration` sniff, which is included in `Extra` can _sort of_ cover the "after" part, as in: it check for exactly one blank line after, which is close to, but not exactly what we want.
    - The `PSR12.Files.FileHeader` sniff can check both "before" and "after", but will also, again, check for exactly one blank line.
        The problem with that sniff is that it currently is "all or nothing", it does not have modular error codes, so we cannot ignore some other things from that sniff (requires blank line between PHP open tag and file docblock), which makes it problematic to include the sniff.
        If upstream PR squizlabs/PHP_CodeSniffer#2729 would (finally) be merged, we could reconsider adding that sniff though.
* As for the "_... it (the namespace declaration) should be at the top of the file_" rule. I do not believe we need to actively check for this.
    With the "_only one namespace declaration per file_" rule being enforced, placing the namespace declaration anywhere else than at the top of the file would be a parse error.

Other notes:
* The "_namespace keyword should be lowercase_" part, as mentioned in the Make post, is already covered by the `Generic.PHP.LowerCaseKeyword` sniff.
* The "_There should be exactly one space between the namespace keyword and the start of the namespace name_" part, as mentioned in the Make post, is already covered by the `Generic.WhiteSpace.LanguageConstructSpacing` sniff, which was moved to `Core` in #2097.
* The "_namespace names in a namespace declaration should not start with a leading backslash \\_", as well as the "_There should be no whitespace or comments within the name part of the namespace declaration._", as mentioned in the Make post, are things which can (and should) be flagged by PHPCompatibility 10.0.

Closes #1763

### DIFF
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -414,6 +414,29 @@
 
 	<!--
 	#############################################################################
+	Handbook: Declare Statements, Namespace, and Import Statements - Namespace declarations.
+	Ref: https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#namespace-declarations
+	#############################################################################
+	-->
+	<!-- Rule: Each part of a namespace name should consist of capitalized words separated by underscores. -->
+
+	<!-- Rule: Namespace declarations should have exactly one blank line before the declaration and at least one blank line after. -->
+
+	<!-- Covers rule: There should be only one namespace declaration per file... -->
+	<rule ref="Universal.Namespaces.OneDeclarationPerFile"/>
+
+	<!-- Rule: ... and it should be at the top of the file.
+		 Note: with only one namespace declaration, it not being at the top of the file would be a parse error. -->
+
+	<!-- Covers rule: Namespace declarations using curly brace syntax are not allowed. -->
+	<rule ref="Universal.Namespaces.DisallowCurlyBraceSyntax"/>
+
+	<!-- Covers rule: Explicit global namespace declaration (namespace declaration without name) are also not allowed. -->
+	<rule ref="Universal.Namespaces.DisallowDeclarationWithoutName"/>
+
+
+	<!--
+	#############################################################################
 	Handbook: Object-Oriented Programming - Only One Object Structure (Class/Interface/Trait) per File.
 	Ref: https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#only-one-object-structure-class-interface-trait-per-file
 	#############################################################################


dingo-d (approved):

GaryJones (approved):
✅

PR #2248, opened 2023-06-14
Move is_use_of_global_constant() utility method to dedicated `ConstantsHelper`

jrfnl:
### Move is_use_of_global_constant() utility method to dedicated `ConstantsHelper`

The `is_use_of_global_constant()` utility method is only used by a small set of sniffs, so is better placed in a dedicated class.

This commit moves the `is_use_of_global_constant()` method to the new `WordPressCS\WordPress\Helpers\ConstantsHelper` class and starts using that class in the relevant sniff.

Related to #1465

The functionality in this new class will very likely be replaced at some point in the future by functions from PHPCSUtils.

This method will be tested via the `WordPress.Security.EscapeOutput` sniff (via pre-existing tests).

### ConstantsHelper::is_use_of_global_constant(): simplify/implement PHPCSUtils

The `$tokens_to_ignore` array was set up using token types (strings) instead of token constants as some of the tokens would not always be available in all supported PHP/PHPCS combinations.

This is no longer needed now PHPCS 3.7.1 is the minimum supported PHPCS version.

This commit:
* Changes the token types to token codes.
* Simplifies the array by leveraging pre-existing token arrays.
    - `T_CLASS`, `T_TRAIT` and `T_INTERFACE` are replaced by `Tokens::$ooScopeTokens`, which also includes the PHP 8.1 `T_ENUM` token.
    - `T_DOUBLE_COLON` and `T_OBJECT_OPERATOR` are replaced by `Collections::objectOperators()`, which also includes the PHP 8.0 `T_NULLSAFE_OBJECT_OPERATOR` token.
    - `T_PUBLIC`, `T_PROTECTED` and `T_PRIVATE` are replaced by `Tokens::$scopeModifiers`.

The PHPCompatibility standard has the same method and the method is expected to be moved at some point to PHPCSUtils.
PHPCompatibility also has dedicated tests for this method, so I'm not adding tests at this time for `T_ENUM` and `T_NULLSAFE_OBJECT_OPERATOR` support.

See: PHPCompatibility/PHPCompatibility#1439

### ConstantsHelper::is_use_of_global_constant(): minor simplification

A non-inline HTML token will always have another non-empty token before it, if nothing else, the PHP open tag, so checking if `$prev` is `false` is redundant.

### DIFF
--- a/WordPress/Helpers/ConstantsHelper.php
+++ b/WordPress/Helpers/ConstantsHelper.php
@@ -11,6 +11,7 @@ namespace WordPressCS\WordPress\Helpers;
 
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Tokens\Collections;
 use PHPCSUtils\Utils\Scopes;
 use WordPressCS\WordPress\Helpers\ContextHelper;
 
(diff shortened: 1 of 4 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2249, opened 2023-06-14
Move escaping functions related functionality to dedicated `EscapingFunctionsTrait`

jrfnl:
Move escaping functions related functionality to dedicated `EscapingFunctionsTrait`

The escaping function lists are only used by a small set of sniffs, so are better placed in a dedicated trait.

The choice for a `trait` over a `class` is due to the `public` properties allowing for adding additional functions to these lists.

Moving both the base function lists + the `public` properties to the same trait will allow us to encapsulate all the functionality related to the use of these lists in one place.

The `$escapingFunctions` and the `$autoEscapedFunctions` properties, containing the base lists, have also been made `private`.

Checking whether or not something is an escaping function should now be done by calling the `EscapingFunctionsTrait::is_escaping_function()` and/or the `EscapingFunctionsTrait::is_auto_escaped_function()` methods.

Related to #1465


GaryJones (approved):
✅

dingo-d (approved):

PR #2250, opened 2023-06-15
Move printing functions related functionality to dedicated `PrintingFunctionsTrait`

jrfnl:
The printing function list is only used by a small set of sniffs, so are better placed in a dedicated trait.

The choice for a `trait` over a `class` is due to the `public` property allowing for adding additional functions to the list.

Moving both the base function list + the `public` property to the same trait will allow us to encapsulate all the functionality related to the use of this list in one place.

The `$printingFunctions` property, containing the base lists, has also been made `private`.

Checking whether or not something is a printing function should now be done by calling the `PrintingFunctionsTrait::is_printing_function()` method.

Related to #1465

### DIFF
--- a/WordPress/Helpers/PrintingFunctionsTrait.php
+++ b/WordPress/Helpers/PrintingFunctionsTrait.php
@@ -1,0 +1,112 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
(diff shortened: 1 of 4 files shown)


dingo-d (approved):
I wanted to leave a comment in the previous PR if these are going to be moved as well, but I kinda had the hunch you already did that 😄

GaryJones (approved):
✅

jrfnl:
Just checking @dingo-d @GaryJones: What with you both approving, can this be merged ?

PR #2251, opened 2023-06-16
Minor doc tweaks

jrfnl:
Follow up on #2249 and #2250.

### DIFF
--- a/WordPress/Helpers/PrintingFunctionsTrait.php
+++ b/WordPress/Helpers/PrintingFunctionsTrait.php
@@ -29,9 +29,9 @@ trait PrintingFunctionsTrait {
 	 * Custom list of functions which print output incorporating the passed values.
 	 *
 	 * @since 0.4.0
-	 * @since 3.0.0 Moved from the EscapeOutput Sniff class to this class.
+	 * @since 3.0.0 Moved from the EscapeOutput Sniff class to this trait.
 	 *
-	 * @var string|string[]
+	 * @var string[]
 	 */
 	public $customPrintingFunctions = array();
 
@@ -40,10 +40,10 @@ trait PrintingFunctionsTrait {
 	 *
 	 * @since 0.5.0
 	 * @since 0.11.0 Changed from public static to protected non-static.
-	 * @since 3.0.0 - Moved from the Sniff class to this class.
+	 * @since 3.0.0 - Moved from the Sniff class to this trait.
 	 *              - Visibility changed from protected to private.
 	 *
-	 * @var array
+	 * @var array<string, bool>
 	 */
 	private $printingFunctions = array(
 		'_deprecated_argument'    => true,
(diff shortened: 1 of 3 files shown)


dingo-d (dismissed):

jrfnl:
Sorry, found some more doc issues and fixed those up now (improved specificity in the `@var` type specifications).

WordPress/Helpers/PrintingFunctionsTrait.php
```diff
@@ -70,10 +70,10 @@ trait PrintingFunctionsTrait {
 	 * @since 0.4.0
 	 * @since 0.11.0 - Changed from public static to protected non-static.
 	 *               - Changed the format from simple bool to array.
-	 * @since 3.0.0  - Moved from the EscapeOutput Sniff class to this class.
+	 * @since 3.0.0  - Moved from the EscapeOutput Sniff class to this trait.
 	 *               - Visibility changed from protected to private.
 	 *
-	 * @var array
+	 * @var string|string[]
```
GaryJones, line 76: What's the advantage to keeping the customPrintingFunction allowed as a string?

dingo-d (approved):

WordPress/Helpers/PrintingFunctionsTrait.php
jrfnl: Good question. Same question actually applies to the `$customEscapingFunctions` and `$customAutoEscapedFunctions` (which already had the above type annotation so don't show up in this PR).

In the [wiki](https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#xss-custom-output-escaping-functions) they all are annotated as expecting an array.

If I remember correctly, the annotation is like that for historic reasons, as people prior to the array notation syntax which was added in PHPCS 3.3.0 would easily pass these incorrectly (forget the `type="array"`).

I'm not even sure it would still work if passed as a `string` at this moment and I doubt there are any tests for it.

As we're at 3.0.0 now, now may actually be a good time to officially remove the tolerance for the value being passed as a `string`, even if only in annotation as I'm not sure any code changes are needed.

What do you think ?

jrfnl: (same also applies to the `$customSanitizingFunctions` and `$customUnslashingSanitizingFunctions`)

GaryJones: > As we're at 3.0.0 now, now may actually be a good time to officially remove the tolerance for the value being passed as a string, even if only in annotation as I'm not sure any code changes are needed.

Agreed on all cases - it seems like an oddity for us to initialise the class property as an array, but still document it as allowing a string in. The `type="array"` changes for PHPCS have been around a while now, and a WPCS 3.0.0 seems like a good time to cut the legacy support.

jrfnl: In that case, shall I squash the two commits so this PR can be merged ?

There's a separate PR upcoming to split off the sanitizing functions list and I'll handle the removal of the `string` bit from the docblock for those there.

GaryJones: Sounds good.

jrfnl: Squashed ;-)

GaryJones (approved):
✅

PR #2253, opened 2023-06-19
Move formatting functions related functionality to dedicated `FormattingFunctionsHelper`

jrfnl:
The formatting function list is only used by a small set of sniffs, so are better placed in a dedicated class.

The `$formattingFunctions` property has also been made `private static`.

Checking whether or not something is a formatting function should now be done by calling the `FormattingFunctionsHelper::is_formatting_function()` method.

Related to #1465


dingo-d (approved):
How come `explode` isn't on the list?

I thought that only the formatting functions that would return a string would be on the list, but `array_fill` returns a filled array.

PR #2256, opened 2023-06-19
Move "array walking" functions related functionality to dedicated `ArrayWalkingFunctionsHelper` + support PHP 8.0+ named parameters

jrfnl:
### Move "array walking" functions related functionality to dedicated `ArrayWalkingFunctionsHelper`

The "array walking" function list is only used by a small set of sniffs, so are better placed in a dedicated class.

The `$arrayWalkingFunctions` property has also been made `private static`.

Checking whether or not something is an "array walking" function should now be done by calling the `ArrayWalkingFunctionsHelper::is_array_walking_function()` method.

Related to #1465

### ArrayWalkingFunctionsHelper: add support for PHP 8.0+ named parameters

* Transform the `$arrayWalkingFunctions` array to a multi-dimensional array to allow for it to include information about parameter names.
* Add new `get_callback_parameter()` method to retrieve the parameter information using the PHPCSUtils `PassedParameters` functionality.
* Adjust the sniff code retrieving a parameter from these functions to use this new method.
* Adjust the `get_array_walking_functions()` method to still return a single-dimensional array.

### DIFF
--- a/WordPress/Helpers/ArrayWalkingFunctionsHelper.php
+++ b/WordPress/Helpers/ArrayWalkingFunctionsHelper.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Helpers;
+
+/**
+ * Helper functions and function lists for checking whether a function applies a callback to an array.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @since   3.0.0 The property in this class was previously contained in the
+ *                `WordPressCS\WordPress\Sniff` class and has been moved here.
+ */
+final class ArrayWalkingFunctionsHelper {
+
+	/**
+	 * List of array functions which apply a callback to the array.
+	 *
+	 * These are often used for sanitization/escaping an array variable.
+	 *
+	 * Note: functions which alter the array by reference are not listed here on purpose.
+	 * These cannot easily be used for sanitization as they can't be combined with unslashing.
+	 * Similarly, they cannot be used for late escaping as the return value is a boolean, not
+	 * the altered array.
+	 *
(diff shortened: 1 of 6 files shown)


WordPress/Helpers/ArrayWalkingFunctionsHelper.php
```diff
@@ -0,0 +1,108 @@
...
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Helpers;
+
+use PHP_CodeSniffer\Files\File;
+use PHPCSUtils\Utils\PassedParameters;
+
+/**
+ * Helper functions and function lists for checking whether a function is one which applies a callback to an array.
```
GaryJones, line 16:
```suggestion
 * Helper methods and function lists for checking whether a function applies a callback to an array.
```

WordPress/Helpers/ArrayWalkingFunctionsHelper.php
```diff
@@ -0,0 +1,108 @@
...
+	 * Similarly, they cannot be used for late escaping as the return value is a boolean, not
+	 * the altered array.
+	 *
+	 * @since 2.1.0
+	 * @since 3.0.0 - Moved from the Sniff class to this class.
+	 *              - Visibility changed from protected to private and property made static.
+	 *              - The value has changed from an integer to an array containing the integer
+	 *                parameter position + its name.
+	 *
+	 * @var array <string function name> => <array>
```
GaryJones, line 40:
```suggestion
	 * @var array<string, array>
```

jrfnl: Fixed up in the relevant commit

WordPress/Helpers/ArrayWalkingFunctionsHelper.php
jrfnl: Re: `methods` - I think the same phrasing was also used in the other helpers. We should either fix them all or leave them all as is... I'm tempted to leave it for now as a `method` is a function after all, just one which is within an OO structure.

dingo-d (approved):

GaryJones (approved):
✅

Issue #1942, opened 2020-07-31
CapitalPDangit doesn't prevent matches on reverse domain name notation


mirka:
## Bug Description

The `CapitalPDangit` sniff can correct case-sensitive identifiers in reverse domain name notation (e.g. `com.wordpress`).

## Minimal Code Snippet

```php
<?php
$stat_id = 'com.wordpress.signups';
```

## Error Code

`WordPress.WP.CapitalPDangit.Misspelled`

## Environment

| Question               | Answer
| ------------------------| -------
| PHP version             | 7.3.11 
| PHP_CodeSniffer version | 3.5.5
| WPCS version            | 2.3.0
| WPCS install type       | Composer project local

## Additional Context (optional)

Similar issue to #1698

I understand it's unreasonable to expect this sniff to prevent every single false positive, but just flagging this particular case since it is a potentially bug-inducing autofix 🙂

## Tested Against `develop` branch?
- [ ] I have verified the issue still exists in the `develop` branch of WPCS.

jrfnl:
@mirka Could you give me some context about the use of these "reverse domain name" notations ? AFAIK this is not something which is supported by WP by default or am I missing something ?

mirka:
Yes, that’s probably correct, I don’t imagine the WP Core codebase uses notation like this anywhere. This is just an arbitrary identifier that was in our team’s codebase.

If it matters, they were IDs for tracking stats, and they had [reverse-DNS](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) prefixes for namespacing.

https://github.com/WordPress/WordPress-Coding-Standards/blob/ac0a94c2f831b9adb633e7af5f000d6a0df650f9/WordPress/Sniffs/WP/CapitalPDangitSniff.php#L244-L256

^ I saw that there was already some false positive prevention code like this. But I wasn’t sure if these were specifically meant for exceptions that could occur within the WP Core code, or for broader applications as well. If it’s the former, and this case is not worth preventing specifically, that’s totally reasonable!

jrfnl:
@mirka The false positive prevention code currently in place is for things which regularly occur in codebases, not for individual usecases.

For the time being, I'd like to suggest for you to use the standard PHPCS ignore annotations to prevent the issue being reported and the auto-fixer from changing the code:
```php
// phpcs:ignore WordPress.WP.CapitalPDangit.Misspelled -- Reverse DNS notation.
$stat_id = 'com.wordpress.signups';
```

[...]

PR #2257, opened 2023-06-19
WP/CapitalPDangit: use more modular error codes

jrfnl:
This splits the `Misspelled` error code into two distinct error codes to allow for more modular ignoring of issues.

The new error codes are based on the context in which the misspelling is found (similar to other modular codes already in place, like `MisspelledNamespaceName`):
* `MisspelledInText`
* `MisspelledInComment`

As per the discussion in #1942.


GaryJones (approved):
✅

dingo-d (approved):

PR #2259, opened 2023-06-20
Move sanitization functions related functionality to dedicated `SanitizingFunctionsTrait`

jrfnl:
The sanitization function lists are only used by a small set of sniffs, so are better placed in a dedicated trait.

The choice for a `trait` over a `class` is due to the `public` properties allowing for adding additional functions to the lists.

Moving both the base function lists + the `public` properties to the same trait will allow us to encapsulate all the functionality related to the use of these lists in one place.

The `$sanitizingFunctions` and the `$unslashingSanitizingFunctions` property, containing the base lists, have also been made `private`.

Checking whether or not something is a sanitization function should now be done by calling the `SanitizingFunctionsTrait::is_sanitizing_function()` or the `SanitizingFunctionsTrait::is_sanitizing_and_unslashing_function()` method.

Related to #1465

### DIFF
--- a/WordPress/Helpers/SanitizationFunctionsTrait.php
+++ b/WordPress/Helpers/SanitizationFunctionsTrait.php
@@ -1,0 +1,238 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Helpers;
+
+use WordPressCS\WordPress\Helpers\RulesetPropertyHelper;
+
+/**
+ * Helper functions and function lists for checking whether a sanitization function is being used.
+ *
+ * Any sniff class which incorporates this trait will automatically support the
+ * following `public` properties which can be changed from within a custom ruleset:
+ * - `customSanitizingFunctions`.
+ * - `customUnslashingSanitizingFunctions`
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @since   3.0.0 The properties in this trait were previously contained partially in the
+ *                `WordPressCS\WordPress\Sniff` class and partially in the `NonceVerificationSniff`
+ *                and the `ValidatedSanitizedInputSniff` classes and have been moved here.
+ */
+trait SanitizationFunctionsTrait {
+
+	/**
+	 * Custom list of functions that sanitize the values passed to them.
+	 *
+	 * @since 0.5.0
+	 * @since 3.0.0 Moved from the NonceVerification and the ValidatedSanitizedInput sniff classes to this trait.
+	 *
+	 * @var string[]
+	 */
+	public $customSanitizingFunctions = array();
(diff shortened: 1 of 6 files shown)


WordPress/Helpers/SanitizationFunctionsTrait.php
```diff
@@ -0,0 +1,238 @@
...
+		'wp_sanitize_redirect'       => true,
+		'wp_strip_all_tags'          => true,
+	);
+
+	/**
+	 * Sanitizing functions that implicitly unslash the data passed to them.
+	 *
+	 * This list is complementary to the `$sanitizingFunctions` list.
+	 * Sanitizing functions should be added to this list if they also
+	 * implicitely unslash data and to the `$sanitizingFunctions` list
```
GaryJones, line 112:
```suggestion
	 * implicitly unslash data and to the `$sanitizingFunctions` list
```

WordPress/Helpers/SanitizationFunctionsTrait.php
```diff
@@ -0,0 +1,238 @@
...
+	private $allUnslashingSanitizingFunctions = array();
+
+	/**
+	 * Retrieve a list of all known sanitization functions.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var array<string, bool>
+	 */
+	public function get_sanitization_functions() {
```
GaryJones, line 176: How come this is function named `sanitization` and not `sanitizING`? Seems inconsistent with the property names as well as "escapING", "array walkING", and other -ING references.

jrfnl: Grammar technically, "sanitization" sounds better to me that "sanitizing", but happy enough to change it (in all the places in the trait).

jrfnl: If we change this, should the trait name also be changed to `SanitizingFunctionsTrait` ?

jrfnl: I've adjusted this in a separate commit now. Let me know what you think and i can squash the commits if okay.

GaryJones (approved):
✅  Good to squash.

WordPress/Helpers/SanitizationFunctionsTrait.php
GaryJones: Looks good - good to squash.

dingo-d (approved):

WordPress/Helpers/SanitizationFunctionsTrait.php
jrfnl: Squashed + updated the commit message and the above PR description to use the new Trait and method names.

jrfnl:
Hmm.. I  force-pushed the branch and the branch (correctly) says there is now one commit, but this PR still shows as if there were 3 commits... ? Didi I just break GH ?

image

dingo-d:
Yeah, I can see 3 commits.

I don't see a force push commit though 🤷🏼‍♂️

jrfnl:
As this PR doesn't seem to want to update, shall we do Squash commit in the PR itself ?

PR #2263, opened 2023-06-23
Move unslashing-only functions related functionality to dedicated `UnslashingFunctionsHelper`

jrfnl:
The unslashing function list is only used by a small set of sniffs, so are better placed in a dedicated class.

The `$unslashingFunctions` property has also been made `private static`.

Checking whether or not something is an unslashing function should now be done by calling the `UnslashingFunctionsHelper::is_unslashing_function()` method.

Includes removing a `@internal` comment from the `Sniff` class docblock as it is now redundant.

Related to #1465


GaryJones (approved):
✅

dingo-d (approved):

PR #2265, opened 2023-06-25
WP/PostsPerPage: simplification

jrfnl:
The WPCS `AbstractArrayAssignmentRestrictionsSniff` expects an array from `getGroups()` which includes a `'message'` key. The `callback()` method subsequently allows to override that message with a higher priority message if needed in specific circumstances.

This sniff, however, doesn't need to override the message, so the `callback()` method can just use the default behaviour of returning `true` or `false` and use the `'message'` as declared in the group.


dingo-d (approved):

GaryJones (approved):
✅

PR #2268, opened 2023-06-28
Improve consistency in method names across Helpers

jrfnl:
### ArrayWalkingFunctionsHelper: improve consistency in method names across Helpers

For the stand-alone Helper classes which only contain one "functions list" array property, name the retrieval method `get_functions()`.

After all, these are not traits, which run a risk of a class `use`-ing the trait overloading the method and as the method is `static`, the class name used in the function call already clearly indicates _which_ function list is being retrieved.

### UnslashingFunctionsHelper: improve consistency in method names across Helpers

For the stand-alone Helper classes which only contain one "functions list" array property, name the retrieval method `get_functions()`.

After all, these are not traits, which run a risk of a class `use`-ing the trait overloading the method and as the method is `static`, the class name used in the function call already clearly indicates _which_ function list is being retrieved.


jrfnl:
Note: The `WPHookHelper` class already uses this same method name.

GaryJones (approved):

dingo-d (approved):

PR #2269, opened 2023-06-28
Helpers: add some more defensive coding

jrfnl:
The `Helper` classes containing stand-alone `static` methods can be used by both WPCS itself as well as external standards extending WPCS.

As these methods have been made stand-alone, they should probably contain a little more defensive coding in select places.

Note: this code will not be "covered" by tests at this moment as the WPCS code as-is will never hit these conditions.

At a later point in time, we may want to add dedicated tests for the Helper methods which aren't moving to PHPCSUtils. When we do, we can make sure this defensive code gets covered too.

### DIFF
--- a/WordPress/Helpers/ContextHelper.php
+++ b/WordPress/Helpers/ContextHelper.php
@@ -129,6 +129,10 @@ final class ContextHelper {
 	 */
 	public static function has_object_operator_before( File $phpcsFile, $stackPtr ) {
 		$tokens = $phpcsFile->getTokens();
+		if ( isset( $tokens[ $stackPtr ] ) === false ) {
+			return false;
+		}
+
 		$before = $phpcsFile->findPrevious( Tokens::$emptyTokens, ( $stackPtr - 1 ), null, true );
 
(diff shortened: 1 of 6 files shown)


GaryJones (approved):

WordPress/Helpers/ContextHelper.php
```diff
@@ -129,6 +129,10 @@ final class ContextHelper {
 	 */
 	public static function has_object_operator_before( File $phpcsFile, $stackPtr ) {
 		$tokens = $phpcsFile->getTokens();
+		if ( isset( $tokens[ $stackPtr ] ) === false ) {
```
dingo-d, line 132: This was written as `isset === false` on purpose? Because 

```suggestion
		if ( !isset( $tokens[ $stackPtr ] ) ) {
```

would yield the same result, no?

jrfnl: Yes, would yield the same result, I'm just used to PHPCS, where comparisons must be explicit and strict, which `!` is not.
I've gotten to like these explicit comparisons as it make the code much easier to scan as the `!` is easily overlooked.

dingo-d (approved):

PR #2271, opened 2023-06-29
ContextHelper::$safe_casts: make `private`

jrfnl:
Follow up on #2232.

This property was previously not made `private` as it is used by the `EscapeOutput` sniff. For consistency with other classes and to better protect the value of the property, I'm proposing to make it `private` now anyway and add a `get_safe_cast_tokens()` method to retrieve the list.

This prevents potential problems if external standards would attempt to adjust the list (which they could while the property was `public static`, even though the class is `final`).


GaryJones (approved):
✅

dingo-d (approved):

PR #2273, opened 2023-06-29
Helpers: make sure function names are checked case-insensitively

jrfnl:
### ArrayWalkingFunctionsHelper: make sure function names are checked case-insensitively

These functions should be self-contained, so should not presume that the sniff has already lowercased the function name before passing it.

This fixes a bug as, in this case, the sniffs didn't actually lowercase the name before passing it to the Helper class methods, so the sniffs would throw false positives for non-lowercase function calls.

Tested by adjusting some pre-existing tests for the `EscapeOutput` sniff and the `ValidatedSanitizedInput` sniff.

### FormattingFunctionsHelper: make sure function names are checked case-insensitively

These functions should be self-contained, so should not presume that the sniff has already lowercased the function name before passing it.

This fixes a bug as, in this case, the sniffs didn't actually lowercase the name before passing it to the Helper class methods, so the sniffs would throw false positives for non-lowercase function calls.

Tested by adjusting some pre-existing tests for the `PreparedSQL` sniff and the `EscapeOutput` sniff.

### UnslashingFunctionsHelper: make sure function names are checked case-insensitively

These functions should be self-contained, so should not presume that the sniff has already lowercased the function name before passing it.

This fixes a bug as, in this case, the sniff didn't actually lowercase the name before passing it to the Helper class methods, so the sniff would throw false positives for non-lowercase function calls.

Tested by adjusting some pre-existing tests for the `ValidatedSanitizedInput` sniff.

### EscapingFunctionsTrait: make sure function names are checked case-insensitively

These functions should be self-contained, so should not presume that the sniff has already lowercased the function name before passing it.

This fixes a bug as, in this case, the sniff didn't actually lowercase the name before passing it to the trait methods, so the sniff would throw false positives for non-lowercase function calls.

Tested by adjusting some pre-existing tests for the `EscapeOutput` sniff.

### PrintingFunctionsTrait: make sure function names are checked case-insensitively

These functions should be self-contained, so should not presume that the sniff has already lowercased the function name before passing it.

This fixes a bug as, in this case, the sniff didn't actually lowercase the name before passing it to the trait method, so the sniff would throw false negatives for non-lowercase function calls.

Tested by adjusting some pre-existing tests for the `EscapeOutput` sniff.

### SanitizingFunctionsTrait: make sure function names are checked case-insensitively

These functions should be self-contained, so should not presume that the sniff has already lowercased the function name before passing it.

This fixes a bug as, in this case, the sniffs didn't actually lowercase the name before passing it to the Helper class methods, so the sniffs would throw false positives for non-lowercase function calls.

Tested by adjusting some pre-existing tests for the `ValidatedSanitizedInput` sniff.

### DIFF
--- a/WordPress/Helpers/ArrayWalkingFunctionsHelper.php
+++ b/WordPress/Helpers/ArrayWalkingFunctionsHelper.php
@@ -71,7 +71,7 @@ final class ArrayWalkingFunctionsHelper {
 	 * @return bool
 	 */
 	public static function is_array_walking_function( $functionName ) {
-		return isset( self::$arrayWalkingFunctions[ $functionName ] );
+		return isset( self::$arrayWalkingFunctions[ strtolower( $functionName ) ] );
 	}
 
 	/**
@@ -93,7 +93,7 @@ final class ArrayWalkingFunctionsHelper {
 			return false;
 		}
 
-		$functionName = $tokens[ $stackPtr ]['content'];
+		$functionName = strtolower( $tokens[ $stackPtr ]['content'] );
 		if ( isset( self::$arrayWalkingFunctions[ $functionName ] ) === false ) {
 			return false;
 		}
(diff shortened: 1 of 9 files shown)


GaryJones (approved):
✅

dingo-d (approved):

PR #2278, opened 2023-07-03
WP/DiscouragedConstants: use the `ConstantsHelper` class

jrfnl:
Cricky... not sure why none of us ever noticed this, but we had a huge chunk of duplicate code here, which can be removed by letting the sniff use the utility method.

Includes one minor tweak to the utility method.

Includes additional test which would previously have failed.


GaryJones (approved):
✅

dingo-d (approved):

PR #2280, opened 2023-07-04
Docs: clarify the `$matched_content` parameter

jrfnl:
When the `$matched_content` param is passed from the `AbstractFunctionRestrictionsSniff` or, by extension, from the `AbstractFunctionParameterSniff``, it will always already be in lowercase.

When the `$matched_content` param is passed from the `AbstractClassRestrictionsSniff` sniff it will be in its original case.

This commit updates the docs of all involved methods to include information about the case of the content of `$matched_content` to clarify this.

Includes removing two unnecessary calls to `strtolower()` in child classes.

Includes overloading the `process_matched_token()` method in the `AbstractClassRestrictionsSniff`, not because the code does anything different, but only to ensure the docblock annotates correctly that in child classes of that abstract the `$matched_content` will be in its original case.

### DIFF
--- a/WordPress/AbstractClassRestrictionsSniff.php
+++ b/WordPress/AbstractClassRestrictionsSniff.php
@@ -192,6 +192,26 @@ abstract class AbstractClassRestrictionsSniff extends AbstractFunctionRestrictio
 	}
 
 	/**
+	 * Process a matched token.
+	 *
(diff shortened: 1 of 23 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2281, opened 2023-07-04
Docs: use correct format for `@internal` tags

jrfnl:
... which have two meanings:
* `@internal` stand-alone without braces means that a construct is intended for internal use only.
* `{@internal ....}` with a comment means this is a comment for other devs.


dingo-d (approved):

GaryJones (approved):
✅

PR #2282, opened 2023-07-04
Docs: remove redundant comment

jrfnl:
Follow up on #2187 which removed the need to reset the `minimum_wp_version` property in test case files to a specific WP version, in favour of allowing a reset to "unset".

This means the version number in the `WP/AlternativeFunctionsUnitTest.inc` test case file doesn't need to be updated anymore when the default value of the `minimum_wp_version` changes, so this comment is now just confusing.


dingo-d (approved):

GaryJones (approved):
✅

PR #2283, opened 2023-07-04
WP/AlternativeFunctionsSniff::is_local_data_stream(): update param name and docs

jrfnl:
PR #2169 switched the sniff from using the `'raw'` parameter value to using the `'clean'` (without comments) parameter value, however, the `is_local_data_stream()` method still referred to the `'raw'` value, even though it is now using the `'clean'` value.

Fixed now to prevent future confusion.


dingo-d (approved):

GaryJones (approved):
✅

PR #2284, opened 2023-07-04
IsUnitTestTrait: make sure namespace and class names are checked case-insensitively

jrfnl:
As PHP treats namespace names and class names case-insensitively, this helper should do so too.

This is a bug fix as the sniff would previously throw false positives for test classes where the case of the class name/extended class name did not match.

Tested by adding an extra test to the `WordPress.Files.FilesName` sniff.


dingo-d (approved):

GaryJones (approved):
✅

Issue #2164, opened 2022-12-20
Files/FileName: check file extensions and stricter file name checking ?


jrfnl:
Originally posted by @GaryJones in https://github.com/WordPress/WordPress-Coding-Standards/pull/2154#issuecomment-1355407781:

> I can see there are some test files with extra `.` in (and not just underscores), but I was wondering if an incorrect file name like `class.my-class.php` might be detected, as the checking for hyphenation only replaces underscores.
>
> I also see `substr( $file_name, 0, -4 )`, which is assuming that a PHP-rendered file will always have a three letter file extension, but I wonder if someone could break that assumption with `--extensions=php3,myphp` or similar.

Response by @jrfnl :

> > I can see there are some test files with extra `.` in (and not just underscores), but I was wondering if an incorrect file name like `class.my-class.php` might be detected, as the checking for hyphenation only replaces underscores.
>
> I don't think it is, but did you test it ? If a test would proof it not to be handled, it might be good to open an issue about that.
>
> > I also see `substr( $file_name, 0, -4 )`, which is assuming that a PHP-rendered file will always have a three letter file extension, but I wonder if someone could break that assumption with `--extensions=php3,myphp` or similar.
>
> They most definitely could, but is that likely for a WP file or plugin ? I mean, for that to work on a server, either additional `.htaccess` rules or nginx rules (or similar) would need to be set up to support that.
> 
> All the same, I agree it may be good to clarify the file name rule in the handbook to be explicit about file extensions. Once that is done, this sniff would need to be updated.

jrfnl:
The most pertinent points from this issue have now been addressed in PR #2285.

The only thing left is whether or not to enforce that files use a `.php` file extension. As this is not a (handbook) rule, this would need a post on Make to decide whether that rule should be added or not.

@GaryJones Do you want to leave this issue open for that ? Or shall we close this issue as fixed and open a separate issue for the file extension check ?

GaryJones:
> shall we close this issue as fixed and open a separate issue for the file extension check ?

Let's do this.

jrfnl:
@GaryJones 👍🏻 As the file extension check was your suggestion, will you open the issue ?

GaryJones:
> As the file extension check was your suggestion, will you open the issue ?

Done: https://github.com/WordPress/WordPress-Coding-Standards/issues/2292

PR #2285, opened 2023-07-04
Files/FileName: various improvements

jrfnl:
### Files/FileName: improve handling of files using non-underscore, non-dash word separators

Addresses Gary's question in #2154:

> I can see there are some test files with extra `.` in (and not just underscores), but I was wondering if an incorrect file name like `class.my-class.php` might be detected, as the checking for hyphenation only replaces underscores.

I have tested this and while the file would already be flagged by the "class name vs file name check", the "is hyphenated" check would not flag the file.

Fixed now. Includes extra tests.

Note: this fix requires that the numbered test files we were using have to be renamed as the `.#.inc` would cause the sniff to throw errors when there shouldn't be any.

Related to issue #2164, but doesn't fix it completely.

### Files/FileName: improve handling of files using different file extensions

Addresses Gary's second question in #2154:

> I also see `substr( $file_name, 0, -4 )`, which is assuming that a PHP-rendered file will always have a three letter file extension, but I wonder if someone could break that assumption with `--extensions=php3,myphp` or similar.

While I still think we should add a rule about requiring the `.php` extension to the handbook (after a Make post), as that rule currently doesn't exist, the sniff should not assume that all files will have a three character file extension.

Fixed now. Includes extra tests.

Related to issue #2164, but doesn't fix it completely. (the "check for a `.php` file extension" is not addressed yet as this is not a handbook rule at this time)

### Files/FileName: update for changes in WP Core

In WP 6.1.0, the classes contained in the files listed in `$class_exceptions` were moved to files which comply with the file name requirements.
The original files still remain, but no longer contain a class, so we no longer need to make an exception for these in the `check_filename_has_class_prefix()` check.

However, with the stricter check for hyphenation-only word separators as introduced in the earlier commit, we do need to continue making an exception for these files for the `check_filename_is_hyphenated()` check.

For that reason, the properties remain.

Having said that, in this commit:
* I remove the check against these properties from the `check_filename_has_class_prefix()` check.
* The properties are renamed to `$*hyphenation_exceptions` to better reflect what they are used for.
* Two additional non-compliant files from WP Core are added to the list.


dingo-d (approved):

GaryJones (approved):
✅

PR #2286, opened 2023-07-04
NamingConventions/PrefixAllGlobals: allow non-prefixed declarations for pluggable functions and classes

jrfnl:
While the `PrefixAllGlobals` sniff already allowed for WP native constants which can be user declared, it did not allow for explicitly pluggable functions and classes being declared in plugins/themes.

I'm quite surprised actually that this was not reported as a bug so far.

Either way, fixed now.

In both cases, a case-insensitive name comparison is done as PHP natively treats function names and class names case-insensitively.

The lists added are based on some prelim sniffs created for issue #1803 and WP Core 6.3-beta2.

Notes:
* When generating these lists, I've explicitly excluded the `wp-includes/compat.php` file (which contains the polyfills for PHP native functionality).
* I've intentionally included the Twenty* theme related functions and classes as WPCS should also allow for being used by child themes of those themes.
* I've _excluded_ polyfills for new WP functionality from the Twenty* themes (which allow the themes to use those functions when running on older WP versions), as those functions aren't intentionally pluggable by WP itself.
* I've marked a number of functions/classes with a `// Deprecated` comment based on manual verification (and the declaration being in the `wp-includes/pluggable-deprecated.php` file). While it would, of course, be better if those would not be "plugged", it is not for this sniff to have an opinion on that and declaring these without a prefix should be allowed.
* In the `$pluggable_functions` list, I've intentionally commented two functions out as those are declared nested within another function and do not look _intentionally_ pluggable. This may need further verification, but at least this way it is documented why these are not included in the list.

Includes tests.

### DIFF
--- a/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
+++ b/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
@@ -196,6 +196,201 @@ final class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	);
 
 	/**
+	 * A list of functions declared in WP core as "Pluggable", i.e. overloadable from a plugin.
+	 *
+	 * Note: deprecated functions should still be included in this list as plugins may support older WP versions.
+	 *
+	 * @since 3.0.0.
+	 *
+	 * @var array<string, bool>
+	 */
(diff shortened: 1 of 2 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2288, opened 2023-07-04
DB/DirectDatabaseQuery: 2 fixes for case-insensitivity (PHP + SQL)

jrfnl:
### DB/DirectDatabaseQuery: SQL keywords are case-insensitive

This was already handled correctly for `ALTER` etc, but not for the new check for `TRUNCATE` (added in #2186).

Includes test.

### DB/DirectDatabaseQuery: method names are case-insensitive

... so the sniff needs to do the name comparisons in a case-insensitive manner.

Includes test.


dingo-d (approved):

GaryJones (approved):
✅

PR #2289, opened 2023-07-04
WP/PostsPerPage: various bug fixes

### LINKED ISSUE #2027: False positive "High pagination limit" when `posts_per_page` is set to a function.

jrfnl:
### WP/PostsPerPage: bug fix - non-numeric values for "posts_per_page" should be disregarded

This expands the tests for the "value capturing" fix in commit 1c969ab6146625fe24cfb7107e900e45142a5755 which was included in #2266.

Additionally, it adds a safeguard to the `WordPress.WP.PostsPerPage` sniff `callback()` method to prevent reporting an error when the value captured is not purely numeric.

This safeguard is needed as the way [text to number comparisons](https://wiki.php.net/rfc/string_to_number_comparison) are executed has changed in PHP 8.0 and that change affects this sniff.
See: https://3v4l.org/nOlYL#veol

Fixes #2027

### WP/PostsPerPage: bug fix - allow for explicitly positive numbers

This prevents false negatives when the number given is explicitly positive.

Includes tests.

### WP/PostsPerPage: allow for non-decimal nrs, PHP 7.4 numeric literals, PHP 8.1 octal literals

While it will probably be exceedingly rare for the `posts_per_page` value to be passed as anything but a decimal number (except to bypass the sniff), the sniff should still treat numbers passed using a different base correctly.

Along the same lines, decimal numbers passed using PHP 7.4 numeric literal notation should also be handled correctly.

Fixed now. Includes safeguard to only handle text string values when they are purely numeric.

Includes tests.


GaryJones (approved):
✅

dingo-d (approved):

PR #2290, opened 2023-07-04
Always use BCFile for find[Start|End]OfStatement()

jrfnl:
All methods in the `BCFile` class have PHPCSUtils native alternatives (which we already use), with the exception of the `findEndOfStatement()` method and the `findStartOfStatement()` method.

To be on the safe side, in case a polyfill is needed in those methods for a future PHP version or for support for PHPCS 4.x, let's consistently use the `BCFile` versions of these methods.


GaryJones (approved):
✅

dingo-d (approved):

PR #2293, opened 2023-07-06
Docs: add `@internal` tag to classes marked as internal API

jrfnl:
This adds _"the other"_ `@internal` tag to these classes, also marking the classes as "internal API" via the tag (on top of the notice in the docblock description).


dingo-d (approved):

GaryJones (approved):
✅

PR #2294, opened 2023-07-06
Docs: add missing `@uses` tags

jrfnl:
Follow up on #2259, which moved these `public` properties to the `SanitizingFunctionsTrait`.


GaryJones (approved):
✅

PR #2297, opened 2023-07-09
Security/NonceVerification: use PHPCSUtils, modern PHP + other bug fixes

### LINKED ISSUE #1981: NonceVerification: ignore unsetting of superglobal values
## Bug Description

When all that's done with a superglobal value is to _unset_ it, no nonceverification should be necessary.

Currently, the sniff will throw a warning for this:
```
 148 | WARNING | Processing form data without nonce verification.
```

## Minimal Code Snippet

```php
// OK.
unset( $_GET['meta-box-loader'] );

// Not okay.
do_something_with( $_GET['meta-box-loader'] );
```

## Error Code
`WordPress.Security.NonceVerification.Recommended`

## Environment

| PHP version             | any
| PHP_CodeSniffer version | `master`
| WPCS install type       | N/A
| IDE (if relevant)       | N/A

Might be a good idea to check in with the security team to confirm whether they agree with the above.

jrfnl:
### Security/NonceVerification: rename test case file

... to allow for splitting the test case file up for properly pure tests and for adding additional test case files.

### Security/NonceVerification: move test involving global scope to separate test case file

This prevents tests which involve the global scope from influencing each other (which was happening for an upcoming bug fix).

### Security/NonceVerification: add some extra tests involving global scope

... to raise code coverage.

### Security/NonceVerification: test with empty()

There were plenty of tests which checked the nonce verification in combination with a superglobal being used in `isset()` and `array_key_exists()`, but none in combination with `empty()`, while the sniff does handle that as well and treats it the same as `isset()`/`array_key_exists()`.

This adjusts a few pre-existing tests to make sure that the handling of `empty()` is covered by tests too.

Note: it is important that these have separate tests as `empty()` has its own token.

### Security/NonceVerification: remove redundant function call

The `NonceVerification` sniff extends the WPCS base `Sniff` class, so has access to `$this->tokens`. No need for the call to `$phpcsFile->getTokens()`.

### Security/NonceVerification: remove redundant local variable

No need to store the sub-array if it's only accessed a couple of times.

### Security/NonceVerification: changes to $addedCustomFunctions property

* As the sniff itself now only handles the nonce functions, we don't need a multi-dimensional array (the `'nonce'` subkey).
* To prevent potential conflicts with properties in traits, make the property name more specific.
* And even though the class has not been made `final`, I do think it makes sense to make this property `private`.
    The property is only intended for use by the (`protected`) `mergeFunctionLists()` method. If a child class overloads that method, they are responsible for whatever properties their overloaded method needs.

### Security/NonceVerification: bug fix - $_FILES not $_FILE

PHP does not have a `$_FILE` superglobal...
Ref: https://www.php.net/manual/en/language.variables.superglobals.php

Looks like this typo has been in this sniff from the start:
* WordPress/WordPress-Coding-Standards#325
* 1480350

A similar typo elsewhere was fixed long time ago: WordPress/WordPress-Coding-Standards#368

Adjusted some existing tests now to safeguard this.

### Security/NonceVerification: move results cache to property

No idea why this was made a function local `static` variable, but it makes more sense (and results in cleaner code) if we cache this in a property.

As the cache wasn't previously accessible from child classes anyway, I'm making this a `private` property.

### Security/NonceVerification: improve caching

As things were, the "cache" would only ever contain information about one search.

Now, while this is fine for files which have all code in the global scope òr all code within function scopes, for mixed files, this makes the cache useless as the cache would be reset based on global or function scope and results would not be re-used as intended.

In this commit, I'm changing the cache to a multi-dimensional array, which can hold information about all searches done within a file, which should improve performance for (large) mixed files.

When I then add some debug code, you can see the difference based on the new test file:
**Before**
```
Start: 0
End: 321
----
Cache re-used
----
Cache re-used
----
Start: 113
End: 150
----
Cache re-used
----
Start: 0
End: 188
----
Cache re-used
----
Start: 212
End: 249
----
Cache re-used
----
Start: 0
End: 321
----
Cache re-used
```

**After**
```
Start: 0
End: 321
----
Cache re-used
----
Cache re-used
----
Start: 113
End: 150
----
Cache re-used
----
Cache re-used
----
Cache re-used
----
Start: 212
End: 249
----
Cache re-used
----
Cache re-used
----
Cache re-used
```

### Security/NonceVerification: add cache getter and setter methods

Move the logic for getting and setting the cache to separate methods.

Includes removing the cache "priming", which isn't actually necessary as the cache will be set, no matter what, before the function returns.

Even though the class isn't `final`, as the cache was previously not accessible from child classes anyway, I'm making these methods `private`.

### Security/NonceVerification: bug fix - ignore OO properties

Properties with the same name as a superglobal should be ignored as these are not the actual superglobal.

Note: I'm not adding a similar check for function parameters as a function parameter with the same name as a superglobal with cause a fatal error since PHP 5.4, so IMO checking for that is not relevant.

Includes tests.

### Security/NonceVerification: bug fix - disregard assignments via list()

Sometimes I hate my brain....

Okay, so, while hopefully not common, assignments to superglobals can also happen via list constructs. As this sniff (correctly) disregards all assignments as safe, we should also disregard list assignments.

This commit makes it so.

Includes tests safeguarding the fix, which would previously lead to false positives.

### Security/NonceVerification: new needs_nonce_check() method

The "does this superglobal use need a nonce ?" check was previously contained in the `process_token()` method and the "must the nonce check be done before or is it allowed after ?" check was done in the `has_nonce_check()` method.

With an eye on expanding the number of checks in this field, I'm moving the logic for these two checks to a new `needs_nonce_check()` method which will either return (string) `'before|after'` or `false` when no nonce check is needed.

As the class isn't `final`, I'm making this a `protected` method (instead of `private`) to allow some flexibility for child classes.

Note: I'm deliberately _not_ moving the (new) "is OO property" check to the `needs_nonce_check()` method as in that case it isn't a superglobal in the first place, so not a variable this sniff should look at.

### Security/NonceVerification: efficiency fix - check for being in a nonce verification method early

As things were, if the superglobal being examined was _in_ a call to one of the nonce verification functions, the sniff would still loop from the start of the scope (start of the function/file) to the current token before coming to that conclusion.
This is, of course, highly inefficient for long functions/large files.

I've been in two minds about whether to add this check to the `needs_nonce_check()` method or the `has_nonce_check()` method.
For optimal efficiency, the best place is at the start of the `needs_nonce_check()` method as it will prevent having to do all the other "is in type check, is in sanitization function" checks, which are redundant if the call is a nonce check itself.
And after all, if the current superglobal is _in_ a nonce check, it doesn't _need_ a nonce check as it already has one.

To allow for caching the found nonce check, the gathering of the basic information needed for the caching has to be moved to the `process_token()` method though, so both the `needs_nonce_check()` method, as well as the `has_nonce_check()` method now get an extra `$cache_keys` parameter.

It also means that the lists of custom and known nonce verification functions have to be merged earlier, so the call to `$this->mergeFunctionLists()` has been moved up.

### Security/NonceVerification: bug fix - allow unsetting of superglobal key without nonce verification

If all that's being done is to unset a value, there is no risk, so let the sniff stay silent.

Includes tests.

Note: I've checked in with the security team to confirm this is the correct behaviour.

Fixes #1981

### Security/NonceVerification: add test with PHP 7.0+ null coalesce operator

This is already being handled correctly as the WPCS `VariableHelper::is_comparison()` method (optionally) treats null coalesce the same as other comparison operators.

This behaviour is now safeguarded with a test.

### Security/NonceVerification: support PHP 7.4+ null coalesce equals

Assignments to a superglobal using the PHP 7.4 null coalesce equals operator should be treated the same as if the variable was used in an `isset()` or with a `??` (PHP 7.0 null coalesce) operator. I.e. as long as there is a nonce check _before_ or _after_, it's fine.

As things were, null coalesce equals was treated the same as an assignment (which does not need nonce verification), even though the value of the superglobal key may not be overwritten due to the use of null coalescence.

To fix this, I'm adding a new `$include_coalesce` parameter to the `VariableHelper::is_assignment()` method to allow the method to include/exclude the null coalesce equal operator based on the value passed to that parameter.
The default value of the parameter is `true` which maintains the pre-existing behaviour of the function.

This mirrors the similar method signature and behaviour in the `VariableHelper::is_comparison()` method.

The `NonceVerificationSniff::needs_nonce_check()` method will now call that method twice, once to exclude variable assignments (with the exception of null coalesce assignments) and once to allow null coalesce equals assignments to set the expectation for the nonce verification to "after".

This may seem inefficient (and is in a way), but the footprint of the `VariableHelper::is_assignment()` method is small and the duplicate code we'd need to handle this differently not worth the effort.

Includes tests safeguarding the change.

### Security/NonceVerification: disregard PHP 7.4 arrow function when determining search scope

As arrow functions are "open", we should not limit a search for a nonce verification to the scope of the arrow function.

Includes tests documenting the behaviour.

### Security/NonceVerification: disregard nonce check in PHP 7.4 arrow function when searching

As an arrow function is a callback and we don't trace whether a callback is called, any nonce check found within an arrow function should be disregarded.
This mirrors the sniff behaviour with regards to closures, meaning nonce checks in nested arrow functions and in nested closures will now be treated the same.

Includes test.

### Security/NonceVerification: add test with PHP 8.0+ match construct

This just adds a test documenting the behaviour in combination with `match` constructs.

The use of `$_POST` in the `default` branch should realistically be flagged as the nonce check will not have been executed (as it is executed conditionally, only for `'value'`.

However, this is a structural flaw in the sniff as a whole, as it doesn't check at all whether a nonce check is done conditionally or not.

Example:
```php
function test() {
	if ( $foo ) {
		wp_verify_nonce( $_POST['nonce'] );
	} else {
		echo $_POST['key']; // Should be flagged as nonce check will definitely not have run, but isn't.
	}

	echo $_POST['key']; // Should potentially be flagged, depending on value of $foo, but isn't.
}
```

If would be very hard to fix this as a lot of conditions involve variables and we don't have access to runtime values, so it would, in most cases, be very hard to determine whether a conditionally run nonce check will have run.

With that in mind, I think allowing the sniff to treat `match` the same as other conditional constructs are currently treated is the correct behaviour for now.

If/when at some point in the future someone would attempt to make the conditionally run nonce check more precise, `match` should be handled the same as all other control structures.

### Security/NonceVerification: add test with PHP 8.1+ nested enum

Just like other nested constructs, the search for a nonce check should limit itself to the current scope and skip over nested closed scopes/not look outside the current scope.

This is already handled correctly for enums due to the sniff using the PHPCSUtils `Collections::closedScopes()` method.

This just adds tests to safeguard this.

### Security/NonceVerification: various documentation tweaks


WordPress/Sniffs/Security/NonceVerificationSniff.php
```diff
@@ -66,193 +87,229 @@ class NonceVerificationSniff extends Sniff {
...
-					break;
-				}
-			}
+		if ( Context::inUnset( $this->phpcsFile, $stackPtr ) ) {
+			// Variable is only being unset, no nonce check needed.
+			return false;
+		}
+
+		if ( VariableHelper::is_assignment( $this->phpcsFile, $stackPtr, false ) ) {
+			// Overwritting the value of a superglobal.
```
GaryJones, line 228:
```suggestion
			// Overwriting the value of a superglobal.
```

WordPress/Sniffs/Security/NonceVerificationSniff.php
```diff
@@ -66,193 +87,229 @@ class NonceVerificationSniff extends Sniff {
...
-				return ( true === $allow_nonce_after || $last['nonce_check'] < $stackPtr );
-			} elseif ( $end <= $last['end'] ) {
-				// If not, we can still go ahead and return false if we've already
-				// checked to the end of the search area.
-				return false;
-			}
+		// Check against the cache.
+		$current_cache = $this->get_cache( $cache_keys['file'], $start );
+		if ( false !== $current_cache['nonce'] ) {
+			// If we have already found an nonce check in this scope, we just
```
GaryJones, line 277:
```suggestion
			// If we have already found a nonce check in this scope, we just
```

WordPress/Tests/Security/NonceVerificationUnitTest.1.inc
```diff
@@ -391,3 +381,108 @@ function disallow_for_non_array_comparison_in_condition_with_named_params() {
...
+	$callback = fn() => $_POST['key']++; // Bad.
+}
+
+function test_disregard_noncecheck_in_nested_arrow_function() {
+	$callback = fn() => check_admin_referer( 'foo' );
+	echo $_POST['foo']; // Bad, we don't know if the callback has been called or not.
+}
+
+function test_match() {
+	$var = match($_POST['key']) { // OK. comparison with nonce check after.
```
GaryJones, line 457:
```suggestion
	$var = match($_POST['key']) { // OK, it's a comparison, with the nonce check after.
```

GaryJones (approved):
✅ (with some typos to fix and merge into that last commit.)

jrfnl:
> with some typos to fix and merge into that last commit

Done.

dingo-d (approved):

PR #2300, opened 2023-07-10
NamingConventions/ValidVariableName: bring back `$catID`

jrfnl:
Looks like between my last WP Core scan and now, `$catID` is back... _sigh_

Essentially reverts #2277, except for the list ordering.


GaryJones (approved):
✅

dingo-d (approved):

PR #2301, opened 2023-07-10
Arrays/ArrayDeclarationSpacing: partially replace the sniff + implement PHPCSUtils in what remains

jrfnl:
### Arrays/ArrayDeclarationSpacing: implement PHPCSUtils [2] / PHP 7.2+ keyed lists

Since PHP 7.2, lists can also contain keys and the double arrow in those will tokenize as `T_DOUBLE_ARROW`, same as for arrays.
An array value can contain a list assignment, so we need to take this into account.

The PHPCSUtils `Arrays::getDoubleArrowPtr()` method does so correctly since PHPCSUtils 1.0.7, so upping the minimum supported PHPCSUtils version in combination with switching to the PHPCSUtils method for retrieving the double arrow will fix this.

Includes removing the "initial check" for a double arrow as that only serves to cause the array to be walked twice when an arrow is found, so it's more efficient to not do this initial check.

Includes tests safeguarding the issue with lists.

👉 Reviewing with _ignore whitespace changes_ recommended.

### Arrays/ArrayDeclarationSpacing: implement PHPCSUtils [3]

Simplify the "associative array should be multi-line" fixer.

The `process_single_line_array()` method would previously loop through the array items and place each array item on a new line.

However, the `process_multi_line_array()` method contains logic to do the same (and better).

This simplifies the `process_single_line_array()` method to only change the single-line array to a multi-line array. This will then trigger the `process_multi_line_array()` method in the next fixer loop, which will do the rest.

Note: there is one small difference in how the file is fixed: comments will no longer be treated as _trailing_ comments for the _previous_ item, but will be treated as belonging with the current item.

In my opinion, this change is acceptable as in all honesty, there is no way for the sniff to know whether the comment belongs with the previous item or the current item, so keeping it with the current item should be fine.

Includes implementing use of the PHPCSUtils `SpacesFixer` to make the actual fix.

### Arrays/ArrayDeclarationSpacing: replace whitespace checks with PHPCSExtra sniff

This commit removes the brace related spacing checks from the `WordPress.Arrays.ArrayDeclarationSpacing` sniff in favour of using the PHPCSExtra `NormalizedArrays.Arrays.ArrayBraceSpacing` sniff.

Includes removing tests which were only testing the brace spacing related issue.
Includes minor updates to the `fixed` file for the remaining tests to account for a fix which will no longer be made by this sniff.

Note: in effect, the combined sniffs will do the same as before, though a few more notices will be thrown.
This is due to the original set-up not throwing errors for "space after opener" and "space before closer", when a "single line array with associative keys must be multi-line" error would be thrown, as the fix would be the same.
These "space after opener" and "space before closer" errors will now show together with the "single line must be multi-line" error and the fixers of the combined sniffs will sort it out between them.

Refs:
* PHPCSStandards/PHPCSExtra#12
* https://github.com/PHPCSStandards/PHPCSExtra/pulls?q=is%3Apr+arraybracespacing+is%3Aclosed

### Arrays/ArrayDeclarationSpacing:process_multi_line_array(): remove redundant parameter

The previous change made passing the array `$closer` token to the `process_multi_line_array()` method redundant, so let's remove the parameter.

### Arrays/ArrayDeclarationSpacing:process_single_line_array(): lower nesting level

As the checks for whitespace around the array opener/closer have now been removed, the code in the `process_single_line_array()` method no longer needs to be nested in condition upon condition.

This commit applies the typical "bow out early" pattern to the code in the method.

👉 This change will be easier to review while ignoring whitespace changes.

### Arrays/ArrayDeclarationSpacing: minor test tweaks


GaryJones (approved):
✅ (I appreciate the notes about reviewing with whitespace changes ignored too!)

dingo-d (approved):

PR #2312, opened 2023-07-20
Rulesets: various minor tweaks

jrfnl:
Double-checked the ruleset against the handbook. Everything currently in the handbook is now in the Core ruleset. :white_checkmark:

* Removed duplicate rule (the `Universal.UseStatements.LowercaseFunctionConst` was previously added here, but subsequently added to the "Use import statement" section as well).
* Removed links to closed issues when the issue has been addressed in the mean time. _(in the Core ruleset, the link is only supposed to be there for things which haven't got a sniff yet)_
* Made sure that rules which haven't got a sniff associated with them, while they are potentially sniffable, have an issue attached to them.
* Made sure that rules which are "covered" by one or more sniffs have the `Covers` prefix.


GaryJones (approved):
✅

dingo-d (approved):

PR #2313, opened 2023-07-21
Docs: various minor tweaks

jrfnl:
... mostly based on things I came across while creating the changelog and when updating the "lists".

(and yes, I have verified the `OperatorSpacing` sniff against the PHPCS native sniff again)

### DIFF
--- a/WordPress/Sniffs/DB/SlowDBQuerySniff.php
+++ b/WordPress/Sniffs/DB/SlowDBQuerySniff.php
@@ -46,14 +46,13 @@ final class SlowDBQuerySniff extends AbstractArrayAssignmentRestrictionsSniff {
 
 	/**
 	 * Callback to process each confirmed key, to check value.
-	 * This must be extended to add the logic to check assignment value.
 	 *
 	 * @param  string $key   Array index / key.
 	 * @param  mixed  $val   Assigned value.
 	 * @param  int    $line  Token line.
 	 * @param  array  $group Group definition.
-	 * @return mixed         FALSE if no match, TRUE if matches, STRING if matches
-	 *                       with custom error message passed to ->process().
+	 *
+	 * @return bool Always returns TRUE as the value is irrelevant.
 	 */
 	public function callback( $key, $val, $line, $group ) {
 		return true;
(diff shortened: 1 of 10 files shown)


WordPress/Sniffs/DB/SlowDBQuerySniff.php
```diff
@@ -46,14 +46,13 @@ public function getGroups() {
...
-	 * This must be extended to add the logic to check assignment value.
 	 *
 	 * @param  string $key   Array index / key.
 	 * @param  mixed  $val   Assigned value.
 	 * @param  int    $line  Token line.
 	 * @param  array  $group Group definition.
-	 * @return mixed         FALSE if no match, TRUE if matches, STRING if matches
-	 *                       with custom error message passed to ->process().
+	 *
+	 * @return bool FALSE if no match, TRUE if matches.
```
GaryJones, line 55: That description doesn't seem correct for this concrete class method - it only ever returns true.

jrfnl: What do you men ? Yes, the method only returns `true`, but the comment for the `@return` now explains what that means.

GaryJones: For this class, this method can never be a false though. The description makes it seem like it could.

jrfnl: I'm still not sure - the current implementation only ever returns `true`, but that implementation may change.

Got a suggestion for what you'd like to see for the description ?

jrfnl: I should probably say: for me, it's not so much about the current implementation as it is that the `callback()` method name + description does not describe what that `true` means, so without a comment like the proposed one, you'd always have to refer back to the parent class to understand what's happening.

dingo-d: But if the meaning is given in the parent class, then it's ok to just say something along the lines:

```php
@return bool TRUE because it will always match the value of the confirmed key.
```

Because as it stands now, it can never be `false`.

jrfnl: Updated now to `@return bool Always returns TRUE as the value is irrelevant`. Hope that satisfies your concerns.

GaryJones (approved):
✅

dingo-d (approved):

PR #2321, opened 2023-07-23
Update the `minimum_wp_version` to WP 6.0

jrfnl:
The minimum version should be three versions behind the latest WP release, so what with 6.3 being in RC, to be released on August 8th, it should now become 6.0.

Includes updating the tests to match.

Previous: #2121


GaryJones (approved):
✅

dingo-d (approved):

PR #2324, opened 2023-07-26
Core/Extra: tweak the inclusion of the `Modernize.FunctionCalls.Dirname` sniff

jrfnl:
Follow up on #2137

As  WP Core has now dropped support for PHP < 7.0, it can start using the `dirname()` `$levels` parameters. Think:
```php
// PHP < 7.0.
$path = dirname( dirname( dirname( __DIR__ ) ) );

// PHP 7.0+.
$path = dirname( __DIR__, 3 );
```

The `Modernize.FunctionCalls.Dirname` sniff we include also includes a check (and fixer) for that, so we can now include that sniff completely in the `Core` ruleset.

For now, I'm proposing to silence the error code related to the PHP 7.0 modernization opportunity (again) for the `Extra` ruleset as not all plugins/themes will have dropped support for PHP < 7.0 yet.

To be on the safe side, I'm explicitly excluding the whole sniff from the ruleset used for WPCS itself as WPCS still has a PHP 5.4 minimum.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -50,6 +50,9 @@
 
 		<!-- Linting is done in a separate CI job, no need to duplicate it. -->
 		<exclude name="Generic.PHP.Syntax"/>
+
+		<!-- WPCS still has a PHP 5.4 minimum. -->
+		<exclude name="Modernize.FunctionCalls.Dirname"/>
 	</rule>
 
 	<!-- Check code for cross-version PHP compatibility. -->
--- a/WordPress-Core/ruleset.xml
+++ b/WordPress-Core/ruleset.xml
@@ -923,8 +923,9 @@
 	<!-- Check that class name references use the correct case. -->
 	<rule ref="WordPress.WP.ClassNameCase"/>
 
-	<!-- Check that __DIR__ is favoured over dirname(__FILE__).
+	<!-- Check that __DIR__ is favoured over dirname(__FILE__)
+		 and that dirname( __DIR__, $levels ) is favoured over nested calls to dirname().
 		 See: https://core.trac.wordpress.org/ticket/48082 -->
-	<rule ref="Modernize.FunctionCalls.Dirname.FileConstant"/>
+	<rule ref="Modernize.FunctionCalls.Dirname"/>
 
 </ruleset>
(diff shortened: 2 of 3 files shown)


dingo-d (approved):

GaryJones (approved):
✅ 

Are there other sniffs where a `minimum_supported_php_version` attribute could be beneficial to configuring them?

jrfnl:
> ✅
> 
> Are there other sniffs where a `minimum_supported_php_version` attribute could be beneficial to configuring them?

Well, sort of.... but it's not that clear-cut...

PHPCS natively supports passing a [`php_version` as a configuration option](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Configuration-Options#setting-the-php-version), but that's a single patch version option, like `70305` (= PHP `7.3.5`), not a minimum/maximum option and the value cannot be changed using properties.

PHPCS natively has a few sniffs which look at that config variable and if it is not available, PHPCS presumes the PHP version on which the sniff is currently being run.

PHPCSExtra also has a few sniffs which look at that config variable, but if it is not available, PHPCSExtra will **not** presume the PHP version on which the sniff is currently being run, but err on the side of caution (as the option is too little known, so that would too easily lead to problems).

The `Modernize.FunctionCalls.Dirname` doesn't actually look at the `php_version` config variable (though maybe it should).
There are three other PHPCSExtra sniffs which do take the `php_version` config variable into account and will adapt their behaviour based on the found value if available.

The [sniff documentation](https://github.com/phpcsstandards/phpcsextra#sniffs) in PHPCSExtra details exactly which sniffs and what the behavioural difference is when the `php_version` is set (or not).

Does that help ?

GaryJones:
Yes, thank you. I don't think I was aware of the `php_version` option.

jrfnl:
@GaryJones Few people are, which is why I'm so careful about not drawing conclusions based on the absence of the setting for PHPCSExtra sniff (and why I've keep pointing to the option in the sniff docs for PHPCSExtra).

jrfnl:
FYI: I've updated the Modernize/Dirname sniff in PHPCSExtra now to also respect the `php_version`, if available: https://github.com/PHPCSStandards/PHPCSExtra/pull/261 (not yet released, but will be in the next version)

Issue #2189, opened 2023-01-09
WordPress 5.9.0 Restored `sanitize_url()` as a Proper Method for Sanitizing URLs


timnolte:
## Bug Description

When using the `sanitize_url` it is not recognized as being a proper sanitization method, and recommends `esc_url_raw()`.

## Minimal Code Snippet

```php
$postid = ! empty( sanitize_url( $_SERVER['REQUEST_URI'] ) ) ? url_to_postid( sanitize_url( $_SERVER['REQUEST_URI'] ) ) : $context['post']->ID;
```

The issue happens when running this command:
```bash
phpcs ...
```

... over a file containing this code:
```php
$postid = ! empty( sanitize_url( $_SERVER['REQUEST_URI'] ) ) ? url_to_postid( sanitize_url( $_SERVER['REQUEST_URI'] ) ) : $context['post']->ID;
```

## Environment

| Question               | Answer
| ------------------------| -------
| PHP version             | 8.0.24
| PHP_CodeSniffer version | 3.7.1
| WPCS version            | 2.3.0
| WPCS install type       | Composer project local
| IDE (if relevant)       | Neovim

## Additional Context (optional)

https://developer.wordpress.org/reference/functions/sanitize_url/#changelog

## Tested Against `develop` branch?
- [x] I have verified the issue still exists in the `develop` branch of WPCS.

Ipstenu:
This is addressed in https://github.com/WordPress/WordPress-Coding-Standards/issues/2031

timnolte:
This issue is not resolve. I am running the latest release 2.3.0 and I'm stilling receiving an error:
```
 265 | ERROR | sanitize_url() has been deprecated since WordPress
     |       | (WordPress.WP.DeprecatedFunctions.sanitize_urlFound)
```

timnolte:
Looks like this is only in the `develop` release. When are we going to get an official release published so that I can be using something stable?

jrfnl:
@timnolte When it's ready...

timnolte:
@jrfnl is there a list of what's outstanding and/or slated to be completed for the next release?

jrchamp:
#2031 and the associated PR #2121 only removed the deprecation warning. `sanitize_url` is not in the list of sanitizing functions in `develop`:
https://github.com/WordPress/WordPress-Coding-Standards/blob/f4d4ff58ddfe1ca79575e60c552543966a5ed7b2/WordPress/Sniff.php#L197-L200

PR #2325, opened 2023-07-27
Update various function lists

### LINKED ISSUE #1765: Add wp_kses_one_attr() to $escapingFunctions
## Bug Description
`wp_kses_one_attr()` should satisfy escaping rules.

## Minimal Code Snippet
```php
<?php $target = 'target="_blank"'; ?>
<a href="#" <?php echo wp_kses_one_attr( $target, 'a' ); ?>>Link</a>
```

## Error Code
WordPress.Security.EscapeOutput.OutputNotEscaped should be cleared, but it's not.

## Environment

| PHP version             | 7.2.1
| WPCS version            | develop
| WPCS install type       | git clone
| IDE (if relevant)       | PhpStorm 2019.1

- [y] I have verified the issue still exists in the `develop` branch of WPCS.

### LINKED ISSUE #1904: Add wp_timezone_choice() to list of autoescaped functions.
## Description

Add `wp_timezone_choice()` to the list of functions that are already escaped.

Docs: https://developer.wordpress.org/reference/functions/wp_timezone_choice/

## Minimal Code Snippet

```php
<select id="event_timezone_string" name="event_timezone_string">
	<?php echo wp_timezone_choice( $timezone, get_user_locale() ); ?>
</select>
```

### LINKED ISSUE #1937: WP 5.5: allow for esc_xml()
WP 5.5 will introduce a new `esc_xml()` escaping function.

At a later point in time (not in WP 5.5), it is expected that the I18n variants `esc_xml__()` and `esc_xml_e()` will be added.

The WPCS sniffs which examine output for being escaped should be updated to allow for escaping using `esc_xml()`.

Note: as this function won't be available until WP 5.5, we may need to take the minimum supported WP version into account when sniffing this for the time being.

Ref: https://make.wordpress.org/core/2020/07/21/new-esc_xml-function-in-wordpress-5-5/

jrfnl:
### EscapingFunctionsTrait: add esc_xml() to the list of escaping functions

This function was introduced in WP 5.5.

As it's been around for about three years now, I'm not going to special case it with a version check.

Ref: https://core.trac.wordpress.org/ticket/50117

Fixes #1937

### EscapingFunctionsTrait: add wp_timezone_choice() to the list of auto-escaped functions

This function was introduced in WP 2.9.0.

I've looked the code of the function over and agree it can be added to the auto-escaped functions list.

Ref: https://developer.wordpress.org/reference/functions/wp_timezone_choice/

Fixes #1904

### EscapingFunctionsTrait: add wp_readonly() to the list of auto-escaped functions

This function was added in WP 5.9 to replace the `readonly()` function as `readonly` became a reserved keyword in PHP 8.1 (though only partially to allow the WP community some time to catch up).

I'm leaving `readonly()` in the list. While it shouldn't be used anymore, if it is, it still is a valid auto-escaped function.

Ref: https://core.trac.wordpress.org/ticket/53858

### SanitizingFunctionsTrait: add sanitize_url() to the list of sanitization functions

This function was previously deprecated, but has been resurrected in WP 6.1.

Refs:
* https://core.trac.wordpress.org/ticket/55852
* WordPress/WordPress-Coding-Standards#2031
* WordPress/WordPress-Coding-Standards#2189

### [SanitizingFunctions|EscapingFunctions]Trait: add wp_kses_one_attr() to function lists

This function was introduced in WP 4.2.3.

I've done a cursory look-over of the code, but I have not gone down the rabbit-hole of examining all functions involved (and yes, `wp_kses_one_attr()` calls a _lot_ of functions under the hood).

Having said that, @anastis did do a more detailed analysis and included their write-up of it in the ticket, so I'm satisfied that this function can be added to the list.

Based on this write-up, I believe the function should be added to both the `$sanitizingFunctions` list as well as the `$escapingFunctions` list.

Refs:
* https://developer.wordpress.org/reference/functions/wp_kses_one_attr/
* WordPress/WordPress-Coding-Standards#1765#issuecomment-512192213

Fixes #1765

### DIFF
--- a/WordPress/Helpers/EscapingFunctionsTrait.php
+++ b/WordPress/Helpers/EscapingFunctionsTrait.php
@@ -71,6 +71,7 @@ trait EscapingFunctionsTrait {
 		'esc_textarea'               => true,
 		'esc_url_raw'                => true,
 		'esc_url'                    => true,
+		'esc_xml'                    => true,
 		'filter_input'               => true,
 		'filter_var'                 => true,
 		'floatval'                   => true,
--- a/WordPress/Helpers/EscapingFunctionsTrait.php
+++ b/WordPress/Helpers/EscapingFunctionsTrait.php
@@ -162,6 +162,7 @@ trait EscapingFunctionsTrait {
 		'wp_login_form'             => true,
 		'wp_loginout'               => true,
 		'wp_nav_menu'               => true,
+		'wp_readonly'               => true,
 		'wp_register'               => true,
 		'wp_tag_cloud'              => true,
 		'wp_timezone_choice'        => true,
--- a/WordPress/Helpers/EscapingFunctionsTrait.php
+++ b/WordPress/Helpers/EscapingFunctionsTrait.php
@@ -164,6 +164,7 @@ trait EscapingFunctionsTrait {
 		'wp_nav_menu'               => true,
 		'wp_register'               => true,
 		'wp_tag_cloud'              => true,
+		'wp_timezone_choice'        => true,
 		'wp_title'                  => true,
 	);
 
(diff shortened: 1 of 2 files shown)


dingo-d (approved):

GaryJones (approved):
✅

PR #2329, opened 2023-08-11
Add release checklist

jrfnl:
While not an officially supported file, in my experience it is helpful to have a release checklist available. This makes the release process more transparent and less dependent on specific maintainers.

The checklist documents the release steps and can be updated via PRs to add additional steps/adjust existing steps whenever needed.

The checklist should be used as the PR description for the release PR from `develop` to `main`.

### DIFF
--- a/.github/release-checklist.md
+++ b/.github/release-checklist.md
@@ -1,0 +1,63 @@
+# Template to use for release PRs from `develop` to `main`
+
+:warning: **DO NOT MERGE (YET)** :warning:
+
+**Please **do** add approvals if you agree as otherwise we won't be able to release.**
+
+PR for tracking changes for the x.x.x release. Target release date: **DOW MONTH DAY YEAR**.
+
+## Release checklist
+
+### General
+
+- [ ] Verify, and if necessary, update the allowed version ranges for various dependencies in the `composer.json` - PR #xxx
+- [ ] PHPCS: check if there have been releases since the last WordPressCS release and check through the changelog to see if there is anything WordPressCS could take advantage of - PR #xxx
+- [ ] PHPCSUtils: check if there have been releases since the last WordPressCS release and update WordPressCS code to take advantage of any new utilities - PR #xxx
+- [ ] PHPCSExtra: check if there have been releases since the last WordPressCS release and check through the changelog to see if there is anything WordPressCS could take advantage of - PR #xxx
+- [ ] Check if the minimum WP version property needs updating in `MinimumWPVersionTrait::$default_minimum_wp_version` and if so, action it - PR #xxx
+- [ ] Check if any of the list based sniffs need updating and if so, action it.
+    List based sniffs:
+    - [ ] `WordPress.WP.ClassNameCase` - PR #xxx
+    - [ ] `WordPress.WP.DeprecatedClasses` - PR #xxx
+    - [ ] `WordPress.WP.DeprecatedFunctions` - PR #xxx
+    - [ ] `WordPress.WP.DeprecatedParameters` - PR #xxx
+    - [ ] `WordPress.WP.DeprecatedParameterValues` - PR #xxx
+- [ ] Check if any of the other lists containing information about WP Core need updating and if so, action it.
+    - [ ] `$allowed_core_constants` in `WordPress.NamingConventions.PrefixAllGlobals` - PR #xxx
+    - [ ] `$pluggable_functions` in `WordPress.NamingConventions.PrefixAllGlobals` - PR #xxx
+    - [ ] `$pluggable_classes` in `WordPress.NamingConventions.PrefixAllGlobals` - PR #xxx
+    - [ ] `$target_functions` in `WordPress.Security.PluginMenuSlug` - PR #xxx
+    - [ ] `$reserved_names` in `WordPress.NamingConventions.ValidPostTypeSlug` - PR #xxx
+    - [ ] `$wp_time_constants` in `WordPress.WP.CronInterval` - PR #xxx
+    - [ ] `$known_test_classes` in `IsUnitTestTrait` - PR #xxx
+    - [ ] ...etc...
+
+### Release prep
+
+- [ ] Add changelog for the release - PR #xxx
+    :pencil2: Remember to add a release link at the bottom!
+- [ ] Update `README` (if applicable) - PR #xxx.
+- [ ] Update wiki (new customizable properties etc) (if applicable)
+
+### Release
+
+- [ ] Merge this PR
+- [ ] Make sure all CI builds are green.
+- [ ] Tag and create a release against `main` (careful, GH defaults to `develop`!) & copy & paste the changelog to it.
+    :pencil2: Check if anything from the link collection at the bottom of the changelog needs to be copied in!
+- [ ] Make sure all CI builds are green.
+- [ ] Close the milestone
+- [ ] Open a new milestone for the next release
+- [ ] If any open PRs/issues which were milestoned for this release did not make it into the release, update their milestone.
+- [ ] Fast-forward `develop` to be equal to `main`
+
+### After release
+
+- [ ] Open a Trac ticket for WordPress Core to update.
+
+### Publicize
+
+- [ ] [Major releases only] Publish post about the release on Make WordPress
+- [ ] Tweet, toot, etc about the release.
+- [ ] Post about it in Slack.
+- [ ] Submit for "Month in WordPress": https://make.wordpress.org/community/month-in-wordpress-submissions/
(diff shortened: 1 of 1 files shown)


.github/release-checklist.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Release checklist
+
+### General
+
+- [ ] Verify, and if necessary, update the allowed version ranges for various dependencies in the `composer.json` - PR #xxx
+- [ ] PHPCS: check if there have been releases since the last WordPressCS release and check through the changelog to see if there is anything WordPressCS could take advantage of - PR #xxx
+- [ ] PHPCSUtils: check if there have been releases since the last WordPressCS release and update WordPressCS code to take advantage of any new utilities - PR #xxx
+- [ ] PHPCSExtra: check if there have been releases since the last WordPressCS release and check through the changelog to see if there is anything WordPressCS could take advantage of - PR #xxx
+- [ ] Check if the minimum WP version property needs updating in `MinimumWPVersionTrait::$default_minimum_wp_version` and if so, action it - PR #xxx
```
GaryJones, line 17: We should note that we default to three versions behind current.

.github/release-checklist.md
```diff
@@ -0,0 +1,63 @@
...
+**Please **do** add approvals if you agree as otherwise we won't be able to release.**
+
+PR for tracking changes for the x.x.x release. Target release date: **DOW MONTH DAY YEAR**.
+
+## Release checklist
+
+### General
+
+- [ ] Verify, and if necessary, update the allowed version ranges for various dependencies in the `composer.json` - PR #xxx
+- [ ] PHPCS: check if there have been releases since the last WordPressCS release and check through the changelog to see if there is anything WordPressCS could take advantage of - PR #xxx
```
GaryJones, line 14: Can you include a link to the relevant releases pages for this and the next two bullets?

.github/release-checklist.md
```diff
@@ -0,0 +1,63 @@
...
+- [ ] PHPCSUtils: check if there have been releases since the last WordPressCS release and update WordPressCS code to take advantage of any new utilities - PR #xxx
+- [ ] PHPCSExtra: check if there have been releases since the last WordPressCS release and check through the changelog to see if there is anything WordPressCS could take advantage of - PR #xxx
+- [ ] Check if the minimum WP version property needs updating in `MinimumWPVersionTrait::$default_minimum_wp_version` and if so, action it - PR #xxx
+- [ ] Check if any of the list based sniffs need updating and if so, action it.
+    List based sniffs:
```
GaryJones, line 19: Can these include links to the relevant files?

.github/release-checklist.md
```diff
@@ -0,0 +1,63 @@
...
+    - [ ] `$wp_time_constants` in `WordPress.WP.CronInterval` - PR #xxx
+    - [ ] `$known_test_classes` in `IsUnitTestTrait` - PR #xxx
+    - [ ] ...etc...
+
+### Release prep
+
+- [ ] Add changelog for the release - PR #xxx
+    :pencil2: Remember to add a release link at the bottom!
+- [ ] Update `README` (if applicable) - PR #xxx.
+- [ ] Update wiki (new customizable properties etc) (if applicable)
```
GaryJones, line 40:
```suggestion
- [ ] Update wiki (new customizable properties etc.) (if applicable)
```

GaryJones (changes requested):
There's also an inconsistent use of a full stop on items.

.github/release-checklist.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Release
+
+- [ ] Merge this PR
+- [ ] Make sure all CI builds are green.
+- [ ] Tag and create a release against `main` (careful, GH defaults to `develop`!) & copy & paste the changelog to it.
+    :pencil2: Check if anything from the link collection at the bottom of the changelog needs to be copied in!
+- [ ] Make sure all CI builds are green.
```
dingo-d, line 48: This is intentional? There is the same check on line 45

dingo-d (commented):
Just one comment, also agree with Gary, we should either add full stops at the end of every list item, or remove them, but not mix.

.github/release-checklist.md
jrfnl: That is documented in the property docblock in the trait. Do we really need to duplicate that information here ?

.github/release-checklist.md
jrfnl: Yes, CI builds run for the merge to `main` and then separate CI builds run for the tag. Builds for both events should be green, though it would be very exceptional if the outcome of these builds wouldn't match each other.

Still, for a release, I rather be sure.

.github/release-checklist.md
jrfnl: They could, but I'm not sure of the relevance ?

By the time the release PR is pulled, these updates should already have been made and the most relevant links would be the PR links confirming the update(s) have been made and the checklist asks for those PR links (which can be replaced with _N/A_ if no update was necessary).

What am I missing ?

.github/release-checklist.md
dingo-d: Ok, that makes sense. I thought that, since `main` merge would be ok, then the tag one should be as well. I mean it should be for all the checks, but could fail for external issues that should be checked for I guess 👍🏻

jrfnl: Yes, and the builds which run for a push event (merge) don't have to be the same as the builds which run for a tag event, so better be sure.
(Example: in PHPCSUtils, a tag also triggers an update & deploy of the website, while a merge/push does not)

.github/release-checklist.md
GaryJones: > the most relevant links would be the PR links confirming the update(s) have been made

... _if_ updates were needed. I see this release checklist as something to create earlier on in the getting-ready-for-a-release period, when the PRs may not be done yet, and so a quick link to the relevant files would make sense. If these "updates should already have been made", because you have other ways of remembering to do them, then there's less benefit in including them in the checklist at all.

Happy to be disagreed with :-)

.github/release-checklist.md
```diff
@@ -0,0 +1,70 @@
...
+- [ ] Update wiki (new customizable properties etc.) (if applicable)
+
+### Release
+
+- [ ] Merge this PR.
+- [ ] Make sure all CI builds are green.
+- [ ] Tag and create a release against `main` (careful, GH defaults to `develop`!) & copy & paste the changelog to it.
+    :pencil2: Check if anything from the link collection at the bottom of the changelog needs to be copied in!
+- [ ] Make sure all CI builds are green.
+- [ ] Close the milestone..
```
GaryJones, line 50:
```suggestion
- [ ] Close the milestone.
```

GaryJones (approved):
One extra full stop snuck in, but otherwise ✅

.github/release-checklist.md
jrfnl: Well, the checklist - in my mind - serves more as a reminder to double-check that everything is ready for a release.

Like for PHPCSUtils and PHPCSExtra (and Requests for that matter), where I've been using these type of checklists for a while, and have been doing rapid releases (release quickly & often), there's always something in the list when I'm getting ready to release which I'm happy to be reminded off, even though most things are done & dusted.

I can add the links, but even when they are there, I would personally never use them as the kind of updates needed to the files listed here, are not the kind of updates I'd ever make using the web-interface, if only as it's never just one file which needs updating - let's not forget that the tests also will need updates -, so what purpose do the links then serve in the list ? (other than give the misleading impression to newcomers that those would be simple updates which _can_ be made through the web-interface ?)

For me, the purpose of the list is to have the reminders and make sure we don't forget any of the many tasks in getting the release out. And on the off-chance that we would occasionally skip a task, that doing so has to be a deliberate choice, which is then annotated in the checklist or via a comment in the PR.

And yes, also happy to be disagreed with ;-). Discussion makes things better.

.github/release-checklist.md
jrfnl: Sneaky, sneaky full stop! I'll give it a stern talking too!

.github/release-checklist.md
```diff
@@ -0,0 +1,70 @@
...
+    - [ ] `$reserved_names` in `WordPress.NamingConventions.ValidPostTypeSlug` - PR #xxx
+    - [ ] `$wp_time_constants` in `WordPress.WP.CronInterval` - PR #xxx
+    - [ ] `$known_test_classes` in `IsUnitTestTrait` - PR #xxx
+    - [ ] ...etc...
+
+### Release prep
+
+- [ ] Add changelog for the release - PR #xxx
+    :pencil2: Remember to add a release link at the bottom!
+- [ ] Update `README` (if applicable) - PR #xxx.
```
dingo-d, line 40:
```suggestion
- [ ] Update `README` (if applicable) - PR #xxx
```

dingo-d (changes requested):
One tiny change

PR #2331, opened 2023-08-11
Docs: remove redundant `@package` tags

jrfnl:
`@package` tags are an arcane manner to group related files as belonging to one project.

For projects using namespaces, the current recommendation is to only have `@package` tags when they have supplemental information to the namespace.

That is only in a very limited way the case in WPCS, so I'm proposing to remove the `@package` tags from the WPCS class docblocks, though leaving them for now in the file docblocks.

At the very least, this removed duplicate information for which there is no reason for the duplication.

Includes cleaning up (normalizing) the tag description alignments in the class docblocks.

:point_right: reviewing with whitespace changes ignored should make it easier to see that the only real change is the removal of the package tags.

Refs:
* https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/package.html#package
* https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#59-package

### DIFF
--- a/WordPress/AbstractArrayAssignmentRestrictionsSniff.php
+++ b/WordPress/AbstractArrayAssignmentRestrictionsSniff.php
@@ -20,13 +20,11 @@ use WordPressCS\WordPress\Sniff;
 /**
  * Restricts array assignment of certain keys.
  *
- * @package WPCS\WordPressCodingStandards
- *
- * @since   0.3.0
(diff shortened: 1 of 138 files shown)


dingo-d (approved):

GaryJones (approved):
✅

jrfnl:
Rebased to get passed the merge conflict.

PR #1736, opened 2019-06-20
Docs/WordPress.WhiteSpace.ControlStructureSpacing

ckanitz:
Adds documentation for the WordPress.WhiteSpace.ControlStructureSpacing

Related to #1722

PR #2334, opened 2023-08-15
WhiteSpace/ControlStructureSpacing: add documentation

jrfnl:
Follow up on closed PR #1736, but now with correct descriptions and standard blocks and code samples for all things being flagged by the sniff.

Related to #1722

Prop to @ckanitz for the initial draft.

### DIFF
--- a/WordPress/Docs/WhiteSpace/ControlStructureSpacingStandard.xml
+++ b/WordPress/Docs/WhiteSpace/ControlStructureSpacingStandard.xml
@@ -1,0 +1,150 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Control Structure Spacing"
+    >
+    <standard>
+    <![CDATA[
+    Put one space on both sides of the opening and closing parentheses of control structures.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: One space on each side of the open and close parentheses.">
+        <![CDATA[
+while<em> ( </em>have_posts()<em> ) </em>{}
+
+// For multi-line conditions,
+// a new line is also accepted.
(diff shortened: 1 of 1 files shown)


WordPress/Docs/WhiteSpace/ControlStructureSpacingStandard.xml
```diff
@@ -0,0 +1,150 @@
...
+
+
+</em>}
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    [Optional, turned off by default]
+	There should be no blank line(s) at the start or end of the control structure body.
```
GaryJones, line 127: Is the indentation on this line incorrect?

GaryJones (approved):
One small query.

WordPress/Docs/WhiteSpace/ControlStructureSpacingStandard.xml
jrfnl: Oh, thanks for catching that! I wonder why that wasn't caught by xmllint... Hmm.. we may need to look into that.

jrfnl: Just checked - it's not checked because it would mean we'd need to add (and continuously maintain) every individual docs file in CI to allow xmllint to check this: https://github.com/WordPress/WordPress-Coding-Standards/blob/2d7f396fcc9fa726710d973b6311e26ae854f596/.github/workflows/basic-qa.yml#L79-L85

Also, we want tabs indents for rulesets, while these docs files need space indents.

We may want to look at moving the docs checks to a separate job and see about doing some bash magic or have some other type of script which loops through the docs files to xmllint them (without us having to manually maintain the list).

jrfnl: I've opened an issue as a reminder: https://github.com/WordPress/WordPress-Coding-Standards/issues/2337

GaryJones: I'm surprised we haven't got an `.editorconfig` file!

PR #2336, opened 2023-08-15
Docs: update the sample ruleset

jrfnl:
* Replace an example of a rule exclusion as excluding security sniffs should be encouraged.
* Remove an example of an "extra rule" as that sniff is now included in WPCS.
* Minor tweak to the example of how to include PHPCompatibilityWP. Note: I've raised the minimum PHP version to 5.6, not yet to 7.0 as plugins/themes (which are the typical target public for this example ruleset) often support multiple WP versions.
* Update property name.
* Give a few examples of legitimate selective directory excludes.

### DIFF
--- a/phpcs.xml.dist.sample
+++ b/phpcs.xml.dist.sample
@@ -52,13 +52,13 @@
 		Once we know the sniff names, we can opt to exclude sniffs which don't
 		suit our project like so.
 
-		The below two examples just show how you can exclude rules.
+		The below two examples just show how you can exclude rules/error codes.
 		They are not intended as advice about which sniffs to exclude.
 		-->
 
 		<!--
 		<exclude name="WordPress.WhiteSpace.ControlStructureSpacing"/>
-		<exclude name="WordPress.Security.EscapeOutput"/>
+		<exclude name="Modernize.FunctionCalls.Dirname.Nested"/>
 		-->
 	</rule>
 
@@ -66,7 +66,6 @@
 	<rule ref="WordPress-Docs"/>
 
 	<!-- Add in some extra rules from other standards. -->
-	<rule ref="Generic.CodeAnalysis.UnusedFunctionParameter"/>
 	<rule ref="Generic.Commenting.Todo"/>
 
 	<!-- Check for PHP cross-version compatibility. -->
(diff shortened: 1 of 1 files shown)


phpcs.xml.dist.sample
```diff
@@ -118,4 +119,35 @@
...
+	<!--
+	Sometimes, you may want to exclude a certain directory, like your tests,
+	for select sniffs.
+	The below examples demonstrate how to do this.
+
+	In the example, the `GlobalVariablesOverride` rule is excluded for test files
+	as it is sometimes necessary to overwrite WP globals in test situations (just
+	don't forget to restore them after the test!).
+
+	Along the same lines, PHPUnit is getting stricter about using PSR4 file names,
```
GaryJones, line 139:
```suggestion
	Along the same lines, PHPUnit is getting stricter about using PSR-4 file names,
```

GaryJones (approved):
Now that the encouragement of having tests is called out, how about an example of defining a custom test case as well?

jrfnl:
> Now that the encouragement of having tests is called out, how about an example of defining a custom test case as well?

What do you mean ?

dingo-d (approved):

dingo-d:
I'm holding on merging this one because I'm curious about this custom test case @GaryJones mentioned 😅

jrfnl:
> > > Now that the encouragement of having tests is called out, how about an example of defining a custom test case as well?
> > 
> > 
> > What do you mean ?
> 
> I was thinking of adding an example of https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#custom-unit-test-classes to the sample file. Not a hard push.

Ah, now I see what you mean. The custom unit test classes option is basically a work-around for the better way of doing things which I'm showing in the example. WPCS is unlikely to be 100% correct when using that property, while excluding the known test directories is a surefire way of preventing those false positives.

This is also noted as such in the docblock of the property:

https://github.com/WordPress/WordPress-Coding-Standards/blob/9518e8e934d2149aea01f317d3d6163f3b65c79f/WordPress/Helpers/IsUnitTestTrait.php#L56-L58

It is also mentioned at the bottom of the wiki section about the property: https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#custom-unit-test-classes

Having both the property & the exclusion in the ruleset would be kind of confusing as the directives are intended to do the same thing, just in a different way.

PR #2338, opened 2023-08-17
Code coverage: initial setup for CodeCov

jrfnl:
Follow up after PR #2225 and other PRs which added more tests and raised code coverage.

The Codecov service is a way to monitor test vs code coverage of a project over time and allows for the code coverage % + delta to be reported in each PR.

This commit:
* Adds CodeCov configuration and ignores it for creating the package file.
    * Note: by default CodeCov reports back with build statuses (in the table at the bottom of a PR) + posts a really extensive comment on a PR.
        I personally find those comments + mail on each PR terribly annoying and unnecessary when the build statuses are set to required statuses anyway, so I've turned the posting of those comments off (via a setting in the configuration file).
    * Another thing to note is the patch - threshold setting being at `3%` (= 3% deviation allowed). The reason for setting that threshold a little wider is that code removals for fully covered code would otherwise quickly fail builds.
* Adds a badge showing off code coverage to the README.
* Adjusts the GH Actions workflows.
    * For the `test` workflow (PRs), the tests will selectively be run with code coverage (high/low PHP x high/low PHPCS).
    * For the `quicktest` workflow (pushes), which only runs high/low PHP/PHPCS, the code coverage will only be run for (merges to) the `develop` branch.
        This will allow us keep track of code coverage over time, while not slowing down the "quick test" for feature branch pushes.
        Note: I've simplified the conditions a little for this workflow as we only run against two known PHP versions anyway.

Note: As this is a public repo, we shouldn't need a repo secret for the CodeCov token. The normal GHA token will be used (and should work fine).

Refs:
* https://docs.codecov.com/docs
* https://docs.codecov.com/docs/codecovyml-reference

Note: as this PR can only partially be tested without it being merged, minor touch-up may be needed after this initial setup, but 🤞🏻  we should be good.

### DIFF
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php: [ '5.4', 'latest' ]
+        php: [ '5.4', '7.4', 'latest' ]
         phpcs_version: [ 'lowest', 'dev-master' ]
 
     name: QTest - PHP ${{ matrix.php }} on PHPCS ${{ matrix.phpcs_version }}
@@ -48,7 +48,7 @@ jobs:
         with:
           php-version: ${{ matrix.php }}
           ini-values: ${{ steps.set_ini.outputs.PHP_INI }}
-          coverage: none
+          coverage:  ${{ github.ref_name == 'develop' && 'xdebug' || 'none' }}
 
       - name: "Set PHPCS version (master)"
         if: ${{ matrix.phpcs_version != 'lowest' }}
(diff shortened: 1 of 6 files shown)


.github/workflows/quicktest.yml
```diff
@@ -76,10 +76,23 @@ jobs:
...
-      - name: Run the unit tests - PHP 5.4 - 8.0
-        if: ${{ matrix.php < '8.1' && matrix.php != 'latest' }}
+      - name: Run the unit tests without code coverage - PHP 5.4 - 8.0
+        if: ${{ matrix.php == '5.4' && github.ref_name != 'develop' }}
         run: composer run-tests
 
-      - name: Run the unit tests - PHP >= 8.1
-        if: ${{ matrix.php >= '8.1' || matrix.php == 'latest' }}
+      # Until PHPCS supports PHPUnit 9, we cannot run code coverage on PHP 8.0+, so run it on PHP 5.4 and 7.4.
+      - name: Run the unit tests with code coverage - PHP 5.4 - 8.0
```
GaryJones, line 84:
```suggestion
      - name: Run the unit tests with code coverage - PHP 5.4 - 7.4
```

GaryJones (approved):
✅

.github/workflows/quicktest.yml
jrfnl: Ah, the `5.4 - 8.0` is not about the fact that code coverage is being run, but about the command used for PHPUnit, which works for PHP 5.4 - 8.0 ;-)

dingo-d (approved):

PR #2345, opened 2023-08-17
Docs: array properties expect arrays

jrfnl:
Follow up on #1584, which was included in WPCS 2.0.0.

While PHPCS still supports the _old_ array syntax (comma-delimited string), WPCS no longer has a (WPCS native) tolerance for those properties being passed with the `type="array"` attribute, so these public properties should always be arrays.


dingo-d (approved):

GaryJones (approved):
✅

PR #2347, opened 2023-08-17
Sniff: move "missing unslashing" handling to callback function

jrfnl:
As things were, the "missing unslashing" error was being thrown from the `Sniff::is_sanitized()` method, based on whether the `$require_unslash` parameter was set to `true` or `false`.

In my opinion, it should not be the responsibility of utility functions to throw errors/warning, but always of the originating sniff.

This commit replaces the boolean `$require_unslash` parameter for the `Sniff::is_sanitized()` method to a nullable callable `$unslash_callback` parameter.

Instead of letting the `Sniff::is_sanitized()` method throw the error/warning, the callback will now be called and as the originating sniff passes the callback, the originating sniff can now handle the throwing of the error/warning.

To remove any presumptions of the callback being passed, being in the same context as the utility method/originating sniff and having access to the `$phpcsFile` via a property, the callback will receive both the `$phpcsFile` as well as the `$stackPtr` as parameters.

The "missing unslash" check can still be skipped by not passing the parameter or by passing anything non-callable. This maintains the original behaviour for the method for the default value/`false` case.

The `Sniff::add_unslash_error()` method can with this change now be moved to the `ValidatedSanitizedInput` sniff and that sniff will now have full control over the error message and error code.

Note: the callback is called using `call_user_func()` instead of `$callable` to allow some tolerance for the _partially supported callables_, which were deprecated in PHP 8.2. While WPCS itself does not use these, the `ValidatedSanitizedInputSniff` is one of the exceptions which has not been made `final`, so an extending class _could_ use one of the _partially supported callables_ and the `Sniff::is_sanitized()` method should not be the reason that doesn't work.

### DIFF
--- a/WordPress/Sniff.php
+++ b/WordPress/Sniff.php
@@ -145,14 +145,24 @@ abstract class Sniff implements PHPCS_Sniff {
 	 * Check if something is being sanitized.
 	 *
 	 * @since 0.5.0
+	 * @since 3.0.0 The second ($require_unslash) parameter has been changed from
+	 *              a boolean toggle to a ?callable $unslash_callback parameter to
+	 *              allow a sniff calling this method to handle their "unslashing"
+	 *              related messaging itself.
 	 *
-	 * @param int  $stackPtr        The index of the token in the stack.
-	 * @param bool $require_unslash Whether to give an error if no unslashing function
-	 *                              is used on the variable before sanitization.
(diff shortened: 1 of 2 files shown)


dingo-d (approved):

GaryJones (approved):
Lovely change ✅

PR #2354, opened 2023-08-17
QA/CS: various minor tweaks

jrfnl:
### QA/CS: various minor tweaks

* Remove some undefined/unused variables.
* Remove some extraneous whitespace from arrays where (longer) entries were removed.

### QA/CS: alphabetize import use statements


GaryJones (approved):

WordPress/Sniffs/WhiteSpace/ObjectOperatorSpacingSniff.php
```diff
@@ -9,8 +9,8 @@
 
 namespace WordPressCS\WordPress\Sniffs\WhiteSpace;
 
-use PHP_CodeSniffer\Standards\Squiz\Sniffs\WhiteSpace\ObjectOperatorSpacingSniff as Squiz_ObjectOperatorSpacingSniff;
 use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Standards\Squiz\Sniffs\WhiteSpace\ObjectOperatorSpacingSniff as Squiz_ObjectOperatorSpacingSniff;
```
dingo-d, line 13: How come this alias is named with `_`? Wouldn't `SquizObjectOperatorSpacingSniff` be more consistent with the naming of the classes?

dingo-d (changes requested):
Left one question.

WordPress/Sniffs/WhiteSpace/ObjectOperatorSpacingSniff.php
jrfnl: This PR isn't changing the alias, just moving the statement.

Other than that, the alias is done in a similar way as aliases used elsewhere in WPCS, though I agree, they could use more consistency:
* `Sniff` class: `use PHP_CodeSniffer\Sniffs\Sniff as PHPCS_Sniff;`
* `ValidVariableNameSniff`: `use PHP_CodeSniffer\Sniffs\AbstractVariableSniff as PHPCS_AbstractVariableSniff;` (looks like this sniff can probably be removed)
* `ObjectOperatorSpacingSniff`: `use PHP_CodeSniffer\Standards\Squiz\Sniffs\WhiteSpace\ObjectOperatorSpacingSniff as Squiz_ObjectOperatorSpacingSniff;`
* `OperatorSpacingSniff`: `use PHP_CodeSniffer\Standards\Squiz\Sniffs\WhiteSpace\OperatorSpacingSniff as PHPCS_Squiz_OperatorSpacingSniff;`

dingo-d: I think we can change alias names later on, that shouldn't be a BC break, no?

jrfnl: Correct, aliases (just like `use` statements) are file specific, so have no BC implications.

dingo-d (approved):

Issue #2272, opened 2023-06-29
[Task] Add dedicated tests for most Helper classes and traits


jrfnl:
The "Helper" classes and traits are currently tested via the sniff (integration) tests.

It would be preferable to have dedicated test for the Helper classes and traits though.

In most cases, setting up these tests shouldn't be hard as when no token walking is needed, we can use the PHPUnit native `TestCase` as the base class, while when token walking is needed, we can use the `UtilityMethodTestCase` form PHPCSUtils as a base class.

I propose we add these tests as a separate `WPCSSniffHelpers` test suite to the PHPUnit configuration.

I propose we add these tests in a `Helpers` subdirectory of the `Tests` directory off the project root.
If a class would need multiple tests classes, a sub-subdirectory can be created using the Helper class/trait name as the directory name.

We may need to add an `autoload-dev` directive to the `composer.json` file specifically for these tests to get things working, as these tests won't run via the PHPCS native test framework.

When adding tests, Helper specific tests which were added to sniff test classes _may_ be removed if they don't have added value anymore, though in most cases, it will be better to just leave them be.

When dedicated tests have been added for a Helper, the `@covers` tag(s) for that Helper in sniff specific tests classes should be removed.

#### Helper classes

- [ ] `ArrayWalkingFunctionsHelper`
- [ ] `FormattingFunctionsHelper`
- [ ] `UnslashingFunctionsHelper`
- [ ] `ValidationHelper`
- [ ] `WPGlobalVariablesHelper`
- [ ] `WPHookHelper`

#### Helper traits
- [ ] `EscapingFunctionsTrait`
- [ ] `IsUnitTestTrait`
- [ ] `MinimumWPVersionTrait`
- [ ] `PrintingFunctionsTrait`
- [ ] `SanitizingFunctionsTrait`
- [ ] `WPDBTrait`

#### Low priority Helper classes

The following Helper classes are low(er) priority for adding dedicated tests as the methods in these are likely to move to PHPCSUtils at some point in the future:
- [ ] `ConstantsHelper`
- [ ] `ContextHelper`
- [ ] `DeprecationHelper`
- [ ] `ListHelper`
- [ ] `RulesetPropertyHelper`
- [ ] `SnakeCaseHelper`
- [ ] `VariableHelper`

PR #2356, opened 2023-08-18
Move sanitization methods to SanitizationHelperTrait + minor improvements

### LINKED ISSUE #1465: Extract re-usable code from Sniff.php into new package

jrfnl:
These methods haven't had extensive work done as they barely do any token walking themselves and the underlying functions used by the methods all have had work done to improve them.

Note: the `is_only_sanitized()` method _could_ be moved to the `NonceVerification` sniff as it's only used by that sniff, but for now, I've elected to keep it together with the `is_sanitized()` method.

---

### Rename SanitizingFunctionsTrait to SanitizationHelperTrait

The `Sniff:;is_*sanitized()` utility methods are about to be moved to this trait, but once they have been, the trait name is no longer correct as all other `*Functions[Helper|Trait]` classes/traits _only_ deal with function lists and checking whether something is in those lists, while this trait will now do more than that.

With that in mind, I propose to rename the trait to `SanitizationHelperTrait`.

### Move sanitization methods to SanitizationHelperTrait

This moves the last utility functions from the `Sniff` class to a dedicated trait.

I am also explicitly marking the trait as internal API to allow us to iterate on this (not so very clean) solution in future 3.x releases without being forced to wait for a 4.0 release.

The methods have been made stand-alone, as in, the presumption that the WPCS `Sniff` class is being extended has been removed by adding the `$phpcsFile` parameter.

Closes #1465

### SanitizationHelperTrait::is_sanitized(): add some defensive coding

As this method is now stand-alone, we'd better make sure the token being examined exists.

Note: no need to do the same for the `is_only_sanitized()` method as the first thing that method calls is this method and if this method returns `false`, the `is_only_sanitized()` method will too.

### SanitizationHelperTrait::is_sanitized(): implement PHPCSUtils

Note: The `$functionPtr` variable is overwritten a few lines later and the other variables being set are effectively unused (other than in these lines), so this code can be safely removed/replaced.

### SanitizationHelperTrait::is_sanitized(): efficiency tweak - move unset check up

A variable which is being unset, doesn't need to be unslashed or sanitized, so bow out early.

### SanitizationHelperTrait::is_sanitized(): efficiency fix/improve performance

Remove the need for a call to the performance negative `array_diff_key()` function by setting up the original arrays in a better way.

### SanitizationHelperTrait::is_sanitized(): minor code tweaks

* Get rid of an unnecessary `else`.
* Improve readability of long condition.
* Get rid of unnecessary true/false condition code.

### SanitizationHelperTrait::is_sanitized(): minor doc tweaks

### DIFF
--- a/WordPress/Helpers/SanitizationHelperTrait.php
+++ b/WordPress/Helpers/SanitizationHelperTrait.php
@@ -11,6 +11,7 @@ namespace WordPressCS\WordPress\Helpers;
 
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Utils\Context;
 use PHPCSUtils\Utils\TextStrings;
 use WordPressCS\WordPress\Helpers\ArrayWalkingFunctionsHelper;
 use WordPressCS\WordPress\Helpers\ContextHelper;
@@ -338,14 +339,8 @@ trait SanitizationHelperTrait {
 			return false;
 		}
 
-		// Get the function that it's in.
-		$nested_parenthesis = $tokens[ $stackPtr ]['nested_parenthesis'];
-		$nested_openers     = array_keys( $nested_parenthesis );
-		$function_opener    = array_pop( $nested_openers );
-		$functionPtr        = $phpcsFile->findPrevious( Tokens::$emptyTokens, ( $function_opener - 1 ), null, true, null, true );
-
 		// If it is just being unset, the value isn't used at all, so it's safe.
-		if ( \T_UNSET === $tokens[ $functionPtr ]['code'] ) {
+		if ( Context::inUnset( $phpcsFile, $stackPtr ) ) {
 			return true;
 		}
 
--- a/WordPress/Helpers/SanitizationHelperTrait.php
+++ b/WordPress/Helpers/SanitizationHelperTrait.php
@@ -9,7 +9,13 @@
 
 namespace WordPressCS\WordPress\Helpers;
 
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Utils\TextStrings;
+use WordPressCS\WordPress\Helpers\ArrayWalkingFunctionsHelper;
+use WordPressCS\WordPress\Helpers\ContextHelper;
 use WordPressCS\WordPress\Helpers\RulesetPropertyHelper;
+use WordPressCS\WordPress\Helpers\UnslashingFunctionsHelper;
 
 /**
  * Helper functions and function lists for checking whether a sanitizing function is being used.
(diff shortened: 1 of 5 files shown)


jrfnl:
Also note: I still have a niggly feeling there is a logic error in the `is_only_sanitized()` method, but I haven't been able to figure out the reason this was originally coded this way, so I'm not touching it for now.

The logic error I suspect is in the following code:
```php
		// The only parentheses should belong to the sanitizing function. If there's
		// more than one set, this isn't *only* sanitization.
		return ( \count( $tokens[ $stackPtr ]['nested_parenthesis'] ) === 1 );
```

This seems to presume that parentheses could only be for function calls, while the extra parentheses may just as well be for control structure conditions.

If the other parentheses _are_ for control structure conditions, I believe the variable should still count as "only sanitized".
It also discounts situations where unslashing is nested within a sanitization function call (which is taken into account in the `NonceVerification` sniff, but that is now inconsistent).
```php
$a = is_email( $_GET['email'] ); // is_ony_sanitized(): true
$a = is_email( wp_unslash( $_GET['email'] ) ); // is_ony_sanitized(): false
if ( is_email( $_GET['email'] ) ) {} // is_ony_sanitized(): false
```

This should probably be further investigated when adding dedicated tests for these methods (#2272).

GaryJones (approved):

jrfnl:
Repushed to fix the order of the remaining `use` import statements in the `Sniff` class.

WordPress/Helpers/SanitizationHelperTrait.php
```diff
@@ -236,4 +252,166 @@ public function is_sanitizing_function( $functionName ) {
...
+
+	/**
+	 * Check if something is only being sanitized.
+	 *
+	 * @since 0.5.0
+	 * @since 3.0.0 - Moved from the Sniff class to this class.
+	 *              - The method visibility was changed from `protected` to `public`.
+	 *              - The `$phpcsFile` parameter was added.
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcsFile The file being scanned.
```
dingo-d, line 264: This is probably something we can address in a separate PR, but since we have the import of this class at the top, I don't think we need to write the entire namespace in the docblock (unless we really want to be explicit, but we can discuss this in an issue maybe).

Just leaving a comment here not to forget about it later on 😄

dingo-d (approved):
Left one comment, mostly as a reminder.

WordPress/Helpers/SanitizationHelperTrait.php
dingo-d: Opened the issue: https://github.com/WordPress/WordPress-Coding-Standards/issues/2358

jrfnl: `use` statements have no effect on docblocks, so it's a matter of preference. Currently we are using FQNs everywhere and it saves having to scroll up in the file to check where something is coming from.

PR #2362, opened 2023-08-18
Security/ValidatedSanitizedInput: use PHPCSutils and bug fixes

### LINKED ISSUE #1720: Does $_FILES really need wp_unslash?
## Bug Description
I'm using `wp_handle_upload` in one of my plugins. PHPCS complains about the `$_FILES` slice not being run through `wp_unslash`. However, [`wp_magic_quotes` does not add slashes to `$_FILES`](https://developer.wordpress.org/reference/functions/wp_magic_quotes/). 

Running an apparently superfluous `wp_unslash` on `$_FILES` is harmless on POSIX platforms, but [breaks file uploading on Windows](https://wordpress.org/support/topic/there-was-an-error-uploading-the-avatar-specified-file-failed-upload-test/).

## Minimal Code Snippet
```php
$mimetypes = [
	'jpg|jpeg|jpe' => 'image/jpeg',
	'gif'          => 'image/gif',
	'png'          => 'image/png',
];
$args = [
	'mimes'                    => $mimetypes,
	'test_form'                => false,
];
$avatar = \wp_handle_upload( \wp_unslash( $_FILES[ 'avatar' ], $args ) ); 
```

## Error Code
`WordPress.Security.ValidatedSanitizedInput.MissingUnslash`

## Environment

| PHP version             | 7.3.6
| PHP_CodeSniffer version | 3.4.2
| WPCS version            | 2.1.1

- [ ] I have verified the issue still exists in the `develop` branch of WPCS.

jrfnl:
### ValidationHelper: bug fix - unsetting is not validation

A superglobal key being unset doesn't need sanitization or validation, however, the unset should _not_ be seen as validation for use of the superglobal key later on.

This bug was introduced long ago (0.3.0 - 64fab4f) and later moved into the `is_validated()` method via ba15db9

Discovered when making some changes to the tests for the `ValidatedSanitizedInput` sniff.

Includes dedicated test for this situation.

### ValidationHelper: bug fix - validation done in another scope

Discovered when making some changes to the tests for the `ValidatedSanitizedInput` sniff.

A number of the tests in that file are in the _global_ scope.
As things were, if validation was encountered for a particular superglobal key in a _closed_ scope (function, class) prior to the use of the superglobal key in the _global_ scope, the superglobal key would be seen as correctly validated, even though the code within the closed scope may not have been executed.

IMO tracking whether the code within the closed scope is executed is outside the scope of this sniff and a "missing validation" error should be thrown for those cases.

And while PHP 7.4+ arrow functions are _open_ scopes, validation done within an arrow function should not be counted as valid for superglobals used _outside the arrow function, though validation done outside an arrow function should IMO be okay for superglobals used _within_ an arrow function.

Fixed now.

Includes a minor tweak to the pre-existing tests for the case which made me discover the bug (superglobal use in global scope).
Includes additional tests with closed scopes within an already closed scope.
Includes additional tests documenting the behaviour around arrow functions.

### Security/ValidatedSanitizedInput: make sure all superglobals being examined are used in tests

The `$_SERVER` and `$_COOKIE` superglobals are being examined by the sniff, but there were no tests safeguarding this.

Fixed now by adjusting some pre-existing tests.

### Security/ValidatedSanitizedInput: use PHPCSUtils for superglobal determination [1]

Previously, the sniff would disregard the `$_SESSION` and `$_ENV` superglobals. This is now no longer the case when examining variables embedded in text strings.

This may have been on purpose, but I could not find any discussion about this and it seems more consistent (and more correct) to flag the unvalidated/unsanitized use of **all** superglobals.

Note: the `$GLOBALS` superglobal is exempt as in that case, any variable may be accessed (`$GLOBALS['var']`), not just superglobals (`$GLOBALS['_GET']`).

Includes tests.

### Security/ValidatedSanitizedInput: use PHPCSUtils for superglobal determination [2]

Previously, the sniff would disregard the `$_SESSION` and `$_ENV` superglobals. This will now no longer be the case when the sniff examines variable tokens.

I suspect these superglobals were originally disregarded as the values in these superglobals are not slashed by WP (see the below refs), but that doesn't mean the values shouldn't be validated and sanitized before use.

The strange thing is that `$_FILES` _was_ already being examined, even though it also isn't slashed by WP.
The sniff, however, would throw an incorrect "requires unslashing" error when the `$_FILES` superglobal was being used. Also see bug report #1720.

Either way, along the same lines of the previous commit, when examining variable tokens, the `ValidatedSanitizedInput` sniff will now take all superglobals into account, with the exception of the `$GLOBALS` superglobal (this last bit is the same as before).

As the values in the `$_FILES`, `$_SESSION` and `$_ENV` superglobals aren't slashed by WP, they get an exception for the "unslash" error.

And as the `Sniff::$input_superglobals` property is now unused, I'm removing it.

Includes tests.

Fixes #1720

Refs:
* https://developer.wordpress.org/reference/functions/wp_magic_quotes/
* https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/load.php#L1229-L1238

### Security/ValidatedSanitizedInput: add extra test for PHP 7.4+ null coalesce equals

PR #1684 (WPCS 2.1.0) already took care of the handling of validation via the null coalesce equals assignment operator, but there was one test case missing.

This adds that extra test as an additional safeguard.

### Security/ValidatedSanitizedInput: add test with PHP 8.0+ match [1]

When a superglobal is used in a match condition, it doesn't need to be unslashed/sanitized.

This is already handled correctly, this just adds a test to safeguard this.

### Security/ValidatedSanitizedInput: add tests with PHP 8.0+ match [2]

When a superglobal is used in the match return expressions, it should be validated, unslashed and sanitized.

This is already handled correctly, this just adds a test to safeguard this.

### Security/ValidatedSanitizedInput: make MissingUnslash message more informative

Previously, the error would read `$_POST data not unslashed...`. Now, the error message will show the exact variable with keys we're talking about like `$_POST['key'] not unslashed...`.

### DIFF
--- a/WordPress/Sniffs/Security/ValidatedSanitizedInputSniff.php
+++ b/WordPress/Sniffs/Security/ValidatedSanitizedInputSniff.php
@@ -11,6 +11,7 @@ namespace WordPressCS\WordPress\Sniffs\Security;
 
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Utils\Context;
 use PHPCSUtils\Utils\TextStrings;
 use WordPressCS\WordPress\Helpers\ContextHelper;
 use WordPressCS\WordPress\Helpers\SanitizationHelperTrait;
@@ -88,6 +89,11 @@ class ValidatedSanitizedInputSniff extends Sniff {
 
 		// Check if this is a superglobal.
 		if ( ! \in_array( $this->tokens[ $stackPtr ]['content'], $superglobals, true ) ) {
+			return;
+		}
+
+		// If the variable is being unset, we don't care about it.
+		if ( Context::inUnset( $this->phpcsFile, $stackPtr ) ) {
 			return;
 		}
 
--- a/WordPress/Sniffs/Security/ValidatedSanitizedInputSniff.php
+++ b/WordPress/Sniffs/Security/ValidatedSanitizedInputSniff.php
@@ -46,6 +46,23 @@ class ValidatedSanitizedInputSniff extends Sniff {
 	public $check_validation_in_scope_only = false;
 
 	/**
+	 * Superglobals for which the values will be slashed by WP.
+	 *
+	 * @link https://developer.wordpress.org/reference/functions/wp_magic_quotes/
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var array<string, true>
+	 */
+	private $slashed_superglobals = array(
+		'$_COOKIE'  => true,
+		'$_GET'     => true,
+		'$_POST'    => true,
+		'$_REQUEST' => true,
+		'$_SERVER'  => true,
+	);
+
+	/**
 	 * Returns an array of tokens this test wants to listen for.
 	 *
 	 * @return array
@@ -66,8 +83,6 @@ class ValidatedSanitizedInputSniff extends Sniff {
 	 * @return void
 	 */
 	public function process_token( $stackPtr ) {
-
-		$superglobals = $this->input_superglobals;
 
 		// Handling string interpolation.
 		if ( \T_DOUBLE_QUOTED_STRING === $this->tokens[ $stackPtr ]['code']
@@ -96,8 +111,12 @@ class ValidatedSanitizedInputSniff extends Sniff {
 			return;
 		}
 
-		// Check if this is a superglobal.
-		if ( ! \in_array( $this->tokens[ $stackPtr ]['content'], $superglobals, true ) ) {
+		/* Handle variables */
+
+		// Check if this is a superglobal we want to examine.
+		if ( '$GLOBALS' === $this->tokens[ $stackPtr ]['content']
+			|| Variables::isSuperglobalName( $this->tokens[ $stackPtr ]['content'] ) === false
+		) {
 			return;
 		}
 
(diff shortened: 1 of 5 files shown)


WordPress/Sniffs/Security/ValidatedSanitizedInputSniff.php
```diff
@@ -65,29 +84,44 @@ public function register() {
...
 
-		$superglobals = $this->input_superglobals;
-
 		// Handling string interpolation.
 		if ( \T_DOUBLE_QUOTED_STRING === $this->tokens[ $stackPtr ]['code']
 			|| \T_HEREDOC === $this->tokens[ $stackPtr ]['code']
 		) {
 			// Retrieve all embeds, but use only the initial variable name part.
 			$interpolated_variables = array_map(
 				function ( $embed ) {
```
dingo-d, line 93:
```suggestion
				static function ( $embed ) {
```

Mircooptimization.

WordPress/Sniffs/Security/ValidatedSanitizedInputSniff.php
```diff
@@ -65,29 +84,44 @@ public function register() {
...
+					return preg_replace( '`^(\{?\$\{?\(?)([a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*)(.*)$`', '$2', $embed );
 				},
 				TextStrings::getEmbeds( $this->tokens[ $stackPtr ]['content'] )
 			);
 
-			foreach ( array_intersect( $interpolated_variables, $superglobals ) as $bad_variable ) {
+			// Filter the embeds down to superglobals only.
+			$interpolated_superglobals = array_filter(
+				$interpolated_variables,
+				function ( $var_name ) {
```
dingo-d, line 102:
```suggestion
				static function ( $var_name ) {
```

Same microoptimization.

dingo-d (changes requested):
Left two minor suggestions.

jrfnl:
@dingo-d 💯 Duh... I should have done that out of habit... fixed now.

dingo-d (approved):

GaryJones (approved):
✅

PR #2364, opened 2023-08-20
Sniff: remove `init()` method

jrfnl:
As pointed out by @GaryJones, as all utility methods have now been made self-contained, it is no longer necessary to have a separate `init()` method in the `Sniff` class.


GaryJones (approved):
✅

dingo-d (approved):

PR #2366, opened 2023-08-20
Tests: Remove redundant placeholder function

### LINKED ISSUE #1915: Remove problematic words from the code

GaryJones:
It's the only error flagged by a non-inclusive terminology checker.

Originally added in https://github.com/WordPress/WordPress-Coding-Standards/commit/3cd319133d6ff20cc8957450326e4cdf8c97f9a6.

Fixes #1915.

### DIFF
--- a/WordPress/Tests/DB/DirectDatabaseQueryUnitTest.inc
+++ b/WordPress/Tests/DB/DirectDatabaseQueryUnitTest.inc
@@ -16,8 +16,8 @@ function bar() {
 	return $listofthings;
 }
 
-function dummy() {
-}
+
+
 
 function baz() {
 	global $wpdb;


WordPress/Tests/DB/DirectDatabaseQueryUnitTest.inc
```diff
@@ -16,8 +16,8 @@ function bar() {
 	return $listofthings;
 }
 
-function dummy() {
```
jrfnl, line 19: This may be removing a test case ? Would keeping the code, but renaming the function be an option ? (just to be on the safe side).

GaryJones: I did originally just rename it, but I couldn't see what case the original was meant to be. Based on the [initial commit](https://github.com/WordPress/WordPress-Coding-Standards/commit/3cd319133d6ff20cc8957450326e4cdf8c97f9a6), and then the [follow up commit](https://github.com/WordPress/WordPress-Coding-Standards/commit/aa57302aa71b8afd1aa2de4e6ec51498df858315), I'm not sure how an empty function was a test case for a direct DB query-related sniff.

jrfnl: Yes, I was struggling to see this as well, but I haven't got the time atm to dive into the code to be sure (which is why I suggested the "on the safe side" change).

If you are confident it is not testing anything, I'm happy for this to be merged as-is.

jrfnl (approved):

dingo-d (approved):

PR #2370, opened 2023-08-21
Escape output sniff fix for static method calls

dingo-d:
As it currently is the following code

```php
echo ClassName::someMethodName( $attributes, $test );
```

Would trigger the EscapeOutput sniff for the class name, but also for any attributes of the class method, except the first one.

The sniff should just trigger the class name, and bow out after that (what is inside the method is of no concern, as the output is all that matters).

I've added a test case and found the part that caused the issue (the `*::class` part didn't take into account static methods, only the `*::class` case).

This should fix it.


jrfnl:
> The sniff should just trigger the class name, and bow out after that (what is inside the method is of no concern, as the output is all that matters).

Is that true though ? 

Shouldn't the sniff skip over the method call, but then continue ?

Test case:
```php
echo My::method($param), $somethingElse; // I'd expect two errors here, one for the method call, one for $somethingElse.
```

dingo-d:
Knew there are must be edge cases I missed 😅 

I'll work on this some more in the following days, I need to get some sleep 🙈

jrfnl:
@dingo-d Should this PR be moved back to draft for the time being ?

dingo-d:
@jrfnl The tests seem to be passing now, only the 2 seem to be stalling (or just taking an unusually long time).

dingo-d:
Closing this PR as this is obviously not going to be moving along.

Issue #2375, opened 2023-09-03
WordPressCS 3.0: Fatal error: Cannot declare class WordPressCS\WordPress\Sniffs\Files\FileNameSniff


dawidurbanski:
## Minimal Code Snippet

I created a repository with [minimal reproduction](https://github.com/dawidurbanski/phpcs3.0-custom-rule-repro).

You can test this using `develop` branch using [use-develop](https://github.com/dawidurbanski/phpcs3.0-custom-rule-repro#testing-using-develop-branch) branch on the repro repository.

## Error Code

```
Fatal error: Cannot declare class WordPressCS\WordPress\Sniffs\Files\FileNameSniff, because the name is already in use in /Users/xxx/Local Sites/xxx/app/public/wp-content/plugins/phpcs3.0-custom-rule-repro/vendor/wp-coding-standards/wpcs/WordPress/Sniffs/Files/FilenameSniff.php on line 36
```

## Custom Ruleset

```xml
<?xml version="1.0"?>
<ruleset name="Test Config">
	<file>.</file>
	<arg name="extensions" value="php"/>
	<rule ref="WordPress"/>
	<rule ref="WordPress-Extra"/>
	<rule ref="WordPress-Docs"/>

	<rule ref="WordPress.Files.Filename">
		<exclude-pattern>/src/*\.php</exclude-pattern>
	</rule>
</ruleset>
```

## Environment

| Question                 | Answer
| ------------------------ | -------
| PHP version              | PHP 8.1.9 (cli) (built: Aug 30 2022 10:24:25) (NTS)
| PHP_CodeSniffer version  | PHP_CodeSniffer version 3.7.2 (stable) by Squiz (http://www.squiz.net)
| WordPressCS version      | 3.0.0
| PHPCSUtils version       | 1.0.8
| PHPCSExtra version       | 1.1.1
| WordPressCS install type | Composer project local
| IDE (if relevant)        | Not relevant

## Tested Against `develop` Branch?
- [x] I have verified the issue still exists in the `develop` branch of WordPressCS.

jrfnl:
@dawidurbanski I have cloned that repo and tried with both the `main` branch as well as the `use-develop` branch and cannot reproduce the error.

I do see some other things which aren't as they should be in your repo, but that still shouldn't cause this issue.

As a side-note, now WordPressCS 3.0 has been released, why would you still want to use `dev-develop` ?


[...]

PR #2376, opened 2023-09-03
Docs: Use `FileName` rule instead of `Filename` in the sample phpcs ruleset

dawidurbanski:
Currently file [phpcs.xml.dist.sample](https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/phpcs.xml.dist.sample#L149) has a case sensitive typo.

It is:

```xml
	<rule ref="WordPress.Files.Filename">
		<exclude-pattern>/path/to/Tests/*Test\.php</exclude-pattern>
	</rule>
```

Should be:

```xml
	<rule ref="WordPress.Files.FileName">
		<exclude-pattern>/path/to/Tests/*Test\.php</exclude-pattern>
	</rule>
```

I stumble across this when having [this issue](https://github.com/WordPress/WordPress-Coding-Standards/issues/2375).

### DIFF
--- a/phpcs.xml.dist.sample
+++ b/phpcs.xml.dist.sample
@@ -146,7 +146,7 @@
 	<rule ref="WordPress.WP.GlobalVariablesOverride">
 		<exclude-pattern>/path/to/Tests/*Test\.php</exclude-pattern>
 	</rule>
-	<rule ref="WordPress.Files.Filename">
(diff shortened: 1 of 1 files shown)


GaryJones (approved):
✅

dingo-d (approved):

Issue #2374, opened 2023-09-01
WordPress.Security.EscapeOutput encourages bad practice wrt exceptions


anomiex:
## Bug Description

Escaping should normally be done close to the point of output (or at least the point where HTML is being generated). The `WordPress.Security.EscapeOutput` sniff instead encourages escaping at the point where exceptions are generated, without knowing if the exception is going to be caught, or output to a CLI, or logged to a text-based log, etc.

[Apparently](https://github.com/WordPress/WordPress-Coding-Standards/issues/884) this is because someone was worried about people having `display_errors` on (which is PHP's default) and `html_errors` off (which is **not** the default). The sensible thing would have been to have WordPress use `set_exception_handler()` to set a default exception handler early in the request, which could easily ensure any output of the exception message to the browser is escaped even on such misconfigured sites and which would also catch errors generated by PHP itself and errors generated by other libraries that don't use WPCS.

I note a similar argument could be made for WPCS's insisting on `trigger_error()` having escaping rather than having WordPress use `set_error_handler()`. Although in that case `html_errors` being on doesn't prevent the problem.

## Minimal Code Snippet

The issue happens when running this command:
```bash
phpcs --standard=WordPress-Extra
```

... over a file containing this code:
```php
throw new Exception( "Text is $text" );
```

## Error Code

WordPress.Security.EscapeOutput.OutputNotEscaped

## Custom Ruleset

Use `--standard=WordPress-Extra`

## Environment

| Question                 | Answer
| ------------------------ | -------
| PHP version              | 8.2.7
| PHP_CodeSniffer version  | 3.7.2
| WordPressCS version      | 3.0.0
| PHPCSUtils version       | 1.0.8
| PHPCSExtra version       | 1.1.1
| WordPressCS install type | Composer project local
| IDE (if relevant)        | None

## Tested Against `develop` Branch?
- [x] I have verified the issue still exists in the `develop` branch of WordPressCS.

jrfnl:
@anomiex I agree that would be the ideal solution, except WP does not do that. I suggest you open a Trac ticket to get WordPress fixed. I.e. implement callbacks to `set_exception_handler()` and `set_error_handler(0` which do the escaping when necessary and do so while playing nice with other handlers from plugins and themes.

Once that fix is in place, I'd be very happy to remove sniffing for both those issues which annoy you.

Until that time, however, the messages from the sniff are perfectly valid IMO and this bug report is not.

[...]

Issue #1864, opened 2020-02-07
trigger_error is counted as an output function for escaping purposes


rmccue:
## Bug Description

`trigger_error` is treated as a HTML output method, and hence requires escaping.

Errors are only output as part of the HTML when `show_errors` is set to true, which is not common in production environments. Error text is not necessarily HTML data, and doesn't require escaping.

## Minimal Code Snippet

```php
trigger_error( 'Undefined config option: ' . $key );
```

## Error Code

`WordPress.Security.EscapeOutput.OutputNotEscaped`

## Environment

| Question               | Answer
| ------------------------| -------
| PHP version             | 7.1.32
| PHP_CodeSniffer version | 3.5.4
| WPCS version            | 1.2.1
| WPCS install type       | Composer project local

## Additional Context (optional)

Problem exists in latest WPCS: https://github.com/WordPress/WordPress-Coding-Standards/blob/dd86ccddab77f4631f005893a8b7786b5311efe3/WordPress/Sniff.php#L459

Given this is environment-dependent, I think the rule makes sense by default, but a way to configure or override it would be good. Currently, it's possible to add additional functions, but not remove the built in ones in Sniff.php (AFAIK?)

## Tested Against `develop` branch?
- [x] I have verified the issue still exists in the `develop` branch of WPCS.

jrfnl:
> Errors are only output as part of the HTML when show_errors is set to true, which is not common in production environments. Error text is not necessarily HTML data

All of these are actually arguments why it *should* be escaped:
* You don't know the environment in which the code is run, so should take all possibilities into account.
* Arbitrary variables passed to `trigger_error()` may contain HTML and - most of the time - that HTML should be stripped for an error message.

> I think the rule makes sense by default, but a way to configure or override it would be good.

That's what the whitelist comments are for:
```php
// phpcs:ignore WordPress.Security.EscapeOutput -- $key is safe.
trigger_error( 'Undefined config option: ' . $key );
```

Does that help ?

rmccue:
> You don't know the environment in which the code is run, so should take all possibilities into account.

We know for certain the environment the code is being run in, so we know that `trigger_error` and `error_log` are not output functions. (We control the production environment as well as local environments.)

We also use these functions extensively for informative logging in production environments. Escaping the content of these plain-text messages would lead to double-encoding bugs.

(Actually, we also know that all the `_deprecated_...` functions won't cause output as well, FWIW.)

[...]

PR #2378, opened 2023-09-03
Security/EscapeOutputSniff: More modular error codes

anomiex:
This splits certain cases out of `OutputNotEscaped` to allow for ignoring certain cases that are looking at error strings at time of generation (where they may eventually be used in both HTML and non-HTML contexts) rather than at time of output.

* `ExceptionNotEscaped` for unescaped strings in throws (cf. #2374).
* `ErrorNotEscaped` for unescaped strings in `trigger_error` (cf. #1864).

### DIFF
--- a/WordPress/Sniffs/Security/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/Security/EscapeOutputSniff.php
@@ -344,7 +344,7 @@ class EscapeOutputSniff extends AbstractFunctionRestrictionsSniff {
 				break;
 		}
 
-		return $this->check_code_is_escaped( $start, $end, 'OutputNotEscaped' );
+		return $this->check_code_is_escaped( $start, $end );
 	}
 
 	/**
@@ -395,7 +395,7 @@ class EscapeOutputSniff extends AbstractFunctionRestrictionsSniff {
 					continue;
 				}
 
-				$this->check_code_is_escaped( $param['start'], ( $param['end'] + 1 ), 'OutputNotEscaped' );
+				$this->check_code_is_escaped( $param['start'], ( $param['end'] + 1 ) );
 			}
 
 			return $end;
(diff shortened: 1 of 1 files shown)


WordPress/Sniffs/Security/EscapeOutputSniff.php
```diff
@@ -414,7 +414,7 @@ public function process_matched_token( $stackPtr, $group_name, $matched_content
 				return $end;
 			}
 
-			return $this->check_code_is_escaped( $message_param['start'], ( $message_param['end'] + 1 ) );
+			return $this->check_code_is_escaped( $message_param['start'], ( $message_param['end'] + 1 ), 'ErrorNotEscaped' );
```
anomiex, line 417: If you don't want to change this yet, since `trigger_error` having `OutputNotEscaped` is a long-standing thing versus `throw` being new in 3.0.0,
```suggestion
			return $this->check_code_is_escaped( $message_param['start'], ( $message_param['end'] + 1 ), 'OutputNotEscaped' );
```

WordPress/Sniffs/Security/EscapeOutputSniff.php
```diff
@@ -446,12 +446,13 @@ public function process_matched_token( $stackPtr, $group_name, $matched_content
...
-	 * @param int $start The position to start checking from.
-	 * @param int $end   The position to stop the check at.
+	 * @param int    $start The position to start checking from.
+	 * @param int    $end   The position to stop the check at.
+	 * @param string $code  Code to use for the PHPCS error.
 	 *
 	 * @return int Integer stack pointer to skip forward.
 	 */
-	protected function check_code_is_escaped( $start, $end ) {
+	protected function check_code_is_escaped( $start, $end, $code ) {
```
jrfnl, line 455: General question: why not set `'OutputNotEscaped'` as the default value for the new `$code` parameter and only pass it when it is going to be different ?

anomiex: I started with that, but once I found there were some recursive calls that needed handling it seemed safer to be explicit on all the calls.

jrfnl: Hmm... I'd prefer to use a default value. If that really wouldn't work, I'd like to suggest adding class constants for the error codes and referring to those constants in the function calls to prevent having the hard-coded codes all over the place with the risk of a typo creeping in.

anomiex: Done, not worth arguing over IMO.

dingo-d (approved):
Looks good to me.

GaryJones (approved):
✅

PR #2381, opened 2023-09-07
Update README.md

alanef:
Move upgrade notice for 3.0.0 to top

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,10 @@
 
 </div>
 
+# Upgrade Notice
+
+> If you are upgrading from an older WordPressCS version to version 3.0.0, please read the [Upgrade guide for ruleset maintainers and end-users](https://github.com/WordPress/WordPress-Coding-Standards/wiki/Upgrade-Guide-to-WordPressCS-3.0.0-for-ruleset-maintainers) first!
+
 
 # WordPress Coding Standards for PHP_CodeSniffer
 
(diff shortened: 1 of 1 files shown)


GaryJones:
The readme was carefully revised and reviewed before the release, including the update references. 

While existing users may need the update information, having the section above the main title doesn't make sense for people new to the project. 

What are you trying to gain with this change?

alanef:
Just trying to help avoid the issue that I, and @wizzyto12 and @zakitekhqs fell into by not realizing that 3.0.0 would break when upgrading, to avoid creating extra support.

An alternative may be to add perhaps a 'Are you upgrading' in template=bug_report.md

alanef:
Somehow Imissed Mika's x-Post from Core to Plugins of https://make.wordpress.org/core/2023/08/21/wordpresscs-3-0-0-is-now-available/#comment-45469

So really just my 'bad'.

GaryJones:
> 3.0.0 would break when upgrading

This project, like many projects (alas, not WP core), follows Semantic Versioning, as linked to at the top of the [Changelog](https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/CHANGELOG.md). As such, a major version like a 3.0.0 is _expected_ to contain breaking changes (in fact, it's arguably THE reason for having major releases).

It's also [mentioned](https://make.wordpress.org/core/2023/08/21/wordpresscs-3-0-0-is-now-available/#upgrading) in the Make post, and the [Release Notes](https://github.com/WordPress/WordPress-Coding-Standards/releases/tag/3.0.0), and at least implied by the long list of error code and other changes, if not explicitly said "breaking", in the multiple wiki upgrade guides. 

Also, media articles like [WPTavern](https://wptavern.com/wordpress-coding-standards-maintainer-warns-maintenance-will-be-halted-without-funding-this-is-an-unsustainable-situation), [WordPress VIP](https://lobby.vip.wordpress.com/2023/09/05/vip-coding-standards-3-0-0-is-now-available/) (about VIPCS), and probably others also mention about it containing breaking changes, so however someone might find out about WordPressCS 3.0, there's already a strong chance that they know it contains breaking changes.

alanef:
Yes, I said I missed it. I'm only human.
Thanks for reminding me to double check in future after composer updates ( any library) for major version ticks.

PR #2382, opened 2023-09-10
Funding: fix format

jrfnl:
This should either be an array of links or a singular link. The `WP PHP` was regarded as an invalid link (not as a link description).

Ref: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/displaying-a-sponsor-button-in-your-repository


dingo-d (approved):

dingo-d:
This is my bad, I thought that the array syntax was something like `[url, name]`, not a list of custom links 😅

PR #2384, opened 2023-09-12
Add dependabot configuration file

jrfnl:
**Context**: I know some updates are needed to the GHA workflows due to new major releases of action runners. May as well set up  Dependabot now to get those updates in (which automatically tests the config).

----

This commit adds an initial Dependabot configuration to:
* Submit pull requests for security updates and version updates for GH Action runner dependencies.

At a later point in time, we could consider enabling it for Composer dependencies as well.

The configuration has been set up to:
* Run weekly (for now).
* Submit a maximum of 5 pull requests at a time. If additional pull requests are needed, these will subsequently be submitted the next time Dependabot runs after one or more of the open pull requests have been merged.
* The commit messages for PRs submitted by Dependabot will be prefixed according the unofficial conventions used in this repo up to now.
* The PRs will automatically be labelled with an appropriate label as already in use in this repo.

Refs:
* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,17 @@
+# Dependabot configuration.
+#
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      time: "09:00"
+    open-pull-requests-limit: 5
+    commit-message:
+      prefix: "GH Actions:"
+    labels:
+      - "Type: Chores/Cleanup"


dingo-d (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
...
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5
```
GaryJones, line 12:
```suggestion
```

The [default value](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit) is 5 already.

.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
...
+#
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
GaryJones, line 11: Let's set a time (default day is Monday, default timezone is UTC), so that we know when the PRs should have arrived if they were coming for that week.

```suggestion
      interval: "weekly"
      time: "09:00"
```

GaryJones (changes requested):

.github/dependabot.yml
jrfnl: True, except most people don't know that and the default may change without notice.
This makes it explicit and easy to check for people. Happy to remove it if you prefer, but at least you now know my reasoning for including it.

.github/dependabot.yml
jrfnl: I've added it, but I previously didn't add it on purpose. If everyone sets a time, especially at a whole hour, this causes a peak burden on the GH servers, which may have unintended side-effects. Leaving the time open, you still know that the PRs should have arrived by midnight UTC on Monday, but it allows GH/Dependabot to sort out their queue with more flexibility.

.github/dependabot.yml
GaryJones: We already rely on other defaults (like it being on a Monday), so this is no different. Having it also suggests that we're limiting it on purpose (I was going to query why we'd want to do that, really), rather than leaving it to the defaults (more effort to look up).

jrfnl: So, what should we do ?

dingo-d: I'm really fine either way. Extra information doesn't hurt  I guess

GaryJones (approved):

PR #2385, opened 2023-09-13
Add changelog for v3.0.1

dingo-d:
Adding the changelog for the v3.0.1 hotfix release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,23 @@ This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a 
 
 _No documentation available about unreleased changes as of yet._
 
+## [3.0.1] - 2023-09-13
+
+### Added
+
+- In WordPressCS 3.0.0, the functionality of the `WordPress.Security.EscapeOutput` sniff was updated to report unescaped message parameters passed to exceptions created in `throw` statements. This specific violation now has a separate error code: `ExceptionNotEscaped`. This will allow users to ignore or exclude that specific error code. Props [@anomiex].
+    The error code(s) for other escaping issues flagged by the sniff remain unchanged.
+
+### Changed
+
+- Updated the CI workflow to test the example ruleset for issues.
+- Funding files and updates in the Readme about funding the project.
+
+### Fixed
+
+- Fixed a sniff name in the `phpcs.xml.dist.sample` file (case-sensitive sniff name). Props [@dawidurbanski].
+
+
 ## [3.0.0] - 2023-08-21
 
 ### Important information about this release:
@@ -1556,6 +1573,7 @@ Initial tagged release.
 [Composer PHPCS plugin]: https://github.com/PHPCSStandards/composer-installer
 
 [Unreleased]: https://github.com/WordPress/WordPress-Coding-Standards/compare/main...HEAD
+[3.0.1]: https://github.com/WordPress/WordPress-Coding-Standards/compare/3.0.0...3.0.1
 [3.0.0]: https://github.com/WordPress/WordPress-Coding-Standards/compare/2.3.0...3.0.0
 [2.3.0]: https://github.com/WordPress/WordPress-Coding-Standards/compare/2.2.1...2.3.0
 [2.2.1]: https://github.com/WordPress/WordPress-Coding-Standards/compare/2.2.0...2.2.1
@@ -1585,8 +1603,10 @@ Initial tagged release.
 [0.3.0]: https://github.com/WordPress/WordPress-Coding-Standards/compare/2013-10-06...0.3.0
 [2013-10-06]: https://github.com/WordPress/WordPress-Coding-Standards/compare/2013-06-11...2013-10-06
 
+[@anomiex]:       https://github.com/anomiex
 [@ckanitz]:       https://github.com/ckanitz
 [@craigfrancis]:  https://github.com/craigfrancis
+[@dawidurbanski]: https://github.com/dawidurbanski
 [@desrosj]:       https://github.com/desrosj
 [@grappler]:      https://github.com/grappler
 [@Ipstenu]:       https://github.com/Ipstenu


CHANGELOG.md
```diff
@@ -8,6 +8,22 @@ This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a
 
 _No documentation available about unreleased changes as of yet._
 
+## [3.0.1] - 2023-09-13
+
+### Added
+
+- Funding files and updates in the Readme about funding the project.
```
jrfnl, line 15: This is not functionality added to this package which affects end-users, so IMO doesn't belong in "Added".

Maybe mention it in "Changed" ?

CHANGELOG.md
```diff
@@ -8,6 +8,22 @@ This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a
...
+
+### Changed
+
+- The `WordPress.Security.EscapeOutput` sniff has more modular error codes, so that developers can ignore certain output escaping cases based on where the errors occur. Props [@anomiex].
+- Updated the CI workflow to test the example ruleset for issues.
+
+### Fixed
+
+- Fixed the sniff name in the `phpcs.xml.dist.sample` file (case-sensitive sniff name). Props [@dawidurbanski].
```
jrfnl, line 24:
```suggestion
- Fixed a sniff name in the `phpcs.xml.dist.sample` file (case-sensitive sniff name). Props [@dawidurbanski].
```

Whether this belongs in "Fixed" or "Changed" is open for debate. It's not a bug in a sniff (end-user functionality), but was a "bug" in the documentation, which _if strictly copied_ could affect end-users, so I'm not sure...

CHANGELOG.md
```diff
@@ -8,6 +8,22 @@ This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a
...
 
+## [3.0.1] - 2023-09-13
+
+### Added
+
+- Funding files and updates in the Readme about funding the project.
+
+### Changed
+
+- The `WordPress.Security.EscapeOutput` sniff has more modular error codes, so that developers can ignore certain output escaping cases based on where the errors occur. Props [@anomiex].
```
jrfnl, line 19: IMO this changelog entry needs to be made more specific and possibly moved to "Added" ?

We need to make it clear that it is only the new functionality added in 3.0.0 regarding flagging parameters passed to exceptions created in a throw statement which is affected + the new error code for those needs to be mentioned. It should also be mentioned that the error code of other things being flagged by the sniff has not changed.

Think of it this way: if you read this as a user of the EscapeOutput sniff, would you know what to do/what you could do after reading this line ?

CHANGELOG.md
```diff
@@ -1597,3 +1614,5 @@ Initial tagged release.
 [@NielsdeBlaauw]: https://github.com/NielsdeBlaauw
 [@slaFFik]:       https://github.com/slaFFik
 [@sandeshjangam]: https://github.com/sandeshjangam
+[@anomiex]:       https://github.com/anomiex
+[@dawidurbanski]: https://github.com/dawidurbanski
```
jrfnl, line 1618: Please keep the list in alphabetic order.

jrfnl (changes requested):
Thanks for setting this up. Left my notes inline.

CHANGELOG.md
dingo-d: It is a change, but also a fix from a bug that kinda affected the users 🤷🏼‍♂️ Not a sniff bug, but issue was reported.

CHANGELOG.md
dingo-d: I will move it to added and explain it a bit more in details 👍🏼

CHANGELOG.md
```diff
@@ -8,6 +8,23 @@ This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a
 
 _No documentation available about unreleased changes as of yet._
 
+## [3.0.1] - 2023-09-13
+
+### Added
+
+- In the WordPressCS v3.0.0, the functionality of the `WordPress.Security.EscapeOutput` sniff was changed to flag the unescaped parameters passed to exceptions in the `throw` statements. This sniff now has more modular error codes based on whether the error happened while checking the exceptions or while outputting unescaped content. The error code has only changed for the exceptions case (now `ExceptionNotEscaped`), and not the rest of the sniff (still `OutputNotEscaped`). This will enable the users to silence the error based on the error code if they need to.
+- Props [@anomiex].
```
jrfnl, line 16:
```suggestion
- In the WordPressCS v3.0.0, the functionality of the `WordPress.Security.EscapeOutput` sniff was updated to also flag unescaped parameters passed to exceptions created in `throw` statements. This specific error now has a separate error code: `ExceptionNotEscaped`. This will allow users to ignore/exclude that specific error based on the error code if they need to. Props [@anomiex].
    The error code(s) for other escaping issues flagged by the sniff remain unchanged.
```

Reasons for my change suggestions:
* The sniff has more (unchanged) error codes, not just `OutputNotEscaped`.
* The other error codes have _not_ been made more modular, this one has just been split off.
* "ignore/exclude" references the terms used in _ignore_ annotations/ruleset _excludes_, while "silence" does not.

jrfnl (changes requested):
Little tweak to the actual description of the change.

CHANGELOG.md
```diff
@@ -8,6 +8,23 @@ This projects adheres to [Semantic Versioning](https://semver.org/) and [Keep a
 
 _No documentation available about unreleased changes as of yet._
 
+## [3.0.1] - 2023-09-13
+
+### Added
+
+- In the WordPressCS v3.0.0, the functionality of the `WordPress.Security.EscapeOutput` sniff was updated to also flag unescaped parameters passed to exceptions created in `throw` statements. This specific error now has a separate error code: `ExceptionNotEscaped`. This will allow users to ignore/exclude that specific error based on the error code if they need to. Props [@anomiex].
```
GaryJones, line 15:
```suggestion
- In WordPressCS 3.0.0, the functionality of the `WordPress.Security.EscapeOutput` sniff was updated to report unescaped message parameters passed to exceptions created in `throw` statements. This specific violation now has a separate error code: `ExceptionNotEscaped`. This will allow users to ignore or exclude that specific error code. Props [@anomiex].
```

GaryJones (changes requested):

jrfnl (approved):

dingo-d:
Added the suggestions, thanks for the review 👍🏼 🙂 

I've squashed all the commits to one and added a co-authored messages.

GaryJones (approved):
✅

PR #2387, opened 2023-09-13
Update PHPCSExtra to v1.1.1

dingo-d:


jrfnl (approved):
The PR is good, but not strictly necessary. The 1.1.1 release doesn't _add_ anything WPCS relies nor fixes bugs in sniffs WPCS uses.

GaryJones (changes requested):
> The 1.1.1 release doesn't add anything WPCS relies on nor fixes bugs in sniffs WPCS uses.

Since WordPressCS doesn't **need** 1.1.1, then we should keep the constraint wider.

Suggest Close.

PR #2391, opened 2023-09-17
Hotfix - Lowercase custom sniff properties

dingo-d:
This PR will address the issue of handling mixed case function/method/class names when those are passed as custom-allowed properties in certain sniffs.

For instance, if we pass a custom sanitization function as a property `my_custom_sanitization` then it doesn't matter if we are calling `My_custom_Sanitization` or `MY_CUSTOM_SANITIZATION` (or any combination of cases for that matter), PHP is case insensitive in that regard and the same function will be called.

So the sniff shouldn't trigger on those variations (in some cases it will do that, as shown in PR #2370 which holds the fix for the custom escaping functions).

I've added tests to ensure that these changes will be caught.

I'm not 100% sure if I've covered all the cases, so any help in pointing out if I've missed something helps.

### DIFF
--- a/WordPress/Helpers/EscapingFunctionsTrait.php
+++ b/WordPress/Helpers/EscapingFunctionsTrait.php
@@ -218,15 +218,11 @@ trait EscapingFunctionsTrait {
 		if ( array() === $this->allEscapingFunctions
 			|| $this->customEscapingFunctions !== $this->addedCustomEscapingFunctions['escape']
 		) {
-			/*
-			 * Lowercase all names, both custom as well as "known", as PHP treats namespaced names case-insensitively.
-			 */
-			$custom_escaping_functions = array_map( 'strtolower', $this->customEscapingFunctions );
-			$escaping_functions        = array_change_key_case( $this->escapingFunctions, \CASE_LOWER );
-
 			$this->allEscapingFunctions = RulesetPropertyHelper::merge_custom_array(
-				$custom_escaping_functions,
-				$escaping_functions
+				$this->customEscapingFunctions,
+				$this->escapingFunctions,
+				true,
+				true
 			);
 
 			$this->addedCustomEscapingFunctions['escape'] = $this->customEscapingFunctions;
@@ -248,15 +244,11 @@ trait EscapingFunctionsTrait {
 		if ( array() === $this->allAutoEscapedFunctions
 			|| $this->customAutoEscapedFunctions !== $this->addedCustomEscapingFunctions['autoescape']
 		) {
-			/*
-			 * Lowercase all names, both custom as well as "known", as PHP treats namespaced names case-insensitively.
-			 */
-			$custom_auto_escaped_functions = array_map( 'strtolower', $this->customAutoEscapedFunctions );
-			$auto_escaped_functions        = array_change_key_case( $this->autoEscapedFunctions, \CASE_LOWER );
-
 			$this->allAutoEscapedFunctions = RulesetPropertyHelper::merge_custom_array(
-				$custom_auto_escaped_functions,
-				$auto_escaped_functions
+				$this->customAutoEscapedFunctions,
+				$this->autoEscapedFunctions,
+				true,
+				true
 			);
 
 			$this->addedCustomEscapingFunctions['autoescape'] = $this->customAutoEscapedFunctions;
--- a/WordPress/Helpers/EscapingFunctionsTrait.php
+++ b/WordPress/Helpers/EscapingFunctionsTrait.php
@@ -218,9 +218,15 @@ trait EscapingFunctionsTrait {
 		if ( array() === $this->allEscapingFunctions
 			|| $this->customEscapingFunctions !== $this->addedCustomEscapingFunctions['escape']
 		) {
+			/*
+			 * Lowercase all names, both custom as well as "known", as PHP treats namespaced names case-insensitively.
+			 */
+			$custom_escaping_functions = array_map( 'strtolower', $this->customEscapingFunctions );
+			$escaping_functions        = array_change_key_case( $this->escapingFunctions, \CASE_LOWER );
+
 			$this->allEscapingFunctions = RulesetPropertyHelper::merge_custom_array(
-				$this->customEscapingFunctions,
-				$this->escapingFunctions
+				$custom_escaping_functions,
+				$escaping_functions
 			);
 
 			$this->addedCustomEscapingFunctions['escape'] = $this->customEscapingFunctions;
@@ -242,9 +248,15 @@ trait EscapingFunctionsTrait {
 		if ( array() === $this->allAutoEscapedFunctions
 			|| $this->customAutoEscapedFunctions !== $this->addedCustomEscapingFunctions['autoescape']
 		) {
+			/*
+			 * Lowercase all names, both custom as well as "known", as PHP treats namespaced names case-insensitively.
+			 */
+			$custom_auto_escaped_functions = array_map( 'strtolower', $this->customAutoEscapedFunctions );
+			$auto_escaped_functions        = array_change_key_case( $this->autoEscapedFunctions, \CASE_LOWER );
+
 			$this->allAutoEscapedFunctions = RulesetPropertyHelper::merge_custom_array(
-				$this->customAutoEscapedFunctions,
-				$this->autoEscapedFunctions
+				$custom_auto_escaped_functions,
+				$auto_escaped_functions
 			);
 
 			$this->addedCustomEscapingFunctions['autoescape'] = $this->customAutoEscapedFunctions;
--- a/WordPress/Helpers/RulesetPropertyHelper.php
+++ b/WordPress/Helpers/RulesetPropertyHelper.php
@@ -55,9 +55,9 @@ final class RulesetPropertyHelper {
 	 */
 	public static function merge_custom_array( $custom, array $base = array(), $flip = true, $lowercaseKeyValues = false ) {
 		if ( $lowercaseKeyValues ) {
-			$base = array_map( 'strtolower', $base );
+			$base   = array_map( 'strtolower', $base );
 			$custom = array_map( 'strtolower', $custom );
-			$base = array_change_key_case( $base );
+			$base   = array_change_key_case( $base );
 			$custom = array_change_key_case( $custom );
 		}
 
--- a/WordPress/Helpers/RulesetPropertyHelper.php
+++ b/WordPress/Helpers/RulesetPropertyHelper.php
@@ -54,6 +54,13 @@ final class RulesetPropertyHelper {
 	 * @return array
 	 */
 	public static function merge_custom_array( $custom, array $base = array(), $flip = true, $lowercaseKeyValues = false ) {
+		if ( $lowercaseKeyValues ) {
+			$base = array_map( 'strtolower', $base );
+			$custom = array_map( 'strtolower', $custom );
+			$base = array_change_key_case( $base );
+			$custom = array_change_key_case( $custom );
+		}
+
 		if ( true === $flip ) {
 			$base = array_filter( $base );
 		}
@@ -68,13 +75,6 @@ final class RulesetPropertyHelper {
 
 		if ( empty( $base ) ) {
 			return $custom;
-		}
-
-		if ( $lowercaseKeyValues ) {
-			$base = array_map( 'strtolower', $base );
-			$custom = array_map( 'strtolower', $custom );
-			$base = array_change_key_case( $base );
-			$custom = array_change_key_case( $custom );
 		}
 
 		return array_merge( $base, $custom );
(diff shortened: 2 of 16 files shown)


WordPress/Helpers/EscapingFunctionsTrait.php
```diff
@@ -218,9 +218,15 @@ final public function is_escaping_function( $functionName ) {
 		if ( array() === $this->allEscapingFunctions
 			|| $this->customEscapingFunctions !== $this->addedCustomEscapingFunctions['escape']
 		) {
+			/*
+			 * Lowercase all names, both custom as well as "known", as PHP treats namespaced names case-insensitively.
+			 */
+			$custom_escaping_functions = array_map( 'strtolower', $this->customEscapingFunctions );
+			$escaping_functions        = array_change_key_case( $this->escapingFunctions, \CASE_LOWER );
```
jrfnl, line 225: Why are you doing the lowercasing _before_ the merge ?

Would it make sense to add a new parameter to the `RulesetPropertyHelper::merge_custom_array()` method (internal) to do the lowercasing in that method ?

WordPress/Sniffs/DB/DirectDatabaseQuerySniff.php
```diff
@@ -230,17 +230,17 @@ public function process_token( $stackPtr ) {
...
 							$cached = true;
 							break;
 						}
-					} elseif ( isset( $this->cacheGetFunctions[ $this->tokens[ $i ]['content'] ] ) ) {
+					} elseif ( isset( $this->cacheGetFunctions[ strtolower( $this->tokens[ $i ]['content'] ) ] ) ) {
 
 						$wp_cache_get = true;
 
-					} elseif ( isset( $this->cacheSetFunctions[ $this->tokens[ $i ]['content'] ] ) ) {
+					} elseif ( isset( $this->cacheSetFunctions[ strtolower( $this->tokens[ $i ]['content'] ) ] ) ) {
```
jrfnl, line 243: Would lowercasing the value once before going the if/elseif's make sense ?

jrfnl (changes requested):
@dingo-d Thanks for setting up.

Note: The tests aren't actually testing the issue. This needs fixing.

WordPress/Helpers/EscapingFunctionsTrait.php
dingo-d: I mean, this is a public helper, so would that be considered a major breaking change?

WordPress/Sniffs/DB/DirectDatabaseQuerySniff.php
dingo-d: Yeah, since it's the same value used in different branches 👍🏼

WordPress/Helpers/EscapingFunctionsTrait.php
jrfnl: 1. A new optional parameter is not a breaking change.
2. I'd keep the default to `false` as this ruleset helper is not just used for case _insensitive_ properties, so let's not break things which shouldn't be broken.

WordPress/Helpers/RulesetPropertyHelper.php
```diff
@@ -43,15 +43,24 @@ final class RulesetPropertyHelper {
 	 * @since 2.0.0  No longer supports custom array properties which were incorrectly
 	 *               passed as a string.
 	 * @since 3.0.0  Moved from the Sniff class to this class.
+	 * @since 3.1.0  Added a new parameter to lowercase array keys and values.
```
jrfnl, line 46: Should this be `3.1.0` or `3.0.2` ? It's a new feature, true, but it is on an internal class which doesn't have any BC promise and the actual patch is about _fixing a bug_.

WordPress/Helpers/RulesetPropertyHelper.php
```diff
@@ -43,15 +43,24 @@ final class RulesetPropertyHelper {
...
-	 * @param array $custom Custom list as provided via a ruleset.
-	 * @param array $base   Optional. Base list. Defaults to an empty array.
-	 *                      Expects `value => true` format when `$flip` is true.
-	 * @param bool  $flip   Optional. Whether or not to flip the custom list.
-	 *                      Defaults to true.
+	 * @param array $custom             Custom list as provided via a ruleset.
+	 * @param array $base               Optional. Base list. Defaults to an empty array.
+	 *                                  Expects `value => true` format when `$flip` is true.
+	 * @param bool  $flip               Optional. Whether or not to flip the custom list.
+	 * @param bool  $lowercaseKeyValues Optional. Whether to lowercase keys and values in the resulting array.
```
jrfnl, line 52:
```suggestion
	 * @param bool  $lowercasekeys Optional. Whether to lowercase keys in the resulting array.
```

Lowercasing the keys for use with an `isset()` lookup is one thing, lowercasing the values is something completely different and should **not** be done.

Using `strtolower()` on the values will change the type for any non-string scalars to string and will throw fatal errors for non-scalars.

WordPress/Helpers/RulesetPropertyHelper.php
```diff
@@ -43,15 +43,24 @@ final class RulesetPropertyHelper {
...
 	 * @return array
 	 */
-	public static function merge_custom_array( $custom, array $base = array(), $flip = true ) {
+	public static function merge_custom_array( $custom, array $base = array(), $flip = true, $lowercaseKeyValues = false ) {
+		if ( $lowercaseKeyValues ) {
+			$base   = array_map( 'strtolower', $base );
+			$custom = array_map( 'strtolower', $custom );
+			$base   = array_change_key_case( $base );
+			$custom = array_change_key_case( $custom );
+		}
```
jrfnl, line 62: Aside from the fact that the values shouldn't be touched, the `$custom` array should not be changed yet at this point, only the `$base` (with `array_change_key_case()`).

The `array_change_key_case()` for `$custom` should be done after the potential flip and before the check for empty `$base`.

Also never a bad thing to explicitly pass the `CASE_LOWER` constant as the second parameter to make the code more readable for the next person (prevents them having to look up what the default is).

jrfnl (changes requested):
Reviewed again. Left feedback.

Also looks like the tests are _better_ than before, but still not completely targeted to this patch.
Basically, the `phpcs:set` directives should only contain mixed/uppercase functions to test this patch.

WordPress/Helpers/RulesetPropertyHelper.php
dingo-d: I think I added that because in the `NoSilencedErrors` sniff the `customAllowedFunctionsList` is having its values lowercased (https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php#L187-L188). 
So I lowercased the values in the helper, seeing how in the majority of cases the non-list arrays have booleans for the values. 

I'll try to see what the values are in every case where this helper is used and see if it's safe to leave it, otherwise, I'll remove that part and revert the change in the `NoSilencedErrors` sniff.

dingo-d:
I've just now realized that this helper is kinda doing too much on its own than its name suggests.

The helper says it will _merge_ the custom array, but can also flip the base array in the case the custom array is empty and the `$flip` parameter is `true` (which is the default). And also do the same for the base array.

Would it make sense to rename it to something else? `manipulate_arrays` is a bit too generic and doesn't really describe what this helper does. Maybe `merge_and_prepare_arrays`, `flip_and_merge_arrays`? Something like that? The second one fits a bit more since the `$flip` parameter is always true.

WordPress/Helpers/RulesetPropertyHelper.php
jrfnl: Values coming from properties are (in most cases) "flipped" to be the keys via the helper function, so the resulting array can be used with `isset()`.

I haven't checked the code, but might well be that what is being done in the NoSilencedErrors sniff is basically the same as you're doing in this PR, it just was already done for that sniff ?

jrfnl:
> I've just now realized that this helper is kinda doing too much on its own than its name suggests.
> 
> The helper says it will _merge_ the custom array, but can also flip the base array in the case the custom array is empty and the `$flip` parameter is `true` (which is the default). And also do the same for the base array.
> 
> Would it make sense to rename it to something else? `manipulate_arrays` is a bit too generic and doesn't really describe what this helper does. Maybe `merge_and_prepare_arrays`, `flip_and_merge_arrays`? Something like that? The second one fits a bit more since the `$flip` parameter is always true.

Hard "no" from me on this. The function merges two arrays, whether flipped or not etc, those are just implementation details for the merge, so functionality-wise, the name is perfectly fine.

The fact that the Helper does not have a BC-promise, still doesn't mean that BC should be broken without good reason and there is no poignant reason to change the method name.

WordPress/Helpers/RulesetPropertyHelper.php
dingo-d: From what I've seen it was just done to this one sniff so it might be an isolated case. I'll need to take a bit closer look.

WordPress/Helpers/RulesetPropertyHelper.php
```diff
@@ -43,15 +43,25 @@ final class RulesetPropertyHelper {
...
-	 * @param array $base   Optional. Base list. Defaults to an empty array.
-	 *                      Expects `value => true` format when `$flip` is true.
-	 * @param bool  $flip   Optional. Whether or not to flip the custom list.
-	 *                      Defaults to true.
+	 * @param array $custom        Custom list as provided via a ruleset.
+	 * @param array $base          Optional. Base list. Defaults to an empty array.
+	 *                             Expects `value => true` format when `$flip` is true.
+	 * @param bool  $flip          Optional. Whether or not to flip the custom list.
+	 * @param bool  $lowercasekeys Optional. Whether to lowercase keys and values in the resulting array.
+	 *                             Defaults to false.
```
GaryJones, line 53: Can we avoid the addition of another boolean parameter?
- https://alexkondov.com/should-you-pass-boolean-to-functions/
- https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/

Consider making it a string, like `$key_value_transform = 'none'` as the default, but 'lowercase_keys_only' or `uppercase_keys_values` or similar strings can be passed to determine the transformation of keys, values, or both.

dingo-d: I'm not sure this is necessary, as this is a binary operation - either we lowercase the keys or not, there is no third option to choose from that would necessitate a string parameter imo.

Having a boolean value, to me, lowers the code complexity.

GaryJones: There's no third option now, but there might be in the future. 

Having descriptive string values makes the code easier to read than `true, true` in function calls, and then having to work out what they mean.

jrfnl: > Having descriptive string values makes the code easier to read than true, true in function calls, and then having to work out what they mean.

Well, named parameters in function calls is the real solution to that. We just have to wait for a PHP 8.0 minimum before we can use that...

I dislike using text labels for something which is a boolean thing. Too easy to have typos.
Which, of course, could be bypassed by using an enum for the values, but that will have to wait till PHP 8.1.

The PHP 5.4 compatible version of all that would be to have (class) constants which contain the valid values, be it boolean or string values, and pass those around instead.

I have no objection to a solution using class constants for the value of this parameter to make it more descriptive, but wonder if it's really necessary.

WordPress/Helpers/RulesetPropertyHelper.php
```diff
@@ -43,15 +43,25 @@ final class RulesetPropertyHelper {
 	 * @since 2.0.0  No longer supports custom array properties which were incorrectly
 	 *               passed as a string.
 	 * @since 3.0.0  Moved from the Sniff class to this class.
+	 * @since 3.0.2  Added a new parameter to lowercase array keys and values.
```
jrfnl, line 46:
```suggestion
	 * @since 3.0.2  Added a new parameter to lowercase array keys for the resulting array.
```

WordPress/Helpers/RulesetPropertyHelper.php
```diff
@@ -43,15 +43,25 @@ final class RulesetPropertyHelper {
...
-	 * @param array $custom Custom list as provided via a ruleset.
-	 * @param array $base   Optional. Base list. Defaults to an empty array.
-	 *                      Expects `value => true` format when `$flip` is true.
-	 * @param bool  $flip   Optional. Whether or not to flip the custom list.
-	 *                      Defaults to true.
+	 * @param array $custom        Custom list as provided via a ruleset.
+	 * @param array $base          Optional. Base list. Defaults to an empty array.
+	 *                             Expects `value => true` format when `$flip` is true.
+	 * @param bool  $flip          Optional. Whether or not to flip the custom list.
+	 * @param bool  $lowercasekeys Optional. Whether to lowercase keys and values in the resulting array.
```
jrfnl, line 52:
```suggestion
	 * @param bool  $lowercasekeys Optional. Whether to lowercase keys in the resulting array.
```

jrfnl:
@dingo-d Is this PR ready for its final review in the mean time ?

Maybe dismiss "requested changes" once they have been addressed and ask for a new review via the interface when you think its ready ? It's now not always clear whether or not it's time to review again.

dingo-d:
Ready for the final review @jrfnl 🙂

GaryJones (approved):

dingo-d:
Closing this PR as this is obviously not going to be moving along.

PR #2408, opened 2023-12-02
Switch to PHPCSStandards/PHP_CodeSniffer & raise minimum supported version

jrfnl:
⚠️  **_This is a DRAFT PR on purpose as it references releases which have not yet been tagged. Once the PHPCS 3.8.0 tag is available ~~(which contains the Composer `replace` directive)~~, this can/should be merged ~~and released ASAP~~._** ⚠️ 

---

### Switch to PHPCSStandards/PHP_CodeSniffer

The Squizlabs repo has been abandoned. The project continues in a fork in the PHPCSStandards organisation.

Ref:
* squizlabs/PHP_CodeSniffer#3932

### 🆕 Revert changes to the Composer package name

In contrast to earlier information, arrangements are now being made to allow the package to continue under its original name on Packagist. The commit (in the new repo) to rename the package [has been reverted](https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/135).

### Composer: raise the minimum supported PHPCS version to 3.8.0

... ... for improved PHP 8.2 support.

Includes updating references to the PHPCS version whenever relevant throughout the codebase.

### Composer: update various version constraints

... after the tooling has also updated to the PHPCSStandards version of PHPCS.

Refs:
* https://github.com/PHPCSStandards/PHPCSUtils/releases
* https://github.com/PHPCSStandards/PHPCSExtra/releases

### 🆕 DeprecatedFunctionsUnitTest: remove work-around for PHPCS bug

... as the bug is now fixed.

PR #2409, opened 2023-12-03
Use two new sniffs from PHPCSExtra 1.2.0

jrfnl:
PHPCSExtra 1.2.0 was released yesterday and contains two new sniffs which IMO would be suitable for adding to WordPressCS.

Includes raising the minimum supported PHPCSExtra version. (will conflict with PR #2408 and will need rebase depending on which is merged first)

Ref:
* https://github.com/PHPCSStandards/PHPCSExtra/releases/tag/1.2.0

--

_Note: adding these sniffs means this PR will need to go into a new minor, while #2408 can go into a patch release._


GaryJones (approved):

dingo-d (approved):

dingo-d:
Yup that is fine with me, we shouldn't block new releases (this one wasn't a draft so it's better to merge it sooner than later imo).

PR #2412, opened 2023-12-08
PHPStan: ignore two new issues

jrfnl:
... caused by improved, but not 100% correct, type information coming from PHPCSUtils.

This incorrect type info will be fixed in the next version of PHPCSUtils.


GaryJones (approved):

dingo-d (approved):

dingo-d:
Are those tests expected to fail or?

Issue #2123, opened 2022-12-04
PHPCS 3.?.?: remove temporary test exclusion on PHP 8.2+


jrfnl:
## Rationale

PR #2121 adds a temporary test exception for a test involving a function call to `readonly()`.

Once the upstream PHPCS tokenizer supports PHP 8.2 - hopefully in PHPCS 3.8.0 -, the minimum supported PHPCS version should be raised and the exclusion should be removed.

While this may _sound_ like a test tweak, it is in actual fact an issue with the sniff as the WP `DeprecatedFunctions` sniff will currently not report on the deprecated `readonly()` WP function, while it is important that it should as it will prevent plugins/themes from running into trouble when using the deprecated WP `readonly()` function with future PHP versions in which the function call will likely become a fatal error, so making this change as early as possible is highly recommended.

## References

* The discussion about this in squizlabs/PHP_CodeSniffer#3667
* php/php-src#9512

## Action Checklist

Once a PHPCS version is available which handles this correctly...
- [ ] Update the minimum supported PHPCS version in `composer.json` and update the README and workflows etc.
- [ ] Remove the test exception from the `DeprecatedFunctions` sniff.

jrfnl:
Closed via https://github.com/WordPress/WordPress-Coding-Standards/pull/2408/commits/105a10cafc6b5c6e37d326b336daec06a7d2cc90 which was included in #2408

PR #2414, opened 2023-12-10
Tests: allow for PHPUnit 8/9

jrfnl:
PHP_CodeSniffer 3.8.0 now allows for running the tests, which are based on the PHPCS native test suite, with PHPUnit 8 and 9.

This commit updates the package to take advantage of that.

Includes:
* Widening the PHPUnit version requirements.
* Adding the PHPUnit 8+ cache file to `.gitignore`.
* Updating the PHPUnit configuration to make sure that deprecations will always show, even when on a high PHPUnit 9.6 version.
* Updating the info in `CONTRIBUTING`.
* Simplifications to the `quicktest` and `test` workflows. Also, the code coverage "high" run can now be run against PHP 8.3.
* Running PHPStan against PHP `latest` (couldn't previously be done due to the old PHPUnit version).
* Minor tweak to the PHPStan config to make sure PHP 8.x specific issues don't get flagged for this codebase.
* Removing a no longer needed `--ignore-platform*` argument.
* Updating the "setUp" method in one test to call the parent "setUp" method via the new method name.

Ref:
* PHPCSStandards/PHP_CodeSniffer#59

### DIFF
--- a/.github/workflows/basic-qa.yml
+++ b/.github/workflows/basic-qa.yml
@@ -127,8 +127,7 @@ jobs:
           custom-cache-suffix: $(date -u "+%Y-%m")
 
       - name: "Set PHPCS version (lowest)"
-        if: ${{ matrix.phpcs_version == 'lowest' }}
-        run: composer update squizlabs/php_codesniffer --prefer-lowest --ignore-platform-req=php+ --no-scripts --no-interaction
+        run: composer update squizlabs/php_codesniffer --prefer-lowest --no-scripts --no-interaction
 
       - name: Test the WordPress-Core ruleset
         run: $(pwd)/vendor/bin/phpcs -ps ./Tests/RulesetCheck/class-ruleset-test.inc --standard=WordPress-Core
@@ -178,7 +177,7 @@ jobs:
       - name: Install PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '7.4'
+          php-version: 'latest'
           coverage: none
           tools: phpstan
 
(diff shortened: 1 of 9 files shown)


dingo-d (approved):

.github/workflows/quicktest.yml
```diff
@@ -76,21 +68,17 @@ jobs:
         if: ${{ matrix.phpcs_version == 'dev-master' }}
         run: composer lint -- --checkstyle
 
-      - name: Run the unit tests without code coverage - PHP 5.4 - 8.0
-        if: ${{ matrix.php == '5.4' && github.ref_name != 'develop' }}
+      - name: Run the unit tests without code coverage
+        if: ${{ github.ref_name != 'develop' }}
         run: composer run-tests
 
       # Until PHPCS supports PHPUnit 9, we cannot run code coverage on PHP 8.0+, so run it on PHP 5.4 and 7.4.
```
GaryJones, line 75: Does this line need updating / removing?

dingo-d:
This closes #2414 issue if I'm not mistaken.

jrfnl:
> This closes #2123 issue if I'm not mistaken.

Ow, nice catch, I missed that issue. Still, that one is not closed by this PR, but was fixed by the last commit in #2408.

GaryJones (approved):

.github/workflows/basic-qa.yml
```diff
@@ -127,8 +127,7 @@ jobs:
           custom-cache-suffix: $(date -u "+%Y-%m")
 
       - name: "Set PHPCS version (lowest)"
-        if: ${{ matrix.phpcs_version == 'lowest' }}
```
jrfnl, line 130: Shoot! This condition shouldn't have been removed... Must have accidentally happened during on of the rebases....

I will fix this up.

PR #2416, opened 2023-12-12
Fix PHPStan (again)

jrfnl:
PHPStan released version 1.10.49 today and managed to break things again... maybe someone should explain semver to them....

Either way, this fixes it.

I've favoured ignoring these issues instead of "fixing" them as we were already ignoring a "Result of method * (void) is used" for level 0, which is about the same thing.

These methods all `return` the return value of a parent/child method, which can either be `int` or `void`. For code consistency and stability, it is a lot clearer to have these `return` statements in place and it will prevent potential future breakage if the parent/child would start returning a stack pointer.

Ref: https://github.com/phpstan/phpstan/releases/tag/1.10.49

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -17,6 +17,16 @@ parameters:
             message: "#^Cannot unset offset 'query' on non-empty-array<int, array<string, int\\|string>>.$#"
             path: WordPress/Sniffs/DB/PreparedSQLPlaceholdersSniff.php
 
+        # Level 3
+        # Consistency preferred - these are methods which return whatever a parent/child method returns.
+        # They shouldn't need to keep track of whether the child method returns int or void.
+        - '#^Method [^ ]+Sniff::process_token\(\) should return int\|void but returns null\.$#'
+        -
+            # False positive. Will be fixed via next version of PHPCSUtils.
+            count: 1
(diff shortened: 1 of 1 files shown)


dingo-d (approved):

ondrejmirtes:
We're going to fix this use-case in PHPStan, looks easy enough to fix.

jrfnl:
Much appreciated @ondrejmirtes !

GaryJones (approved):
Approved, but I won't merge in case there's a thought that phpstan itself may be updated with a new release quick enough.

jrfnl:
> Approved, but I won't merge in case there's a thought that phpstan itself may be updated with a new release quick enough.

That's fine, let's give it a little, but I do have another PR ready to go, which would have a broken build without this, so let's not leave it too long as I think it would probably be good to get 3.1.0 out of the door soonish.

ondrejmirtes:
Fixed in PHPStan 1.10.50: https://github.com/phpstan/phpstan/releases/tag/1.10.50

jrfnl:
Thank you @ondrejmirtes !

I've triggered a [build on `develop`](https://github.com/WordPress/WordPress-Coding-Standards/actions/runs/7194355028) and can confirm this PR is no longer needed. Closing.

PR #2367, opened 2023-08-20
Update documentation (README, CONTRIBUTING) for release of WordPressCS 3.0.0

jrfnl:
### README: updates for WordPressCS 3.0.0

This commit updates the README guide to be in line with WordPressCS 3.0.0.

This commit:
* Removes the "Last commit to stable" and "Nr of contributors" badges.
* Updates the "Tested on PHP..." badges.
* Updates the table of contents to match the new content of the file.
* Removes the "Project history" section (has been moved to the wiki).
* Makes the pre-requisites for the package more explicit and puts them in a separate section.
* Updates the installation instructions and adds update instructions.
    - Removes the instructions related to stand-alone installs and using `composer create-project`.
    - The new installation instructions take the changes in Composer 2.2 (permission required for plugins to run) into account.
    - The new installation instructions include a link to the end user/ruleset maintainer upgrade guide for WPCS 3.0.0 as the initial upgrade may need extra work.
* Updates references to the Composer PHPCS plugin to make sure these reference the new name and link to the correct GH repo (as the repo has moved).
* Removes the reference to the VIP ruleset which was removed in WPCS 2.0.0 from the "Rulesets" section.
* Adds links to the "customizable properties" section for where to find customizable properties for PHPCS native sniffs and sniffs from PHPCSExtra.
* Adds the `VariableAnalysis` standard and the `VIP Coding Standards` to the list of optional additional rulesets to use.
* Updates the command and output for the example run.
* Removes the links to IDE related tutorials in favour of a link to the wiki (to save us having to maintain this list in multiple places).
* Ensures the text consistently uses `WordPressCS` instead of `WPCS`.

Closes #481
Closes #1821
Closes #1901

### CONTRIBUTING: updates for WordPressCS 3.0.0

This commit updates the CONTRIBUTING guide to be in line with WordPressCS 3.0.0.

This commit:
* Updates the information about usage of upstream sniffs and reporting upstream issues.
* Updates the information on creating a pull request.
* Removes the information about the use of a long running release PR (as we don't do this anymore in practice).
* Updates the information on running the unit tests, including the information on how to get set up with cloned dependencies.
* Updates the examples test run output and removes the outdated the asciinema image.
* Updates the code fragments and example output used in the tutorial part to be in line with the current code and run instructions.
* Ensures the text consistently uses `WordPressCS` instead of `WPCS`.
* Replaced some ablist language.

### Docs: miscellaneous changes

Change `WPCS` to `WordPressCS` in miscellaneous files.

PR #2418, opened 2023-12-15
Fix the test case file path used in a CONTRIBUTING.md example

rodrigoprimo:
I believe the path of the test case file used in the example of how to generate the list of expected lines with errors is wrong. This PR fixes it.

Here is the commit that introduced the current version of the line that is changed in this PR: https://github.com/WordPress/WordPress-Coding-Standards/pull/2367/commits/4a33d468821f805a1cd222990c46ad9d632b0bd7#diff-98e64bc1cd2db9333c6effe87bbe0d6dfe8714aba4c6bde45aa037fe0796e44cR141.

If we want to keep the current path for some reason, we might want to change the part that says "from the root directory of your WordPressCS clone".

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -138,7 +138,7 @@ Also note the class name convention. The method `getErrorList()` MUST return an 
 If you run the following from the root directory of your WordPressCS clone:
 
 ```sh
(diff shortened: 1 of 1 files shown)


jrfnl (approved):
Good catch!

PR #2424, opened 2024-02-16
GH Actions: minor tweaks

jrfnl:
### GH Actions: minor tweaks

* Remove `--checkstyle` for a lint job which doesn't display with `cs2pr`.
* Remove a stray condition.

### GH Actions: move code coverage build without MbString to PHP 7.2

PHP 7.2 should be using PHPUnit 8, which doesn't use PHP Parser, so _fingers crossed_, this should solve the build failure.

### DIFF
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -57,7 +57,7 @@ jobs:
 
       - name: Lint PHP files against parse errors
         if: ${{ matrix.dependencies == 'stable' }}
-        run: composer lint -- --checkstyle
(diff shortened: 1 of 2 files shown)


jrfnl:
The test build failure is related to this issue: https://github.com/sebastianbergmann/php-code-coverage/issues/1025.

First thing to try is adding cache warming for PHPUnit 9.3+. If that doesn't work... well, we'll have to have a think about that.

GaryJones (approved):

dingo-d (approved):

PR #2443, opened 2024-04-23
Composer: avoid writing a lock file

fredden:
When working with this repository as a developer, we should be using the latest compatible packages. By writing a lock file for Composer, we may get into a state where we are "stuck" on an older version of a dependency. This change avoids such a situation by telling Composer to not write out a lock file in the project.


dingo-d (approved):

jrfnl:
Rebased without changes to get a passing build.

PR #2466, opened 2024-07-14
Introduce new auto escaping functions

widoz:
Add the following functions to the auto escaped functions

- `get_block_wrapper_attributes` Render the block wrapper tag attributes
- `wp_interactivity_data_wp_context` Render the interactivity json context value


jrfnl (commented):
@widoz Thanks for the PR, but I'm going to reject these changes as neither function is actually properly safe.

Also see a previous discussion about `get_block_wrapper_attributes` in #2085.

PR #2471, opened 2024-08-06
Documentation: Update PHPCompatibilityWP version recommendation to 7.2

johnjago:
As of WordPress 6.6, released on 16 July 2024, WordPress no longer supports PHP 7.0 and 7.1, so I’ve updated the recommended testVersion for PHPCompatibilityWP to be 7.2 in the README.

References:
- https://make.wordpress.org/core/2024/04/08/dropping-support-for-php-7-1/
- https://wordpress.org/news/2024/07/dorsey/ (6.6 released)


jrfnl:
I'm in two minds about this change. For plugins/themes, WPCS, by default, presumes support for the last three versions of WP Core and two of those do not have a minimum supported version of PHP `7.2` yet.

At the same time, the phrasing in the README says "the same PHP versions as WordPress Core supports", so leaving that recommendation at PHP `7.0` is not correct either.

Also note that any update like this should probably also update the `phpcs.xml.dist.sample` file.

jrfnl:
I've taken the liberty to merge in the dev branch to get this into a mergable state. I will (squash)-merge this once the build has passed.

PR #2472, opened 2024-08-07
Fix typos

szepeviktor:
Found few misspellings.

Remaining ones: "post_statusses" and [cachable](https://github.com/search?q=repo%3AWordPress%2FWordPress-Coding-Standards%20cachable&type=code)

https://github.com/crate-ci/typos

### DIFF
--- a/WordPress/AbstractClassRestrictionsSniff.php
+++ b/WordPress/AbstractClassRestrictionsSniff.php
@@ -102,7 +102,7 @@ abstract class AbstractClassRestrictionsSniff extends AbstractFunctionRestrictio
 			return;
 		}
 
-		if ( true === $this->is_targeted_token( $stackPtr ) ) {
+		if ( true === $this->is_targetted_token( $stackPtr ) ) {
 			return $this->check_for_matches( $stackPtr );
 		}
 	}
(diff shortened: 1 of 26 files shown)


WordPress/AbstractClassRestrictionsSniff.php
```diff
@@ -116,7 +116,7 @@ public function process_token( $stackPtr ) {
 	 *
 	 * @return bool
 	 */
-	public function is_targetted_token( $stackPtr ) {
+	public function is_targeted_token( $stackPtr ) {
```
szepeviktor, line 119: There is a public method name change here.

dingo-d: From a quick search on the GH there are some external standards which seem to be using this abstract class, so changing this public method would be a major BC break IIRC.

dingo-d (changes requested):
The typos changes in other files are ok, but the changing of the public method would be a BC break so I'm not sure this should be changed.

WordPress/AbstractClassRestrictionsSniff.php
szepeviktor: Okay, reverting ...

szepeviktor: Done 🍏

dingo-d (approved):

jrfnl (approved):

Issue #2467, opened 2024-07-20
PrefixAllGlobals: sniff should check for four-letter prefixes


davidperezgar:
The PrefixAllGlobals sniff show that 3 letters are the minimum to make prefixes for functions and variables, but in the Plugin Review Team, We're using a minimum of 4.

We ask developers to use these minimum letters and this is our argument:

> Don't try to use two (2) or three (3) letter prefixes anymore. We host nearly 100-thousand plugins on WordPress.org alone. There are tens of thousands more outside our servers. Believe us, you’re going to run into conflicts.

So, let's update this sniff.

It solves updating the number to 4 in [line of the sniff](https://github.com/WordPress/WordPress-Coding-Standards/blob/d77b0204f49e4b4a2bbac791e3bbfbfe603edf06/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php#L67).

Related with #1733 and and the [update of Developer documentation](https://github.com/WordPress/developer-plugins-handbook/issues/18).

jrfnl:
The [proposed change to the plugin handbook](https://github.com/WordPress/developer-plugins-handbook/pull/19) needs to be approved and merged before this issue becomes actionable.

Once that's happened, feel free to submit a PR.

jrfnl:
I've left a comment with my thoughts on the proposed change in the [plugin handbook ticket](https://github.com/WordPress/developer-plugins-handbook/issues/18#issuecomment-2241046249) as the proposed change appears to be based on false logic.

jrfnl:
Looks like the [plugin handbook](https://github.com/WordPress/developer-plugins-handbook/pull/19) has been updated now, so a PR to make this change could now be pulled.

davidperezgar:
I've made the PR ready to review.

PR #2479, opened 2024-08-29
Update 4 letters for min prefix length

davidperezgar:
Related #2467

### DIFF
--- a/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
+++ b/WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
@@ -58,13 +58,13 @@ final class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	/**
 	 * Minimal number of characters the prefix needs in order to be valid.
 	 *
-	 * @since 2.2.0
+	 * @since n.e.x.t.
 	 *
-	 * @link https://github.com/WordPress/WordPress-Coding-Standards/issues/1733 Issue 1733.
+	 * @link https://github.com/WordPress/WordPress-Coding-Standards/issues/2467 Issue 2467.
 	 *
 	 * @var int
 	 */
-	const MIN_PREFIX_LENGTH = 3;
+	const MIN_PREFIX_LENGTH = 4;
 
 	/**
 	 * Target prefixes.


WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -58,13 +58,13 @@ final class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	/**
 	 * Minimal number of characters the prefix needs in order to be valid.
 	 *
-	 * @since 2.2.0
+	 * @since n.e.x.t.
```
dingo-d, line 61: This is not a correct syntax.

The `@since 2.2.0` tag should be left as is, and the next version where this update is going to be released has to be added under it with the explanation of what was changed, possibly link to the plugin handbook where this change is mentioned.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -58,13 +58,13 @@ final class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
...
 	 *
-	 * @link https://github.com/WordPress/WordPress-Coding-Standards/issues/1733 Issue 1733.
+	 * @link https://github.com/WordPress/WordPress-Coding-Standards/issues/2467 Issue 2467.
```
dingo-d, line 63: Same here, I'd add this as a separate `@link`.

dingo-d (changes requested):
I've added some comments about docblock changes. Not sure why the tests are failing tho 🤔

jrfnl (changes requested):
Thanks for PR @davidperezgar.

Can you please do things properly though ? It looks like you don't understand how docs work.

* The `@since` tag should NOT be changed as it indicated when the constant was added. A second `@since` tag can be added to in explain that the value of the constant changed in WPCS 3.2.0.
* The `@link` tag should NOT be changed as it is a reference to explain why the constant was added. A secondary `@link` tag can be added to the new ticket to explain why it was raised.

Also, this is a functional change, so the tests will need to be updated to match, as can also be seen by the failing build.

jrfnl:
Ha, looks like @dingo-d and me were reviewing at the same time, but had the same feedback ;-)

davidperezgar:
Ok! I'll check it , thanks.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
davidperezgar: ok! I've reverted.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
davidperezgar: ok!

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -58,8 +58,9 @@ final class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	/**
 	 * Minimal number of characters the prefix needs in order to be valid.
 	 *
-	 * @since n.e.x.t.
+	 * @since 2.2.0
```
jrfnl, line 61: Still missing a second tag `@since 3.2.0 The minimum number of characters was raised from 3 to 4`.

jrfnl (changes requested):
Still missing a test with a three character prefix which should be flagged as not acceptable.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
davidperezgar: ok!

jrfnl: @davidperezgar You've now removed the original `@since` tag (again...) and still not added the extra `@since` tag with explanation of what's changed.
The extra test needed is also still missing.

If something is unclear: read documentation, check existing code for how it's done and ask questions. In the mean time, this is now the third time I've reviewed and this tiny PR is still not mergable.

davidperezgar:
I don't understand how it works the unit testing in the sniff.. I'd need some help...

jrfnl:
> I don't understand how it works the unit testing in the sniff.. I'd need some help...

Did you read the [CONTRIBUTING guide](https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/.github/CONTRIBUTING.md#unit-testing) ?

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.1.inc
```diff
@@ -674,3 +674,6 @@ class WP_Atom_Server {
 }
 
 // phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[]
+
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] ^%&
+function ^%&_do_something() {}
```
jrfnl, line 679: 1. This is not a valid test for the minimum char limit as this is an invalid prefix no matter what.
2. The test is placed incorrectly, it needs to be **above** the reset of the prefixes on line 676. Custom properties always need to be reset at the end of each test file.
3.. Always have a new line at the end of every file.

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
davidperezgar: I didn't know that it has to be both. I update the tag.

davidperezgar:
> Did you read the [CONTRIBUTING guide](https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/.github/CONTRIBUTING.md#unit-testing) ?

Yes!

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -58,13 +58,15 @@ final class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	/**
 	 * Minimal number of characters the prefix needs in order to be valid.
 	 *
+	 * @since 3.2.0
 	 * @since 2.2.0
```
GaryJones, line 62:
```suggestion
	 * @since 2.2.0
	 * @since 3.2.0 The minimum number of characters was raised from 3 to 4.
```

- Keep the since tags in order from top to bottom (then the history reads in a more logical order). 
- Add the description of what was changed in the latest amendment.

davidperezgar: Thanks Gary

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.1.inc
```diff
@@ -673,4 +673,7 @@ class WP_Atom_Server {
 	}
 }
 
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] tod
+function tod_something() {}
```
GaryJones, line 677: If you look at the [action test failure](https://github.com/WordPress/WordPress-Coding-Standards/actions/runs/10708736147/job/29691753365?pr=2479), you'll see that line 677 is NOT generating an error.

Why?

See the note on line 502:

> The errors are triggered on LINE 1 for the unit-test, because it's the phpcs:set command that is wrong, not the implementing code.

You may also see that lines 506-520 also handle the short prefixes stuff already - so consider updating one of the two-character prefixes (e.g. the two emojis) to a three-character prefix, and the `aaa` prefix test on line 514 to be `aaaa`. Then remove this line 676-678, and reset the line 1 error count.

davidperezgar: gotcha!!!! thankssss

GaryJones:
I see that this sniff is included in WordPress-Extra.

I also see the history of this change is related to plugins, but what about themes and the theme review team? This change will affect those folks who are writing themes with custom functions too - has that impact been considered?

davidperezgar:
Hello, yes it will impact to themes as well, but we have been considered the benefits of this measure, and that's why is now in the Developers HandBook:
https://developer.wordpress.org/plugins/plugin-basics/best-practices/

We were using this number of prefixes for a some time ago and we are unifying everywhere.

We could ping Themes Theme to be aware of it.

GaryJones:
There are several mentions of "prefix" in the theme developers handbook, but:

- https://developer.wordpress.org/themes/getting-started/reading-this-handbook/ and
- https://developer.wordpress.org/themes/basics/theme-functions/

... seem to be the missed opportunities to clarify/define the specific minimum length of such prefixes.

It seems right (to me) to get their buy-in on this before we merge this PR; the Plugins Review Team can continue to require 4 characters as they have apparently have for a while anyway).

davidperezgar:
I've spoken with Team Rep and they will check with the Team. I don't see a problem for that. Thanks @GaryJones

kafleg:
@GaryJones Thank you for sharing the link. Once we decided to have 4 letter for min prefix length, we'll discuss within the team for updating the handbook.

davidperezgar:
Thanks!  @kafleg !

rodrigoprimo (changes requested):
I believe the sniff documentation needs to be updated as well to reflect the changes proposed in this PR:

https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Docs/NamingConventions/PrefixAllGlobalsStandard.xml#L102-L118

davidperezgar:
I believe now is ready to be reviewed. Thanks.

rodrigoprimo (approved):
Thanks for implementing the changes, @davidperezgar! Looks good to me now.

@WordPress/wpcs, as far as I can check, all the points raised were addressed and this PR is ready to be merged.

Edit: @jrfnl (as we discussed this via DM), it seems that mentioning teams is not working in this repository. Maybe due to the way permissions are configured in this GH organization. Mentioning the other maintainers as well directly as the team mention above didn't work: @GaryJones and @dingo-d.

rodrigoprimo:
@davidperezgar and @kafleg, any updates on including the minimum four-letter prefix rule in the theme handbook? @GaryJones mentioned he'd prefer this information to be added to the theme handbook before merging this PR.

kafleg:
Updated the guidelines doc, https://make.wordpress.org/themes/handbook/review/required/ mentioning in the prefix section.

dingo-d (approved):

WordPress/Sniffs/NamingConventions/PrefixAllGlobalsSniff.php
```diff
@@ -58,13 +58,15 @@ final class PrefixAllGlobalsSniff extends AbstractFunctionParameterSniff {
 	/**
 	 * Minimal number of characters the prefix needs in order to be valid.
 	 *
+	 * @since 3.2.0 The minimum number of characters was raised from 3 to 4.
 	 * @since 2.2.0
```
jrfnl: The order of the `@since` tags is incorrect.

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.1.inc
```diff
@@ -511,13 +511,13 @@ function aa_do_something(){}
...
-function aaa_do_something(){}
+function aaaa_do_something(){}
 
 // phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] 😊
 function 😊_do_something(){}
 
-// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] 😊😊
-function 😊😊_do_something(){}
+// phpcs:set WordPress.NamingConventions.PrefixAllGlobals prefixes[] 😊😊😊
+function 😊😊😊_do_something(){}
```
jrfnl, line 520: I don't understand why this test was changed ? This line isn't flagged with or without this change.

jrfnl (changes requested):
Sorry that it took me so long to have another look at this, but the sheer amount of churn and feedback which has been needed to get this PR to the finish line was making me dread having to have another look.

Unfortunately, the PR is **_still_** not correct.

I'm inclined to just fix it up myself now and get it merged and be done with it. @GaryJones @dingo-d Objections ?

WordPress/Tests/NamingConventions/PrefixAllGlobalsUnitTest.1.inc
GaryJones: I assume this was updated to test the new boundary.

jrfnl: So not to test the line, but to test the warning (on line 1) about the prefix being too short ? Fair enough.

davidperezgar: Should I change it back? I'm fine with that.

PR #2492, opened 2024-09-17
Add documentation for FileName sniff

Nic-Sevic:
Related to #1722

- docs/Files/FileNameStandard.xml

### DIFF
--- a/WordPress/Docs/Files/FileNameStandard.xml
+++ b/WordPress/Docs/Files/FileNameStandard.xml
@@ -1,0 +1,32 @@
+<documentation title="File Name Standards">
+<standard>
+    <![CDATA[
+        Ensures that file names meet the following standards:
+        - Filenames should be all lowercase with hyphens separating words.
+        - Class file names should be based on the class name with "class-" prepended.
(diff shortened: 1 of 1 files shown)


rodrigoprimo (commented):
Thanks for your contribution, @Nic-Sevic! Usually, a `<standard>` and `<code_comparison>` block should be created for each of the error/warnings generated by the sniff, but there are some exceptions. Did you intentionally combine everything in a single `<standard>` and `<code_comparison>` pair? Without thinking too much about it, I'm more inclined to think that it would be better to separate the error/warning messages.

I just noticed that what I shared above is not mentioned in #1722, so there is a big change you were not aware of that (cc @jrfnl in case we should add this information in the issue description or maybe link to https://github.com/PHPCSStandards/PHP_CodeSniffer/blob/master/.github/CONTRIBUTING.md#writing-sniff-documentation). That is my bad.

rodrigoprimo:
@Nic-Sevic, I was just wondering if you'll have a chance to finish this off in the near future. It would be great if this PR could be included in the next WPCS release.

PR #2493, opened 2024-09-17
Adding documentation to WordPress.WP.DiscouragedConstants

RafaelFunchal:
Related to #1722

### DIFF
--- a/WordPress/Docs/WP/DiscouragedConstantsStandard.xml
+++ b/WordPress/Docs/WP/DiscouragedConstantsStandard.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Discouraged Constants"
+    >
+    <standard>
+    <![CDATA[
+    The usage of WordPress Constants is discouraged.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: A function is used in the code to retrieve the theme directory.">
+        <![CDATA[
+<img src="
+<?php echo <em>get_template_directory_uri()</em>; ?>
+/img/logo.svg" />
+        ]]>
+        </code>
+        <code title="Invalid: The TEMPLATEPATH constant is used in the code.">
+        <![CDATA[
+<img src="
+<?php echo <em>TEMPLATEPATH</em>;?>
+/img/logo.svg" />
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Defining WordPress Constants is discouraged.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: A function is used in the code to retrieve the theme directory.">
+        <![CDATA[
+<img src="
+<?php echo <em>get_template_directory_uri()</em>; ?>
+/img/logo.svg" />
+        ]]>
+        </code>
+        <code title="Invalid: The TEMPLATEPATH constant is set by a theme.">
+        <![CDATA[
+<em>define("TEMPLATEPATH", "foo");</em>
+<img src="
+<?php echo <em>$foo</em>; ?>
+/img/logo.svg" />
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>


WordPress/Docs/WP/DiscouragedConstantsStandard.xml
```diff
@@ -0,0 +1,49 @@
...
+    >
+    <standard>
+    <![CDATA[
+    The usage of WordPress Constants is discouraged.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: A function is used in the code to retrieve the theme directory.">
+        <![CDATA[
+<img src="
```
rodrigoprimo, line 14: AFAIK, it is not common to indent an HTML tag like it is done here. Maybe instead of fixing the indentation, we can simplify the example and make it easier to understand what this sniff is about? I believe the code sample could be trimmed down to `echo <em>TEMPLATEPATH</em>;`.

WordPress/Docs/WP/DiscouragedConstantsStandard.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Discouraged Constants"
+    >
+    <standard>
+    <![CDATA[
+    The usage of WordPress Constants is discouraged.
```
rodrigoprimo, line 8:
```suggestion
    The usage of WordPress constants is discouraged.
```

WordPress/Docs/WP/DiscouragedConstantsStandard.xml
```diff
@@ -0,0 +1,49 @@
...
+        <![CDATA[
+<img src="
+<?php echo <em>TEMPLATEPATH</em>;?>
+/img/logo.svg" />
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Defining WordPress Constants is discouraged.
```
rodrigoprimo, line 29:
```suggestion
    Defining WordPress constants is discouraged.
```

WordPress/Docs/WP/DiscouragedConstantsStandard.xml
rodrigoprimo, line 8: I wonder if it makes sense to list here and in the other `<standard>` block the discouraged constants, as this sniff triggers a warning only for a specific set of WP constants and not all constants. What do you think?

WordPress/Docs/WP/DiscouragedConstantsStandard.xml
```diff
@@ -0,0 +1,49 @@
...
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Discouraged Constants"
+    >
+    <standard>
+    <![CDATA[
+    The usage of WordPress Constants is discouraged.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: A function is used in the code to retrieve the theme directory.">
```
rodrigoprimo, line 12: This is not explicitly mentioned in #1722, but my understanding, based on checking the documentation from other sniffs, is that the description of the valid and invalid examples should be generic and describe what constitutes a valid example for this particular check that the sniff performs instead of describing the code sample provided.

Maybe here, the valid description should mention that the code does not use one of the discouraged WP constants or something like that?

rodrigoprimo (changes requested):
Thanks for creating this PR, and sorry for the delay in reviewing it, @RafaelFunchal! I left some comments with points I believe need to be addressed before it can be merged.

WordPress/Docs/WP/DiscouragedConstantsStandard.xml
rodrigoprimo: Actually just `echo <em>TEMPLATEPATH</em>;` is not a realistic example. Maybe something like the example below is better?

```php
$subdir = TEMPLATEPATH . '/subdir/';
```

WordPress/Docs/WP/DiscouragedConstantsStandard.xml
```diff
@@ -0,0 +1,49 @@
...
+        <![CDATA[
+<img src="
+<?php echo <em>get_template_directory_uri()</em>; ?>
+/img/logo.svg" />
+        ]]>
+        </code>
+        <code title="Invalid: The TEMPLATEPATH constant is set by a theme.">
+        <![CDATA[
+<em>define("TEMPLATEPATH", "foo");</em>
```
rodrigoprimo, line 42: I would suggest using a different discouraged constant here, so that we have two examples of constants that trigger this sniff. And possibly wrapping it with a if defined check to illustrate that the sniff is triggered even in this case.

rodrigoprimo:
@RafaelFunchal, I was just wondering if you'll have a chance to finish this off in the near future. It would be great if this PR could be included in the next WPCS release.

WordPress/Docs/WP/DiscouragedConstantsStandard.xml
paulopmt1: That makes sense. I'm also adding two examples.

WordPress/Docs/WP/DiscouragedConstantsStandard.xml
paulopmt1: Sounds even better, thanks!

paulopmt1:
I'm continuing this work here: https://github.com/WordPress/WordPress-Coding-Standards/pull/2589

PR #2495, opened 2024-09-17
Add documentation for WordPress.PHP.NoSilencedErrors

gogdzl:
Related to #1722

### DIFF
--- a/WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
+++ b/WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
@@ -1,0 +1,36 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Discourage the use of the PHP error silencing operator"
+    >
+    <standard>
+    <![CDATA[
+    Silencing errors is strongly discouraged. Use proper error checking instead.
+
+    Using the error operator with certain functions is allowed because no amount of error checking can fully prevent PHP from throwing errors when these functions are executed. The functions where this is permitted include, but are not limited to:
+
+    - `file_exists()`
+    - `file_get_contents()`
+    - `filesize()`
+    - `filetype()`
+    - `fopen()`
+    - `ftp_login()`
+    - `ftp_rename()`
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using proper error checking when calling functions not in the allowed list.">
+        <![CDATA[
+$conn_id = ftp_connect( $ftp_server );
+if ( ! $conn_id ) {
+  // Handle it as needed.
+}
+        ]]>
+        </code>
+        <code title="Invalid: Using the error operator to silence errors.">
+        <![CDATA[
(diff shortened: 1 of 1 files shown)


WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Discourage the use of the PHP error silencing operator"
+    >
+    <standard>
+    <![CDATA[
+    Silencing errors is strongly discouraged. Use proper error checking instead.
+
+    Using the error operator with certain functions is allowed because no amount of error checking can fully prevent PHP from throwing errors when these functions are executed. The functions where this is permitted include, but are not limited to:
```
rodrigoprimo, line 10: I agree that we should not include all the allowed functions, as the list is very long. But I'm unsure about adding just a few. I don't really know what criteria would be good for deciding which functions to include. What criteria did you use?

I'd be more inclined to just mention that some functions are allowed and point users to search for the `$allowedFunctionsList` property in https://github.com/WordPress/WordPress-Coding-Standards/blob/b0fc0c082a07da248fe767c10486a7fa27cc7c57/WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php. That being said, I was not able to find other sniff documentation doing what I'm suggesting both in the WPCS and the PHPCS repository. So, I advise you to wait to hear what the maintainers think is the best approach here.

In the WPCS repository, there is one sniff documentation that uses an external link in a different context: https://github.com/WordPress/WordPress-Coding-Standards/blob/cdb29b5a7c145ea30a84bcdf254b4aafaccf349c/WordPress/Docs/NamingConventions/PrefixAllGlobalsStandard.xml#L6-L16

WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
```diff
@@ -0,0 +1,36 @@
...
+    - `file_get_contents()`
+    - `filesize()`
+    - `filetype()`
+    - `fopen()`
+    - `ftp_login()`
+    - `ftp_rename()`
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using proper error checking when calling functions not in the allowed list.">
```
rodrigoprimo, line 22: To me, the valid message should be about what the sniff considers valid. The sniff doesn't check for proper error checking. Instead, it checks whether the `@` operator is present.

For the same reason, I'm unsure if error handling should be included in the code sample. Maybe checking the documentation of other sniffs might help you decide which approach you want to take? Also, feel free to wait for the input of one of the maintainers regarding this.

rodrigoprimo (changes requested):
Thanks for your contribution, @gogdzl! I left some inline comments. As always, as I'm not a maintainer, those are just suggestions from someone who created sniff documentation in the past.

rodrigoprimo:
@gogdzl, I was just wondering if you'll have a chance to finish this off in the near future. It would be great if this PR could be included in the next WPCS release.

rodrigoprimo:
@gogdzl, please let us know within a week if you are still interested in finishing this PR. If we don't hear back from you, we will presume you don't have time, and we will see if we can find someone else to take over and finish it. Thanks for your work so far!

PR #2504, opened 2024-11-04
Add a check for the leading and trailing spaces

### LINKED ISSUE #2501: I18n: Report the translatable strings with leading and trailing spaces

At translate.w.org, sometimes we get translations that start or end with one or more empty spaces, and we want to avoid this. We have [added some flags](https://github.com/GlotPress/GlotPress/pull/1500) to GlotPress to show them, but ideally they should not appear in the English strings (the strings to be translated), as it is indicated in the “Best practices” of [How to Internationalize Your Plugin](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/): `Do not leave leading or trailing whitespace in a translatable phrase`.

Some examples of incorrect strings in the translation system: [1](https://translate.wordpress.org/projects/wp-plugins/elementor/dev/gl/default/?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=15824298&filters%5Btranslation_id%5D=105222672), [2](https://translate.wordpress.org/projects/wp-plugins/elementor/dev/gl/default/?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=15824297&filters%5Btranslation_id%5D=105222304), [3](https://translate.wordpress.org/projects/wp-plugins/elementor/dev/gl/default/?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=15824296&filters%5Btranslation_id%5D=105222669), [4](https://translate.wordpress.org/projects/wp-plugins/elementor/dev/gl/default/?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=15824295&filters%5Btranslation_id%5D=105222664).

When a user will execute `phpcs`, I would like to see these errors, with an option to resolve them with `phpcbf`.

```
FILE: /Users/my-user/wordpress/wp-content/plugins/my-plugin/index.php
---------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
---------------------------------------------------------------------
 1 | ERROR | [x] The translatable string can't start with one or more empty spaces
 2 | ERROR | [x] The translatable string can't end with one or more empty spaces
---------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
---------------------------------------------------------------------
```

- [x] I intend to create a pull request to implement this feature.

amieiro:
As I described in https://github.com/WordPress/WordPress-Coding-Standards/issues/2501, at translate.w.org, sometimes we get translations that start or end with one or more empty spaces, and we want to avoid this.

This PR adds a new check to detect if a translatable string has leading or trailing spaces.

Fixes https://github.com/WordPress/WordPress-Coding-Standards/issues/2501.


jrfnl (commented):
Hi @amieiro, thanks for this PR.

I haven't tested it yet, but based on a code review alone, this is looking good.

I wonder if the tests can be made a little more varied ? I.e. not just use leading/trailing _spaces_, but also have some leading/trailing tabs, new lines and combinations of spaces, tabs, new lines.

What do you think ?

jrfnl:
@amieiro Just checking in as you never responded to the question I posed in my initia review.... Do you still intend to finish off this PR ?

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+		// Check for leading spaces.
+		if ( preg_match( $pattern_leading_spaces, $content_without_quotes ) ) {
+			$this->phpcsFile->addError(
+				'Translatable string should not have leading spaces. Found: %s',
+				$first_non_empty,
+				'LeadingSpaces',
+				array( $param_info['clean'] )
+			);
+		}
+
```
GaryJones:
```suggestion

```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+		// Check for trailing spaces.
+		if ( preg_match( $pattern_trailing_spaces, $content_without_quotes ) ) {
+			$this->phpcsFile->addError(
+				'Translatable string should not have trailing spaces. Found: %s',
+				$first_non_empty,
+				'TrailingSpaces',
+				array( $param_info['clean'] )
+			);
+		}
+
```
GaryJones:
```suggestion

```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+		// Check for leading tabs.
+		if ( preg_match( $pattern_leading_tabs, $content_without_quotes ) ) {
+			$this->phpcsFile->addError(
+				'Translatable string should not have leading tabs. Found: %s',
+				$first_non_empty,
+				'LeadingTabs',
+				array( $param_info['clean'] )
+			);
+		}
+
```
GaryJones:
```suggestion

```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+		// Check for trailing tabs.
+		if ( preg_match( $pattern_trailing_tabs, $content_without_quotes ) ) {
+			$this->phpcsFile->addError(
+				'Translatable string should not have trailing tabs. Found: %s',
+				$first_non_empty,
+				'TrailingTabs',
+				array( $param_info['clean'] )
+			);
+		}
+
```
GaryJones:
```suggestion

```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+		// Check for leading vertical tabs.
+		if ( preg_match( $pattern_leading_vtabs, $content_without_quotes ) ) {
+			$this->phpcsFile->addError(
+				'Translatable string should not have leading vertical tabs. Found: %s',
+				$first_non_empty,
+				'LeadingVTabs',
+				array( $param_info['clean'] )
+			);
+		}
+
```
GaryJones:
```suggestion

```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+		// Check for leading new lines.
+		if ( preg_match( $pattern_leading_newlines, $content_without_quotes ) ) {
+			$this->phpcsFile->addError(
+				'Translatable string should not have leading new lines. Found: %s',
+				$first_non_empty,
+				'LeadingNewLines',
+				array( $param_info['clean'] )
+			);
+		}
+
```
GaryJones:
```suggestion

```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+		$first_non_empty        = $this->phpcsFile->findNext( Tokens::$emptyTokens, $param_info['start'], ( $param_info['end'] + 1 ), true );
+
+		// Define regex patterns.
+		$pattern_leading_spaces  = '/^[\x20]+/u';
+		$pattern_trailing_spaces = '/[\x20]+$/u';
+		$pattern_leading_tabs    = '/^\x09+/u';
+		$pattern_trailing_tabs   = '/\x09+$/u';
+		$pattern_leading_vtabs   = '/^\x0B+/u';
+		$pattern_trailing_vtabs  = '/\x0B+$/u';
+		$pattern_leading_newlines = '/^\x0A+/u';
```
GaryJones:
```suggestion
		$pattern_leading_newlines  = '/^\x0A+/u';
```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+		$content_without_quotes = TextStrings::stripQuotes( $param_info['clean'] );
+		$first_non_empty        = $this->phpcsFile->findNext( Tokens::$emptyTokens, $param_info['start'], ( $param_info['end'] + 1 ), true );
+
+		// Define regex patterns.
+		$pattern_leading_spaces  = '/^[\x20]+/u';
+		$pattern_trailing_spaces = '/[\x20]+$/u';
+		$pattern_leading_tabs    = '/^\x09+/u';
+		$pattern_trailing_tabs   = '/\x09+$/u';
+		$pattern_leading_vtabs   = '/^\x0B+/u';
+		$pattern_trailing_vtabs  = '/\x0B+$/u';
```
GaryJones:
```suggestion
		$pattern_trailing_vtabs    = '/\x0B+$/u';
```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+		// Strip surrounding quotes.
+		$content_without_quotes = TextStrings::stripQuotes( $param_info['clean'] );
+		$first_non_empty        = $this->phpcsFile->findNext( Tokens::$emptyTokens, $param_info['start'], ( $param_info['end'] + 1 ), true );
+
+		// Define regex patterns.
+		$pattern_leading_spaces  = '/^[\x20]+/u';
+		$pattern_trailing_spaces = '/[\x20]+$/u';
+		$pattern_leading_tabs    = '/^\x09+/u';
+		$pattern_trailing_tabs   = '/\x09+$/u';
+		$pattern_leading_vtabs   = '/^\x0B+/u';
```
GaryJones:
```suggestion
		$pattern_leading_vtabs     = '/^\x0B+/u';
```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+	private function check_string_has_no_leading_trailing_spaces( $matched_content, $param_name, $param_info ) {
+		// Strip surrounding quotes.
+		$content_without_quotes = TextStrings::stripQuotes( $param_info['clean'] );
+		$first_non_empty        = $this->phpcsFile->findNext( Tokens::$emptyTokens, $param_info['start'], ( $param_info['end'] + 1 ), true );
+
+		// Define regex patterns.
+		$pattern_leading_spaces  = '/^[\x20]+/u';
+		$pattern_trailing_spaces = '/[\x20]+$/u';
+		$pattern_leading_tabs    = '/^\x09+/u';
+		$pattern_trailing_tabs   = '/\x09+$/u';
```
GaryJones:
```suggestion
		$pattern_trailing_tabs     = '/\x09+$/u';
```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+	 */
+	private function check_string_has_no_leading_trailing_spaces( $matched_content, $param_name, $param_info ) {
+		// Strip surrounding quotes.
+		$content_without_quotes = TextStrings::stripQuotes( $param_info['clean'] );
+		$first_non_empty        = $this->phpcsFile->findNext( Tokens::$emptyTokens, $param_info['start'], ( $param_info['end'] + 1 ), true );
+
+		// Define regex patterns.
+		$pattern_leading_spaces  = '/^[\x20]+/u';
+		$pattern_trailing_spaces = '/[\x20]+$/u';
+		$pattern_leading_tabs    = '/^\x09+/u';
```
GaryJones:
```suggestion
		$pattern_leading_tabs      = '/^\x09+/u';
```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+	 * @return void
+	 */
+	private function check_string_has_no_leading_trailing_spaces( $matched_content, $param_name, $param_info ) {
+		// Strip surrounding quotes.
+		$content_without_quotes = TextStrings::stripQuotes( $param_info['clean'] );
+		$first_non_empty        = $this->phpcsFile->findNext( Tokens::$emptyTokens, $param_info['start'], ( $param_info['end'] + 1 ), true );
+
+		// Define regex patterns.
+		$pattern_leading_spaces  = '/^[\x20]+/u';
+		$pattern_trailing_spaces = '/[\x20]+$/u';
```
GaryJones:
```suggestion
		$pattern_trailing_spaces.  = '/[\x20]+$/u';
```

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,115 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+	 *
+	 * @return void
+	 */
+	private function check_string_has_no_leading_trailing_spaces( $matched_content, $param_name, $param_info ) {
+		// Strip surrounding quotes.
+		$content_without_quotes = TextStrings::stripQuotes( $param_info['clean'] );
+		$first_non_empty        = $this->phpcsFile->findNext( Tokens::$emptyTokens, $param_info['start'], ( $param_info['end'] + 1 ), true );
+
+		// Define regex patterns.
+		$pattern_leading_spaces  = '/^[\x20]+/u';
```
GaryJones:
```suggestion
		$pattern_leading_spaces    = '/^[\x20]+/u';
```

GaryJones (dismissed):

jrfnl:
@amieiro Seeing as you've been working on this again - could you please explain why you have set things up to have separate errors for spaces vs tabs vs vtabs vs newlines ?

My previous question was only about having tests in place with these and the sniff handling all types of whitespace, not about splitting that up into multiple errors.

I cannot think of any situation where one type of (leading/trailing) whitespace would be acceptable, but another would not be, so having different error codes doesn't make sense to me.

Is there a reason for this distinction based on your experiences in the Polyglots team ? I'd be interested to hear your reasoning for this.

jrfnl:
> @jrfnl Sorry, this PR is not yet ready for review and I forgot to convert it to draft. As we [return to contribute](https://automattic.com/2025/05/29/returning-to-core/) with WordPress.org, I resumed work on this PR.

Okay, thanks for letting me know. I've converted it to draft now and won't do an in-detail review until it has been marked as "Ready to review".


> The horizontal tabulations are not often seen at the beginning or end of sentences, but they are seen in the middle, so I wanted to prevent its use, since I don't think it is correct to use it. You can see a use case inside a string in the core [here](https://translate.wordpress.org/projects/wp/dev/gl/default/?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=9144351&filters%5Btranslation_id%5D=70644590). The same applies to the new lines. You can see an example in the core [here](https://translate.wordpress.org/projects/wp/dev/gl/default/?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=5143989&filters%5Btranslation_id%5D=52619651).

I think that this PR should probably focus on the original problem it was trying to solve - leading and trailing whitespace.

Expanding the scope to mid-string whitespace characters feels like scope-creep and may have different implications (i.e., would involve a different decision point, potentially delaying the merging of the code related to the original problem).

I suggest opening a follow-up issue for that to discuss if this needs flagging and if so, what would need flagging. (I mean, looking at the examples, these seem like perfectly valid use-cases for mid string vertical whitespace, especially if you also consider RTL and BTT languages).

> And we have another thing to check: the different whitespace characters. The regular space characters is U+0020, but there are a lot of different [whitespace characters](https://en.wikipedia.org/wiki/Whitespace_character#Unicode): U+00A0, U+2000, U+2001, U+2002,...
> 
> The most common problem is the U+0020 space problem (in fact it is the only one I usually see), so the others are really edge situations. I can continue this PR in two different approaches:
> 
> * Contemplate only the U+0020 case and add the other cases to the tests.
> * Add all the commented situations.
> 
> Which approach do you prefer?

For leading and trailing whitespace, IMO the new errors should flag all whitespace characters, but without differentiating in the error message or code.
What I mean by that is:
* Change the error message(s) to use the term "whitespace" instead of explicitly saying what type of whitespace you're flagging.
* Change the error code(s) to also use the term "whitespace" - this would mean you'd end up with two error codes: `LeadingWhiteSpace` and `TrailingWhiteSpace`.

Does that help ?

jrfnl:
@amieiro Just checking in - will you have time to finish this PR in the near future ? It's been over a year since you opened it and it's still not marked as ready for review, while the base premise of the PR seems like a valid and valuable addition, which would help WPCS users.

amieiro:
Thanks for the feedback. I just pushed a refactor that addresses all of it.

What changed:
- Collapsed the eight error codes (Leading/Trailing × Spaces/Tabs/VTabs/NewLines) into just two: `LeadingWhiteSpace` and `TrailingWhiteSpace`.
- Reworded both error messages to use the generic term "whitespace".
- Replaced the eight hex-code regexes with a single `\s` match on each side. It already covers space, tab, vertical tab, newline, CR, and form feed in one pass. 
- Added four combination test cases (leading tab + space, trailing space + tab, mixed leading/trailing, and multi-line with newline + tab on each side).
- Bumped the `@since` annotation to 3.4.0.

Out of scope here, happy to open follow-up issues: mid-string whitespace and Unicode whitespace characters ([U+00A0](https://www.compart.com/en/unicode/U+00A0), [U+2000](https://www.compart.com/en/unicode/U+2000), [U+200B](https://www.compart.com/en/unicode/U+200B), etc.).

rodrigoprimo:
@amieiro, thanks for updating this PR! Could you please rebase it to fix the conflicts?

amieiro:
> @amieiro, thanks for updating this PR! Could you please rebase it to fix the conflicts?

Done!

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,42 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+	 * @param array|false $param_info      Parameter info array for an individual parameter,
+	 *                                     as received from the PassedParameters class.
+	 *
+	 * @return void
+	 */
+	private function check_string_has_no_leading_trailing_whitespace( $matched_content, $param_name, $param_info ) {
+		$content_without_quotes = TextStrings::stripQuotes( $param_info['clean'] );
+		$first_non_empty        = $this->phpcsFile->findNext( Tokens::$emptyTokens, $param_info['start'], ( $param_info['end'] + 1 ), true );
+
+		if ( preg_match( '/^\s+/u', $content_without_quotes ) ) {
```
rodrigoprimo, line 825: Nitpick: I believe the `+` quantifier is redundant here and in the regex below. A single surrounding whitespace character is enough to trigger these errors.
```suggestion
		if ( preg_match( '/^\s/u', $content_without_quotes ) ) {
```

rodrigoprimo, line 825: I wonder about the use of the `u` modifier. I assume it was added so that `\s` also matches non-ASCII whitespace characters? I worry a bit about unintended consequences of this modifier, as PHP's UTF-8 handling can be subtle, but I'm not sure whether any such edge cases apply here.

If it is important to match non-ASCII whitespace, I'd suggest adding tests to cover the cases the sniff should catch (presumably the ones you mentioned [here](https://github.com/WordPress/WordPress-Coding-Standards/pull/2504#issuecomment-4297916642)). Currently, the tests pass if the `u` modifier is removed.

WordPress/Sniffs/WP/I18nSniff.php
```diff
@@ -804,6 +805,42 @@ private function check_string_has_no_html_wrapper( $matched_content, $param_name
...
+	 *
+	 * @param string      $matched_content The token content (function name) which was matched
+	 *                                     in lowercase.
+	 * @param string      $param_name      The name of the parameter being examined.
+	 * @param array|false $param_info      Parameter info array for an individual parameter,
+	 *                                     as received from the PassedParameters class.
+	 *
+	 * @return void
+	 */
+	private function check_string_has_no_leading_trailing_whitespace( $matched_content, $param_name, $param_info ) {
```
rodrigoprimo, line 821: This method accepts `$matched_content` and `$param_name` but doesn't use either. The sibling `check_string_has_no_html_wrapper()` and other related methods (introduced in c9e8dbd0) have the same pattern.

@jrfnl, is there a reason for the uniform signature with unused parameters that I'm missing?

If not, I lean toward dropping the unused parameters for `check_string_has_no_leading_trailing_whitespace()`.

rodrigoprimo, line 821: Nitpick: maybe `check_string_has_no_surrounding_whitespace()` is a clearer name?

WordPress/Tests/WP/I18nUnitTest.1.inc
```diff
@@ -328,4 +328,56 @@ MyNamespace\__( 'foo', 'my-slug' ); // Ok.
 namespace\esc_html_e( 'foo', '' ); // The sniff should start flagging this once it can resolve relative namespaces.
 namespace\Sub\translate( 'foo', 'my-slug' ); // Ok.
 
+__( ' string', 'my-slug' ); // Bad: leading space.
```
rodrigoprimo, line 331: Looking at the new test cases for this check, I wonder if there is a reason to have duplicate tests that differ only in function name? They seem redundant to me because the method's logic doesn't vary based on the function name.

To me, the tests should cover different whitespace positions, whitespace types, and potentially parameter names, plus a few variations in whitespace quantity to guard against future regressions. What do you think?

rodrigoprimo (commented):
Thanks for updating this PR, @amieiro. I confirmed that you addressed the points raised in previous reviews. Besides that, I also left some inline comments for us to discuss.

WordPress/Sniffs/WP/I18nSniff.php
jrfnl: > @jrfnl, is there a reason for the uniform signature with unused parameters that I'm missing?

IIRC, in the original sniff (before extensive refactoring), most, if not all, the "sub-functions" were called via `call_user_func()`, which meant the expected parameters of these "sub-functions" needed to be the same.

You may very well be correct in that the various refactoring done since that time have made that requirement redundant. This would need to be verified and confirmed and if so, any method signatures for `private` methods can be cleaned up.

rodrigoprimo: Thanks for the additional details, @jrfnl.

@amieiro, considering the above, I think it is better to remove the unused parameters from `check_string_has_no_leading_trailing_whitespace()`. No need to change the other methods.

PR #2506, opened 2024-11-08
Various sniffs: simplify skipping the rest of the file

jrfnl:
This commit updates various sniffs to use `return $phpcsFile->numTokens` instead of `return ($phpcsFile->numTokens + 1)`.

If a sniff file contains 50 tokens, the last `$stackPtr` will be 49, so returning `50` will already get us passed the end of the token stack and the `+ 1` is redundant.

Includes changing a few plain `return` statements to `return $phpcsFile->numTokens` for efficiency.


dingo-d (approved):

GaryJones:
> If a sniff file contains 50 tokens, the last `$stackPtr` will be 49

Because it's a 0-indexed count?

GaryJones (approved):

PR #2513, opened 2024-11-29
I18nTextDomainFixer: remove unnecessary variable initialization

rodrigoprimo:
While working on #2512, I noticed an unnecessary variable initialization, so I'm creating this PR to suggest that it be removed. Its value is immediately overridden in an if/else below.

https://github.com/rodrigoprimo/WordPress-Coding-Standards/blob/690f05a51d190586e0bb0426a0f04242224d0723/WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php#L727

https://github.com/rodrigoprimo/WordPress-Coding-Standards/blob/690f05a51d190586e0bb0426a0f04242224d0723/WordPress/Sniffs/Utils/I18nTextDomainFixerSniff.php#L749

PR #2514, opened 2024-12-09
Remove unused variables from a few sniffs

rodrigoprimo:
As discussed in https://github.com/WordPress/WordPress-Coding-Standards/pull/2513#issuecomment-2510860338, this PR removes unused variables detected automatically using PHPStorm code inspection from a few sniffs.

I have checked that the removed variables do not trigger errors at PHPStan levels higher than what we are currently using and that the modified code has at least minimal test coverage.

@dingo-d, in the screenshot that you shared in https://github.com/WordPress/WordPress-Coding-Standards/pull/2513#issuecomment-2510860338, `EnqueuedResourcesSniff.php` is included, but I was not able to find any unused variables in this file.

### DIFF
--- a/WordPress/Sniffs/Security/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/Security/EscapeOutputSniff.php
@@ -185,7 +185,6 @@ class EscapeOutputSniff extends AbstractFunctionRestrictionsSniff {
 	 */
 	public function process_token( $stackPtr ) {
 		$start = ( $stackPtr + 1 );
-		$end   = $start;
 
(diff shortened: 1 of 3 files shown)


jrfnl (approved):
LGTM. Thanks @rodrigoprimo!

For context regarding the `$matches` parameters in `preg_match()` calls: prior to PHP 5.4, the `$matches` parameter was a required parameter when calling `preg_match_all()`, so this is likely a left-over loosely related to that (from when PHP 5.3 was still supported).

dingo-d (approved):

PR #2520, opened 2025-02-26
✨ New WordPress.WP.OptionAutoload sniff

### LINKED ISSUE #2473: Add sniff to warn about missing `$autoload` parameter when calling relevant WordPress option functions
Several option functions (e.g. `add_option(), `update_option()`) have been supporting an `$autoload` parameter that is optional and has had a default value of `'yes'`. This has led to notable problems where far too many options that plugins or themes add are being autoloaded, even if they're only needed on a few specific pages of the application.

WordPress 6.6 introduced changes in the API and how it should be used. While we can't require a parameter that used to be optional, it is now more than ever encouraged to always pass it. Going forward, not passing the parameter will be interpreted by core as implicit, i.e. core is able to determine whether or not to autoload the option by itself. So far this effectively still means an option without any provided value will be autoloaded, unless it is a particularly large option (the only condition implemented in WordPress core so far). But this current behavior to still _mostly_ autoload by default is only for backward compatibility and could change in the future.

See https://make.wordpress.org/core/2024/06/18/options-api-disabling-autoload-for-large-options/ for additional context. And somewhat related: https://make.wordpress.org/core/2023/10/17/new-option-functions-in-6-4/

`WordPress-Extra` should include a sniff that warns whenever the `$autoload` parameter is missing on `add_option()` or `update_option()` calls.

While they should typically use either `true` or `false`, passing `null` is fine too as explicit acknowledgement that core will be able to make the decision.

### Nice to have
It used to be formally documented that the strings `'yes'` and `'no'` could be passed alternatively to `true` and `false` respectively. This is still supported and not formally deprecated, but it's also no longer encouraged, so it would be nice to clean that up over time. There are however no functional differences between passing one or the other variant, so it's certainly not as important as the main solution described above. But if it's trivial to include in the new sniff, it's worth considering.

Just some additional context how to decide whether an option should be autoloaded or not:
* Use `true` if the option is retrieved in the majority of page loads, including frontend requests.
* Use `false` if the option is only retrieved in particular page loads, e.g. a specific page in the frontend or a specific admin screen.
* Use `null` to leave the decision whether or not to autoload the option to WordPress core.

One additional explanation for a somewhat common situation:
* If a plugin or theme uses several options only in a shared place (e.g. a specific admin screen), not autoloading them may intuitively feel like the wrong decision because it would make several database requests for them instead of a combined one. However, in this situation the options should still not be autoloaded because it would lead those options to be fetched in all kinds of other areas where they are never used.
* Instead, for such a scenario plugin developers should rely on the functions `wp_prime_option_caches()` or `wp_prime_option_caches_by_group()`, which were introduced in WordPress 6.4. With these functions, multiple options can be retrieved ("pre-loaded") from the database with a single request, without "polluting" the list of autoloaded options, which should remain reserved for options that are loaded across most of the WordPress site.

rodrigoprimo:
This sniff warns users about missing `$autoload` parameter when calling relevant WordPress option functions.

It includes tests and sniff documentation in the XML format.

Closes #2473 

### Sniff behavior

The sniff runs when the following functions are found in the code:
  - add_option( $option, $value = '', $deprecated = '', $autoload = null)
  - update_option( $option, $value, $autoload = null )
  - wp_set_options_autoload( $options, $autoload )
  - wp_set_option_autoload( $option, $autoload )
  - wp_set_option_autoload_values( $options )

For `add_option()` and `update_option()`, if the `$autoload` parameter is not passed, it simply recommends passing it. For the other three functions, the `$autoload` parameter (or `$options` parameter in the case of `wp_set_option_autoload_values()`) is mandatory, so it is not needed to check if the parameter is omitted.

If the `$autoload` parameter is passed and is `true|false|null` for `add_option()` and `update_option()` or `true|false` for `wp_set_options_autoload()`, `wp_set_option_autoload()` and `wp_set_option_autoload_values()`, the sniff stays silent as those are valid values.

If the `$autoload` parameter is passed and is `'yes'|'no'`, the sniff flags as discouraged parameter values and auto-fixes to `true|false`.

For the internal-use only values of the `$autoload` parameter, if the value is `'auto'|'auto-on'|'auto-off'` the sniff flags it as unsupported without auto-fixing. If the value is `'on'|'off'`, the sniff flags it as unsupported and auto-fixes to `true|false`.

Any other value passed is flagged as unsupported, and a list of supported values is provided.

### Open questions

I'm unsure about a few choices that I made in the XML documentation and could use some feedback on the following:

- I opted to include a link to a document explaining how to decide which value to use for the `$autoload` parameter. Since I could not find something wp.org, I ended up linking to a post written by a core committer in their blog. I'm not sure if this is a problem or not.
- Maybe combining the `<standard>` blocks for invalid, deprecated, and internal-only values in a single block would be better? I created a block for each, but there is some repetition in the block description and the valid examples.
- I added a `<standard>` block without code examples at the top with a general description of what the sniff does before creating blocks for each of the warnings generated by the sniff.


jrfnl (changes requested):
@rodrigoprimo Thank you for working on this.

I've so far only done a first pass reading through it (without running the sniff/docs etc), but seeing the amount of remarks I've already left, I figured I should give you the chance to iterate on the current state before continuing my review.

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,135 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Option Autoload"
+    >
+    <standard>
+    <![CDATA[
+    When using `add_option()`, `update_option()`, `wp_set_options_autoload()`,
+    `wp_set_option_autoload()`, or `wp_set_option_autoload_values()`, it is recommended to
+    explicitly set the autoload value to ensure predictable behavior. This value determines whether
```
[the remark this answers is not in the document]
rodrigoprimo: I tried not to use `parameter` as for `wp_set_option_autoload_values()` autoload is not a parameter. But I agree that value is probably worst. I will accept your suggestion, and I will let you know if I can think of a better approach.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,416 @@
...
+ *
+ * @link https://github.com/WordPress/WordPress-Coding-Standards/issues/2473
+ *
+ * @since 3.2.0
+ */
+final class OptionAutoloadSniff extends AbstractFunctionParameterSniff {
+
+	/**
+	 * The phrase to use for the metric recorded by this sniff.
+	 *
```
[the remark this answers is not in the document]
rodrigoprimo: Done

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,416 @@
...
+
+	/**
+	 * List of valid values for the `$autoload` parameter in the add_option() and update_option()
+	 * functions.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @var string[]
+	 */
+	protected static $valid_values_add_and_update = array( 'true', 'false', 'null' );
```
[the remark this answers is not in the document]
rodrigoprimo: I changed all the `protected` properties and methods to `private`, with the exception of `$group_name` and `$target_functions`, as their access level should be the same or weaker than the corresponding properties in the parent class.

I also changed the `static` properties to regular properties.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,416 @@
...
+	 * @param int    $stackPtr      The position of the current token in the stack.
+	 * @param string $function_name The name of the function being checked.
+	 *
+	 * @return void
+	 */
+	protected function maybe_display_missing_autoload_warning( $stackPtr, $function_name ) {
+		$this->phpcsFile->recordMetric( $stackPtr, self::METRIC_NAME, 'param missing' );
+
+		// Only display a warning for the functions in which the `$autoload` parameter is optional.
+		if ( in_array( $function_name, self::$autoload_is_optional, true ) ) {
```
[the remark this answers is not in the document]
rodrigoprimo: Good point. I believe we discussed this already in the past and I failed to take it into account while creating this sniff. I will think a bit more about how to implement this in the other places and get back to you.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,416 @@
...
+	 *
+	 * @since 3.2.0
+	 *
+	 * @param int    $stackPtr      The position of the current token in the stack.
+	 * @param string $function_name The name of the function being checked.
+	 *
+	 * @return void
+	 */
+	protected function maybe_display_missing_autoload_warning( $stackPtr, $function_name ) {
+		$this->phpcsFile->recordMetric( $stackPtr, self::METRIC_NAME, 'param missing' );
```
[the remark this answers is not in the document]
rodrigoprimo: I'm not sure about it as well.

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,135 @@
...
+    the option is automatically loaded on every page load, which can impact site performance.
+
+    Check https://felix-arntz.me/blog/autoloading-wordpress-options-efficiently-and-responsibly/ for
+    more information on when to set autoload to `true` or `false`.
```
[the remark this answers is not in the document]
rodrigoprimo: Removed

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,135 @@
...
+    )
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Only `true` and `false` are valid values for the autoload parameter, with one exception: `null`
+    is also valid for `add_option()` and `update_option()`.
+    ]]>
```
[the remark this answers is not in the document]
rodrigoprimo: In a separate commit, I tried again to structure the XML documentation. I kept the "introductory" standard block and the block for the presence of the autoload parameter. The other three blocks were combined into a single one that describes the autoload parameter's valid and invalid values. Any input on this new structure is welcome. I'm unsure if I should combine everything into a single `<standard>` and if I'm insisting too much on separating the parameter is missing warning from the other warnings.

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,135 @@
...
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    It is not recommended to use internal-only values ('auto', 'auto-on', 'auto-off', 'on', 'off')
+    in plugin or theme code. Instead, use `true`|`false`|`null` if using `add_option()` or
+    `update_option()` or `true`|`false` if using `wp_set_option_autoload_values()`,
+    `wp_set_option_autoload()`, or `wp_set_options_autoload()`.
```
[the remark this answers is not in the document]
rodrigoprimo: See https://github.com/WordPress/WordPress-Coding-Standards/pull/2520#discussion_r1975996912

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,135 @@
...
+    ''
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Using 'yes' or 'no' is not recommended. Those values are deprecated. Instead, use
+    `true`|`false`|`null` if using `add_option()` or `update_option()` or `true`|`false` if using
+    `wp_set_option_autoload_values()`, `wp_set_option_autoload()`, or `wp_set_options_autoload()`.
```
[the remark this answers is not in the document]
rodrigoprimo: See https://github.com/WordPress/WordPress-Coding-Standards/pull/2520#discussion_r1975996912

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,135 @@
...
+    Check https://felix-arntz.me/blog/autoloading-wordpress-options-efficiently-and-responsibly/ for
+    more information on when to set autoload to `true` or `false`.
+    ]]>
+    </standard>
+    <standard>
+    <![CDATA[
+    Even though the autoload parameter is optional for `add_option()` and `update_option()`, it is
+    recommended to always explicitly set it.
+    ]]>
+    </standard>
```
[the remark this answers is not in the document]
rodrigoprimo: See https://github.com/WordPress/WordPress-Coding-Standards/pull/2520#discussion_r1975996912

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,416 @@
...
+	 * @since 3.2.0
+	 *
+	 * @param array  $autoload_info Information about the autoload value (start and end tokens and
+	 *                              the clean and raw value).
+	 * @param string $function_name The name of the function being checked.
+	 *
+	 * @return void
+	 */
+	protected function check_autoload_value( array $autoload_info, $function_name ) {
+		// Find the first and second param non-empty tokens (the second token might not exist).
```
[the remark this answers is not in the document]
rodrigoprimo: Removed

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,416 @@
...
+			&& 'null' !== strtolower( $this->tokens[ $param_first_token ]['content'] )
+		) {
+			// Bail early if the first non-empty token in the parameter is T_VARIABLE or T_STRING as
+			// this means it is not possible to determine the value.
+			$this->phpcsFile->recordMetric( $param_first_token, self::METRIC_NAME, 'undetermined value' );
+			return;
+		}
+
+		if ( $param_second_token
+			&& ! in_array( $this->tokens[ $param_first_token ]['code'], array( T_ARRAY, T_OPEN_SHORT_ARRAY ), true )
```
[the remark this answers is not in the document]
rodrigoprimo: Done

WordPress/Tests/WP/OptionAutoloadUnitTest.inc.fixed
```diff
@@ -0,0 +1,214 @@
...
+ */
+$ok = \add_option( 'my_option', 'my_value', '', true );
+$ok = add_option( 'my_option', array('some', 'big', 'array'), '', false );
+$ok = add_option( 'my_option', new stdClass(), '', null );
+$ok = add_option( 'my_option', autoload: FALSE, value: 'my_value' );
+$ok = add_option( 'my_option', autoload: false );
+$ok = add_option( autoload: false, deprecated: '', value: 'my_value', option: 'my_option' );
+$ok = \update_option( 'my_option', 'new_value', \true );
+$ok = update_option( 'my_option', 'new_value', \false );
+$ok = update_option( 'my_option', 'new_value', \NULL );
```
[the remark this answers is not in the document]
rodrigoprimo: While scanning the WP.org plugins codebase, I found code samples passing a "namespaced" boolean as the value of the `$autoload` parameter of some of the option functions. As far as I can test, PHP treats `\true` as `true` (and so on):

https://3v4l.org/KmT8M

So, I thought it was worth adding those test cases and also supporting it in the sniff. What do you think?

WordPress/Tests/WP/OptionAutoloadUnitTest.inc.fixed
```diff
@@ -0,0 +1,214 @@
...
+$ok = update_option( 'my_option', autoload: TrUe, value: 'new_value' );
+$ok = update_option( autoload: false, value: 'my_value', option: 'my_option' );
+$ok = update_option( $option_name, get_option( $option_name ), false );
+$ok = wp_set_options_autoload( array( 'option1', 'option2' ), \false );
+$ok = wp_set_options_autoload( array( 'option1', 'option2' ), \true );
```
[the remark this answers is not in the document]
rodrigoprimo: See https://github.com/WordPress/WordPress-Coding-Standards/pull/2520#discussion_r1975999191

WordPress/Tests/WP/OptionAutoloadUnitTest.inc.fixed
```diff
@@ -0,0 +1,214 @@
...
+        'option6' => yes,
+        'option7' => \my_function(),
+        'option8' => \SOME_CONSTANT,
+        'option9' => \YES,
+    )
+);
+
+/*
+ * These are ignored because it is not easy to determine the option names and the autoload values or
+ * the `$options` parameter is empty.
```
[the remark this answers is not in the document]
rodrigoprimo: Ok

WordPress/Tests/WP/OptionAutoloadUnitTest.inc.fixed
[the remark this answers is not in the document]
rodrigoprimo: I will create more tests and push a new commit in the next few days.

rodrigoprimo:
Thanks for your review, @dingo-d and @jrfnl! I believe I addressed all of your comments except two that are still pending, and I will need a bit more time (https://github.com/WordPress/WordPress-Coding-Standards/pull/2520#discussion_r1974686099 and https://github.com/WordPress/WordPress-Coding-Standards/pull/2520#discussion_r1974641122).

> The only problem is that that is not official documentation. And if their site goes down, you have a dead link in the documentation.

That makes sense. As @jrfnl suggested, I removed the paragraph that included this link.

> Does the validation passes against the [phpcsdocs.xsd](https://github.com/PHPCSStandards/PHPCSDevTools/blob/develop/DocsXsd/phpcsdocs.xsd)?

I believe so. This is how I tested:

```
$ xmllint --schema phpcsdocs.xsd WordPress/Docs/WP/OptionAutoloadStandard.xml --noout 
WordPress/Docs/WP/OptionAutoloadStandard.xml validates
```

Let me know if I'm missing something.

WordPress/Tests/WP/OptionAutoloadUnitTest.inc.fixed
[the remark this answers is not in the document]
dingo-d: Shouldn't these be flagged by the Generic.PHP.LowerCaseConstant sniff?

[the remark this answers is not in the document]
jrfnl: > So, I thought it was worth adding those test cases and also supporting it in the sniff. What do you think?

@rodrigoprimo Fair point. Agreed.

> Shouldn't these be flagged by the Generic.PHP.LowerCaseConstant sniff?

@dingo-d The `NULL` will be flagged by that sniff, but the `\` prefix will not be flagged as that sniff doesn't look for that.
Might not be a bad idea to create a new sniff for that (PHPCSExtra is probably the best place for that). Happy to see an issue about this in that repo.

Aside from that, these are valid test cases for _this_ sniff, independently of whether another sniff would flag this as sniffs should function independently of each other.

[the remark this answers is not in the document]
jrfnl: Looking at this now... there looks to be a tokenizer issue with `\true` etc when used on PHP 8.0 or higher.... hmm... I think that needs a fix in PHPCS.
I haven't checked how this sniff handles this....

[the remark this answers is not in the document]
jrfnl: Oh sod it, wrote a sniff now to forbid this: https://github.com/PHPCSStandards/PHPCSExtra/pull/327
Once merged & released, we may want to consider adding this new sniff to WPCS.

As for the PHPCS tokenizer issue, I've looked into that and decided against fixing it as things will change again in PHPCS 4.0, so fixing it now will only make things more difficult for PHPCS 4.0.

WordPress/Tests/WP/OptionAutoloadUnitTest.inc.fixed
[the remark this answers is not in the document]
jrfnl: I don't think you need _more_ tests. Probably just updating some of the existing tests to include those variations would be enough.

WordPress/Docs/WP/OptionAutoloadStandard.xml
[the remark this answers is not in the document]
jrfnl: I do think the "missing parameter" and the "invalid parameter value" issues are two separate issues, but I'm not convinced the changed structure of the docs makes this clear enough.

If I look at the latest version, my recommendation would be:
* Move the last `code_comparison` to be directly under the first `standard` block (and drop the third `standard` block which was originally associated with that `code_comparison`.

WordPress/Tests/WP/OptionAutoloadUnitTest.inc.fixed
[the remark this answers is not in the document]
rodrigoprimo: I updated some of the tests to add more variation in whitespace, comments/annotations in unexpected places, and function names using different cases.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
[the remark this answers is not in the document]
rodrigoprimo: > Please change the $autoload_is_optional array to 'name' => true format and check using isset() for a small performance improvement.
> ☝🏻 This second point should be fixed in quite a few places in this sniff.

I implemented this change in all the places that seemed relevant to me. For `$valid_values_add_and_update` and `$valid_values_other_functions`, I used the same format as what is used in the `\PHPCSUtils\Tokens\Collections` properties and repeated the value in the index of each element of the array. I opted to do this to avoid calling `array_keys()` when passing `$valid_values` to `array_map()`:

https://github.com/WordPress/WordPress-Coding-Standards/blob/fc98b8eb338a18e71b8bab9fd26674bbbaa9733f/WordPress/Sniffs/WP/OptionAutoloadSniff.php#L404-L409

rodrigoprimo:
@jrfnl, I addressed the two points that were pending. Could you please take another look at this PR when you get a chance?

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,102 @@
...
+    - For `add_option()` and `update_option()`: `true`, `false`, or `null`.
+    - For `wp_set_option_autoload_values()`, `wp_set_option_autoload()`, and
+    `wp_set_options_autoload()`: `true` or `false`.
+
+    Using 'yes' or 'no' is deprecated since WordPress 6.6.0. Using 'auto', 'auto-on', 'auto-off',
+    'on', or 'off' in plugin/theme code is strongly discouraged as those values are meant for
+    internal-use only.
+
+    Any other values, including `null` for functions other than `add_option()` and `update_option()`
+    are invalid.
```
rodrigoprimo:
```suggestion
    Any other values, including `null` for functions other than `add_option()` and `update_option()`,
    are invalid.
```

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,102 @@
...
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Option Autoload"
+    >
+    <standard>
+    <![CDATA[
+    When using `add_option()`, `update_option()`, `wp_set_options_autoload()`,
+    `wp_set_option_autoload()`, or `wp_set_option_autoload_values()`, it is recommended to
+    explicitly set the autoload parameter to ensure predictable behavior. This parameter determines
+    whether the option is automatically loaded on every page load, which can impact site performance.
```
jrfnl:
```suggestion
    explicitly set the autoload parameter to ensure predictable behavior.
    The autoload flag determines whether the option is automatically loaded on every page load,
    which can impact site performance.
```

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,102 @@
...
+    'on', or 'off' in plugin/theme code is strongly discouraged as those values are meant for
+    internal-use only.
+
+    Any other values, including `null` for functions other than `add_option()` and `update_option()`
+    are invalid.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using a boolean value or `null`">
+        <![CDATA[
```
jrfnl, line 59: Should the valid example have an example from both function groups ?

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,102 @@
...
+    <![CDATA[
+    The following values should be used when setting the autoload parameter depending on the function:
+
+    - For `add_option()` and `update_option()`: `true`, `false`, or `null`.
+    - For `wp_set_option_autoload_values()`, `wp_set_option_autoload()`, and
+    `wp_set_options_autoload()`: `true` or `false`.
+
+    Using 'yes' or 'no' is deprecated since WordPress 6.6.0. Using 'auto', 'auto-on', 'auto-off',
+    'on', or 'off' in plugin/theme code is strongly discouraged as those values are meant for
+    internal-use only.
```
jrfnl: I wonder if we're not giving people ideas by explicitly listing the internal values.... ?
Maybe the second sentence should be removed ? Or maybe just mention something along the lines of _"Plugin/theme code should also not use values for the parameter which are for WP-Core internal use only."_

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,102 @@
...
+    array('option1', 'option2'),
+    <em>1</em>
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Even though the autoload parameter is optional for `add_option()` and `update_option()`, it is
+    recommended to always explicitly set it.
```
jrfnl: Maybe ? Isn't this what the sniff is mostly about ?

```suggestion
    strongly recommended to always explicitly set it.
```

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,102 @@
...
+);
+
+wp_set_options_autoload(
+    array('option1', 'option2'),
+    <em>1</em>
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
```
jrfnl: Would it make sense to move this `<standard>` block up ? This is about the parameter being missing, while the first block also talks about that, so it kind of feels like having this above the block about the values reads more fluently.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+use PHPCSUtils\Utils\PassedParameters;
+use PHPCSUtils\Utils\TextStrings;
+use WordPressCS\WordPress\AbstractFunctionParameterSniff;
+
+/**
+ * Warns when calls to add_option(), update_option(), wp_set_options_autoload(),
+ * wp_set_option_autoload(), wp_set_option_autoload_values() are missing the `$autoload` param or
+ * contain an invalid, internal or deprecated value.
+ *
+ * @link https://github.com/WordPress/WordPress-Coding-Standards/issues/2473
```
jrfnl, line 24: Would it make sense to add that link to the external blog post about the principle of this here ?

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+		$param_second_token = $this->phpcsFile->findNext(
+			Tokens::$emptyTokens,
+			$param_first_token + 1,
+			( $autoload_info['end'] + 1 ),
+			true
+		);
+
+		$normalized_value = strtolower( $autoload_info['clean'] );
+
+		if ( \T_NS_SEPARATOR === $this->tokens[ $param_first_token ]['code'] && $param_second_token ) {
```
rodrigoprimo: Note: this needs to be changed when adding support for PHPCS 4.0. Not something that should be addressed in this PR.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+	 * The phrase to use for the metric recorded by this sniff.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @var string
+	 */
+	const METRIC_NAME = 'Value of the `$autoload` parameter in the option functions';
+
+	/**
+	 * Valid values for the `$autoload` parameter in the add_option() and update_option() functions.
```
jrfnl: Would it make sense to also add backticks around the function calls (for consistency) ? (here and possibly in other docblocks)

```suggestion
	 * Valid values for the `$autoload` parameter in the `add_option()` and `update_option()` functions.
```

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+
+	/**
+	 * Valid values for the `$autoload` parameter in the wp_set_options_autoload(),
+	 * wp_set_option_autoload(), and wp_set_option_autoload_values() functions.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @var array<string, string>
+	 */
+	private $valid_values_other_functions = array(
```
jrfnl: Other functions ? How about this ?
```suggestion
	private $valid_values_wp_set_option_functions = array(
```

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+		'true'  => 'true',
+		'false' => 'false',
+	);
+
+	/**
+	 * Deprecated values for the `$autoload` parameter.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @var array<string, true>
```
jrfnl: When the array has been set up for optimization with `isset()` and the array value has no meaning, we normally try to have a comment along those lines in the `@var` tag to prevent someone scratching their head over what the values mean when looking at the sniff in the future. (here and in a few other places)
```suggestion
	 * @var array<string, true> Key is the deprecated value, the array value is unused/irrelevant.
```

WordPress/Sniffs/WP/OptionAutoloadSniff.php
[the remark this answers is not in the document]
jrfnl: What about moving it into the condition ?

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+		}
+	}
+
+	/**
+	 * Check the autoload value for possible violations.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @param array  $autoload_info Information about the autoload value (start and end tokens and
+	 *                              the clean value).
```
jrfnl:
```suggestion
	 * @param array  $autoload_info Information about the autoload value (start and end tokens,
	 *                              the clean value and potentially the "raw" value - which isn't used).
```

Also, should the `array` be made more specific ? Maybe `array<string, int|string>` ?

WordPress/Sniffs/WP/OptionAutoloadSniff.php
jrfnl: What are you testing with `$param_second_token` ?

Maybe there should also be a test with `\` being the only content of the parameter ? Not that that would be valid code, but that's the case which you are testing for with this condition. Or am I missing something ?

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+	}
+
+	/**
+	 * Handle the `wp_set_option_autoload_values()` function. It requires special treatment as it
+	 * takes an array of option names and autoload values instead of the autoload value of a single
+	 * option as a separate parameter.
+	 *
+	 * @since 3.2.0
+	 *
+	 * @param array $options_param Options parameter information.
```
jrfnl: Please describe the array structure in the param notation.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+	 * @param int   $stackPtr      The position of the current token in the stack.
+	 *
+	 * @return void
+	 */
+	private function handle_wp_set_option_autoload_values( array $options_param, $stackPtr ) {
+		$array_token = $this->phpcsFile->findNext(
+			Tokens::$emptyTokens,
+			$options_param['start'],
+			$options_param['end'],
```
jrfnl:
```suggestion
			( $options_param['end'] + 1 ),
```

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+			true
+		);
+
+		if ( false === isset( Collections::arrayOpenTokensBC()[ $this->tokens[ $array_token ]['code'] ] ) ) {
```
jrfnl:
```suggestion
		if ( false === $array_token || false === isset( Collections::arrayOpenTokensBC()[ $this->tokens[ $array_token ]['code'] ] ) ) {
```

WordPress/Tests/WP/OptionAutoloadUnitTest.inc
```diff
@@ -0,0 +1,229 @@
...
+        'option8' => \SOME_CONSTANT,
+        'option9' => \YES,
+    )
+);
+
+/*
+ * These are ignored because the option names and the autoload values can't be reliably determined
+ * or the `$options` parameter is empty.
+ */
+$ignored = wp_set_option_autoload_values( $options );
```
jrfnl, line 94: You need a second test here without the whitespace to test the condition on line 247 in the sniff correctly.
```
$ignored = wp_set_option_autoload_values($options);
```

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+			$valid_values = $this->valid_values_other_functions;
+		}
+
+		if ( isset( $valid_values[ $normalized_value ] ) ) {
+			$this->phpcsFile->recordMetric( $param_first_token, self::METRIC_NAME, $normalized_value );
+			return;
+		}
+
+		if ( in_array( $this->tokens[ $param_first_token ]['code'], array( \T_VARIABLE, \T_STRING ), true )
+			&& 'null' !== strtolower( $this->tokens[ $param_first_token ]['content'] )
```
jrfnl, line 361: While in a call, we looked into why `null` was special cased and not `true` and `false` as this condition was mystifying me.

Conclusion (after spending way more time on this that we should have wanted to): `true` and `false` are always valid, so will already be filtered out as valid via the `isset( $valid_values[ $normalized_value ] )` condition above.
`null` is not always valid, so may not have been filtered out yet, so when a FQN `\null` is used, the `null` may be tokenized as `T_STRING` on PHP 8.0+, which is why it needs special casing here.

May I suggest documenting this via an inline comment ?

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+		) {
+			// Bail early if the first non-empty token in the parameter is T_VARIABLE or T_STRING as
+			// this means it is not possible to determine the value.
+			$this->phpcsFile->recordMetric( $param_first_token, self::METRIC_NAME, 'undetermined value' );
+			return;
+		}
+
+		if ( $param_second_token
```
jrfnl: Again, what are you testing with `$param_second_token` ?

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+			&& false === isset( Collections::arrayOpenTokensBC()[ $this->tokens[ $param_first_token ]['code'] ] )
+		) {
+			// Bail early if the parameter has two or more non-empty tokens and the second token is
```
jrfnl:
```suggestion
			// Bail early if the parameter has two or more non-empty tokens and the first token is
```

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+			// not an array opener as this means an undetermined param value or a value that is not
+			// easy to determine.
+			$this->phpcsFile->recordMetric( $param_first_token, self::METRIC_NAME, 'undetermined value' );
+			return;
+		}
+
+		$autoload_value = TextStrings::stripQuotes( $autoload_info['clean'] );
+
+		$known_discouraged_values = array_merge( $this->deprecated_values, $this->internal_values_non_fixable, $this->internal_values_fixable );
```
jrfnl: Any particular reason to use `array_merge()` instead of array union (`+`) ? The arrays being merged are perfectly suitable for use with array union.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+
+		$this->phpcsFile->recordMetric( $param_first_token, self::METRIC_NAME, $metric_value );
+
+		if ( isset( $this->deprecated_values[ $autoload_value ] ) ) {
+			$message    = 'The use of `%s` as the value of the `$autoload` parameter is deprecated. Use `%s` instead.';
+			$error_code = 'Deprecated';
+			$data       = array( $autoload_info['clean'], $this->fixable_values[ $autoload_value ] );
+		} elseif ( isset( $this->internal_values_fixable[ $autoload_value ] ) ) {
+			$message    = 'The use of `%s` as the value of the `$autoload` parameter is discouraged. Use `%s` instead.';
+			$error_code = 'InternalUseOnly';
```
jrfnl: I'm not sure I like the same error code being used for two - closely related, but still different - errors here.... (line 397 + 401)

Maybe the second error code (line 401) could be something like `InternalUseOnlyAuto` (as they values which will be flagged all start with "auto") ?

And maybe this one should then be `InternalUseOnlyOnOff` ? (to keep things consistent)

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+			$data       = array( $autoload_info['clean'] );
+		} else {
+			$valid_values        = array_map(
+				function ( $value ) {
+					return '`' . $value . '`';
+				},
+				$valid_values
+			);
+			$valid_values_string = implode( ', ', $valid_values );
```
jrfnl: For your consideration:

```suggestion
			$valid_values_string = '`' . implode( '`, `', $valid_values ) . '`';
```

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,439 @@
...
+
+		if ( isset( $this->fixable_values[ $autoload_value ] ) ) {
+			$fix = $this->phpcsFile->addFixableWarning(
+				$message,
+				$param_first_token,
+				$error_code,
+				$data
+			);
+
+			if ( $fix ) {
```
jrfnl:
```suggestion
			if ( true === $fix ) {
```

jrfnl (changes requested):
Hi @rodrigoprimo thanks for making those initial updates.

I've now done a proper code review (with the exception of the tests). Hope this helps!

WordPress/Docs/WP/OptionAutoloadStandard.xml
rodrigoprimo: I opted to add the sentence that you suggested to mention that plugin/theme code should not use internal values, while not mentioning the values directly to avoid giving people ideas.

WordPress/Docs/WP/OptionAutoloadStandard.xml
rodrigoprimo: Done

WordPress/Sniffs/WP/OptionAutoloadSniff.php
rodrigoprimo: Done

WordPress/Sniffs/WP/OptionAutoloadSniff.php
rodrigoprimo: Done

WordPress/Sniffs/WP/OptionAutoloadSniff.php
rodrigoprimo: Done

WordPress/Sniffs/WP/OptionAutoloadSniff.php
rodrigoprimo: Done

WordPress/Sniffs/WP/OptionAutoloadSniff.php
rodrigoprimo: Done for this @param tag and the others where the param type is array and the structure was not described.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
rodrigoprimo: Done, I also added a test for this.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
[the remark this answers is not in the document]
rodrigoprimo: Done

rodrigoprimo:
Thanks for taking another look at this PR, @jrfnl. I started replying to your feedback and implementing the changes, but I was unable to finish everything today. I won't be able to look into this next week. The following week, I will address the comments that I was unable to address today.

rodrigoprimo:
Rebased without changes to fix an unrelated failing test.

WordPress/Docs/WP/OptionAutoloadStandard.xml
rodrigoprimo: I'm not sure I understand your suggestion. As far as I can see, the first block does not mention the parameter being missing for `add_option()` and `update_option()`, so I'm not sure why it would read better if this `<standard>` block were moved up.

My reasoning for ordering the blocks the way they are was:

- First block: introduces what the sniff is about and what the autoload parameter is.
- Second block: explains the rule about the parameter value that applies to all the functions that this sniff cares about.
- Third block: explains the rule about the parameter missing that only applies to `add_option()` and `update_option()`.

I think I prefer the current order, but I'm happy to reconsider if I'm missing something.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
rodrigoprimo: Done

WordPress/Sniffs/WP/OptionAutoloadSniff.php
rodrigoprimo: Done

WordPress/Sniffs/WP/OptionAutoloadSniff.php
rodrigoprimo: I updated the if condition to explicitly test that `$param_second_token` is an integer and added a test with just `\` as the content of the parameter.

WordPress/Tests/WP/OptionAutoloadUnitTest.inc
rodrigoprimo: I went ahead and added this test. To be honest, it is not clear to me why it is necessary, and this test is testing that was not already being tested.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
rodrigoprimo: I expanded the existing comment to include information about why `null` requires special treatment and `false` and `true` don't.

WordPress/Sniffs/WP/OptionAutoloadSniff.php
rodrigoprimo: I made the test more explicit by checking if `$param_second_token` is an integer.

rodrigoprimo:
> What about moving it into the condition ?

Regarding [the comment quoted above](https://github.com/WordPress/WordPress-Coding-Standards/pull/2520#discussion_r2185578684), I'm not sure why I cannot reply directly to it, but it was addressed, and now the `param missing` metric will only be recorded for `add_option()` and `update_option()`.

rodrigoprimo:
This will need to be updated to work on PHPCS 4.0 before it is reviewed again.

rodrigoprimo:
Force pushed to rebase the branch on the latest `develop` without other changes.

rodrigoprimo:
Just documenting here that I pushed two new commits to this PR:

1. Bump the version number in the `@since` tags
2. Add tests for namespaced names in preparation for PHPCS 4.0

WordPress-Extra/ruleset.xml
```diff
@@ -193,6 +193,13 @@
 		 https://github.com/WordPress/WordPress-Coding-Standards/issues/2459 -->
 	<rule ref="WordPress.WP.GetMetaSingle"/>
 
+	<!-- Flags calls to add_option(), update_option(), wp_set_options_autoload(),
+		 wp_set_option_autoload(), wp_set_option_autoload_values() functions when the
+		 `$autoload` parameter is missing or contain an invalid, internal-only or
+		 deprecated value.
```
GaryJones, line 199:
```suggestion
	<!-- Flags calls to add_option(), update_option(), wp_set_options_autoload(),
		 wp_set_option_autoload(), wp_set_option_autoload_values() functions when the
		 `$autoload` parameter is missing or contains an invalid, internal-only, or
		 deprecated value.
```

WordPress/Docs/WP/OptionAutoloadStandard.xml
```diff
@@ -0,0 +1,100 @@
...
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Option Autoload"
+    >
+    <standard>
+    <![CDATA[
+    When using `add_option()`, `update_option()`, `wp_set_options_autoload()`,
+    `wp_set_option_autoload()`, or `wp_set_option_autoload_values()`, it is recommended to
+    explicitly set the autoload parameter to ensure predictable behavior.
+    The autoload flag determines whether the option is automatically loaded on every page load,
```
GaryJones:
```suggestion
    The autoload value determines whether the option is automatically loaded on every page load,
```

WordPress/Sniffs/WP/OptionAutoloadSniff.php
```diff
@@ -0,0 +1,442 @@
...
+	 *
+	 * @var array<string, true> The key is the name of the deprecated value. The array value is irrelevant.
+	 */
+	private $deprecated_values = array(
+		'yes' => true,
+		'no'  => true,
+	);
+
+	/**
+	 * Internal-use only values for `$autoload` that cannot be fixed automatically by the sniff.
```
GaryJones:
```suggestion
	 * Internal-use only values for `$autoload` that will not be fixed automatically by the sniff.
```

GaryJones (changes requested):
Nothing jumps out at me besides a couple of string fixes.

rodrigoprimo:
Thanks for taking the time to review this PR, @GaryJones. I have accepted the three suggestions you made.

GaryJones (approved):

mukeshpanchal27:
@rodrigoprimo The PR has already been approved. Is there a plan to get it merged soon?

PR #2525, opened 2025-03-24
Add do_blocks to the $autoEscapedFunctions list

### LINKED ISSUE #2524: Add `do_blocks` to the `$autoEscapedFunctions` list

It's not a good habit to have to keep adding `phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped` before my `echo do_blocks( ... );`

I'd like `do_blocks` added to [the `$autoEscapedFunctions` list](https://github.com/WordPress/WordPress-Coding-Standards/blob/38168bc542ec3565c04f2a88498a2024cdfd9ba1/WordPress/Helpers/EscapingFunctionsTrait.php#L100-L171), just like `do_shortcode` is.

- [x] I intend to create a pull request to implement this feature.

dossy:
Add `do_blocks` to the `$autoEscapedFunctions` list.

Closes #2524.

### DIFF
--- a/WordPress/Helpers/EscapingFunctionsTrait.php
+++ b/WordPress/Helpers/EscapingFunctionsTrait.php
@@ -117,6 +117,7 @@ trait EscapingFunctionsTrait {
 		'comment_class'           => true,
 		'count'                   => true,
 		'disabled'                => true,
+		'do_blocks'               => true,
 		'do_shortcode'            => true,
 		'do_shortcode_tag'        => true,
 		'get_archives_link'       => true,


jrfnl (commented):
The [`do_blocks()`](https://developer.wordpress.org/reference/functions/do_blocks/) function doesn't produce output, but returns instead, so it doesn't belong in the `$autoEscapedFunctions` list.

I also don't see the function escaping the output, so as far as I can see, it should also not be considered an escaping function. It's a re-rendering function, not an escaping function.

dossy:
> The [`do_blocks()`](https://developer.wordpress.org/reference/functions/do_blocks/) function doesn't produce output, but returns instead, so it doesn't belong in the `$autoEscapedFunctions` list.
> 
> I also don't see the function escaping the output, so as far as I can see, it should also not be considered an escaping function. It's a re-rendering function, not an escaping function.

The exact same things can be said about [`do_shortcode()`](https://developer.wordpress.org/reference/functions/do_shortcode/), and yet [it is in the `$autoEscapedFunctions` list](https://github.com/WordPress/WordPress-Coding-Standards/blob/38168bc542ec3565c04f2a88498a2024cdfd9ba1/WordPress/Helpers/EscapingFunctionsTrait.php#L120).

`do_blocks()` should be added for the exact same reasons `do_shortcode()` was.

PR #2529, opened 2025-05-19
Rulesets: update schema URL

jrfnl:
PHPCS now offers permalinks for the schema.

Refs:
* PHPCSStandards/PHP_CodeSniffer 1094
* https://github.com/PHPCSStandards/schema.phpcodesniffer.com

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="WordPress Coding Standards" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/PHPCSStandards/PHP_CodeSniffer/master/phpcs.xsd">
+<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="WordPress Coding Standards" xsi:noNamespaceSchemaLocation="https://schema.phpcodesniffer.com/phpcs.xsd">
 
 	<description>The Coding standard for the WordPress Coding Standards itself.</description>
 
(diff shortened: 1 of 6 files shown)


dingo-d (approved):

GaryJones:
Should the schema URL be versioned?

jrfnl:
> Should the schema URL be versioned?

That depends. There are permalinks available for each (minor) PHPCS version which includes the `phpcs.xsd` file, i.e. PHPCS 3.2 and higher. And then there is the permalink for the "current"/latest version, which is what I've used now.

Now, even though the minors each have there own permalink, the `phpcs.xsd` file hasn't changed that often. If you look at [the file history](https://github.com/PHPCSStandards/PHP_CodeSniffer/commits/master/phpcs.xsd), you will see that the file:
* Was introduced in PHPCS 3.2.0.
* Was updated for the new array property syntax in 3.3.2 (even though the array syntax was introduced in 3.3.0).
* Was updated for extending array properties in 3.4.0.
* Was fixed for an oversight (missing support for `phpcs-only` and `phpcbf-only`) in 3.5.0.
* ... and hasn't changed since.

Now, we could use the permalink for the `3.9` version to align with the current minimum supported PHPCS version of WPCS, but, as per the above, there is no difference between the "current" (3.13) version and the `3.9` version.
And if we'd do that, it would also require for us to keep that version number updated whenever we update the minimum PHPCS version, which is error-prone.

In my opinion, it only makes sense to use a versioned permalink if PHPCS adds support for a new ruleset feature and WPCS doesn't use this ruleset feature (yet) as the minimum supported PHPCS version is too low.
In that case, we  should set the permalink to a versioned one _before_ the upstream change - i.e. to the version which aligns with our minimum supported PHPCS version.

However, that doesn't currently apply and the `phpcs.xsd` file doesn't change that often (unlike the PHPUnit one), which is why I've pulled this with the permalink to the "latest" `phpcs.xsd` file.

@GaryJones Does that answer your question ?

GaryJones:
> In that case, we should set the permalink to a versioned one before the upstream change - i.e. to the version which aligns with our minimum supported PHPCS version.

That's the manual task I'd be minorly concerned about (though, I know you will remember!).

GaryJones (approved):

jrfnl:
> That's the manual task I'd be minorly concerned about (though, I know you will remember!).

And if I don't, I'm sure you'll remind me 😜

PR #2532, opened 2025-06-14
Composer: raise the minimum supported PHPCSUtils version to 1.1.0 + fix CI

jrfnl:
This PR is just to unblock WPCS/fix the failing test run.

Fixing the failure test run without these version bumps would mean having to polyfill support for PHP 8.4 properties in interfaces, which is not something we should want to do (that's why we use PHPCSUtils, which does the polyfill for PHPCS 3.x as PHPCS natively will only start supporting this in PHPCS 4.0).

This PR does **NOT**:
* Start using the new features from PHPCSUtils 1.1.0.
* Add support for all the new PHP syntaxes for which support was added to PHPCS + PHPCSUtils since PHPCS 3.9.0.

---

### Composer: raise the minimum supported PHPCSUtils version to 1.1.0 

... to benefit from new functionality.

This also automatically raises the minimum PHPCS version to `3.13.0` for PHP 8.2 DNF type support and tokenizer support for PHP 8.4 `final` properties and asymmetric visibility.

Includes syncing the PHPCSExtra version to a version which also has a minimum PHPCSUtils version of 1.1.0.

Ref:
* https://github.com/PHPCSStandards/PHPCSUtils/releases/tag/1.1.0
* https://github.com/PHPCSStandards/PHPCSExtra/releases/tag/1.4.0

### NamingConventions/ValidVariableName: allow for PHP 8.4 properties in interfaces

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -39,9 +39,9 @@ When you introduce new `public` sniff properties, or your sniff extends a class 
 
 ## Pre-requisites
 * WordPress-Coding-Standards
-* PHP_CodeSniffer 3.9.0 or higher
-* PHPCSUtils 1.0.10 or higher
-* PHPCSExtra 1.2.1 or higher
+* PHP_CodeSniffer 3.13.0 or higher
+* PHPCSUtils 1.1.0 or higher
+* PHPCSExtra 1.4.0 or higher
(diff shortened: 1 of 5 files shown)


westonruter (approved):
LGTM

dingo-d (approved):

PR #2540, opened 2025-07-14
Fix the class description for ConstantsHelper and ListHelper

rodrigoprimo:
While working on an unrelated task, I noticed that the class description in the docblock for the `ConstantsHelper` and `ListHelper` classes was incorrect. The previous description contained information about the `ContextHelper` class. This PR fixes that.

### DIFF
--- a/WordPress/Helpers/ConstantsHelper.php
+++ b/WordPress/Helpers/ConstantsHelper.php
@@ -16,7 +16,7 @@ use PHPCSUtils\Utils\Scopes;
 use WordPressCS\WordPress\Helpers\ContextHelper;
 
 /**
- * Helper utilities for checking the context in which a token is used.
+ * Helper utilities for identifying constants in PHP code.
  *
  * ---------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


WordPress/Helpers/ConstantsHelper.php
```diff
@@ -16,7 +16,7 @@
 use WordPressCS\WordPress\Helpers\ContextHelper;
 
 /**
- * Helper utilities for checking the context in which a token is used.
+ * Helper utilities for identifying constants in PHP code.
```
jrfnl:
```suggestion
 * Helper utilities for identifying the use of global constants in PHP code.
```

jrfnl (commented):
@rodrigoprimo Good catch on the copy/pasta ! I just have one suggestion.

GaryJones:
Not merged as the Review Ready tag not applied.

jrfnl:
> Not merged as the `Status: Review ready` label not applied.

@GaryJones Maybe we should document things like that in the CONTRIBUTING guide ?

As far as I'm concerned and with three approvals, this is ready, so I'll merge it.

PR #2544, opened 2025-07-17
AbstractFunctionParameterSniff: don't ignore first class callables

jrfnl:
... but pass them to a dedicated `process_first_class_callable()` method instead to allow sniffs to decide whether to flag these or not.

Typical use-case for why first class callables should not be ignored by default:
A sniff which ALWAYS flags the use of a certain function, but has different error messages depending on whether parameters are passed or not.

In that situation, I believe first class callables should be treated the same as other uses of the target function.
First class callables are basically syntax sugar for closures (example: https://3v4l.org/cra8s)  and if the sniff would flag the use of the target function within a closure, it is only reasonable to also flag the use of the target function as a first class callable.

This commit implements this.

The commit does not include tests as there are no sniffs in WPCS for which the above would apply. However, I have manually tested this change via a sniff in an external standard for which this change is relevant.

Follow up on #2518, also see: https://github.com/WordPress/WordPress-Coding-Standards/pull/2518#issuecomment-3084890977

### DIFF
--- a/WordPress/AbstractFunctionParameterSniff.php
+++ b/WordPress/AbstractFunctionParameterSniff.php
@@ -78,8 +78,8 @@ abstract class AbstractFunctionParameterSniff extends AbstractFunctionRestrictio
 			 * No need for extensive defensive coding as the `is_targetted_token()` method has already
 			 * validated the open and close parentheses exist.
 			 */
-			$next          = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $stackPtr + 1 ), null, true );
-			$firstNonEmpty = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $next + 1 ), null, true );
+			$openParens    = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $stackPtr + 1 ), null, true );
+			$firstNonEmpty = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $openParens + 1 ), null, true );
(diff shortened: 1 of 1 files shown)


GaryJones (approved):

WordPress/AbstractFunctionParameterSniff.php
```diff
@@ -147,4 +150,20 @@ abstract public function process_parameters( $stackPtr, $group_name, $matched_co
...
+	 *
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param string $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched
+	 *                                in lowercase.
+	 *
+	 * @return int|void Integer stack pointer to skip forward or void to continue
+	 *                  normal file processing.
+	 */
+	public function process_first_class_callable( $stackPtr, $group_name, $matched_content ) {}
```
dingo-d, line 168: Is the `public` keyword intentional? 

Seeing that the intended use is to be overloaded in a child class, I'd have thought it'd be `protected`.

Just a question, nothing blocking.

jrfnl: Fair question. And yes, it's intentional as the other methods to be overloaded by child classes are also `public`, so it would be inconsistent to have this method be an exception.

dingo-d: Makes sense, thanks for the clarification :)

Issue #2546, opened 2025-07-18
Support for PHP 8.3 typed class constants


rodrigoprimo:
Typed class constants were introduced in PHP 8.3 (https://wiki.php.net/rfc/typed_class_constants). PHPCS started supporting it in version 3.9.0 (https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/321). Since WPCS requires PHPCS 3.13.x, with @jrfnl's help, I investigated what needed to be updated in the WPCS repository to accommodate this new syntax.

I already added safeguarding tests to two sniffs in #2547.

As far as I can check, no further changes to the sniffs are necessary, but two parts of the WPCS code still require changes. I'm creating this issue to document it.

### ConstantsHelper::is_use_of_global_constant()

This helper method determines whether an arbitrary `T_STRING` token is the use of a global constant. It currently does not account for typed class constant syntax when checking if a variable is a class constant, leading to false positives.

For example, it incorrectly flags the following class constant as a global constant:

```php
class MyClass {
    const string STYLESHEETPATH = 'something';
}
```

The method assumes that the first non-empty token before the constant name must be `T_CONST` for a constant to be a class constant, but this is not the case for typed class constants. The example above results in a false positive for the `WordPress.WP.DiscouragedConstants` as it discourages the use of the global `STYLESHEETPATH` constant.

A similar version of this method also exists in `PHPCompatibility` (https://github.com/PHPCompatibility/PHPCompatibility/blob/develop/PHPCompatibility/Helpers/MiscHelper.php#L45), and it will require similar changes. Since the version on `PHPCompatibility` is covered by tests, those changes will likely be implemented first in that repository and then ported here.

### AbstractClassRestrictionsSniff

This abstract sniff class needs to be updated to check if the restricted classes are used as class constant types (and property types as well). Currently, it does not check for the use of the restricted classes in those places.

PR #2547, opened 2025-07-18
PHP 8.3 typed class constants: add tests to two sniffs

rodrigoprimo:
Typed class constants were introduced in PHP 8.3 (https://wiki.php.net/rfc/typed_class_constants). PHPCS started supporting it in version 3.9.0 (https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/321). Since WPCS requires PHPCS 3.13.x, with @jrfnl's help, I investigated what needed to be updated in the WPCS repository to accommodate this new syntax.

In this PR, I'm adding tests to `WordPress.NamingConventions.PrefixAllGlobals` and `WordPress.WP.CapitalPDangit` to safeguard that those two sniffs continue to handle type class constants correctly, as both sniffs use the `T_CONST` token. No changes were required to the code of those sniffs.

`ConstantsHelper::is_use_of_global_constant()` and `AbstractClassRestrictionsSniff` will need to be changed as well to support typed class constants. This will be addressed in another PR. For now, I documented this task in #2546.

### DIFF
--- a/WordPress/Tests/WP/CapitalPDangitUnitTest.1.inc
+++ b/WordPress/Tests/WP/CapitalPDangitUnitTest.1.inc
@@ -231,3 +231,16 @@ echo Some\Enum\WordPress::ENUM_CONSTANT;
 
 // Safeguard that the sniff doesn't act on anonymous classes.
 $anon = new class() {};
+
+/*
+ * Safeguard that PHP 8.3+ typed class constants are handled correctly.
+ */
+class TypeClassConstants {
+    public const string WORDPRESS = 'wordress';
+
+    public const ?string ANOTHER = 'value',
+        WORDPRESS_SOMETHING = 'wordpress';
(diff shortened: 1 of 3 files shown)


WordPress/Tests/WP/CapitalPDangitUnitTest.1.inc
```diff
@@ -231,3 +231,16 @@ echo Some\Enum\WordPress::ENUM_CONSTANT;
 
 // Safeguard that the sniff doesn't act on anonymous classes.
 $anon = new class() {};
+
+/*
+ * Safeguard that PHP 8.3+ typed class constants are handled correctly.
```
jrfnl:
```suggestion
 * Safeguard that PHP 8.3+ typed class constants are handled correctly (i.e. ignored).
```

PR #2548, opened 2025-07-18
Tests: remove unnecessary trailing spaces from the test files

rodrigoprimo:
While adding a new test to CapitalPDangitUnitTest.1.inc in an unrelated commit, I noticed unnecessary trailing spaces in this file. Instead of removing them just for this one test, I opted to create this PR removing unnecessary trailing spaces from all test files using `Squiz.WhiteSpace.SuperfluousWhitespace`. I'm including only the files where it was easy to check that the spaces were not added intentionally.


jrfnl (approved):
Thanks for doing this clean-up @rodrigoprimo !

PR #2549, opened 2025-07-18
AbstractClassRestrictions: improve how anonymous classes are handled

rodrigoprimo:
Before this PR, `AbstractClassRestrictionsSniff::is_targetted_token()` would name an anonymous class as "\class" or "\readonly" (in the case of readonly anonymous classes). While this would not cause false positives as it is not possible for a class to be named "class" or "readonly", it is not the correct behavior.

This commit improves how this method handles anonymous classes by returning `false` early, as those classes don't have a name.

Includes tests for readonly anonymous classes. It was not necessary to include more tests for normal anonymous classes, as the existing tests are already enough.

### DIFF
--- a/WordPress/AbstractClassRestrictionsSniff.php
+++ b/WordPress/AbstractClassRestrictionsSniff.php
@@ -123,6 +123,15 @@ abstract class AbstractClassRestrictionsSniff extends AbstractFunctionRestrictio
 
 		if ( \in_array( $token['code'], array( \T_NEW, \T_EXTENDS, \T_IMPLEMENTS ), true ) ) {
 			if ( \T_NEW === $token['code'] ) {
+				$nextNonEmpty = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $stackPtr + 1 ), null, true );
+
+				if ( false === $nextNonEmpty
+					|| \in_array( $this->tokens[ $nextNonEmpty ]['code'], array( \T_READONLY, \T_ANON_CLASS, \T_ATTRIBUTE ), true )
+				) {
+					// Live coding or anonymous class (bow out for anonymous classes as they don't have a name).
+					return false;
+				}
+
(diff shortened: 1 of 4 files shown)


WordPress/AbstractClassRestrictionsSniff.php
```diff
@@ -123,6 +123,15 @@ public function is_targetted_token( $stackPtr ) {
 
 		if ( \in_array( $token['code'], array( \T_NEW, \T_EXTENDS, \T_IMPLEMENTS ), true ) ) {
 			if ( \T_NEW === $token['code'] ) {
+				$nextNonEmpty = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $stackPtr + 1 ), null, true );
+
+				if ( false === $nextNonEmpty
+					|| \in_array( $this->tokens[ $nextNonEmpty ]['code'], array( \T_READONLY, \T_ANON_CLASS, \T_ATTRIBUTE ), true )
+				) {
+					// Live coding or anonymous class (bow out for anonymous classes as they don't have a name).
```
jrfnl:
```suggestion
					// Live coding or anonymous class. Bow out.
```

(it's not as if you 're not bailing out for live coding, now is it ?)

WordPress/Tests/DB/RestrictedClassesUnitTest.1.inc
```diff
@@ -104,3 +104,14 @@ $anon = new class extends PDOStatement {}; // Error.
...
 enum MysqliEnum implements mysqli {} // Error.
+
+/*
+ * Safeguard handling of PHP 8.3+ readonly anonymous classes.
+ */
+$anon = new readonly class {
+    public function PDO() {} // OK.
+};
+
+$anon = new readonly class extends PDOStatement {}; // Error.
```
jrfnl: Let's have at least one of the tests use parentheses for the class instantiation ;-)
```suggestion
$anon = new readonly class() extends PDOStatement {}; // Error.
```

jrfnl (approved):
Thanks @rodrigoprimo ! Two nitpicky remarks, but nothing blocking.

rodrigoprimo:
Thanks for your review, @jrfnl. I just added your suggestions.

PR #2551, opened 2025-07-18
PHP 8.4 asymmetric visibility properties: add tests to four sniffs

rodrigoprimo:
Asymmetric visibility properties were introduced in PHP 8.4 (https://wiki.php.net/rfc/asymmetric-visibility-v2). PHPCS started supporting it in version 3.13.0 (https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/871). Since WPCS requires PHPCS 3.13.x, with @jrfnl's help, I investigated what needed to be updated in the WPCS repository to accommodate this new syntax.

In this PR, I'm adding tests to `WordPress.NamingConventions.PrefixAllGlobals`, `WordPress.NamingConventions.ValidVariableName`, `WordPress.Security.NonceVerification`, and `WordPress.WP.GlobalVariablesOverride` to safeguard that those four sniffs continue to handle asymmetric visibility properties correctly, as those sniffs have code related to class properties or parameters passed to the class constructor (in the case of asymmetric visibility combined with constructor promoted properties). No changes were required to the code of those sniffs.

As far as I could check, no further changes are required in the WPCS codebase to accommodate the new asymmetric visibility properties syntax.

To be able to add tests to `WordPress.WP.GlobalVariablesOverride`, I had to move a test with an intentional syntax error to a separate file. While doing that, I made a small improvement to the code comment in the sniff related to the moved test.

### DIFF
--- a/WordPress/Sniffs/WP/GlobalVariablesOverrideSniff.php
+++ b/WordPress/Sniffs/WP/GlobalVariablesOverrideSniff.php
@@ -174,7 +174,7 @@ final class GlobalVariablesOverrideSniff extends Sniff {
 	protected function process_list_assignment( $stackPtr ) {
 		$list_open_close = Lists::getOpenClose( $this->phpcsFile, $stackPtr );
 		if ( false === $list_open_close ) {
-			// Short array, not short list.
+			// Live coding or short array, not short list.
 			return;
 		}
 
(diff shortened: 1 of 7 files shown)


rodrigoprimo:
I had to force push without changes to rerun the build because of an unrelated `phpstan 403: rate limit exceeded` error.

jrfnl (approved):
Thanks for setting this up @rodrigoprimo !

The test for the `ValidVariableName` sniff is actually testing something.

The other tests aren't, but don't do any harm either. No objections to those going in, but they also don't add much value.

PR #2559, opened 2025-07-22
Security/EscapeOutput: fix false negatives when handling anonymous classes

### LINKED ISSUE #2552: Security/EscapeOutput: false negatives when handling anonymous classes
While reviewing #2550, @jrfnl reported false negatives when the `WordPress.Security.EscapeOutput` sniff handles anonymous classes:

> - `EscapeOutput` - code in the case `\T_THROW` block will yield false negatives for readonly anonymous classes and anonymous classes with attributes for code like `throw new [...] class( $unescaped ) {};`.
> 
> My approval is for the PR as-is and presumes the missing tests for EscapeOutput will be addressed in a separate PR.

I'm opening this issue to check a few details instead of discussing them on #2550.

I can see how an anonymous class with attributes will yield false negatives. For example, the code snippets below should trigger errors, but that doesn't happen:

```php
throw new #[MyAttribute] class( $unescaped ) extends Exception {};
throw new #[MyAttribute1(FirstParam::CONST, 'another parameter'), MyAttribute2] class( $unescaped ) extends Exception {};
```

I'm unsure about readonly anonymous classes. I can see how something like the snippets below are false negatives:

```php
throw new readonly class( $unescaped ) {};
throw new readonly class( $unescaped ) extends Exception {};
```

But they are also fatal errors. The anonymous class needs to extend `Exception` or `Error`, but that is not possible for an anonymous readonly class, as `Exception` and `Error` are non-readonly.

@jrfnl, should the sniff trigger an error when handling readonly anonymous classes in a throw statement, even though it is a fatal error? Or is there an example that is not a fatal error that I'm missing?

rodrigoprimo:
This PR fixes false negatives when the sniff `WordPress.Security.EscapeOutput` handles readonly anonymous classes and anonymous classes with attributes that are part of a throw statement.

When stepping over tokens after `T_THROW` to find the `T_OPEN_PARENTHESIS` of the exception creation function call/class instantiation, the sniff was not considering that it might need to step over `T_READONLY` tokens or attribute declarations when dealing with anonymous classes.

Fixes #2552

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -68,6 +68,7 @@
 		<exclude name="PHPCompatibility.Constants.NewConstants.t_coalesceFound"/>
 		<exclude name="PHPCompatibility.Constants.NewConstants.t_coalesce_equalFound"/>
 		<exclude name="PHPCompatibility.Constants.NewConstants.t_yield_fromFound"/>
+		<exclude name="PHPCompatibility.Constants.NewConstants.t_readonlyFound"/>
 	</rule>
 
 	<!-- Enforce PSR1 compatible namespaces. -->
--- a/WordPress/Sniffs/Security/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/Security/EscapeOutputSniff.php
@@ -216,41 +216,35 @@ class EscapeOutputSniff extends AbstractFunctionRestrictionsSniff {
 
 			case \T_THROW:
 				// Find the open parentheses, while stepping over the exception creation tokens.
-				$ignore               = Tokens::$emptyTokens;
-				$ignore              += Collections::namespacedNameTokens();
-				$ignore              += Collections::functionCallTokens();
-				$ignore              += Collections::objectOperators();
-				$ignore[ T_READONLY ] = T_READONLY;
+				$ignore                = Tokens::$emptyTokens;
+				$ignore               += Collections::namespacedNameTokens();
+				$ignore               += Collections::functionCallTokens();
+				$ignore               += Collections::objectOperators();
+				$ignore[ \T_READONLY ] = \T_READONLY;
 
-				$next_relevant = $this->phpcsFile->findNext( $ignore, ( $stackPtr + 1 ), null, true );
-				if ( false === $next_relevant ) {
-					return;
-				}
-
-				if ( \T_NEW === $this->tokens[ $next_relevant ]['code'] ) {
+				$next_relevant = $stackPtr;
+				do {
 					$next_relevant = $this->phpcsFile->findNext( $ignore, ( $next_relevant + 1 ), null, true );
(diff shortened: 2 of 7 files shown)


jrfnl:
> The examples that I find there use a `do/while` loop. I opted to implement the solution here using a `while` loop as the resulting code seems more straightforward to me (one less nesting level). Please let me know if there is a reason to use `do/while` that I'm missing.

The reason is to avoid the code duplication which is now being introduced....

rodrigoprimo:
> The reason is to avoid the code duplication which is now being introduced....

I'm sorry, but it is not clear to me which code duplication you are referring to.

Here is how I implemented the code using `while`:

```php
while ( \T_ATTRIBUTE === $this->tokens[ $next_relevant ]['code'] ) {
	if ( isset( $this->tokens[ $next_relevant ]['attribute_closer'] ) === false ) {
		return;
	}

	$next_relevant = $this->phpcsFile->findNext(
		$ignore,
		( $this->tokens[ $next_relevant ]['attribute_closer'] + 1 ),
		null,
		true
	);

	if ( false === $next_relevant ) {
		return;
	}
}
```

And here is how I would implement it using `do/while`:

```php
do {
	if ( \T_ATTRIBUTE === $this->tokens[ $next_relevant ]['code'] ) {
		if ( isset( $this->tokens[ $next_relevant ]['attribute_closer'] ) === false ) {
			return;
		}

		$next_relevant = $this->phpcsFile->findNext(
			$ignore,
			( $this->tokens[ $next_relevant ]['attribute_closer'] + 1 ),
			null,
			true
		);

		if ( false === $next_relevant ) {
			return;
		}

		continue;
	}

	break;
} while ( true );
```

What is the code duplication that you are referring to? Or maybe there is a simpler way to implement this solution using `do/while` that I'm not seeing?

jrfnl:
> > The reason is to avoid the code duplication which is now being introduced....
> 
> I'm sorry, but it is not clear to me which code duplication you are referring to.
>
> What is the code duplication that you are referring to? Or maybe there is a simpler way to implement this solution using `do/while` that I'm not seeing?

Line 225-228 is basically duplicated in the new code block. With a `do-while`, the line 225-235 would move into the control structure and should remove the code duplication.

rodrigoprimo:
Thanks for your help during the call, @jrfnl. The commit you added looks good to me as is. I just added another commit as we discussed including a new test that will make the condition in the `do/while` loop evaluate to false. This PR should be ready for a final review.

jrfnl (approved):
Thanks @rodrigoprimo ! This looks good to me.

Notes for other reviewers:
* The actual code change will be easier to review while ignoring whitespace changes.
* Please SQUASH-MERGE this PR!

GaryJones (approved):

dingo-d (approved):

PR #2566, opened 2025-07-26
EscapingFunctionsTrait: remove `wp_kses_allowed_html()` from escaping functions

jrfnl:
The `wp_kses_allowed_html()` function is not an escaping function, but retrieves an array of allowed HTML tags and attributes for a given context.

Ref: https://developer.wordpress.org/reference/functions/wp_kses_allowed_html/


jrfnl:
I've rebased & force-pushed the PR in hopes of getting CodeCov to report back. Merging once the build passes.

PR #2568, opened 2025-08-05
GH Actions/unit-tests: don't run code coverage in forks

rodrigoprimo:
The `unit-tests` workflow runs code coverage for pushes to `main` and pull requests, but wasn't limited to this repo.

This resulted in failing builds for contributors who have actions enabled on their fork. The reason for the failure is that they don't have access to the `CODECOV_TOKEN`.

This commit changes the conditions on when to run the tests normally and when to run with code coverage to take the organisation where the repo lives into account. That should prevent the failing builds in forks.

Similar to #2542, which applied the same change to the `quicktest`workflow.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -135,15 +135,15 @@ jobs:
         run: composer lint -- --checkstyle | cs2pr
 
       - name: Run the unit tests without code coverage
-        if: ${{ matrix.coverage == false }}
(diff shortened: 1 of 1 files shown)


jrfnl (approved):
Thanks @rodrigoprimo ! LGTM.

I wonder if we shouldn't improve the workflow organisation to only have code coverage in the `unit-test.yml` workflow - probably via a separate job, but that's for a future iteration.

PR #2572, opened 2025-08-12
Security/NonceVerification: handle function names case correctly

rodrigoprimo:
The sniff was incorrectly treating function names as case-sensitive when checking for nonce verification functions. This led to false positives when developers used valid nonce verification functions with mixed or uppercase letters.

The fix ensures function names are properly converted to lowercase before comparing against the allowed nonce verification functions list, respecting PHP's case-insensitive function name behavior.

As discussed with Juliette during the PR review, that also means we need to ensure that the `$customNonceVerificationFunctions` are all lowercased before doing a comparison.

### DIFF
--- a/WordPress/Sniffs/Security/NonceVerificationSniff.php
+++ b/WordPress/Sniffs/Security/NonceVerificationSniff.php
@@ -413,12 +413,12 @@ class NonceVerificationSniff extends Sniff {
 	 */
 	protected function mergeFunctionLists() {
 		if ( $this->customNonceVerificationFunctions !== $this->addedCustomNonceFunctions ) {
-			$this->customNonceVerificationFunctions = array_map( 'strtolower', $this->customNonceVerificationFunctions );
-
 			$this->nonceVerificationFunctions = RulesetPropertyHelper::merge_custom_array(
 				$this->customNonceVerificationFunctions,
 				$this->nonceVerificationFunctions
 			);
+
+			$this->nonceVerificationFunctions = array_change_key_case( $this->nonceVerificationFunctions );
 
 			$this->addedCustomNonceFunctions = $this->customNonceVerificationFunctions;
 		}


jrfnl (changes requested):
Good catch.

> ... respecting PHP's case-insensitive function name behavior.

PHP isn't completely case-insensitive for function names. Only for the ASCII characters and as this sniff allows for end-users to provide a list of `customNonceVerificationFunctions` via a public property, we cannot rely on all nonce-verification functions we need to check against being completely ASCII-based.

In other words, if someone would add a custom nonce verification function called `déjà_vu()`,  this function would now never be recognized as a valid nonce verification function.

So, we can only use `strtolower()` to compare the names case-insensitively if we have a **_known_** list of functions we compare against and we are 100% sure all those function names are ASCII-based names.

When the list of functions to compare against is not known in advance and/or doesn't consist of all ASCII-based functions we need to loop through all "allowed functions" using the PHPCSUtils `NamingConventions::isEqual()` method - which does a comparison between two names in the way comparable to how PHP does it.

rodrigoprimo:
Thanks for your reply, @jrfnl. I just added two new commits, one fixing the sniff to handle custom nonce verification functions correctly regardless of the case, and another adding tests safeguarding that non-ASCII characters are handled properly.

WordPress/Sniffs/Security/NonceVerificationSniff.php
```diff
@@ -413,6 +413,8 @@ private function set_cache( $filename, $start, $end, $nonce ) {
 	 */
 	protected function mergeFunctionLists() {
 		if ( $this->customNonceVerificationFunctions !== $this->addedCustomNonceFunctions ) {
+			$this->customNonceVerificationFunctions = array_map( 'strtolower', $this->customNonceVerificationFunctions );
+
```
jrfnl: Why are you changing the user-provided value ? instead of changing the merge result `$this->nonceVerificationFunctions` ?

rodrigoprimo: I had in my mind that the user-provided value is the only one that required change, and that is why I opted to change it. After seeing your question, I imagine you are suggesting changing `$this->nonceVerificationFunctions` to ensure the sniff behaves correctly if, in the future, a new function is added to this property without all lowercase letters, and also to ensure it behaves correctly for sniffs extending this one. Is that the case?

I went ahead and pushed a new commit changing `$this->nonceVerificationFunctions` instead of `$this->customNonceVerificationFunctions`.

GaryJones (approved):
All looks sensible - and I learned something too!

- Case-insensitive for ASCII letters only.
- Accented letters (é, É, à) are byte-sensitive.

jrfnl:
@rodrigoprimo Could you please rebase the PR to make it mergable ?

rodrigoprimo:
@jrfnl, I force-pushed without changes, but now the PHPStan check is failing. I'm checking to see if I can reproduce this problem locally.

rodrigoprimo:
PHPStan is failing due to an upstream issue: https://github.com/shivammathur/setup-php/issues/1000

`shivammathur/setup-php` installed the incorrect PHPStan version (`1.9.9` instead of `1.12.29`). There is a comment in the issue saying that it was fixed, but I'm still seeing this error. I will investigate a bit more.

jrfnl:
@rodrigoprimo I retriggered PHPStan after the release of setup-php 2.35.5 and things are okay again.

When I look at the commits, I think they should all be squashed. Agreed ? I can do that on merge.

PR #2573, opened 2025-08-13
PHP 8.5 | WP/EnqueuedResourceParameters: prevent deprecation notice for non-standard casts

jrfnl:
### PHP 8.5 | WP/EnqueuedResourceParameters: prevent deprecation notice for non-standard casts

Four non-standard type casts are going to be deprecated in PHP 8.5.

When any of these type casts would be used in the parameter this sniff examines in the "code under scan" and the sniff would be run on PHP 8.5, the non-standard type cast would be executed via the `eval()`, leading to a deprecation notice, which would stop the scan of the file.

This commit works around this by always using the "standard" type cast syntax in the code which would be passed to `eval()`.

Includes tests.

### DIFF
--- a/WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
+++ b/WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
@@ -227,6 +227,30 @@ final class EnqueuedResourceParametersSniff extends AbstractFunctionParameterSni
 				continue;
 			}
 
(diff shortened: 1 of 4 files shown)


GaryJones (approved):

jrfnl:
I've retrigger the PHP 8.5 tests now the Composer issue has been fixed. Tests pass, so all good.

PR #2574, opened 2025-08-16
Encourage use of `wp_safe_remote_*` functions

### LINKED ISSUE #1288: Security: Recommend wp_safe_remote_*() when wp_remote_*() is used
The `wp_remote_*()` functions (like `wp_remote_get()`) do not validate the passed URL, resulting in the possibility for malicious requests if the `$url` is user-controlled. The `wp_safe_remote_*()` functions should be instead, since "the URL is validated to avoid redirection and request forgery attacks."

johnbillion:
Fixes #1288

### DIFF
--- a/WordPress/Docs/Security/SafeRemoteRequestStandard.xml
+++ b/WordPress/Docs/Security/SafeRemoteRequestStandard.xml
@@ -1,0 +1,62 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Safe Remote Request"
+    >
+    <standard>
+    <![CDATA[
+    The wp_safe_remote_* functions should be used instead of wp_remote_* functions when dealing with URLs that might be user-controlled. The wp_remote_* functions do not validate URLs, which can lead to potential malicious requests. The wp_safe_remote_* functions validate URLs to avoid redirection and request forgery attacks (SSRF - Server-Side Request Forgery).
+
+    This is particularly important when:
+    - The URL comes from user input ($_GET, $_POST, etc.)
+    - The URL is stored in the database and could have been modified
+    - The URL comes from any untrusted source
+
+    The safe alternatives are:
+    - wp_safe_remote_get() instead of wp_remote_get()
+    - wp_safe_remote_post() instead of wp_remote_post()
+    - wp_safe_remote_head() instead of wp_remote_head()
+    - wp_safe_remote_request() instead of wp_remote_request()
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using safe remote request functions.">
+        <![CDATA[
+// GET request with URL validation
+$response = <em>wp_safe_remote_get</em>( $url );
+
+// POST request with URL validation
+$response = <em>wp_safe_remote_post</em>( $url, array(
+    'body' => $data
+) );
+
+// HEAD request with URL validation
+$response = <em>wp_safe_remote_head</em>( $url );
+
+// Generic request with URL validation
+$response = <em>wp_safe_remote_request</em>( $url, array(
+    'method' => 'DELETE'
+) );
+        ]]>
+        </code>
+        <code title="Invalid: Using unsafe remote request functions.">
+        <![CDATA[
+// GET request without URL validation
+$response = <em>wp_remote_get</em>( $url );
+
+// POST request without URL validation
+$response = <em>wp_remote_post</em>( $url, array(
+    'body' => $data
+) );
+
+// HEAD request without URL validation
+$response = <em>wp_remote_head</em>( $url );
+
+// Generic request without URL validation
+$response = <em>wp_remote_request</em>( $url, array(
+    'method' => 'DELETE'
+) );
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>
--- a/WordPress/Sniffs/Security/SafeRemoteRequestSniff.php
+++ b/WordPress/Sniffs/Security/SafeRemoteRequestSniff.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Sniffs\Security;
+
+use WordPressCS\WordPress\AbstractFunctionRestrictionsSniff;
+
+/**
+ * Encourages use of wp_safe_remote_* functions to avoid potential security issues.
+ *
+ * The wp_remote_* functions do not validate URLs, which can lead to potential
+ * malicious requests if the URL is user-controlled. The wp_safe_remote_* functions
+ * validate URLs to avoid redirection and request forgery attacks.
+ *
+ * @since x.y.z
+ *
+ * @link https://github.com/WordPress/WordPress-Coding-Standards/issues/1288
+ */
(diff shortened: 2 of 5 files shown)


WordPress/Sniffs/Security/SafeRemoteRequestSniff.php
```diff
@@ -0,0 +1,69 @@
...
+	public function process_matched_token( $stackPtr, $group_name, $matched_content ) {
+		// Extract the function type (get, post, head, request) from the matched function name.
+		$function_type = str_replace( 'wp_remote_', '', $matched_content );
+
+		// Build the custom message with the specific safe alternative.
+		$message = sprintf(
+			$this->groups[ $group_name ]['message'],
+			$matched_content,
+			$function_type
+		);
```
jrfnl: PHPCS has a build-in sprintf mechanism. Use that instead by passing replacements in a fourth `$data` array parameter to `addWarning()`.

WordPress/Docs/Security/SafeRemoteRequestStandard.xml
```diff
@@ -0,0 +1,62 @@
...
+    title="Safe Remote Request"
+    >
+    <standard>
+    <![CDATA[
+    The wp_safe_remote_* functions should be used instead of wp_remote_* functions when dealing with URLs that might be user-controlled. The wp_remote_* functions do not validate URLs, which can lead to potential malicious requests. The wp_safe_remote_* functions validate URLs to avoid redirection and request forgery attacks (SSRF - Server-Side Request Forgery).
+
+    This is particularly important when:
+    - The URL comes from user input ($_GET, $_POST, etc.)
+    - The URL is stored in the database and could have been modified
+    - The URL comes from any untrusted source
```
jrfnl:
```suggestion
    - The URL comes from user input ($_GET, $_POST, etc.);
    - The URL is stored in the database and could have been modified;
    - The URL comes from any untrusted source.
```

WordPress/Docs/Security/SafeRemoteRequestStandard.xml
```diff
@@ -0,0 +1,62 @@
...
+
+    The safe alternatives are:
+    - wp_safe_remote_get() instead of wp_remote_get()
+    - wp_safe_remote_post() instead of wp_remote_post()
+    - wp_safe_remote_head() instead of wp_remote_head()
+    - wp_safe_remote_request() instead of wp_remote_request()
```
jrfnl:
```suggestion
    - wp_safe_remote_get() instead of wp_remote_get().
    - wp_safe_remote_post() instead of wp_remote_post().
    - wp_safe_remote_head() instead of wp_remote_head().
    - wp_safe_remote_request() instead of wp_remote_request().
```

WordPress/Docs/Security/SafeRemoteRequestStandard.xml
```diff
@@ -0,0 +1,62 @@
...
+$response = <em>wp_remote_head</em>( $url );
+
+// Generic request without URL validation
+$response = <em>wp_remote_request</em>( $url, array(
+    'method' => 'DELETE'
+) );
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>
```
jrfnl: Please end the file on a new line character.

WordPress/Tests/Security/SafeRemoteRequestUnitTest.inc
```diff
@@ -0,0 +1,39 @@
...
+wp_REMOTE_request( $url, $args ); // Warning.
+
+// Other wp_remote_* functions that don't have safe alternatives shouldn't trigger warnings.
+wp_remote_retrieve_body( $response ); // OK - This is for parsing responses, not making requests.
+wp_remote_retrieve_response_code( $response ); // OK.
+wp_remote_retrieve_headers( $response ); // OK.
+wp_remote_retrieve_header( $response, 'Content-Type' ); // OK.
+wp_remote_retrieve_response_message( $response ); // OK.
+wp_remote_retrieve_cookie( $response, 'session' ); // OK.
+wp_remote_retrieve_cookies( $response ); // OK.
```
jrfnl: Please end the file on a new line character.

WordPress/Sniffs/Security/SafeRemoteRequestSniff.php
```diff
@@ -0,0 +1,69 @@
...
+	/**
+	 * Groups of functions to restrict.
+	 *
+	 * @return array
+	 */
+	public function getGroups() {
+		return array(
+			'unsafe_remote_requests' => array(
+				'type'      => 'warning',
+				'message'   => '%s() found. Using wp_safe_remote_%s() instead can help avoid redirection and request forgery attacks when dealing with user-controlled URLs.',
```
jrfnl:
```suggestion
				'message'   => 'Found: %s(). Using wp_safe_remote_%s() instead can help avoid redirection and request forgery attacks when dealing with user-controlled URLs.',
```

(Reads better IMO)

WordPress/Docs/Security/SafeRemoteRequestStandard.xml
```diff
@@ -0,0 +1,62 @@
...
+    - The URL comes from any untrusted source
+
+    The safe alternatives are:
+    - wp_safe_remote_get() instead of wp_remote_get()
+    - wp_safe_remote_post() instead of wp_remote_post()
+    - wp_safe_remote_head() instead of wp_remote_head()
+    - wp_safe_remote_request() instead of wp_remote_request()
+    ]]>
+    </standard>
+    <code_comparison>
```
jrfnl, line 22: The fourth code sample in each block does not comply with the max line length requirements for the docs.

Run `phpcs --generator=Text --standard=WordPress --sniffs=WordPress.Security.SafeRemoteRequest` to see the text breaking the table layout.

This needs to be fixed.

jrfnl (changes requested):
Hi @johnbillion, thanks for this PR!

From what I understand based on the original ticket, this is predominantly about preventing issues with arbitrary/user controlled URLs.
With that in mind, I think this sniff will result in preventable false positives as it does not examine the function parameters.

Think:
```
wp_remote_get('https://example.com/my-file');
wp_remote_get(MYCLASS::URL);
```

In both the above cases, we can be sure the URL is not user controlled or from an untrusted source, but these would both be flagged by the sniff as-is.

I'd like to suggest running the sniff over both WP Core as well as a significant number of plugins, then to examine all results and extrapolate patterns which should be considered false positives and adjust the sniff to guard against those.
_Pro-tip: use `--report=code` to get quick access to the code which is being flagged from the CLI._


I'd also like to hear your thoughts on this comment which I previously posted on the ticket:

> ... the WP.AlternativeFunctions sniff already contain some checks for wp_remote_*(). Those may need to be removed/adjusted when this issue gets addressed.

The point here is that some of the checks in the `WP.AlternativeFunctions` sniff _recommend_ using `wp_remote_get()` and I imagine that recommendation may need to be revisited.

As for the tests, these are mostly looking good, though they do need to be expanded a bit more:
* Fully qualified function calls
* Some test code to safeguard that method calls/namespaced functions/constants with the same names are not flagged.
    Yes, this is handled by the abstract, but should still be safeguarded via tests for the sniff to allow for future refactoring.
* Comments in unexpected places (though this will be more relevant once checking of the parameters is put in place).

Other than that, I find the docs a bit wordy. As this is a security sniff, that's not necessarily a bad thing, though it does mean that the docs will need to be kept up to date with the latest security information and I worry that this won't be done.

johnbillion:
No, nobody can do that.

johnbillion:
The more I think about it the more I think this is not going to be achievable to a usable standard with PHPCS. Even basic logic that populates the `$url` parameter cannot reasonably be detected without constructing an AST and having awareness of all the functions that are considered "safe" for a URL to pass through or be constructed from. Example:

```php
$url = add_query_arg( 'arg', $value, home_url( 'foo.php' ) );
$result = wp_remote_post( $url );
```

Running this over WordPress core results in mostly false positives, which is not useful. It only works for narrow use cases.

I'm going to close this PR and another time I'll have a go at writing a PHPStan rule for this.

Cheers!

PR #2578, opened 2025-08-21
DB/RestrictedClasses: remove redundant call to `parent::setUpPrerequisites()`

rodrigoprimo:
Calling `AbstractSniffUnitTest::setUpPrerequisites()` manually inside `RestrictedClassesUnitTest::enhanceGroups()` is not necessary, as this method already executes via its `@before` tag and results in the method running twice.

Originally, `RestrictedClassesUnitTest::enhanceGroups()` was named `RestrictedClassesUnitTest::setUp()` and called `parent::setUp()` ([WordPress/Tests/DB/RestrictedClassesUnitTest.php#L34-L35](https://github.com/WordPress/WordPress-Coding-Standards/blob/3b9ae92607295548df357e108fd27090cf89d1d7/WordPress/Tests/DB/RestrictedClassesUnitTest.php#L34-L35)). But later it was renamed, and then it probably didn't make sense to continue calling `parent::setUp()` (which was later renamed to `setUpPrerequisites()`): https://github.com/WordPress/WordPress-Coding-Standards/commit/00b895c92f41a49923c37df58b370b801080a204.

I found this in the context of the work to prepare this codebase for PHPCS 4.0 as `AbstractSniffUnitTest::setUpPrerequisites()` was renamed in 4.0 (https://github.com/PHPCSStandards/PHP_CodeSniffer/commit/92cf10f65c2ff1486e28eb1e359680eea5c54224#diff-91ba2a2a923bbfdaa61dcd31051d783bd0e7ff99f979f6e1a2917cae7dde3eb2L58).

### DIFF
--- a/WordPress/Tests/DB/RestrictedClassesUnitTest.php
+++ b/WordPress/Tests/DB/RestrictedClassesUnitTest.php
@@ -37,8 +37,6 @@ final class RestrictedClassesUnitTest extends AbstractSniffUnitTest {
 	 * @return void
 	 */
 	protected function enhanceGroups() {
-		parent::setUpPrerequisites();
-
 		AbstractFunctionRestrictionsSniff::$unittest_groups = array(
 			'test' => array(
 				'type'    => 'error',


jrfnl (approved):
Good find @rodrigoprimo and good explanation.

While this fix is valid as-is for the current codebase, I want to point out that this will need to be brought back again once support for PHPCS 3.x will be dropped.

Reasoning:
* Drop PHPCS 3.x means => drop support for PHP 7.2 means => no need to support PHPUnit 4-7 anymore.
* Once support for PHPUnit 4-7 has been dropped, we can remove the `@before`/`@after` annotations and start using the normal `setUp()`/`tearDown()` methods again (as the required return type `void` is no longer a problem), in which case, we will need to call `parent::setUp()` again.

Not sure if you already have some commits in your WIP branch for the follow ups (add cross-version PHPCS 3/4 support => drop support for PHP < 7.2 => drop support for PHPUnit < 8 => drop support for PHPCS 3.x => add support for PHPUnit 10+11). If so, please make sure you add it back in the appropriate follow-up commit.
Otherwise, we should maybe make a note of this somewhere ?

rodrigoprimo:
Thanks for the review, @jrfnl!

> Not sure if you already have some commits in your WIP branch for the follow ups

Thanks for checking. Yes, I have a commit ready for this in a WIP branch.

PR #2579, opened 2025-08-21
ConstantsHelper::is_use_of_global_constant(): do not treat `use const` alias as global constant usage

rodrigoprimo:
`ConstantsHelper::is_use_of_global_constant()` was incorrectly identifying a constant alias as a global constant.

### DIFF
--- a/WordPress/Helpers/ConstantsHelper.php
+++ b/WordPress/Helpers/ConstantsHelper.php
@@ -79,6 +79,7 @@ final class ConstantsHelper {
 			\T_INSTANCEOF => true,
 			\T_INSTEADOF  => true,
 			\T_GOTO       => true,
+			\T_AS         => true,
 		);
 		$tokens_to_ignore += Tokens::$ooScopeTokens;
 		$tokens_to_ignore += Collections::objectOperators();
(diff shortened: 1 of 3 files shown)


jrfnl:
Considering I introduced this (even though, I did it when I was much less experienced) and that I can still come up with a reason why it may be good to continue flagging this, I believe a second opinion is warranted on this issue.

@GaryJones @dingo-d Either of you have an opinion on the above ?

GaryJones:
I'm struggling to comprehend exactly what's being proposed here - @dingo-d I'll let you be the tie-breaker if needed.

dingo-d:
Tbh I am not sure anybody is still using WPThemeReview sniffs (the standard hasn't been updated in 4 years from what I can see). I'd like to see if there is anybody from the themes team still using this when checking themes on .org.

To me, it's ok to fix the 'faulty' behavior here, and then make exceptions downstream (in WPThemeReview for example).

I'm also having a hard time understanding why this was added (let alone trying to understand what was done back in 2016 🙈 ).

jrfnl:
Given the opinions on this, let's remove the support for const aliases. If/when the "is this a global constant" check moves to PHPCSUtils this may need another look, but that's for later.

rodrigoprimo:
Sure, I just rebased this branch and squashed the two commits.

jrfnl:
I'm now in two minds... should this be allowed to go into the next patch release (and regarded as a bug fix) or does this need to go into the next minor or even next major as we are removing support for something....

@GaryJones @dingo-d Opinions ?

dingo-d:
I really don't think this will have a major impact to users 🤷🏼‍♂️ 

So it can be a patch or minor imo

rodrigoprimo:

PR #2580, opened 2025-08-21
WP/GlobalVariablesOverride: more tests covering unresolvable `$GLOBALS` keys

rodrigoprimo:
This PR adds tests for cases where the  `WordPress.WP.GlobalsVariableOverride` sniff is not able to resolve the value of `$GLOBALS` keys.

### DIFF
--- a/WordPress/Tests/WP/GlobalVariablesOverrideUnitTest.1.inc
+++ b/WordPress/Tests/WP/GlobalVariablesOverrideUnitTest.1.inc
@@ -320,3 +320,15 @@ class AsymmetricVisibilityProperties {
 
     public function __construct(public protected(set) int $page = 0) {} // Ok.
 }
+
+/*
+ * $GLOBALS with non-string key, unable to determine if it is a global variable override or not.
+ */
+$GLOBALS[ "some{$name}" ] = 'test'; // Ok.
+$GLOBALS[ $obj->global_name ] = 'test'; // Ok.
+$GLOBALS[ MyClass::GLOBAL_NAME ] = 'test'; // Ok.
+$GLOBALS[ GLOBAL_NAME_CONSTANT ] = 'test'; // Ok.
+$GLOBALS[ get_my_global_name() ] = 'test'; // Ok.
+$GLOBALS[ MyNamespace\get_my_global_name() ] = 'test'; // Ok.
+$GLOBALS[ \MyNamespace\get_my_global_name() ] = 'test'; // Ok.
+$GLOBALS[ namespace\get_my_global_name() ] = 'test'; // Ok.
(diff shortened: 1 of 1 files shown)


WordPress/Tests/WP/GlobalVariablesOverrideUnitTest.1.inc
```diff
@@ -311,6 +311,14 @@ list(
 	get($year, $day) => &$not_a_wp_global[$year]
 ] = $array;
 
-// Live coding/parse error.
-// This has to be the last test in the file!
-list( $tab, $tabs
+/*
+ * $GLOBALS with non-string key, unable to determine if it is a global variable override or not.
```
jrfnl: Ahum... it's definitely a global variable override, we just don't know whether it is one we should flag.

```suggestion
 * $GLOBALS with non-string key, unable to determine which variable is being overridden.
```

WordPress/Tests/WP/GlobalVariablesOverrideUnitTest.1.inc
```diff
@@ -311,6 +311,14 @@ list(
...
+ */
+$GLOBALS[ "some{$name}" ] = 'test'; // Ok.
+$GLOBALS[ $obj->global_name ] = 'test'; // Ok.
+$GLOBALS[ MyClass::GLOBAL_NAME ] = 'test'; // Ok.
+$GLOBALS[ GLOBAL_NAME_CONSTANT ] = 'test'; // Ok.
+$GLOBALS[ get_my_global_name() ] = 'test'; // Ok.
+$GLOBALS[ MyNamespace\get_my_global_name() ] = 'test'; // Ok.
+$GLOBALS[ \MyNamespace\get_my_global_name() ] = 'test'; // Ok.
+$GLOBALS[ namespace\get_my_global_name() ] = 'test'; // Ok.
```
jrfnl: Please wrap this test code within a simple function declaration to prevent it potentially affecting other tests in the file. (I vaguely remember we had issues with this in the past, better to prevent new ones in the future ;-) )

rodrigoprimo:
> You may want to update the PR description as I think one commit was dropped from this PR

Done

dingo-d (approved):
I know this sniff is related to WP globals, but what about the PHP 8.1 change about restricting changing the $GLOBALS entirely?

Do we have a sniff for that?

jrfnl:
> I know this sniff is related to WP globals, but what about the PHP 8.1 change about restricting changing the $GLOBALS entirely?
> 
> Do we have a sniff for that?

@dingo-d There's a sniff in PHPCompatibility for that and IMO that's the right place for it.

PR #2581, opened 2025-08-21
AbstractFunctionRestrictions sniffs: add tests for namespaced names

rodrigoprimo:
In preparation for PHPCS 4.0, which changes the tokenization of namespaced names, this PR adds tests with all variants of namespace names (unqualified, partially qualified, fully qualified, and namespace relative) to the sniffs that extend the `AbstractFunctionRestrictions` class when necessary (some commits already contained part of the required tests).

The only sniff that is missing in this PR is `WordPress.Security.EscapeOutput`, which will be addressed separately. 

Before adding the tests for the `WordPress.WP.DeprecatedFunctions` sniff, I opted to rename the test case file, as it works differently from the other test case files, and adding the namespaced names to it seemed incorrect.

Note: there will be more sniffs which will need additional tests, so this is only one of the first of a range of PRs adding more tests to help get ready for PHPCS 4.0.


WordPress/Tests/PHP/DiscouragedPHPFunctionsUnitTest.inc
```diff
@@ -41,3 +41,11 @@ $handle = popen( '/bin/ls', 'r' );
 // Issue #898.
 class System {}
 class Serialize {}
+
+/*
```
jrfnl: Could these tests have re-used the currently empty block of lines at the top of the file ?

WordPress/Tests/WP/AlternativeFunctionsUnitTest.inc
```diff
@@ -147,3 +147,11 @@ file_get_contents(
...
 	$url
 ); // Warning.
+
+/*
+ * Safeguard support for PHP 8.0+ tokenization of namespaced "names".
+ */
+\curl_init();
+MyNamespace\curl_init();
+\MyNamespace\curl_init();
+namespace\curl_init(); // The sniff should start flagging this once it can resolve relative namespaces.
```
jrfnl: These tests need to be expanded a lot more.

This sniff also looks at the parameters for a number of target functions and for those check for certain (global) constants and/or function calls in the parameters, so tests are needed to cover those being fully qualified as well.

rodrigoprimo: Good point!

If I understand the sniff correctly, besides expanding the tests, the sniff also needs to be changed.

It checks for function calls and constants in the first parameter of `file_get_contents()` without differentiating global and non-global function calls and constants. The examples below result in false negatives:

```php
file_get_contents(MyNamespace\WP_Upload_Dir()['path'] . 'subdir/file.inc');
file_get_contents(\MyNamespace\ABSPATH . 'wp-admin/css/some-file.css');
```

Besides that, the sniff accepts resources passed to the first parameter of `file_get_contents()`, `file_put_contents()`, `fopen()`, and `readfile()`, but those four functions only accept a string. Unless I'm missing something, examples like the one below will always result in a fatal error, and the sniff probably does not need code to handle it:

https://github.com/WordPress/WordPress-Coding-Standards/blob/f6965eed3db6393bfc89e975bcd3fb27186c74c2/WordPress/Tests/WP/AlternativeFunctionsUnitTest.inc#L46

https://github.com/WordPress/WordPress-Coding-Standards/blob/f6965eed3db6393bfc89e975bcd3fb27186c74c2/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php#L358

I believe fixing this is not high priority and should not block the work to prepare WPCS for PHPCS 4.0. My suggestion is that we don't expand the tests for this sniff in this PR, and create two issues to document the problems described above to be addressed at a later point. What do you think?

jrfnl: I think it would be better to remove the test improvements for this particular sniff from this PR to unblock this PR and address improving the `AlternativeFunctions` tests (and sniff) separately.

WordPress/Tests/PHP/DiscouragedPHPFunctionsUnitTest.inc
rodrigoprimo: I pushed a new commit updating the test to reuse the empty block at the top of the file.

WordPress/Tests/WP/AlternativeFunctionsUnitTest.inc
rodrigoprimo: I removed the commits related to `AlternativeFunctions` from this PR. I will create two separate issues documenting the problems that I found.

rodrigoprimo: Here are the issues: https://github.com/WordPress/WordPress-Coding-Standards/issues/2602 and https://github.com/WordPress/WordPress-Coding-Standards/issues/2603

rodrigoprimo: @jrfnl, on a second thought, I'm wondering if it makes sense to include back the original commit that added tests for namespaced function calls to this PR? My reasoning for this is that I don't think #2603 should block the work to make WPCS compatible with PHPCS 4.0, and the original tests already add some value. Currently, even on PHPCS 3.x, the sniff does not handle namespaced function calls/constants correctly when passed as a parameter to one of the functions it checks.

jrfnl: Discussed this and suggested to still do this in a separate PR.

PR #2494, opened 2024-09-17
adds documentation for sniff WordPress.PHP.DiscouragedPHPFunctions

tikifez:
Related to https://github.com/WordPress/WordPress-Coding-Standards/issues/1722

PR #2584, opened 2025-08-26
jasonkenison:Docs/word press.php.discouraged php functions

jasonkenison:
Related to[ #2494](https://github.com/WordPress/WordPress-Coding-Standards/pull/2494)

Updated indentation, added <em> tags and a line break at the end of the file.


WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,113 @@
...
+    </standard>
+    <code_comparison>
+        <code title="Valid: Not changing configuration at runtime.">
+        <![CDATA[
+// Configuration not changed at runtime.
+        ]]>
+        </code>
+        <code title="Invalid: Changing configuration at runtime">
+        <![CDATA[
+error_reporting( 0 );
```
rodrigoprimo: This block and most of the blocks below are missing `<em>` tags.

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,113 @@
...
+    <standard>
+    <![CDATA[
+Use JSON instead of serialized data, which has known vulnerability problems with object injection.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using JSON for serialized data.">
+        <![CDATA[
+$serialized = <em>json_encode</em>( $array );
+$serialized = <em>wp_json_encode</em>( $array );
```
rodrigoprimo: I think that just a single valid example encoding `$array` is enough. I don't see a reason for two. Maybe keep this line and remove the one above?

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,113 @@
...
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Discouraged PHP Functions"
+    >
+    <standard>
+    <![CDATA[
+Use JSON instead of serialized data, which has known vulnerability problems with object injection.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using JSON for serialized data.">
```
rodrigoprimo: I don't have a specific suggestion, but I think this title can be improved. Maybe `Using JSON to encode data` or `JSON encoding data`. As a non-English native speaker, I'm not sure, but I don't think the examples below are about `Using JSON for serialized data` as the data is not necessarily serialized before it is passed to the JSON functions. Makes sense?

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,113 @@
...
+    </code_comparison>
+    <standard>
+    <![CDATA[
+Functions often used for obfuscating code are strongly discouraged. Make sure the function is used for benign reasons.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using functions for benign reasons.">
+        <![CDATA[
+base64_encode( $string );
```
rodrigoprimo: Please double-check, but I believe the examples provided here actually trigger the sniff and thus are invalid. If I'm correct, the sniff triggers a warning if one of those functions is used, regardless of how it is used. It is up to a developer to determine if the function is used for benign purposes if they want to keep it. Let me know if you need help testing the sniff.

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
rodrigoprimo: I'm unsure (here and in other `<code_comparison>` blocks) if it is necessary to include an example for every single function that triggers the sniff. Please check other sniff XML docs if you can, but I believe it is more common to just include one example (if all the others are the same except for the function name). Then, if the list is not too long, you can consider mentioning all the functions in the corresponding `<standard>` description. What do you think?

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,113 @@
...
+        <![CDATA[
+$serialized = <em>json_encode</em>( $array );
+$serialized = <em>wp_json_encode</em>( $array );
+$unserialized = <em>json_decode</em>( $array );
+        ]]>
+        </code>
+        <code title="Invalid: Using serialized data strings.">
+        <![CDATA[
+$serialized = <em>serialize</em>( $array );
+$unserialized = <em>unserialize</em>( $array );
```
rodrigoprimo: Maybe it is better to use a variable name other than `$array` here as this function takes an `string` as its first parameter? The same applies to the `json_decode()` example above.

Please check if other variables could be renamed as well for the same reason in the other examples.

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,113 @@
...
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+URLs should now be encoded using rawurlencode(). Only legacy applications should use urlencode().
```
rodrigoprimo: Maybe it is not necessary to include the second sentence? The sniff will trigger anyway if it finds `urlencode()`. If there is a reason to include the second sentence, maybe it makes sense to add another sentence explaining what determines the cases where legacy applications should use `urlencode()`. I don't remember off the top of my head.

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,113 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Discouraged PHP Functions"
+    >
+    <standard>
+    <![CDATA[
+Use JSON instead of serialized data, which has known vulnerability problems with object injection.
```
rodrigoprimo: Here and in all other `<standard>` blocks, the description should be indented four spaces.
```suggestion
    Use JSON instead of serialized data, which has known vulnerability problems with object injection.
```

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,113 @@
...
+Avoid using functions which change configuration values at runtime.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Not changing configuration at runtime.">
+        <![CDATA[
+// Configuration not changed at runtime.
+        ]]>
+        </code>
+        <code title="Invalid: Changing configuration at runtime">
```
rodrigoprimo: Missing period

```suggestion
        <code title="Invalid: Changing configuration at runtime.">
```

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,113 @@
...
+restore_include_path();
+magic_quotes_runtime( $new_setting );
+set_magic_quotes_runtime( $new_setting );
+dl( $extension_filename );
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+Do not use PHP system calls. They are often disabled by server admins.
```
rodrigoprimo: Since this sniff triggers a warning and not an error, I suggest rephrasing this description to make it explicit that it is a recommendation. The description above is a code example. Maybe something like "Avoid using PHP system calls as they are often disabled by server administrators."

rodrigoprimo (commented):
Thanks for creating this PR, @jasonkenison!

I left some comments with suggestions. Let me know if you have questions.

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
jasonkenison: Sorry I missed these. Line reworded as requested.

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
jasonkenison: Punctuation added.

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,95 @@
...
+    <code_comparison>
+        <code title="Valid: Using functions for benign reasons.">
+        <![CDATA[
+// It is up to the developer to determine if the function is used for benign purposes or if they want to keep it.
+        ]]>
+        </code>
+        <code title="Invalid: Using functions to obfuscate code.">
+        <![CDATA[
+<em>eval( </em>base64_decode( $code_str )<em> )</em>;
+<em>convert_uudecode( $uuencoded )</code>em>;
```
rodrigoprimo, line 90: Besides the point that I raised above about the placement of the `<em>` tags, I believe a typo was introduced in this particular line in the recent commits: `code>`.

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,95 @@
...
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using functions for benign reasons.">
+        <![CDATA[
+// It is up to the developer to determine if the function is used for benign purposes or if they want to keep it.
+        ]]>
+        </code>
+        <code title="Invalid: Using functions to obfuscate code.">
+        <![CDATA[
+<em>eval( </em>base64_decode( $code_str )<em> )</em>;
```
rodrigoprimo, line 89: I believe `eval()` is not flagged by the sniff so it should not be highlighted. More broadly, I'm not sure if it needs to be included as the sniff will trigger anyway for `base64_decode()` regardless of whether `eval()` is used.

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,95 @@
...
+<em>convert_uudecode( $uuencoded )</code>em>;
+<em>str_rot13( $rot13_encoded )</em>;
```
rodrigoprimo, line 91: I migth be missing something but the name of this variable is a bit misleading to me as the function does not expect a ROT13 version of a string as it first parameter. Instead it will return a ROT13 version of a string.

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
rodrigoprimo: In some cases the `<em>` tag is used to highlight just the function name and in other cases it is used to highlight the whole function call. Is there a reason for that? If not, I suggest highlighting just the function name in all examples.

rodrigoprimo: I can see you addressed this point for most `<standard>` blocks except for the one related to functions used to obfuscate code. Is there a reason for that? Would it make sense to list the functions that trigger this warning in the `<standard>` description as you did for the `<standard>` blocks?

WordPress/Docs/PHP/DiscouragedPHPFunctionsStandard.xml
rodrigoprimo: Apologies if my previous message was not clear. I can see you changed the variable name from `$array` to `$my_array` for all the function calls in this `<code_comparison>` block, but that is not exactly what I meant.

I think it is fine to use `$array` (or `$my_array`) for the functions that accept an array as the first parameter (`wp_json_encode()` and `serialize()`). But I don't think we should use `$array` (or `$my_array`) for functions that don't accept an array as the first parameter (`json_decode()` and `unserialize()`). `$string` (or something similar) would be fine in those two cases from my perspective. Do you agree?

rodrigoprimo:
@jasonkenison, just checking if you are still available/interested in finishing this PR? Let me know if you have any questions about the feedback that I left. Thanks!

rodrigoprimo:
@jasonkenison, please let us know within a week if you are still interested in finishing this PR. If we don't hear back from you, we will presume you don't have time, and we will see if we can find someone else to take over and finish it. Thanks for your work so far!

jasonkenison:
@rodrigoprimo I think some of the PHP stuff I don't ever use (that this sniff is saying we shouldn't use) is beyond me. Might be ideal to pass this off to someone more familiar with these function. Thanks.

rodrigoprimo:
No worries, and thanks for your response and the work you did, @jasonkenison!

PR #2586, opened 2025-08-26
Add GlobalVariablesOverride standard doc

paulopmt1:

We're adding a new xml standard file to [GlobalVariablesOverrideSniff](https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/WP/GlobalVariablesOverrideSniff.php)

image

### DIFF
--- a/WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
+++ b/WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
@@ -5,36 +5,27 @@
     >
     <standard>
     <![CDATA[
-    WordPress global variables should not be overridden. This includes variables imported with 
-    the `global` statement within functions, variables assigned via the `$GLOBALS` superglobal array,
-    and variables assigned in the global namespace.
+    WordPress global variables should not be overridden.
 
-    A few select WordPress global variables like `$content_width` and `$wp_cockneyreplace` are 
+    Two WordPress global variables, `$content_width` and `$wp_cockneyreplace`, are 
     specifically intended to be overridden by themes/plugins and are exempt from this rule.
     ]]>
     </standard>
     <code_comparison>
         <code title="Valid: Different variable name.">
         <![CDATA[
-$my_query = new WP_Query( $args );
+<em>$my_query</em> = new WP_Query( $args );
+
+
+<em>$GLOBALS['my_data']</em> = "some data";
         ]]>
         </code>
         <code title="Invalid: Global variable overridden.">
         <![CDATA[
 <em>global $wp_query;</em>
-<em>$wp_query = new WP_Query( $args );</em>
-        ]]>
-        </code>
-    </code_comparison>
-    <code_comparison>
-        <code title="Valid: Custom Superglobal variable.">
-        <![CDATA[
-$GLOBALS['my_data'] = array( 'key' => 'value' );
-        ]]>
-        </code>
-        <code title="Invalid: WordPress Superglobal variable overridden.">
-        <![CDATA[
-<em>$GLOBALS['wp_query'] = new WP_Query( $args );</em>
+<em>$wp_query</em> = new WP_Query( $args );
+
+<em>$GLOBALS['wp_query']</em> = new WP_Query( $args );
         ]]>
         </code>
     </code_comparison>


WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,41 @@
...
+    <code_comparison>
+        <code title="Valid: Different variable name.">
+        <![CDATA[
+$my_query = new WP_Query( $args );
+        ]]>
+        </code>
+        <code title="Invalid: Global variable overridden.">
+        <![CDATA[
+<em>global $wp_query;</em>
+<em>$wp_query = new WP_Query( $args );</em>
```
rodrigoprimo: For this code example, I would suggest highlighting with the `<em>` tag only the `$wp_query` variable instead of both lines of the code example.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,41 @@
...
+    and variables assigned in the global namespace.
+
+    A few select WordPress global variables like `$content_width` and `$wp_cockneyreplace` are 
+    specifically intended to be overridden by themes/plugins and are exempt from this rule.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Different variable name.">
+        <![CDATA[
+$my_query = new WP_Query( $args );
```
rodrigoprimo: Following the pattern that I suggested below, I think that here you can highlight the `$my_query` variable. Typically, the `<em>` tags are used to highlight both the "dos" as well as the "don'ts".

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,41 @@
...
+        <![CDATA[
+<em>global $wp_query;</em>
+<em>$wp_query = new WP_Query( $args );</em>
+        ]]>
+        </code>
+    </code_comparison>
+    <code_comparison>
+        <code title="Valid: Custom Superglobal variable.">
+        <![CDATA[
+$GLOBALS['my_data'] = array( 'key' => 'value' );
```
rodrigoprimo: See my remarks above regarding the `<em>` tag.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,41 @@
...
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Global Variables Override"
+    >
+    <standard>
+    <![CDATA[
+    WordPress global variables should not be overridden. This includes variables imported with 
+    the `global` statement within functions, variables assigned via the `$GLOBALS` superglobal array,
+    and variables assigned in the global namespace.
+
+    A few select WordPress global variables like `$content_width` and `$wp_cockneyreplace` are
```
rodrigoprimo: Since those are the two only WP global variables that can be overriden, I think this paragraph could be rephrased to remove the "like".

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Global Variables Override"
+    >
+    <standard>
+    <![CDATA[
+    WordPress global variables should not be overridden. This includes variables imported with
```
rodrigoprimo: In my opinion, just saying "WordPress global variables should not be overridden." is enough and second phrase is not necessary, but other reviewers might disagree with me, so feel free to leave it for now if you prefer.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,41 @@
...
+        </code>
+        <code title="Invalid: Global variable overridden.">
+        <![CDATA[
+<em>global $wp_query;</em>
+<em>$wp_query = new WP_Query( $args );</em>
+        ]]>
+        </code>
+    </code_comparison>
+    <code_comparison>
+        <code title="Valid: Custom Superglobal variable.">
```
rodrigoprimo: The superglobal is `$GLOBALS` and not `$GLOBALS['my_data']`. I believe the titles in this `code_comparison` block need to be rephrased. Alternatively, I suggest considering using just a single `code_comparison` block and presenting two code examples. One with a global override using `global $wp_variable_name` and another usigin `$GLOBALS['wp_variable_name']`.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
rodrigoprimo: This line has 49 characters and the maximum is 48.

rodrigoprimo (commented):
Thanks for working on this, @paulopmt1!

I left a few comments. Please let me know if you have any questions.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
paulopmt1: I ended up simplifying it.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,32 @@
...
+    <code_comparison>
+        <code title="Valid: Different variable name.">
+        <![CDATA[
+<em>$my_query</em> = new WP_Query( $args );
+
+
+<em>$GLOBALS['my_data']</em> = "some data";
+        ]]>
+        </code>
+        <code title="Invalid: Global variable overridden.">
```
rodrigoprimo, line 23:
```suggestion
        <code title="Invalid: WordPress global variable overridden.">
```

I think it is important to mention that this is about WordPress globals and not any globals.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,32 @@
...
+    <standard>
+    <![CDATA[
+    WordPress global variables should not be overridden.
+
+    Two WordPress global variables, `$content_width` and `$wp_cockneyreplace`, are 
+    specifically intended to be overridden by themes/plugins and are exempt from this rule.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Different variable name.">
```
rodrigoprimo, line 15: I'm not sure about this title. Maybe `Valid: Using custom variable/key names instead of WordPress globals.` or `Valid: Using custom variable/key names.` is a bit better and more descriptive? Happy to hear other suggestions you might have.

The concern that I have with `Valid: Different variable name.` is that it reads a bit vaguely and doesn't clearly communicate what makes the examples below valid.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,32 @@
...
+
+<em>$GLOBALS['my_data']</em> = "some data";
+        ]]>
+        </code>
+        <code title="Invalid: Global variable overridden.">
+        <![CDATA[
+<em>global $wp_query;</em>
+<em>$wp_query</em> = new WP_Query( $args );
+
+<em>$GLOBALS['wp_query']</em> = new WP_Query( $args );
```
rodrigoprimo, line 28:
```suggestion
<em>$GLOBALS['post']</em> = get_post( 1 );
```

Nitpick: maybe use a different global here to illustrate one more WP global that is flagged by this sniff?

rodrigoprimo (commented):
Thanks for addressing the points in my first review, @paulopmt1! I think this is much better now, I left three more small comments with a few things that occurred to me only after checking the new version of the documentation.

PR #2589, opened 2025-08-26
Add DiscouragedConstants standard doc

paulopmt1:

Surpases [RafaelFunchal](https://github.com/RafaelFunchal)'s change made here: https://github.com/WordPress/WordPress-Coding-Standards/pull/2493

### DIFF
--- a/WordPress/Docs/WP/DiscouragedConstantsStandard.xml
+++ b/WordPress/Docs/WP/DiscouragedConstantsStandard.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Discouraged Constants"
+    >
+    <standard>
+    <![CDATA[
+    The usage of WordPress Constants is discouraged.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: A function is used in the code to retrieve the theme directory.">
+        <![CDATA[
+<img src="
+<?php echo <em>get_template_directory_uri()</em>; ?>
+/img/logo.svg" />
+        ]]>
+        </code>
+        <code title="Invalid: The TEMPLATEPATH constant is used in the code.">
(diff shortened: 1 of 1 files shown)


WordPress/Docs/WP/DiscouragedConstantsStandard.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Discouraged Constants"
+    >
+    <standard>
+    <![CDATA[
+    The use of certain WordPress constants is discouraged. Examples include `TEMPLATEPATH` and `STYLESHEETPATH`.
```
jrfnl, line 8:
```suggestion
    The use and (re-)declaration of certain WordPress native constants is discouraged. Examples include `TEMPLATEPATH` and `STYLESHEETPATH`.
```

WordPress/Docs/WP/DiscouragedConstantsStandard.xml
```diff
@@ -0,0 +1,40 @@
...
+        <![CDATA[
+$subdir = <em>TEMPLATEPATH</em> . '/subdir';
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    The use of certain WordPress constants is discouraged. Examples include `TEMPLATEPATH` and `STYLESHEETPATH`.
+    ]]>
+    </standard>
```
jrfnl, line 27: Duplicate `<standard>` block is unnecessary.

A `<standard>` can have multiple `<code_comparison>` blocks associated with it, so this block should be removed.

jrfnl (commented):
@paulopmt1 Thank you for working on this documentation! Looking good.

Two small remarks:
* I wonder if it would make sense to combine the current code samples into one `<code_comparison>` as they are basically demonstrating the same thing (use of discouraged constant).
* I also wonder whether there should be a code sample demonstrating that (re-)declaring the constants is discouraged.
    I imagine that may warrant it's own `<code_comparison>` block and possibly its own `<standard>` block. (also see my suggestion inline).

PR #2590, opened 2025-08-26
[Update] Documentation for sniff WordPress.Files.FileName

jasonkenison:
Continuing updates from https://github.com/WordPress/WordPress-Coding-Standards/pull/2492

1. Grouped all checks into individual standards.
2. Included additional standards not previously included
3. Added more check valid and invalid examples

Closes #2492

### DIFF
--- a/WordPress/Docs/Files/FileNameStandard.xml
+++ b/WordPress/Docs/Files/FileNameStandard.xml
@@ -30,16 +30,18 @@ other<em>+</em>punctuation<em>#</em>.inc
     ]]>
     </standard>
     <code_comparison>
-        <code title="Valid: File name matches prefix and class.">
+        <code title="Valid: File name matches class name with proper prefix.">
         <![CDATA[
-<em>class-bulk-updater</em>.php // For a class named `Bulk_Updater`.
+<em>class-bulk-updater</em>.php // `Bulk_Updater`.
+<em>class-user-profile</em>.php // `User_Profile`.
+<em>class-export-data</em>.php // `Export_Data`.
         ]]>
         </code>
-        <code title="Invalid: File name missing prefix or class.">
+        <code title="Invalid: File name does not match class name or the prefix is incorrect.">
         <![CDATA[
-class<em>.</em>bulk-updater.php
-<em>bulk-updater</em>.php
-bulk-updater<em>-class</em>.php
+class-<em>update-in-bulk</em>.php // `Bulk_Updater`.
+<em></em>user-profile.php // `User_Profile`.
+<em>class.</em>export-data.php // `Export_Data`.
         ]]>
         </code>
     </code_comparison>
--- a/WordPress/Docs/Files/FileNameStandard.xml
+++ b/WordPress/Docs/Files/FileNameStandard.xml
@@ -54,7 +54,7 @@ bulk-updater<em>-class</em>.php
 general<em>-template</em>.php
         ]]>
         </code>
-        <code title="Invalid: File name missing prefix/suffix.">
+        <code title="Invalid: File name missing suffix.">
         <![CDATA[
 <em>general</em>.php
         ]]>
--- a/WordPress/Docs/Files/FileNameStandard.xml
+++ b/WordPress/Docs/Files/FileNameStandard.xml
@@ -11,9 +11,9 @@
     <code_comparison>
         <code title="Valid: Lowercase and `-` separators.">
         <![CDATA[
-taxonomy<em>-</em>my<em>-</em>term.inc
-<em>some</em>-<em>file</em>.inc
-other<em>-</em>punctuation.inc
+<em>taxonomy-my-term</em>.inc
+<em>some-file</em>.inc
+<em>other-punctuation</em>.inc
         ]]>
         </code>
         <code title="Invalid: Other punctuation or mixed case.">
--- a/WordPress/Docs/Files/FileNameStandard.xml
+++ b/WordPress/Docs/Files/FileNameStandard.xml
@@ -5,11 +5,32 @@
     >
     <standard>
     <![CDATA[
-    Filenames should be all lowercase with hyphens separating words. If `$is_theme` property is set to `true` certain theme specific exceptions are made for underscores.
+File name must not be a mimetype sublevel only file names, such as `plain.php`.
     ]]>
     </standard>
     <code_comparison>
-        <code title="Valid: All lowercase with only hyphens as word separators.">
+        <code title="Valid: Is not a mimetype only file name.">
+        <![CDATA[
+<em>taxonomy-term</em>.inc
+<em>class-test-sample</em>.php
+<em>other-punctuation</em>.inc
+        ]]>
+        </code>
+        <code title="Invalid: Is a mimetype only file name.">
+        <![CDATA[
+<em>text-plain</em>.php
+<em>html</em>.inc
+<em>richtext</em>.php
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+Filenames should be all lowercase with hyphens separating words. If `$is_theme` property is set to `true` certain theme specific exceptions are made for underscores.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Lowercase and `-` separators.">
         <![CDATA[
 taxonomy<em>-</em>my<em>-</em>term.inc
 <em>s</em>ome<em>f</em>ile.inc
@@ -17,12 +38,31 @@ taxonomy<em>-</em>my<em>-</em>term.inc
 other<em>-</em>punctuation.inc
         ]]>
         </code>
-        <code title="Invalid: Use of other punctuation or mixed case.">
+        <code title="Invalid: Other punctuation or mixed case.">
         <![CDATA[
-taxonomy-my<em>_</em>term.inc.
+taxonomy-my<em>_</em>term.inc
 <em>S</em>ome<em>F</em>ile.inc
-<em>testsample</em>-unit-2.inc
+<em>testsample</em>-Unit2.inc
 other<em>+</em>punctuation<em>#</em>.inc
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Class file names should be based on the class name with `class-` prepended.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: File name matches prefix and class.">
+        <![CDATA[
+<em>class-bulkupdater</em>.php
+        ]]>
+        </code>
+        <code title="Invalid: File name missing prefix or class.">
+        <![CDATA[
+<em>class.</em>bulk-updater.php
+<em>bulkupdater-class</em>.php
+class-<em>CustomFormatter</em>.php
         ]]>
         </code>
     </code_comparison>
--- a/WordPress/Docs/Files/FileNameStandard.xml
+++ b/WordPress/Docs/Files/FileNameStandard.xml
@@ -5,30 +5,40 @@
     >
     <standard>
     <![CDATA[
-    Ensures that file names meet the following standards:
-    - Filenames should be all lowercase with hyphens separating words.
-    - Class file names should be based on the class name with "class-" prepended.
-    - Files in `wp-includes` containing template tags should end in `-template`.
-
-    If `$is_theme` property is set to `true` certain theme specific exceptions are made for underscores.
+    Filenames should be all lowercase with hyphens separating words. If `$is_theme` property is set to `true` certain theme specific exceptions are made for underscores.
     ]]>
     </standard>
     <code_comparison>
-        <code title="Valid: All lowercase with only hyphens as word separators and proper use of prefix/suffix.">
+        <code title="Valid: All lowercase with only hyphens as word separators.">
         <![CDATA[
 taxonomy<em>-</em>my<em>-</em>term.inc
 <em>s</em>ome<em>f</em>ile.inc
 <em>class</em>-testsample.inc
 other<em>-</em>punctuation.inc
-general<em>-template</em>.php
         ]]>
         </code>
-        <code title="Invalid: Use of other punctuation, uppercase, or missing prefix/suffix.">
+        <code title="Invalid: Use of other punctuation or mixed case.">
         <![CDATA[
 taxonomy-my<em>_</em>term.inc.
 <em>S</em>ome<em>F</em>ile.inc
 <em>testsample</em>-unit-2.inc
 other<em>+</em>punctuation<em>#</em>.inc
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+   Files in `wp-includes` containing template tags should end in `-template`.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: File name with proper prefix/suffix.">
+        <![CDATA[
+general<em>-template</em>.php
+        ]]>
+        </code>
+        <code title="Invalid: File name missing prefix/suffix.">
+        <![CDATA[
 <em>general</em>.php
         ]]>
         </code>
--- a/WordPress/Docs/Files/FileNameStandard.xml
+++ b/WordPress/Docs/Files/FileNameStandard.xml
@@ -1,0 +1,32 @@
+<documentation title="File Name Standards">
+<standard>
+    <![CDATA[
+        Ensures that file names meet the following standards:
+        - Filenames should be all lowercase with hyphens separating words.
+        - Class file names should be based on the class name with "class-" prepended.
+        - Files in `wp-includes` containing template tags should end in `-template`.
+
+        If `$is_theme` property is set to `true` certain theme specific exceptions are made for underscores.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: All lowercase with only hyphens as word separators and proper use of prefix/suffix.">
+        <![CDATA[
+echo taxonomy<em>-</em>my<em>-</em>term.inc
+echo <em>s</em>ome<em>f</em>ile.inc
+echo <em>class</em>-testsample.inc
+echo other<em>-</em>punctuation.inc
+echo general<em>-template</em>.php
+        ]]>
+        </code>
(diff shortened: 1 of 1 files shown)


WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,86 @@
...
+        <![CDATA[
+<em>class.</em>bulk-updater.php
+<em>bulkupdater-class</em>.php
+class-<em>CustomFormatter</em>.php
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+   Files in `wp-includes` containing template tags should end in `-template`.
```
rodrigoprimo: I believe this description could detail a bit more which files in the `wp-includes` directory should end in `-template`. What do you think of the suggestion below? Feel free to improve it or create something different (note that this line should start with four spaces and not three).

```suggestion
    Non-class files in the directory `wp-includes` with a "@subpackage Template" tag should use the `-template` suffix.
```

WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,86 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd" 
+    title="File Name Standards"
```
rodrigoprimo: The title of the document should match the sniff name.

```suggestion
    title="File Name"
```

WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,86 @@
...
+        ]]>
+        </code>
+        <code title="Invalid: Is a mimetype only file name.">
+        <![CDATA[
+<em>text-plain</em>.php
+<em>html</em>.inc
+<em>richtext</em>.php
+        ]]>
+        </code>
+    </code_comparison>
```
rodrigoprimo: Please correct me if I'm mistaken, but I don't think the sniff has this rule.

This can be tested using one of the invalid examples provided here to see that it does not trigger an error:

```bash
$ vendor/bin/phpcs -s --standard=WordPress --sniffs=WordPress.Files.FileName text-plain.php
$
```

WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,86 @@
...
+    <standard>
+    <![CDATA[
+Filenames should be all lowercase with hyphens separating words. If `$is_theme` property is set to `true` certain theme specific exceptions are made for underscores.
```
rodrigoprimo: This is something that needs to be improved in how sniffs are documented, but currently we only mention the default behavior of the sniff, so I don't think we should mention what happens in `$is_theme` is set to `true`. Also, there should be four spaces at the start of this line.

```suggestion
    Filenames should be all lowercase with hyphens separating words.
```

WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,86 @@
...
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Lowercase and `-` separators.">
+        <![CDATA[
+taxonomy<em>-</em>my<em>-</em>term.inc
+<em>s</em>ome<em>f</em>ile.inc
```
rodrigoprimo: Considering the invalid example related to this one, I believe the valid example should separate the two words with a hyphen, no? This would also fix the error in CI that is a false positive in my opinion, as `ome` is incorrectly being flagged as a typo.

```suggestion
some<em>-</em>file.inc
```

WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,86 @@
...
+<em>class</em>-testsample.inc
```
rodrigoprimo: I'm not sure if I understand this specific example. Why is `class` highlighted, considering that this specific error (`NotHyphenatedLowercase`) does not check for the `class` prefix? If this example does not add anything that is not already covered in the other examples, maybe it can be removed together with the corresponding invalid example?

WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,86 @@
...
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Class file names should be based on the class name with `class-` prepended.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: File name matches prefix and class.">
+        <![CDATA[
+<em>class-bulkupdater</em>.php
```
rodrigoprimo: It depends on the class name, but considering that the idea is to separate words using hyphens and also considering the name of the corresponding invalid example, maybe this one should be `class-bulk-updater.php`?

Related to the above, maybe you can add a code comment with the class name to those examples? And then include an example where the name is invalid because the file name does not match the class name, despite it following the other formatting rules?

Then it might also make sense to include a valid example for each of the invalid examples. Maybe there should be one invalid example without the `class-` prefix and one where the file name is not based on the class name? Maybe there is value to add more that I'm missing.

WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,86 @@
...
+        ]]>
+        </code>
+        <code title="Invalid: File name missing prefix or class.">
+        <![CDATA[
+<em>class.</em>bulk-updater.php
```
rodrigoprimo: I wonder if here it is not better to highlight just the dot as this is the exact part that would cause the error?

```suggestion
class<em>.</em>bulk-updater.php
```

WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,86 @@
...
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+   Files in `wp-includes` containing template tags should end in `-template`.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: File name with proper prefix/suffix.">
```
rodrigoprimo: Maybe remove `prefix` here and in the title below as this particular error is checking just for the suffix of the file and not a prefix?

rodrigoprimo (changes requested):
Thanks for working on this PR, @jasonkenison! I left some comments with suggestions. Let me know if you have any questions.

WordPress/Docs/Files/FileNameStandard.xml
jasonkenison: @rodrigoprimo my mistake. The THEME_EXCEPTIONS regex comment on line 43 seemed to indicate it would "not allow for mimetype sublevel only file names". I'll remove it.

WordPress/Docs/Files/FileNameStandard.xml
jasonkenison: @rodrigoprimo might be unrelated to this sniff, but if a class is called "BulkUpdater" this would still include a '-' in the filename? Updated as requested, but can make adjustments as needed.

Examples elsewhere in wp-includes, showing 3 instances where these code standard rules aren't followed:
- class-phppass.php contains class "PasswordHash" (does not match filename)
- class-avif-info.php contains class "Avifinfo" (not camel cased)
- class-walker-category-dropdown.php contains class "Walker_CategoryDropdown" (underscore in class)

WordPress/Docs/Files/FileNameStandard.xml
jasonkenison: @rodrigoprimo these were tricky, was trying to highlight where non-lowercase letters were used. To get past the spelling error, it now highlights `<em>Some</em>` as a complete word instead of just the character.

WordPress/Docs/Files/FileNameStandard.xml
rodrigoprimo: No worries.

I believe this regex is only used to allow for some exceptions when checking theme files when `$is_theme` is set to `true`. Since the default value of this property is `false`, we don't need to document it here. The convention is to document only the default behavior of the sniff.

I believe this should change, and we should include the non-default behavior as well, but that is a separate discussion (just mentioning it to give you more context).

WordPress/Docs/Files/FileNameStandard.xml
rodrigoprimo: > @rodrigoprimo might be unrelated to this sniff, but if a class is called "BulkUpdater" this would still include a '-' in the filename? Updated as requested, but can make adjustments as needed.

@jasonkenison, good question. If the class is called `BulkUpdater`, the expected file name is `class-bulkupdater.php` (without the hyphen). I suggested including the hyphen because the corresponding invalid example already has a hyphen separating `bulk` and `updater`. Adding a code comment with the class name, as suggested, should also help clarify this. This is more or less what I have in mind:

```
<em>class-bulk-updater</em>.php // For a class named `Bulk_Updater`.
```

You can run PHPCS to test it. For example, given the following file named `class-bulk-updater.php`:

```php
<?php

class BulkUpdater {}
```

We get the following error when running this sniff:

```
vendor/bin/phpcs -s --standard=WordPress --sniffs=WordPress.Files.FileName class-bulk-updater.php 

FILE: class-bulk-updater.php
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 1 | ERROR | Class file names should be based on the class name with "class-" prepended. Expected class-bulkupdater.php, but found class-bulk-updater.php. (WordPress.Files.FileName.InvalidClassFileName)
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Time: 107ms; Memory: 12MB

```

> Examples elsewhere in wp-includes, showing 3 instances where these code standard rules aren't followed:

I was unable to find `class-phppass.php` in the WP repository. WP excludes the file `class-avif-info.php` from the PHPCS checks (https://github.com/WordPress/wordpress-develop/blob/1139a51e9c6608dc235938d9d234ed185c5d5706/phpcs.xml.dist#L55). `class-walker-category-dropdown.php` does trigger a warning (it is a warning and not an error because WP changes the configuration of this particular sniff: https://github.com/WordPress/wordpress-develop/blob/1139a51e9c6608dc235938d9d234ed185c5d5706/phpcs.xml.dist#L140-L142):

```
$ ./vendor/squizlabs/php_codesniffer/bin/phpcs ./src/wp-includes/class-walker-category-dropdown.php                                      
W 1 / 1 (100%)



FILE: src/wp-includes/class-walker-category-dropdown.php
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 1 | WARNING | Class file names should be based on the class name with "class-" prepended. Expected class-walker-categorydropdown.php, but found class-walker-category-dropdown.php. (WordPress.Files.FileName.InvalidClassFileName)
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Time: 476ms; Memory: 134MB

```

WordPress/Docs/Files/FileNameStandard.xml
rodrigoprimo: For the valid examples of the first two errors, as they check if the whole name matches what is expect, maybe we can highlight the full name (excluding the period and the file extension)?

For the third error, I think it is better to highlight just `-template` as it is already the case.

For the invalid examples, I think you could highlight just the part that is causing the error.

rodrigoprimo:
@jasonkenison, I was just wondering if you'll have a chance to finish this off in the near future. It would be great if this PR could be included in the next WPCS release.

WordPress/Docs/Files/FileNameStandard.xml
jasonkenison: remove'ed 'prefix' from the invalid line too.

WordPress/Docs/Files/FileNameStandard.xml
jasonkenison: @rodrigoprimo I'm not sure how to solve this one, but am happy to discuss or try recommendations.

WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,63 @@
...
+        ]]>
+        </code>
+        <code title="Invalid: File name missing prefix or class.">
+        <![CDATA[
+class<em>.</em>bulk-updater.php
+<em>bulk-updater</em>.php
+bulk-updater<em>-class</em>.php
+        ]]>
+        </code>
+    </code_comparison>
```
rodrigoprimo, line 47: > @rodrigoprimo I'm not sure how to solve this one, but am happy to discuss or try recommendations.

Regarding your question [here](https://github.com/WordPress/WordPress-Coding-Standards/pull/2590#discussion_r2461022013), this is more or less what I have in mind. What do you think?

```suggestion
    <code_comparison>
        <code title="Valid: File name matches class name with proper prefix.">
        <![CDATA[
<em>class-bulk-updater</em>.php // `Bulk_Updater`.
<em>class-user-profile</em>.php // `User_Profile`.
<em>class-export-data</em>.php // `Export_Data`.
        ]]>
        </code>
        <code title="Invalid: File name does not match class name or the prefix is incorrect.">
        <![CDATA[
class-<em>update-in-bulk</em>.php // `Bulk_Updater`.
<em></em>user-profile.php // `User_Profile`.
<em>class.</em>export-data.php // `Export_Data`.
        ]]>
        </code>
    </code_comparison>
```

jasonkenison: @rodrigoprimo Change made in the latest commit

rodrigoprimo: Thanks, @jasonkenison. This one can be marked as resolved, but I believe https://github.com/WordPress/WordPress-Coding-Standards/pull/2590#discussion_r2323107260 still need to be addressed?

WordPress/Docs/Files/FileNameStandard.xml
jasonkenison: @rodrigoprimo that makes sense. Highlighted the full valid filename string minus the .ext.

rodrigoprimo (approved):
Thanks for addressing the last comment and for all your work, @jasonkenison! This PR looks good to me.

rodrigoprimo:
@dingo-d, @GaryJones, and @jrfnl, just noting that @jasonkenison addressed all the points that I raised a while ago, and this PR is ready for review whenever any of you have the time to check it. Thanks!

WordPress/Docs/Files/FileNameStandard.xml
jrfnl: Just FYI: the `class-phppass.php` file is also excluded from the CS check via configuration as this is an externally maintained dependency for WP.

WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,65 @@
...
+    <![CDATA[
+    Class file names should be based on the class name with `class-` prepended.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: File name matches class name with proper prefix.">
+        <![CDATA[
+<em>class-bulk-updater</em>.php // `Bulk_Updater`.
+<em>class-user-profile</em>.php // `User_Profile`.
+<em>class-export-data</em>.php // `Export_Data`.
```
jrfnl, line 37:
```suggestion
// File: <em>class-bulk-updater</em>.php
<?php
class Bulk_Updater {
    ...
}

// File: <em>class-user-profile</em>.php
<?php
class User_Profile{
    ...
}

// File: <em>class-export-data</em>.php
<?php
class Export_Data{
    ...
}
```

WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,65 @@
...
+        ]]>
+        </code>
+        <code title="Invalid: File name does not match class name or the prefix is incorrect.">
+        <![CDATA[
+class-<em>update-in-bulk</em>.php // `Bulk_Updater`.
+<em></em>user-profile.php // `User_Profile`.
+<em>class.</em>export-data.php // `Export_Data`.
```
jrfnl, line 44:
```suggestion
// File: class-<em>update-in-bulk</em>.php
<?php
class Bulk_Updater {
    ...
}

// File: <em></em>user-profile.php
<?php
class User_Profile{
    ...
}

// File: <em>class.</em>export-data.php
<?php
class Export_Data{
    ...
}
```

WordPress/Docs/Files/FileNameStandard.xml
```diff
@@ -0,0 +1,65 @@
...
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Non-class files in the directory `wp-includes` with a "@subpackage Template" tag should use the `-template` suffix.
```
jrfnl:
```suggestion
    WP Core only: Non-class files in the directory `wp-includes` with a "@subpackage Template" tag should use the `-template` suffix.
```

jrfnl (commented):
Thanks @jasonkenison for continuing to work on this and @rodrigoprimo for doing the initial reviews!

This is mostly looking good.

I'm not 100% convinced that the `// Bulk_Updater.` comments after the file names for the second `<standard>` block actually make it clear what is expected.

I've left some suggestions in-line on how I believe these code samples can be improved, but if I'm honest, I think splitting the single `<code_comparison>` block into three separate `<code_comparison>` blocks with descriptive valid/invalid titles would make this even clearer.

Other than that, I think it may be worthwhile to make it clear that the `-template` suffix rule only applies to specific files in WP Core and is not applicable for plugins/themes etc.

Hope this feedback helps!

rodrigoprimo:
@jasonkenison, I was just wondering if you'll have a chance to finish this off in the near future. It would be great if this PR could be included in the next WPCS release.

rodrigoprimo:
Hi, @jasonkenison. I see you made a commit after my last comment, but there are still some outstanding items from the last review. Just checking if you are still planning to finish this. If you haven't got time or lost interest, please let us know, and we'll see if we can find someone to take over. Thanks!

PR #2591, opened 2025-08-26
Docs: Add documentation for WordPress.PHP.TypeCasts

brentwilson-clariio:
Related to #1722

This PR adds a new documentation file:
`WordPress/Docs/PHP/TypeCastsStandard.xml`

### What it covers
Documents the rule enforced by the `WordPress.PHP.TypeCasts` sniff:
- Normalizes legacy float casts (`(double)`, `(real)`) to `(float)`.

### Notes
Other type-cast related rules in the WordPress standard are enforced by separate sniffs:
- Short form keywords (`(int)`, `(bool)`, `(string)`): PSR12.Keywords.ShortFormTypeKeywords
- No spaces inside parentheses: Squiz.WhiteSpace.CastSpacing
- Exactly one space after the cast: Generic.Formatting.SpaceAfterCast

### DIFF
--- a/WordPress/Docs/PHP/TypeCastsStandard.xml
+++ b/WordPress/Docs/PHP/TypeCastsStandard.xml
@@ -1,37 +1,55 @@
 <?xml version="1.0"?>
-<documentation title="Type Casts">
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Type Casts"
+>
     <standard>
     <![CDATA[
-      This sniff normalizes float type cast keywords.
+      Enforce normalized and safe type casts.
 
-      Use (float) instead of legacy synonyms:
-      - Disallowed: (double), (real)
-      - Allowed:    (float)
-
-      Note: Other type-cast rules in the WordPress standard
-      are enforced by different sniffs:
-      - Short forms like (int), (bool), (string):
-        PSR12.Keywords.ShortFormTypeKeywords
-      - No spaces inside parentheses:
-        Squiz.WhiteSpace.CastSpacing
-      - Exactly one space after the cast:
-        Generic.Formatting.SpaceAfterCast
+      • Use (float) instead of legacy float casts.
+      • Do not use the (unset) cast. Use unset().
+      • Avoid the (binary) cast (discouraged).
     ]]>
     </standard>
 
     <code_comparison>
-        <code title="Invalid: legacy float casts">
+        <code title="Invalid: Using legacy float casts.">
         <![CDATA[
-<?php
 $price = <em>(double)</em> $value;
 $size  = <em>(real)</em>   $value;
         ]]>
         </code>
-        <code title="Valid: normalized float cast">
+        <code title="Valid: Using normalized float casts.">
         <![CDATA[
-<?php
 $price = <em>(float)</em> $value;
 $size  = <em>(float)</em> $value;
+        ]]>
+        </code>
+    </code_comparison>
+
+    <code_comparison>
+        <code title="Invalid: Using the (unset) cast.">
+        <![CDATA[
+$result = <em>(unset)</em> $value;
+        ]]>
+        </code>
+        <code title="Valid: Use unset().">
+        <![CDATA[
+unset( $value );
+        ]]>
+        </code>
+    </code_comparison>
+
+    <code_comparison>
+        <code title="Invalid: Using the (binary) cast.">
+        <![CDATA[
+$bytes = <em>(binary)</em> $value;
+        ]]>
+        </code>
+        <code title="Valid: Prefer a clear alternative.">
+        <![CDATA[
+$bytes = (string) $value;
         ]]>
         </code>
     </code_comparison>
--- a/WordPress/Docs/PHP/TypeCastsStandard.xml
+++ b/WordPress/Docs/PHP/TypeCastsStandard.xml
@@ -14,42 +14,42 @@
     </standard>
 
     <code_comparison>
-        <code title="Invalid: Using legacy float casts.">
-        <![CDATA[
-$price = <em>(double)</em> $value;
-$size  = <em>(real)</em>   $value;
-        ]]>
-        </code>
         <code title="Valid: Using normalized float casts.">
         <![CDATA[
 $price = <em>(float)</em> $value;
 $size  = <em>(float)</em> $value;
         ]]>
         </code>
+        <code title="Invalid: Using legacy float casts.">
+        <![CDATA[
+$price = <em>(double)</em> $value;
+$size  = <em>(real)</em>   $value;
+        ]]>
+        </code>
     </code_comparison>
 
     <code_comparison>
-        <code title="Invalid: Using the (unset) cast.">
-        <![CDATA[
-$result = <em>(unset)</em> $value;
-        ]]>
-        </code>
         <code title="Valid: Use unset().">
         <![CDATA[
 unset( $value );
         ]]>
         </code>
+        <code title="Invalid: Using the (unset) cast.">
+        <![CDATA[
+result = <em>(unset)</em> $value;
+        ]]>
+        </code>
     </code_comparison>
 
     <code_comparison>
-        <code title="Invalid: Using the (binary) cast.">
-        <![CDATA[
-$bytes = <em>(binary)</em> $value;
-        ]]>
-        </code>
         <code title="Valid: Prefer a clear alternative.">
         <![CDATA[
-$bytes = (string) $value;
+bytes = (string) $value;
+        ]]>
+        </code>
+        <code title="Invalid: Using the (binary) cast.">
+        <![CDATA[
+bytes = <em>(binary)</em> $value;
         ]]>
         </code>
     </code_comparison>
--- a/WordPress/Docs/PHP/TypeCastsStandard.xml
+++ b/WordPress/Docs/PHP/TypeCastsStandard.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0"?>
+<documentation title="Type Casts">
+    <standard>
+    <![CDATA[
+      This sniff normalizes float type cast keywords.
+
+      Use (float) instead of legacy synonyms:
+      - Disallowed: (double), (real)
+      - Allowed:    (float)
+
+      Note: Other type-cast rules in the WordPress standard
+      are enforced by different sniffs:
+      - Short forms like (int), (bool), (string):
+        PSR12.Keywords.ShortFormTypeKeywords
+      - No spaces inside parentheses:
+        Squiz.WhiteSpace.CastSpacing
+      - Exactly one space after the cast:
+        Generic.Formatting.SpaceAfterCast
+    ]]>
+    </standard>
+
+    <code_comparison>
+        <code title="Invalid: legacy float casts">
+        <![CDATA[
+<?php
+$price = <em>(double)</em> $value;
+$size  = <em>(real)</em>   $value;
+        ]]>
+        </code>
+        <code title="Valid: normalized float cast">
+        <![CDATA[
+<?php
+$price = <em>(float)</em> $value;
+$size  = <em>(float)</em> $value;
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>


WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,38 @@
...
+      - Allowed:    (float)
+
+      Note: Other type-cast rules in the WordPress standard
+      are enforced by different sniffs:
+      - Short forms like (int), (bool), (string):
+        PSR12.Keywords.ShortFormTypeKeywords
+      - No spaces inside parentheses:
+        Squiz.WhiteSpace.CastSpacing
+      - Exactly one space after the cast:
+        Generic.Formatting.SpaceAfterCast
```
jrfnl: This whole block doesn't belong here. Each sniff functions independently of other sniffs. That also means that the docs should be independent of each other and should just focus on what *this* sniff does.

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,38 @@
...
+    ]]>
+    </standard>
+
+    <code_comparison>
+        <code title="Invalid: legacy float casts">
+        <![CDATA[
+<?php
```
jrfnl: PHP open tag is not needed here. (Same for the "invalid" code sample)

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0"?>
+<documentation title="Type Casts">
+    <standard>
+    <![CDATA[
+      This sniff normalizes float type cast keywords.
```
jrfnl: This sentence seems redundant.

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0"?>
+<documentation title="Type Casts">
```
jrfnl:
```suggestion
<?xml version="1.0"?>
<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
    title="Type Casts"
    >
```

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,38 @@
...
+    <standard>
+    <![CDATA[
+      This sniff normalizes float type cast keywords.
+
+      Use (float) instead of legacy synonyms:
+      - Disallowed: (double), (real)
+      - Allowed:    (float)
```
jrfnl: This is incorrect/incomplete.

The sniff:
* Forbids the use of non-standard casts for floats.
* Forbids the use of the `(unset)` cast.
* Discourages the use of the `(binary)` cast.


Considering the above, I believe the code samples also need work.

jrfnl (changes requested):
@brentwilson-clariio Thanks for picking this up! I've just had a quick look and left some feedback for you to consider.

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,38 @@
...
+        PSR12.Keywords.ShortFormTypeKeywords
+      - No spaces inside parentheses:
+        Squiz.WhiteSpace.CastSpacing
+      - Exactly one space after the cast:
+        Generic.Formatting.SpaceAfterCast
+    ]]>
+    </standard>
+
+    <code_comparison>
+        <code title="Invalid: legacy float casts">
```
jrfnl:
```suggestion
        <code title="Invalid: Using legacy float casts.">
```

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,38 @@
...
+        <![CDATA[
+<?php
+$price = <em>(double)</em> $value;
+$size  = <em>(real)</em>   $value;
+        ]]>
+        </code>
+        <code title="Valid: normalized float cast">
```
jrfnl:
```suggestion
        <code title="Valid: Using normalized float casts.">
```

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,38 @@
...
+      - Short forms like (int), (bool), (string):
+        PSR12.Keywords.ShortFormTypeKeywords
+      - No spaces inside parentheses:
+        Squiz.WhiteSpace.CastSpacing
+      - Exactly one space after the cast:
+        Generic.Formatting.SpaceAfterCast
+    ]]>
+    </standard>
+
+    <code_comparison>
```
jrfnl, line 16: The code comparison should have "valid" on the left (first code sample), "invalid" on the right (second code sample).

WordPress/Docs/PHP/TypeCastsStandard.xml
brentwilson-clariio: Thanks! I’ve removed the paragraph referencing other sniffs. The docs now only describe what WordPress.PHP.TypeCasts enforces.

WordPress/Docs/PHP/TypeCastsStandard.xml
brentwilson-clariio: Removed <?php from all code samples as suggested.

WordPress/Docs/PHP/TypeCastsStandard.xml
brentwilson-clariio: Reworded the opener to be shorter and clearer. It now states the rules enforced without repeating itself.

WordPress/Docs/PHP/TypeCastsStandard.xml
brentwilson-clariio: Updated the standard text and the code examples to cover all three behaviors:
	•	(double)/(real) → error (normalize to (float))
	•	(unset) → error (removed in PHP 8.0; suggest unset())
	•	(binary) → warning (discouraged)

WordPress/Docs/PHP/TypeCastsStandard.xml
brentwilson-clariio: Added the schema header and xsi:noNamespaceSchemaLocation pointing to phpcsdocs.xsd.

WordPress/Docs/PHP/TypeCastsStandard.xml
brentwilson-clariio: Adjusted the code block titles to your suggested wording.

WordPress/Docs/PHP/TypeCastsStandard.xml
brentwilson-clariio: Adjusted the code block titles to your suggested wording.

WordPress/Docs/PHP/TypeCastsStandard.xml
brentwilson-clariio: Thanks for the feedback! I’ve reordered the code examples so that Valid is on the left and Invalid is on the right, as suggested.

WordPress/Docs/PHP/TypeCastsStandard.xml
brentwilson-clariio: Updated code block titles to “Valid: Using normalized float casts.” (and similar phrasing for the other examples), as suggested.

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Type Casts"
+>
```
rodrigoprimo, line 5:
```suggestion
    >
```

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,56 @@
...
+
+    <code_comparison>
+        <code title="Valid: Use unset().">
+        <![CDATA[
+unset( $value );
+        ]]>
+        </code>
+        <code title="Invalid: Using the (unset) cast.">
+        <![CDATA[
+result = <em>(unset)</em> $value;
```
rodrigoprimo, line 39:
```suggestion
$result = <em>(unset)</em> $value;
```

`bytes` below is also missing a `$`.

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,56 @@
...
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Type Casts"
+>
+    <standard>
+    <![CDATA[
+      Enforce normalized and safe type casts.
+
+      • Use (float) instead of legacy float casts.
+      • Do not use the (unset) cast. Use unset().
+      • Avoid the (binary) cast (discouraged).
```
rodrigoprimo, line 12:
```suggestion
    Enforce normalized and safe type casts.

    • Use (float) instead of legacy float casts.
    • Do not use the (unset) cast. Use unset().
    • Avoid the (binary) cast (discouraged).
```

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,56 @@
...
+        </code>
+        <code title="Invalid: Using the (unset) cast.">
+        <![CDATA[
+result = <em>(unset)</em> $value;
+        ]]>
+        </code>
+    </code_comparison>
+
+    <code_comparison>
+        <code title="Valid: Prefer a clear alternative.">
```
rodrigoprimo, line 45: Maybe "Valid: Using (string) cast instead."? This way the title clearly communicates what is the alternative like the other titles.

That being said, I noticed that the sniff also detects the binary string prefix notation `b"string"` which is also tokenized as `T_BINARY_CAST`, but in this case using `(string)` is not the correct alternative. I believe we should explicitly mention in the documentation that `b"string"` also trigger the sniff both here with an example and in the standard description above.

In that case, the title could be something like "Valid: (binary) is not used." or "Valid: Use (string) or regular strings."

Another option is to create a separate `<code_comparison>` block for the binary string. Tipically, sniff documentation uses a single `<code_comparison>` block per warning/error message that is triggered by the sniff. But in this case it might make sense to have a separate block.

 What do you think?

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,56 @@
...
+        <![CDATA[
+bytes = (string) $value;
```
rodrigoprimo, line 47:
```suggestion
bytes = <em>(string)</em> $value;
```

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Type Casts"
+>
+    <standard>
+    <![CDATA[
+      Enforce normalized and safe type casts.
```
rodrigoprimo, line 8: I'm unsure about a few things in this paragraph:

- Mentioning that the casts that are not recommended are not safe. Are they all really not safe? I can see why using the casts that were removed is not safe, but not all casts discouraged by the sniff were removed.
- Maybe we could explicitly mention the legacy float casts?
- I don't think it is necessary to mention in parenthesis that `(binary)` is discouraged. If we want to mention that maybe we can include it in the main sentence?

Here is a suggestion of a possible alternative for description. Feel free to create your own version.

```
Type casts must use normalized keywords.

    • Use (float) not (double) or (real).
    • Do not use the (unset) cast. Use unset().
    • Do not use the (binary) cast or b"string" prefix. Use (string) casts or regular string literals instead.
```

WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,56 @@
...
+$price = <em>(double)</em> $value;
+$size  = <em>(real)</em>   $value;
+        ]]>
+        </code>
+    </code_comparison>
+
+    <code_comparison>
+        <code title="Valid: Use unset().">
+        <![CDATA[
+unset( $value );
```
rodrigoprimo, line 34:
```suggestion
<em>unset</em>( $value );
```

rodrigoprimo (changes requested):
Thanks for this PR, @brentwilson-clariio! I'm genuinely sorry for the delay to respond. There was some miscommunication and I didn't realize the WPCS maintainers were waiting for me to post a review. I will make sure to respond faster going forward.

I checked and confirmed that you addressed the points that Juliette raised. I left a few other remarks for us to discuss. Let me know what you think and if you have any questions.

Thanks again.

rodrigoprimo:
@brentwilson-clariio , I was just wondering if you'll have a chance to finish this off in the near future. It would be great if this PR could be included in the next WPCS release.

If you haven't got time or lost interest, please let us know, and we'll see if we can find someone to take over. Thanks!

brentwilson-clariio:
@rodrigoprimo Let me see if I can circle back to this in the next week or two to get it finished up.

rodrigoprimo:
@brentwilson-clariio, please let us know within a week if you are still interested in finishing this PR. If we don't hear back from you, we will presume you don't have time, and we will see if we can find someone else to take over and finish it. Thanks for your work so far!

brentwilson-clariio:
@rodrigoprimo Sorry for the delay. Please feel free to pass this task along to someone else.

rodrigoprimo:
No worries and thanks for the work you put into this, @brentwilson-clariio!

brentwilson-clariio:
When I worked on this back on Contributor Day at WCUS 2025, I did so using AI. I believe it was before the ban on using AI for contributions was announced. But I just want to be transparent about that.

PR #2592, opened 2025-08-26
[Update] Documentation for sniff WordPress.Security.PluginMenuSlug

jasonkenison:
Related to #1722

Initial Docs/Security/PluginMenuSlugStandard.xml

Describes check for `__FILE__` in plugin menu slugs.


WordPress/Docs/Security/PluginMenuSlugStandard.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Plugin Menu Slug"
+    >
+    <standard>
+    <![CDATA[
+    Plugin functions which can be used to add pages to the WP Admin menu should not include `__FILE__` for page registration.
```
rodrigoprimo: I suggest expanding a bit the description. Maybe something like the suggestion below:

```suggestion
    WordPress functions that can be used to add pages to the WP Admin menu should not use `__FILE__` for the menu slug (or parent menu slug) parameter to avoid revealing system paths.
```

I replaced "Plugin functions" with "WordPress functions" as I imagine those functions can be used in a theme as well, and I also included the reason why `__FILE__` should not be used, plus a few other smaller changes. I'm not a native English speaker, so please feel free to improve my suggestion if something doesn't read very well.

WordPress/Docs/Security/PluginMenuSlugStandard.xml
```diff
@@ -0,0 +1,35 @@
...
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Slug does not include `__FILE__`.">
+        <![CDATA[
+add_menu_page(
```
rodrigoprimo, line 14: Maybe it is worth adding a second valid/invalid example with `add_submenu_page()` as this is the only function that has a different signature than the others, and where `__FILE__` should not be used for the `parent_slug` as well as the `menu_slug`?

There is no need to create new `<code>` blocks, you can add the new example in the `<code>` blocks that already exist.

WordPress/Docs/Security/PluginMenuSlugStandard.xml
```diff
@@ -0,0 +1,35 @@
...
+    'My Plugin Main Page',
+    'My Plugin',
+    'manage_options',
+    <em>my-plugin-main</em>,
```
rodrigoprimo: I believe it should be (missing quotes)

```suggestion
    <em>'my-plugin-main'</em>,
```

rodrigoprimo (commented):
Thanks for working on this @jasonkenison! It is looking good. I left a few comments with suggestions to improve it.

WordPress/Docs/Security/PluginMenuSlugStandard.xml
jasonkenison: @rodrigoprimo I updated the description as requested. As a native English speaker, your suggestion reads clear to me.

WordPress/Docs/Security/PluginMenuSlugStandard.xml
jasonkenison: @rodrigoprimo Good call. I added `add_submenu_page()` examples to the existing code blocks with similar `__FILE__` examples.

WordPress/Docs/Security/PluginMenuSlugStandard.xml
jasonkenison: @rodrigoprimo You are correct, quotes are required here. Updated in my latest commit.

WordPress/Docs/Security/PluginMenuSlugStandard.xml
rodrigoprimo: Thanks!

WordPress/Docs/Security/PluginMenuSlugStandard.xml
```diff
@@ -0,0 +1,53 @@
...
+add_menu_page(
+    'My Plugin Main Page',
+    'My Plugin',
+    'manage_options',
+    <em>__FILE__</em>,
+    'my_plugin_main_page'
+);
+
+add_submenu_page(
+    'my_plugin_main_page',
```
rodrigoprimo: @jasonkenison, one final thing for consideration now that you added an example with the `add_submenu_page()`, maybe it is worth highlighting the first parameter of this function in the valid example, and then changing the value passed to this parameter (and highlighting it as well) to use `__FILE__`? This should make it clearer why an example with this function was added to the documentation (the sniff checks what is passed to both `$parent_slug` and `$menu_slug` instead of just `$menu_slug` as it happens with the other functions).

If you are going to include an invalid example in this parameter, I would suggest adding `__FILE__` plus string concatenation or some other approach to illustrate a different case that also triggers the sniff. For example, the tests for this sniff contain `__FILE__ . 'parent'` ([WordPress/Tests/Security/PluginMenuSlugUnitTest.inc#L9](https://github.com/WordPress/WordPress-Coding-Standards/blob/569c7541808fdae01fc1f8292b73491290f70ea5/WordPress/Tests/Security/PluginMenuSlugUnitTest.inc#L9)).

What do you think?

jasonkenison: @rodrigoprimo agreed. I'll make this change.

WordPress/Docs/Security/PluginMenuSlugStandard.xml
```diff
@@ -0,0 +1,53 @@
...
+);
+        ]]>
+        </code>
+        <code title="Invalid: Slug includes `__FILE__`.">
+        <![CDATA[
+add_menu_page(
+    'My Plugin Main Page',
+    'My Plugin',
+    'manage_options',
+    <em>__FILE__ . </em>'my-plugin-subpage',
```
rodrigoprimo: I apologize if my original comment was unclear. What I had in mind was to keep one example with just `__FILE__`, and since you were going to add another example using `add_submenu_page()`, use it as an opportunity to show that the sniff is triggered as well when `__FILE__` is concatenated with something else.

So, I would suggest reverting the change in this line to keep it as it was with just `__FILE__`. What do you think?

```suggestion
    <em>__FILE__</em>,
```

WordPress/Docs/Security/PluginMenuSlugStandard.xml
```diff
@@ -0,0 +1,53 @@
...
+    'my_plugin_main_page'
+);
+
+add_submenu_page(
+    <em>__FILE__ . </em>'my_plugin_main_page',
```
rodrigoprimo: I would suggest highlighting just `__FILE__` as the concatenation operator is not something that triggers the sniff.

```suggestion
    <em>__FILE__</em> . 'my_plugin_main_page',
```

rodrigoprimo (commented):
Thanks for your work, @jasonkenison! This is looking good.

I left just two final comments with nitpicks. Let me know if you agree or not.

WordPress/Docs/Security/PluginMenuSlugStandard.xml
jasonkenison: @rodrigoprimo I believe I already made this change. Which change requests are still outstanding? I'm happy to finish them up today.

WordPress/Docs/Security/PluginMenuSlugStandard.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Plugin Menu Slug"
+    >
+    <standard>
+    <![CDATA[
+    Wordpress functions that can be used to add pages to the WP Admin menu should not include `__FILE__` for the menu slug (or parent menu slug) parameter to avoid revealing system paths.
```
rodrigoprimo:
```suggestion
    WordPress functions that can be used to add pages to the WP Admin menu should not include `__FILE__` for the menu slug (or parent menu slug) parameter to avoid revealing system paths.
```

jasonkenison: Typo fixed

rodrigoprimo:
@dingo-d, @GaryJones, and @jrfnl, just noting that @jasonkenison addressed all the points that I raised a while ago, and this PR is ready for review whenever any of you have the time to check it. Thanks!

jrfnl (approved):
@jasonkenison Thank you for working on this. Merging now. 🎉

PR #2593, opened 2025-08-26
Docs/array declaration spacing

matt-galdino:
Related to PR https://github.com/WordPress/WordPress-Coding-Standards/issues/1722

Surpasses [RafaelFunchal's](https://github.com/RafaelFunchal) changes made here: #2489

image

### DIFF
--- a/WordPress/Docs/Arrays/ArrayDeclarationSpacingStandard.xml
+++ b/WordPress/Docs/Arrays/ArrayDeclarationSpacingStandard.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Array Declaration Spacing"
+    >
+    <standard>
+    <![CDATA[
+    When an array uses associative keys, each pair of key and value must start on a new line.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: There is only one pair of key and value per line.">
+        <![CDATA[
+$args = array(
+    <em>'post_id' => 22,
+    'category' => 1</em>
+);
+        ]]>
+        </code>
+        <code title="Invalid: More than one pair of key and value per line.">
+        <![CDATA[
+$args = array(
+    <em>'post_id' => 22, 'category' => 1</em>
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Each item in a multi-line array must be on a new line.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Only one array item per line.">
+        <![CDATA[
+$args = array(
+    <em>'post_id',
+    'comment_count',
+    'post_type'</em>
+);
+        ]]>
+        </code>
+        <code title="Invalid: Multiple array items in the same line.">
+        <![CDATA[
+$args = array(
+    <em>'post_id', 'comment_count', 'post_type'</em>
+);
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>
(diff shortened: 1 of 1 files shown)


WordPress/Docs/Arrays/ArrayDeclarationSpacingStandard.xml
```diff
@@ -0,0 +1,75 @@
...
+        </code>
+        <code title="Invalid: Multiple key-value pairs on the same line in a multi-line array.">
+        <![CDATA[
+$settings = array(
+    <em>'width' => 300, 'height' => 200</em>,
+    <em>'color' => 'blue'</em>,
+);
+        ]]>
+        </code>
+    </code_comparison>
```
rodrigoprimo, line 74: I'm sorry if my comment in the original PR was not clear, but in https://github.com/WordPress/WordPress-Coding-Standards/pull/2489/files#r1775277204, I was not suggesting to add a new `<standard>/<code_comparison>` block. I don't think this is necessary. Instead, I was suggesting adding a new valid/invalid example to the `<code_comparison>` block that already exists. `<code_comparison>` blocks can have multiple examples when needed. Or do you see a reason to have a separate `<standard>/<code_comparison>` block in this case?

WordPress/Docs/Arrays/ArrayDeclarationSpacingStandard.xml
```diff
@@ -0,0 +1,75 @@
...
+    <em>'post_id'  => 22</em>,
+    <em>'category' => 1</em>,
+);
+        ]]>
+        </code>
+        <code title="Invalid: More than one key/value pair per line.">
+        <![CDATA[
+$args = array(
+    <em>'post_id' => 22, 'category' => 1</em>,
+);
```
rodrigoprimo, line 24: One thing that I missed in my original review is that in order for this example to trigger the `AssociativeArrayFound` error, which I believe is the intention of this example, everything should be in the same line. Currently, this example triggers the `ArrayItemNoNewLine` error that is already covered in the `<standard>/<code_comparison>` block below. Could you please look into that? Let me know if you need any help. You might need to adjust the example so that it fits in a single line because of the 48 characters per line limit (not counting the `<em>` tags).

WordPress/Docs/Arrays/ArrayDeclarationSpacingStandard.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Array Declaration Spacing"
+    >
+    <standard>
+    <![CDATA[
+    When an array uses keys, each key/value pair must start on a new line.
```
rodrigoprimo, line 8: There is something else that I missed in my original review. Using the default configuration of the sniff, which is what we want to document the XML file, the error  `ArrayItemNoNewLine` is only triggered for multi-item single-line arrays, so it might be worth including this information in the description and in the code comparison titles.

```suggestion
    When a multi-item array uses keys, each key/value pair must start on a new line.
```

WordPress/Docs/Arrays/ArrayDeclarationSpacingStandard.xml
```diff
@@ -0,0 +1,75 @@
...
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: There is only one key/value pair per line.">
```
rodrigoprimo, line 12:
```suggestion
        <code title="Valid: Only one key/value pair per line on a multi-item array.">
```

WordPress/Docs/Arrays/ArrayDeclarationSpacingStandard.xml
```diff
@@ -0,0 +1,75 @@
...
+        <![CDATA[
+$args = array(
+    <em>'post_id'  => 22</em>,
+    <em>'category' => 1</em>,
+);
+        ]]>
+        </code>
+        <code title="Invalid: More than one key/value pair per line.">
```
rodrigoprimo, line 20:
```suggestion
        <code title="Invalid: Single line multi-item array using keys.">
```

rodrigoprimo:
Sounds good. Thanks, @matt-galdino.

PR #2605, opened 2025-09-10
Security/EscapeOutput: fix namespaced ::class detection

rodrigoprimo:
# Description

The sniff now correctly identifies namespaced class references like `\MyNamespace\ClassName::class` and `namespace\ClassName::class` as safe output that doesn't require escaping. Support for `ClassName::class` was added in 4ee9469ae56081359589f9b461c27d075f86e09c (PR #2326), but I believe namespaced classes were not considered at the time.

I included a separate commit to fix what I believe to be an unintentional syntax error in one of the tests added in 4ee9469ae56081359589f9b461c27d075f86e09c when support for `ClassName::class` was introduced.

## Suggested changelog entry

Fixed: `WordPress.Security.EscapeOutput`: namespaced class references using `::class` are now correctly recognized as safe output that doesn't require escaping.

### DIFF
--- a/WordPress/Tests/Security/EscapeOutputUnitTest.1.inc
+++ b/WordPress/Tests/Security/EscapeOutputUnitTest.1.inc
@@ -538,7 +538,7 @@ _deprecated_function( __METHOD__, 'x.x.x', ClassName::class ); // OK.
 die( self::CLASS . ' has been abandoned' ); // OK.
 _deprecated_function( __METHOD__, 'x.x.x', parent::Class ); // OK.
 _deprecated_function( __METHOD__, 'x.x.x', static::class ); // OK.
-echo 'Do not use ' . $object::class ); // OK.
+echo 'Do not use ' . $object::class; // OK.
 
 /*
  * Examine the parameters passed for exception creation via throw.


jrfnl:
> > 💡 _Maybe I should write a sniff to forbid that.... (using `::class` on a fully qualified or namespace relative name)_

Done. Will publish it to PHPCSExtra once Extra drops support for PHPCS 3.x (as writing that sniff for only 4.x is so much simpler 😉 )

WordPress/Sniffs/Security/EscapeOutputSniff.php
```diff
@@ -615,8 +615,13 @@ protected function check_code_is_escaped( $start, $end, $code = 'OutputNotEscape
 			if ( \T_STRING === $this->tokens[ $i ]['code']
 				|| \T_VARIABLE === $this->tokens[ $i ]['code']
 				|| isset( Collections::ooHierarchyKeywords()[ $this->tokens[ $i ]['code'] ] )
+				|| \T_NAMESPACE === $this->tokens[ $i ]['code']
 			) {
-				$double_colon = $this->phpcsFile->findNext( Tokens::$emptyTokens, ( $i + 1 ), $end, true );
+				$skip_tokens                    = Tokens::$emptyTokens;
+				$skip_tokens[ \T_STRING ]       = \T_STRING;
+				$skip_tokens[ \T_NS_SEPARATOR ] = \T_NS_SEPARATOR;
```
jrfnl, line 622: For the record: these extra tokens are only relevant when the `$i` token is `T_STRING` or `T_NAMESPACE`.
I can't think of a valid code sample with `T_VARIABLE` or the hierarchy tokens where it would be problematic though, but give PHP a few years and no doubt it will be ;-)
(but no need to change it until PHP changes)

jrfnl (approved):

Issue #2606, opened 2025-09-11
Dropping support for PHP < 7.2


jrfnl:
## Setting the scene

WordPressCS currently has a minimum PHP requirement of PHP 5.4 and supports PHP_CodeSniffer 3.x (with a minimum of PHPCS 3.13.0).

PHP_CodeSniffer 3.x has a minimum PHP requirement of PHP 5.4.
PHP_CodeSniffer 4.0 (expected soon) will have a minimum PHP requirement of PHP 7.2.

Also keep in mind, that even though PHPCS 4.x will need PHP 7.2 or higher _to run_, it will still be able to analyse code which needs to run on older PHP versions.

## The proposal

With the above in mind and considering that:
* The WordPress project has already dropped support for PHP < 7.2;
* We keep running into issues in CI with PHP 5.5;
* The fast majority of the **users of WordPressCS** do not use PHP < 7.2 anymore: https://packagist.org/packages/wp-coding-standards/wpcs/php-stats (only 0.6% of users still use PHP < 7.2)
* And setting up the CI scripts to test both PHPCS 3.x as well as 4.x against all their respective supported PHP versions adds additional complexity;

**I'm proposing to drop support for PHP < 7.2 in the next minor of WordPressCS.**

Any objections ?

PR #2611, opened 2025-09-15
GH Actions: update for upstream branch rename

jrfnl:
# Description

PHPCS has renamed the `master` branch to `3.x`.

This commit account for that.

* Change `dev-master` to `3.x-dev` in workflows.
* Update links referencing the upstream code to use `HEAD` instead of the branch name.

## Suggested changelog entry
_N/A_

### DIFF
--- a/.github/workflows/basic-qa.yml
+++ b/.github/workflows/basic-qa.yml
@@ -13,7 +13,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  PHPCS_DEV: 'dev-master'
+  PHPCS_DEV: '3.x-dev'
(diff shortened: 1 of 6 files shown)


jrfnl:
Rebased without changes to get a passing build after the merge of #2606

GaryJones (approved):
> "In the case of a branch with a version-like name (v1, in this case), you append -dev as a suffix, rather than using dev- as a prefix."
> — https://getcomposer.org/doc/articles/versions.md#branches

i.e. to tell Composer to require the head of a branch that looks like a tag, and not a specific tag, then add the `-dev` suffix. TIL.

PR #2613, opened 2025-09-15
DB/DirectDatabaseQuery: make cache function name matching case-insensitive

rodrigoprimo:
# Description

The sniff was checking cache function names incorrectly, as it was not considering that PHP function names are case-insensitive, leading to false positives.

Also includes two preparatory commits reorganizing the tests.

## Suggested changelog entry

Fixed: WordPress.DB.DirectDatabaseQuery: ensure the sniff handles cache function names in a case-insensitive manner.


jrfnl (approved):
LGTM! Good catch @rodrigoprimo !

dingo-d (approved):

dingo-d:
Should this be re-milestoned due to #2614?

PR #2615, opened 2025-09-16
DB/DirectDatabaseQuery: fix false negatives when cache function names are not a function call

rodrigoprimo:
# Description

Before, the sniff was checking if the content of a `T_STRING` token matched one of the names of the expected cache functions without checking if it was actually a function call. This could lead to some false negatives that are addressed in this commit.

As suggested by Juliette, I'm including a test to document that method names that match the cache function names are deliberately not flagged by this sniff, as they will be most likely valid custom cache functions.

## Suggested changelog entry

Fixed: WordPress.DB.DirectDatabaseQuery: prevents false negatives when cache function names are not a function call.


jrfnl (approved):
Looks good to me! Thanks @rodrigoprimo !

dingo-d (approved):

Issue #2603, opened 2025-09-09
WP/AlternativeFunctions:


rodrigoprimo:
The sniff `WordPress.WP.AlternativeFunctions` checks for function calls and constants in the first parameter of `file_get_contents()` without differentiating global and non-global function calls and constants. It also fails to differentiate between functions/constants, and class methods/class constants. This leads to false negatives in some cases.

All the examples below don't trigger the sniff, but they should:

```
file_get_contents(MyNamespace\WP_Upload_Dir()['path'] . 'subdir/file.inc');
file_get_contents(\MyNamespace\WP_Upload_Dir()['path'] . 'subdir/file.inc');
file_get_contents(namespace\WP_Upload_Dir()['path'] . 'subdir/file.inc');
file_get_contents(MyNamespace\ABSPATH . 'wp-admin/css/some-file.css');
file_get_contents(\MyNamespace\ABSPATH . 'wp-admin/css/some-file.css');
file_get_contents(namespace\ABSPATH . 'wp-admin/css/some-file.css');

file_get_contents(MyClass::wp_upload_dir() . 'subdir/file.inc');
file_get_contents($this->wp_upload_dir() . 'subdir/file.inc');
file_get_contents(MyClass::ABSPATH . 'subdir/file.inc');
file_get_contents($this->ABSPATH . 'subdir/file.inc');
```

PR #2617, opened 2025-09-16
WP/AlternativeFunctions: add tests for namespaced names tests and fix handling of class functions/constants/properties

rodrigoprimo:
# Description

## WP/AlternativeFunctions: add tests for namespaced names

Add tests for namespaced names, including some that currently are not flagged by the sniff but will be flagged once #2603 is fully addressed. As discussed in that issue, it will be easier to address this part once support for PHPCS 3.x is dropped.

## WP/AlternativeFunctions: improve regex to distinguish global WP constants/functions from non-WP class-based ones

This commit changes two regexes used to identify global WP constants and functions to prevent the sniff from incorrectly identifying class constants/methods with the same names as WordPress globals as being WordPress globals. This is done by adding a negative lookbehind to ensure the searched strings are not preceded by an object operator, null-safe object operator, or scope resolution operator.

## Suggested changelog entry

Fixed: WordPress.WP.AlternativeFunctions: prevents false negatives when class functions/constants/properties use the same name as select global WP constants/functions.

Partially addresses #2603


WordPress/Tests/WP/AlternativeFunctionsUnitTest.inc
```diff
@@ -147,3 +147,27 @@ file_get_contents(
...
+ * https://github.com/WordPress/WordPress-Coding-Standards/issues/2603.
+ */
+file_get_contents(\wp_upload_dir()['path'] . 'subdir/file.inc');
+file_get_contents(MyNamespace\wp_upload_dir()['path'] . 'subdir/file.inc');
+file_get_contents(\MyNamespace\wp_upload_dir()['path'] . 'subdir/file.inc');
+file_get_contents(namespace\wp_upload_dir()['path'] . 'subdir/file.inc');
+file_get_contents(\ABSPATH . 'wp-admin/css/some-file.css');
+file_get_contents(MyNamespace\ABSPATH . 'wp-admin/css/some-file.css');
+file_get_contents(\MyNamespace\ABSPATH . 'wp-admin/css/some-file.css');
+file_get_contents(namespace\ABSPATH . 'wp-admin/css/some-file.css');
```
jrfnl: I wonder if it wouldn't be better to add these tests in the commit which addresses #2603 ? That way, the tests and the fix would be in the same commit (=atomic), making it more straight-forward in the future to understand the history of changes in the sniff.

WordPress/Tests/WP/AlternativeFunctionsUnitTest.inc
```diff
@@ -171,3 +171,14 @@ file_get_contents(\ABSPATH . 'wp-admin/css/some-file.css');
 file_get_contents(MyNamespace\ABSPATH . 'wp-admin/css/some-file.css');
 file_get_contents(\MyNamespace\ABSPATH . 'wp-admin/css/some-file.css');
 file_get_contents(namespace\ABSPATH . 'wp-admin/css/some-file.css');
+
+/*
+ * Safeguard that the sniff does not incorrectly ignore class methods/constants with the same
+ * name as WordPress global functions/constants when used in file_get_contents().
+ */
```
jrfnl, line 163: These tests cover the change, but might be good to vary the constants/functions being referenced in the tests ? And for the functions to include some case variation ?

WordPress/Tests/WP/AlternativeFunctionsUnitTest.inc
rodrigoprimo: That was my original plan. Maybe I misunderstood what you said when we discussed this on Monday, but I thought you had suggested including those tests together with the rest of the namespaced names tests above. I'm happy to remove them from this PR and leave them for when #2603 is addressed.

WordPress/Tests/WP/AlternativeFunctionsUnitTest.inc
rodrigoprimo: I'm unsure about this. I understand your point, but at the same time, to me, it is easier to understand the tests if there is no variation in parts of the code that are irrelevant to what is being tested. That is why I prefer to use the same constants/functions and all lowercase letters. There are already other tests that cover those variations for this specific part of the code.

jrfnl: Yeah, but depending on the sniff, some functions may have special handling, while others do not. If all "namespaced name" tests only relate to the one function, it can easily be missed that special handling for another function is broken in the context of namespaced names (typically FQN global function).

I'm not trying to make your life more difficult, I'm trying to make it easier by having tests which will actually show what needs fixing for PHPCS 4.0.

rodrigoprimo: That is fair. I went ahead and added variation for the tests that safeguard that the sniff does not incorrectly ignore class methods/constants with the same name as WordPress global functions/constants when used in file_get_contents(). I did not do the same for the tests added in the other commit, as I imagine they will be removed from this PR pending what we decide in the other comment thread in this PR. I do plan to add variation to those tests as well.

WordPress/Tests/WP/AlternativeFunctionsUnitTest.inc
rodrigoprimo: I went ahead and those tests.

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -283,7 +283,7 @@ public function process_matched_token( $stackPtr, $group_name, $matched_content
 				}
 
 				$contains_wp_path_constant = preg_match(
-					'`\b(?:ABSPATH|WP_(?:CONTENT|PLUGIN)_DIR|WPMU_PLUGIN_DIR|TEMPLATEPATH|STYLESHEETPATH|(?:MU)?PLUGINDIR)\b`',
+					'`(?<!\?->|->|::)\b(?:ABSPATH|WP_(?:CONTENT|PLUGIN)_DIR|WPMU_PLUGIN_DIR|TEMPLATEPATH|STYLESHEETPATH|(?:MU)?PLUGINDIR)\b`',
```
jrfnl: Simplification suggestion (here and in the second regex):

```suggestion
					'`(?<!->|::)\b(?:ABSPATH|WP_(?:CONTENT|PLUGIN)_DIR|WPMU_PLUGIN_DIR|TEMPLATEPATH|STYLESHEETPATH|(?:MU)?PLUGINDIR)\b`',
```

Reasoning: `?->` will already be matched when "looking behind" for `->`, so doesn't need to be made explicit in the regex.

Note: we should make sure there are tests with `?->` though to safeguard this for the future (which you have added ✅ ).

rodrigoprimo: Good catch, I implemented this change in both regexes. While I was reviewing it, I noticed that the tests that I added to safeguard this change contained a WP coding standard violation (missing spaces after/before opening/closing parentheses). So I fixed that as well.

jrfnl: > Good catch, I implemented this change in both regexes.

👍🏻 

> While I was reviewing it, I noticed that the tests that I added to safeguard this change contained a WP coding standard violation (missing spaces after/before opening/closing parentheses). So I fixed that as well.

That's fair, but not always strictly necessary/desirable. Sniffs should function independently of each other, so as this is a functional sniff, this sniff should function code style independently, so whether the code is space rich or space poor shouldn't matter to the sniff and it's good to make sure that the sniff functions code style independently.

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
```diff
@@ -354,6 +354,11 @@ public function process_matched_token( $stackPtr, $group_name, $matched_content
 	protected function is_local_data_stream( $clean_param_value ) {
 
 		$stripped = TextStrings::stripQuotes( $clean_param_value );
+
+		if ( strpos( $clean_param_value, '\\' ) === 0 ) {
+			$clean_param_value = substr( $clean_param_value, 1 );
+		}
```
jrfnl: Could this be simplified ? `ltrim( $clean_param_value, '\\')` ?

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
rodrigoprimo: > That's fair, but not always strictly necessary/desirable. Sniffs should function independently of each other, so as this is a functional sniff, this sniff should function code style independently, so whether the code is space rich or space poor shouldn't matter to the sniff and it's good to make sure that the sniff functions code style independently.

:+1: 

I tend to default to following the coding standard, but I agree that it is important to have some variation to ensure the sniff works regardless of the spacing. In the particular case of this sniff, it made sense to me to update this block of tests, considering that there are already quite a few tests without spaces before/after the opening/closing parentheses.

WordPress/Sniffs/WP/AlternativeFunctionsSniff.php
rodrigoprimo: Yes, I don't know why this did not occur to me when preparing this fix. It is done now. Thanks for the suggestion.

rodrigoprimo:
Looking into the failing quick test build.

jrfnl:
> Looking into the failing quick test build.

I'm not seeing any failures ?

rodrigoprimo:
> I'm not seeing any failures ?

That is interesting. The ["Basic QA Tests"](https://github.com/rodrigoprimo/WordPress-Coding-Standards/actions/runs/19675630670) and the ["Quick Tests"](https://github.com/rodrigoprimo/WordPress-Coding-Standards/actions/runs/19675630660) actions that ran on my fork failed, and they were showing to me here as a red X next to the last commit. That is what I was referring to. But now I don't see it anymore. Instead, I see that they passed with a link to the actions in this repository. So I guess this can be ignored.

rodrigoprimo:
> Once you finish reviewing this PR, I can squash the commits before another maintainer reviews it. I'm thinking this PR should contain three commits (the ones mentioned in the PR description).

I went ahead, squashed the commits, and force-pushed without changes.

rodrigoprimo:

dingo-d (approved):

dingo-d:
I added a milestone to the PR based on previously merged PRs that handled namespaced names (3.3.x).

PR #2618, opened 2025-09-18
Security/EscapeOutput: fix false positive and add tests for namespaced names

rodrigoprimo:
# Description

In preparation for PHPCS 4.0, which changes the tokenization of namespaced names, this PR adds tests with all variants of namespace names (unqualified, partially qualified, fully qualified, and namespace relative) to the `WordPress.Security.EscapeOutput` sniff. This is a continuation of #2581, where tests were added for all sniffs extending `AbstractFunctionRestrictions` except `EscapeOutput`.

This PR also includes tests to ensure the `basename( __FILE__ )` pattern recognition in `_deprecated_file()` only applies to global `basename()` function calls, not to other constructs that might look similar (such as namespaced variants or class methods).

Besides that, it fixes in a separate commit a false positive bug that was discovered during the PR review, where the sniff would incorrectly handle FQN and/or non-standard case `true` passed to `get_search_query()`.

## Suggested changelog entry

Fixed:
- `WordPress.Security.EscapeOutput`: false positive for `get_search_query()` when the `$escaped` parameter was passed as fully qualified `true` or as `true` using a non-standard case.

### DIFF
--- a/WordPress/Tests/Security/EscapeOutputUnitTest.1.inc
+++ b/WordPress/Tests/Security/EscapeOutputUnitTest.1.inc
@@ -801,3 +801,13 @@ echo implode( '<br>', MyNamespace\array_map( 'esc_html', $items ) ); // Bad x 2.
 echo implode( '<br>', \MyNamespace\map_deep( $items, 'esc_html' ) ); // Bad.
 echo implode( '<br>', namespace\array_map( 'esc_html', $items ) ); // Bad x 2. The sniff should stop flagging this once it can resolve relative namespaces.
 echo implode( '<br>', namespace\Sub\map_deep( $items, 'esc_html' ) ); // Bad.
+
+/*
+ * Safeguard correct handling of get_search_query() with FQN and non-standard case booleans for the $escaped parameter.
+ */
+echo \get_search_query( TRUE ); // Ok.
+echo \get_search_query( \true ); // Ok.
+echo \get_search_query( \True ); // Ok.
+echo \get_search_query( FaLsE ); // Bad.
+echo \get_search_query( \false ); // Bad.
+echo \get_search_query( \FALSE ); // Bad.
--- a/WordPress/Tests/Security/EscapeOutputUnitTest.php
+++ b/WordPress/Tests/Security/EscapeOutputUnitTest.php
@@ -205,6 +205,9 @@ final class EscapeOutputUnitTest extends AbstractSniffUnitTest {
 					801 => 1,
 					802 => 2,
 					803 => 1,
+					811 => 1,
+					812 => 1,
+					813 => 1,
 				);
 
 			case 'EscapeOutputUnitTest.6.inc':
--- a/WordPress/Sniffs/Security/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/Security/EscapeOutputSniff.php
@@ -741,7 +741,7 @@ class EscapeOutputSniff extends AbstractFunctionRestrictionsSniff {
 					// Special case get_search_query() which is unsafe if $escaped = false.
 					if ( 'get_search_query' === strtolower( $functionName ) ) {
 						$escaped_param = PassedParameters::getParameter( $this->phpcsFile, $ptr, 1, 'escaped' );
-						if ( false !== $escaped_param && 'true' !== $escaped_param['clean'] ) {
+						if ( false !== $escaped_param && 'true' !== strtolower( ltrim( $escaped_param['clean'], '\\' ) ) ) {
 							$this->phpcsFile->addError(
 								'Output from get_search_query() is unsafe due to $escaped parameter being set to "false".',
 								$ptr,


rodrigoprimo:
I'm investigating the test that is failing in the PHPCS lowest build.

WordPress/Tests/Security/EscapeOutputUnitTest.1.inc
```diff
@@ -661,7 +661,7 @@ exit( status: esc_html( $foo ) ); // Ok.
 die( status: esc_html( $foo ) ); // Ok.
 
 exit( status: $foo ); // Bad.
-die( status: $foo ); // Bad.
+\die( status: $foo ); // Bad.
```
jrfnl: Maybe ?
```suggestion
\Die( status: $foo ); // Bad.
```

WordPress/Tests/Security/EscapeOutputUnitTest.1.inc
```diff
@@ -687,3 +687,68 @@ _deprecated_function( __METHOD__, 'x.x.x', \ClassName::class ); // OK.
...
+
+/*
+ * Safeguard correct handling of get_search_query() as the sniff has special logic to check the $escaped parameter.
+ */
+echo \get_search_query( true ); // Ok.
+echo \get_search_query( false ); // Bad.
+echo MyNamespace\get_search_query( true ); // Bad.
+echo \MyNamespace\get_search_query( true ); // Bad.
+echo namespace\get_search_query( true ); // Bad. The sniff should stop flagging this once it can resolve relative namespaces.
+echo namespace\Sub\get_search_query( true ); // Bad.
```
jrfnl, line 741: Should there be tests here with FQN `true`/`false` and case-variations ? I.e. `\TRUE` and `\False` ?

WordPress/Tests/Security/EscapeOutputUnitTest.php
```diff
@@ -160,10 +163,34 @@ public function getErrorList( $testFile = '' ) {
 					655 => 1,
 					657 => 1,
 					663 => 1,
-					664 => 1,
+					// PHPCS 3.13.3 changed the tokenization of FQN exit/die it impacts directly how this test case
+					// behaves (see https://github.com/PHPCSStandards/PHP_CodeSniffer/issues/1201).
+					664 => version_compare( $phpcs_version, '3.13.3', '>=' ) ? 1 : 0,
```
jrfnl: I seem to remember we dropped support for PHPCS < 3.13.4 ? In which case, this toggle should no longer be needed.

WordPress/Tests/Security/EscapeOutputUnitTest.1.inc
```diff
@@ -258,7 +258,7 @@ echo esc_html_x( $some_nasty_var, 'context' ); // Ok.
 	<input type="hidden" name="some-action" value="<?php echo esc_attr_x( 'none', 'context' ); ?>" /><!-- OK. -->
 <?php
 
-echo PHP_VERSION_ID, PHP_VERSION, PHP_EOL, PHP_EXTRA_VERSION; // OK.
+echo PHP_VERSION_ID, PHP_VERSION, \PHP_EOL, PHP_EXTRA_VERSION; // OK.
```
jrfnl, line 261: Should there be tests with these "safe constants" in namespaced variants ?

WordPress/Tests/Security/EscapeOutputUnitTest.1.inc
```diff
@@ -687,3 +687,82 @@ _deprecated_function( __METHOD__, 'x.x.x', \ClassName::class ); // OK.
...
+ * Safeguard that the basename( __FILE__ ) pattern recognition in _deprecated_file() only applies to
+ * the global basename() function and not to other constructs.
+ */
+_deprecated_file( $obj->basename( __FILE__ ), '1.3.0' ); // Bad.
+_deprecated_file( $obj?->basename( __FILE__ ), '1.3.0' ); // Bad.
+_deprecated_file( MyClass::basename( __FILE__ ), '1.3.0' ); // Bad.
+_deprecated_file( BASENAME, __FILE__ ); // Bad.
+_deprecated_file( MyNamespace\basename( __FILE__ ), '1.3.0' ); // Bad.
+_deprecated_file( \MyNamespace\basename( __FILE__ ), '1.3.0' ); // Bad.
+_deprecated_file( namespace\basename( __FILE__ ), '1.3.0' ); // Bad. We might want to update the regex so that the sniff stop flagging this once it can resolve relative namespaces.
```
jrfnl:
```suggestion
_deprecated_file( namespace\basename( __FILE__ ), '1.3.0' ); // Bad. We might want to update the regex so that the sniff stops flagging this once it can resolve relative namespaces.
```

WordPress/Tests/Security/EscapeOutputUnitTest.1.inc
jrfnl, line 1: Would it make sense to also have some tests like the below ? (mind: I've not checked whether the function names in the second line (short echo) are the correct WP function names, this will need verification)
```php
	?>
	<a href="<?php \the_permalink(); ?>" title="<?php \the_title_attribute(); ?>"><?php \the_title(); ?></a>
	<a href="<?= \get_permalink(); ?>" title="<?= \get_title_attribute(); ?>"><?= \get_the_title(); ?></a>
	<?php
```

jrfnl: And how about this ?
```php
echo \true, \False, \NULL;
```

I know, it's kind of silly code, but the tokens are in the "safe tokens" list and with the fully qualified bit, the behaviour may change ?

jrfnl: I also see a separate category of array walking functions being handled. Do these need tests with the namespaced name variations too ?

jrfnl (commented):
@rodrigoprimo Very good effort, needs a little more work, but these tests will help a lot!

WordPress/Tests/Security/EscapeOutputUnitTest.1.inc
rodrigoprimo:
```php
<a href="<?php \the_permalink(); ?>" title="<?php \the_title_attribute(); ?>"><?php \the_title(); ?></a>
```

@jrfnl, correct me if I'm missing something, but I'm not sure what the value of adding the test above would be. The sniff bails early when it encounters any of those three functions, as they are not in the list of functions that this sniff looks for.  `the_title_attribute()` is an auto-escaping function, and `the_permalink()` and `the_title()` used to be (removed in https://github.com/WordPress/WordPress-Coding-Standards/pull/1547). But the code above does not trigger the check for auto-escaping functions, and there are already tests for this group of functions. Maybe the original test on line 9 needs to be removed or updated? Or am I missing its purpose?

Regarding the second line you suggested, I made a minor modification, and I believe it makes sense to add it to test multiple short open tags and fully qualified functions.

```php
<a href="<?= \get_permalink(); ?>" title="<?= \the_title_attribute( array( 'echo' => false ) ); ?>"><?= \get_the_title(); ?></a><!-- Bad x 2. -->
```

WordPress/Tests/Security/EscapeOutputUnitTest.1.inc
rodrigoprimo: This uncovered a bug in how the sniff was handling FQN and non-standard case `true`. I added the tests and the fix in a separate commit from the rest of the changes. I also updated the PR description and title to mention this fix and added a changelog entry.

rodrigoprimo:

I responded to all of your comments and implemented your suggestions. There is just one item pending for us to discuss regarding the fully qualified tests for functions like `\the_permalink()`. Please take a look at my remark above (https://github.com/WordPress/WordPress-Coding-Standards/pull/2618#discussion_r2603989780).

WordPress/Tests/Security/EscapeOutputUnitTest.1.inc
jrfnl: @rodrigoprimo I was asking, not telling. If your assessment is that adding that extra test adds no value and/or is covered already sufficiently elsewhere, I'm going to trust you on that.

PR #2619, opened 2025-09-18
DB/DirectDatabaseQuery: fix handling of namespaced calls to cache functions

rodrigoprimo:
# Description

Before this change, the sniff was not properly differentiating between fully qualified calls to global WP cache functions and namespaced calls to non-WP functions with the same name as the WP functions.

In other words, the sniff was correctly identifying `\wp_cache_get()` as a WP cache function, but it was incorrectly identifying `MyNamespace\wp_cache_get()` as a WP cache function as well. This PR fixes this issue.

## Suggested changelog entry

Fixed: `WordPress.DB.DirectDatabaseQuery`: handling of calls to namespaced non-global functions with the same name as WP cache functions.


rodrigoprimo:
Thanks for your input, @jrfnl. I have pushed a new commit reverting the changes to the sniff code and updating the tests to safeguard against regressions for the current behavior of the sniff. Is this more or less what you had in mind? Do you prefer that I update the title and description of this PR to reflect that now it only adds `DirectDatabaseQuery` tests or that I close this PR and open a new one?

dingo-d (approved):

PR #2626, opened 2025-10-02
ContextHelper::is_in_function_call(): add basic tests

rodrigoprimo:
# Description

As part of the work to prepare WPCS for PHPCS 4.0, this PR adds basic tests for `ContextHelper::is_in_function_call()`. Those tests should help safeguard this method against regressions when updating it to be compatible with PHPCS 4.0.

Since there were no previous tests for utility methods, it was also necessary to modify the test structure to enable running utility tests. They are different from the sniff tests as they extend `UtilityMethodTestCase` and don't need to run via `./vendor/squizlabs/php_codesniffer/tests/AllTests.php`.

After this one is approved, I plan to open another PR, adding tests for more utility methods. I opted to open this one first with tests for a single method to get feedback on the way that I organized the tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,10 +53,12 @@
 			"@php ./vendor/squizlabs/php_codesniffer/bin/phpcbf"
 		],
 		"run-tests": [
-			"@php ./vendor/phpunit/phpunit/phpunit --filter WordPress ./vendor/squizlabs/php_codesniffer/tests/AllTests.php --no-coverage"
+			"@php ./vendor/phpunit/phpunit/phpunit --filter WordPress ./vendor/squizlabs/php_codesniffer/tests/AllTests.php --no-coverage",
+            "@php ./vendor/phpunit/phpunit/phpunit WordPress/Util/Tests/ --no-coverage"
 		],
 		"coverage": [
-			"@php ./vendor/phpunit/phpunit/phpunit --filter WordPress ./vendor/squizlabs/php_codesniffer/tests/AllTests.php"
+			"@php ./vendor/phpunit/phpunit/phpunit --filter WordPress ./vendor/squizlabs/php_codesniffer/tests/AllTests.php",
+            "@php ./vendor/phpunit/phpunit/phpunit WordPress/Util/Tests/"
 		],
 		"check-complete": [
 			"@php ./vendor/phpcsstandards/phpcsdevtools/bin/phpcs-check-feature-completeness -q ./WordPress"
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -13,7 +13,10 @@
 	forceCoversAnnotation="true">
 
 	<testsuites>
-		<testsuite name="WordPress">
+        <testsuite name="WordPress Utilities Tests">
+            <directory suffix="UnitTest.php">./WordPress/Util/Tests/</directory>
+        </testsuite>
+		<testsuite name="WordPress Sniffs Tests">
 			<directory suffix="UnitTest.php">./WordPress/Tests/</directory>
 		</testsuite>
 	</testsuites>
--- a/Tests/bootstrap.php
+++ b/Tests/bootstrap.php
@@ -45,6 +45,21 @@ if ( false !== $phpcsDir
 ) {
 	require_once $phpcsDir . $ds . 'autoload.php';
 	require_once $phpcsDir . $ds . 'tests' . $ds . 'bootstrap.php'; // PHPUnit 6.x+ support.
+
+	spl_autoload_register(
+		function ( $className ) {
+			// Only try & load our own classes.
+			if ( stripos( $className, 'WordPressCS' ) !== 0 ) {
+				return;
+			}
+
+			$file = realpath( dirname( __DIR__ ) ) . DIRECTORY_SEPARATOR . strtr( str_replace( 'WordPressCS\\', '', $className ), '\\', DIRECTORY_SEPARATOR ) . '.php';
+
+			if ( file_exists( $file ) ) {
+				include_once $file;
+			}
+		}
+	);
 } else {
 	echo 'Uh oh... can\'t find PHPCS.
 
--- a/WordPress/Util/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.inc
+++ b/WordPress/Util/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.inc
@@ -1,0 +1,57 @@
+<?php
+
+/*
+ * Make sure that tokens inside a function call are correctly identified using the default values
+ * for the optional parameters (`$global_function = true` and `$allow_nested = false`).
+ *
+ * The below should *NOT* be recognized as inside a function call to one of the valid functions.
+ */
+
+$a = /* test return false 1 */ 'foo';
+another_function( /* test return false 2 */ $a );
+MyNamespace\my_function( /* test return false 3 */ $a );
+\MyNamespace\my_function( /* test return false 4 */ $a );
+namespace\MyNamespace\my_function( /* test return false 5 */ $a );
+my_function( another_function( /* test return false 6 */ $a ) );
+$anon_func = function() {
+    /* test return false 7 */ $a = 'foo';
+};
+$my_function( /* test return false 8 */ $a );
+
+/*
+ * Make sure that tokens inside a function call are correctly identified using the default values
+ * for the optional parameters (`$global_function = true` and `$allow_nested = false`).
+ *
+ * The below should be recognized as inside a function call to one of the valid functions.
+ */
+
+/* test function call 1 */ my_function( /* test inside function pointer 1 */ $a );
+/* test function call 2 */ MY_FUNCTION( /* test inside function pointer 2 */ $a );
+/* test function call 3 */ \my_function( /* test inside function pointer 3 */ $a );
+
+/*
+ * Make sure that tokens inside a function call are correctly identified when `$global_function` is
+ * set to false.
+ *
+ * The below should be recognized as inside a function call to one of the valid functions.
+ */
+
+/* test function call 4 */ my_function( /* test inside function pointer 4 */ $a );
+/* test function call 5 */ MY_FUNCTION( /* test inside function pointer 5 */ $a );
+/* test function call 6 */ \my_function( /* test inside function pointer 6 */ $a );
+MyNamespace\/* test function call 7 */my_function( /* test inside function pointer 7 */ $a );
+\MyNamespace\/* test function call 8 */my_function( /* test inside function pointer 8 */ $a );
+namespace\MyNamespace\/* test function call 9 */my_function( /* test inside function pointer 9 */ $a );
+MyClass::/* test function call 10 */my_function( /* test inside function pointer 10 */ $a );
+/* test function call 11 */ $obj->my_function( /* test inside function pointer 11 */ $a );
+/* test function call 12 */ $obj?->my_function( /* test inside function pointer 12 */ $a );
+
+/*
+ * Make sure that tokens inside a function call are correctly identified when `$allow_nested` is
+ * set to true.
+ *
+ * The below should be recognized as inside a function call to one of the valid functions.
+ */
+/* test function call 13 */ my_function( another_function( /* test inside function pointer 13 */ $a ) );
+another_function( /* test function call 14 */ my_function( /* test inside function pointer 14 */ $a ) );
+/* test function call 15 */ my_function( middle_function( inner_function( /* test inside function pointer 15 */ $a ) ) );
--- a/WordPress/Util/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.php
+++ b/WordPress/Util/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.php
@@ -1,0 +1,165 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Util\Tests\Helpers\ContextHelper;
+
+use PHPCSUtils\Tokens\Collections;
+use WordPressCS\WordPress\Helpers\ContextHelper;
+use PHPCSUtils\TestUtils\UtilityMethodTestCase;
+
+/**
+ * Tests for the `ContextHelper::is_in_function_call()` utility method.
+ *
+ * @since 3.3.0
+ *
+ * @covers \WordPressCS\WordPress\Helpers\ContextHelper::is_in_function_call()
+ */
+final class IsInFunctionCallUnitTest extends UtilityMethodTestCase {
+
+
+	/**
+	 * Test is_in_function_call() returns false if given token is not inside a function call
+	 * or not inside one of the expected function calls.
+	 *
+	 * @dataProvider dataIsInFunctionCallShouldReturnFalse
+	 *
+	 * @param string     $commentString The comment which prefaces the target token in the test file.
+	 * @param int|string $tokenType     The token type to search for.
+	 *
+	 * @return void
+	 */
+	public function testIsInFunctionCallShouldReturnFalse( $commentString, $tokenType ) {
+		$stackPtr = $this->getTargetToken( $commentString, $tokenType );
+		$result   = ContextHelper::is_in_function_call( self::$phpcsFile, $stackPtr, array( 'my_function' => true ) );
+		$this->assertFalse( $result );
+	}
+
+	/**
+	 * Data provider.
+	 *
+	 * @return array
+	 * @see testIsInFunctionCallShouldReturnFalse()
+	 */
+	public static function dataIsInFunctionCallShouldReturnFalse() {
+		return array(
+			array( '/* test return false 1 */', \T_CONSTANT_ENCAPSED_STRING ),
+			array( '/* test return false 2 */', \T_VARIABLE ),
+			array( '/* test return false 3 */', \T_VARIABLE ),
+			array( '/* test return false 4 */', \T_VARIABLE ),
+			array( '/* test return false 5 */', \T_VARIABLE ),
+			array( '/* test return false 6 */', \T_VARIABLE ),
+			array( '/* test return false 7 */', \T_VARIABLE ),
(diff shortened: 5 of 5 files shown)


composer.json
```diff
@@ -53,10 +53,12 @@
 			"@php ./vendor/squizlabs/php_codesniffer/bin/phpcbf"
 		],
 		"run-tests": [
-			"@php ./vendor/phpunit/phpunit/phpunit --filter WordPress ./vendor/squizlabs/php_codesniffer/tests/AllTests.php --no-coverage"
+			"@php ./vendor/phpunit/phpunit/phpunit --filter WordPress ./vendor/squizlabs/php_codesniffer/tests/AllTests.php --no-coverage",
+            "@php ./vendor/phpunit/phpunit/phpunit WordPress/Util/Tests/ --no-coverage"
```
jrfnl: Here and below:

1. Please fix the indent.
2. This may not work as you expect. I've run into trouble with multi-command composer scripts before.

phpunit.xml.dist
```diff
@@ -13,7 +13,10 @@
 	forceCoversAnnotation="true">
 
 	<testsuites>
-		<testsuite name="WordPress">
+        <testsuite name="WordPress Utilities Tests">
+            <directory suffix="UnitTest.php">./WordPress/Util/Tests/</directory>
+        </testsuite>
```
jrfnl: Please fix the indent (use tabs, not spaces).

jrfnl (commented):
While I understand the desire to have separate tests for this particular utility function, this PR disregards the fact that the method is already being tested.

Additionally:
* I have a feeling that the separate test suite is unnecessary.
* I find the placement of the tests completely illogical.
    I can understand the tests being placed in a `WordPress/Tests/Helpers/...` subdirectory or a `/WordPress/Helpers/Tests/...` subdirectory, but I do not understand why a new `WordPress/Util/Tests` directory is being created with only these tests.
* I find the test markers used, i.e. `/* test return false 4 */` very non-descript and unhelpful. These tests are not self-describing.
* Same for the data providers in the test class, which doesn't make things any better.

phpunit.xml.dist
rodrigoprimo: Done

composer.json
rodrigoprimo: > Please fix the indent.

Done

> This may not work as you expect. I've run into trouble with multi-command composer scripts before.

Do you happen to remember more details, or what you used instead? This seems to be working as I expect. Composer is running both PHPUnit commands. It will not run the second command if the first one fails, but I guess that is not a problem, or is this what you are referring to for some reason that I'm missing?

Running the tests in two separate commands is creating problems for the code coverage report. I'm looking into this now, and I will continue this part of the conversation when I reply to your other remark about creating two separate test suites.

WordPress/Tests/Security/NonceVerificationUnitTest.php
jrfnl, line 1: The `@covers \WordPressCS\WordPress\Helpers\ContextHelper::is_in_function_call` on line 21 needs to be removed.

WordPress/Tests/Security/NonceVerificationUnitTest.1.inc
```diff
@@ -291,18 +291,18 @@ function function_containing_nested_closure() {
...
-	$var = MyClass::stripslashes_from_strings_only( $_POST['foo'] ); // Bad.
-	wp_verify_nonce( $var );
-	echo $var;
-}
 
-function disallow_custom_unslash_before_noncecheck_via_namespaced_function() {
-	$var = MyNamespace\stripslashes_from_strings_only( $_POST['foo'] ); // Bad.
-	wp_verify_nonce( $var );
-	echo $var;
-}
```
jrfnl, line 305: These tests don't necessarily need to be removed if they still have value in the context of the `NonceVerification` sniff.

rodrigoprimo: I misunderstood your comment and thought you were suggesting that those tags are removed. I missed the `@covers` tag. This is now fixed.

WordPress/Tests/Security/NonceVerificationUnitTest.php
rodrigoprimo: Done

rodrigoprimo:
@GaryJones and @dingo-d, do you have any input on Juliette's question?

> That also explains my reluctance to approve this PR as I don't want it to become a blocker for PHPCSUtils improvements and as far as I can see, it currently would be (with me running the legal risks).

I'm ok with not adding the tests for `ContextHelper::is_in_function_call()` to WPCS and closing this PR. But from what I remember from my conversation with @jrfnl about this a couple of weeks ago, the fact that I created the PR already causes the issue, so closing the PR won't solve the problem.

dingo-d:
I mean, if the work has been done here, I see no reason for not adding it to WPCS. For me, the most demotivating thing that can happen when contributing to OSS is doing the work, spending time, only for it to be totally disregarded (when the addition is valid ofc).

I partially understand the problem with moving it to PHPCSUtils, but only partially. Why?
Because if I'm not mistaken (I could be so feel free to correct me if I am), new tests for the new feature can be added in the PHPCSUtils. The method there may be named differently, and structured differently. Plus, seeing how this feature was written by Juliette, and the PHPCSUtils helper will most likely be written by Juliette ,I really don't see the problem (I mean, the main problem I can think of is that she will have to lose time by writing it partially from scratch, at least the test will have to be written from zero).

Again, I don't have context you two have from your call, or why that could have some deep legal issues (I don't know who would be the person instigating any legal problems here, etc.) 🤷🏼‍♂️

rodrigoprimo:
Just documenting here that I force pushed without changes to be able to rebase this branch with the latest from the `develop` branch.

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.inc
```diff
@@ -0,0 +1,79 @@
...
+$my_function( /* testVariableFunction */ $a );
+if ( /* testIfCondition */ true ) {}
+
+/*
+ * The below should be recognized as inside a function call to one of the valid functions regardless of the value
+ * of the optional parameters (`$global_function` and `$allow_nested`).
+ */
+
+/* testLowercaseName */
+valid_function1( /* testLowercaseNameInsideCall */ );
```
jrfnl: Just a note that the first thing I did when looking at this test code was scroll to the data provider to understand what token the marker was referring to - which turned out to be `T_WHITESPACE`.

While perfectly valid as a test case, it is not very realistic (why would anyone care whether whitespace is used inside a function call to a certain specific function ?) and as demonstrated by the above, makes the test case more confusing than necessary for anyone reading the tests.

Maybe let it search for a `T_VARIABLE` (`$a`) instead ?

**Note**: this is **not** a blocker and can be fixed up/improved in a future iteration of these tests.

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.inc
```diff
@@ -0,0 +1,79 @@
...
+/* testUppercaseName */
+VALID_FUNCTION2( /* testUppercaseNameInsideCall */ 'string' );
+/* testFullyQualified */
+\valid_function2( /* testFullyQualifiedInsideCall */ 42 );
+
+/*
+ * The below might be recognized as inside a function call to one of the valid functions or not depending on the values
+ * passed to the optional parameters (`$global_function` and `$allow_nested`).
+ */
```
jrfnl, line 32: I understand that the below is a mix & match of test cases for variation of `$global_function` and `$allow_nested` parameter values.

However, the first test case I looked at (line 33) is not subject to any changes in `$allow_nested`, so that straight away caused some confusion.

Similar to the other comment - not a blocker -, but it may be an idea to split these tests into three blocks:
* Tests covering the `$global_function` param value.
* Tests covering the `$allow_nested` param value.
* Tests covering various combinations of values for these two parameters.

That would make the tests more straight forward to grep/scan/read as a human.

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.inc
```diff
@@ -0,0 +1,79 @@
...
+	MyNamespace\other_function(
+		/* testNestedBothNamespacedOuterInsideCall */ 'value' . $var
+	)
+);
+/* testComplexParametersAlwaysMatch */
+valid_function1(
+	array( 'key1' => 'value1' ),
+	other_function( 'nested' ),
+	/* testComplexParametersAlwaysMatchInsideCall */ $var
+);
```
jrfnl: What is "complex" about the `$var` parameter ?

(similar for the other tests below)

Also, when looking at the tests below, the "complex" parameters always seem to be _before_ the "target" token. Would it make sense/add value to also have some tests where the "complex" parameters are _after_ the "target" token ?

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.php
```diff
@@ -0,0 +1,420 @@
...
+ */
+final class IsInFunctionCallUnitTest extends UtilityMethodTestCase {
+
+	/**
+	 * Expected results: when a test case uses this constant, `is_in_function_call()` should return `false` regardless
+	 * of the value of the parameters `$global_function` and `$allow_nested`.
+	 *
+	 * @var array<string, bool>
+	 */
+	const EXPECT_NO_MATCH = array(
```
jrfnl: Please add constant visibility modifiers for all these OO constants.

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.php
```diff
@@ -0,0 +1,420 @@
...
+		$tokenType,
+		$expectedResults,
+		$expectedFunctionMarker = null
+	) {
+		$this->runIsInFunctionCallTest(
+			$testMarker,
+			$tokenType,
+			$expectedResults,
+			$expectedFunctionMarker,
+			'global_only'
```
jrfnl: This would necessitate some more changes, but I have a feeling, it will make the code more maintainable and easier to reason about.
```suggestion
			self::PARAMETER_MAP['global_only']
```

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.php
```diff
@@ -0,0 +1,420 @@
...
+	 * @param string|null         $expectedFunctionMarker Optional. The comment for the expected function (if match expected).
+	 *
+	 * @return void
+	 */
+	public function testIsInFunctionCallWithDefaultParams(
+		$testMarker,
+		$tokenType,
+		$expectedResults,
+		$expectedFunctionMarker = null
+	) {
```
jrfnl: Just a suggestion (as I don't see the need for the long param names and multi-line signatures):

```suggestion
	public function testIsInFunctionCallWithDefaultParams( $marker, $tokenType, $expected, $expectedMarker = null ) {
```

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.php
```diff
@@ -0,0 +1,420 @@
...
+				'valid_function1' => true,
+				'valid_function2' => true,
+			),
+			$globalFunction,
+			$allowNested
+		);
+
+		$expected = $expectedResults[ $expectedKey ]
+			? $this->getTargetToken( $expectedFunctionMarker, \T_STRING )
+			: false;
```
jrfnl: I don't understand this ternary. As far as I can see the `false` is impossible to reach with the current code, so the ternary is unnecessary ?

rodrigoprimo: The `false` is reachable. It is hit for every no-match test case across all 4 test methods. For example, `testIsInFunctionCallWithDefaultParams` with `plain_assignment`, where `$expectedResults['global_only']` is `false`.

Without the ternary, the code would call `$this->getTargetToken( null, \T_STRING )` since `$expectedFunctionMarker` is `null` for no-match test cases.

That said, the ternary condition was renamed from `$expectedResults[$expectedKey]` to `$shouldMatch` in the latest changes, which should hopefully make the intent clearer.

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.inc
rodrigoprimo: I believe I added these tests to ensure the method wouldn't get confused by extra parentheses (from arrays, nested function calls, or closures) between the target token and the enclosing function call. But thinking more about it, I'm inclined to think those tests aren't necessary, since they test PHPCS's construction of the `nested_parenthesis` array rather than anything specific to `is_in_function_call()`. On top of that, naming the tests "complex parameters" didn't help to communicate their purpose.

For the same reason, I don't see value in adding tests with other parameters after the target token. The position of other parameters is irrelevant to how the method works.

I'd suggest removing these test cases. What do you think?

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.php
rodrigoprimo: Done.

I refactored `runIsInFunctionCallTest()` to receive a `PARAMETER_MAP` entry directly instead of a string key. The method now also receives $shouldMatch (a boolean) instead of the full $expectedResults array.

As a side effect, the `$expectedKey` variable no longer exists, so I removed the custom assertion message. PHPUnit's default output already includes the information from the custom message.

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.inc
jrfnl: I agree these tests are largely superfluous and can be removed.

I'd say, let's just keep one test with something like `function_name( array param, target param, closure param);` to ensure that, even if the method is refactored, the code does not get confused by parentheses in other parameters within the same function call.

But just that one test should suffice.

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.php
```diff
@@ -263,15 +260,15 @@ private function runIsInFunctionCallTest(
...
-			$globalFunction,
-			$allowNested
+			$params['global_function'],
+			$params['allow_nested']
 		);
 
-		$expected = $expectedResults[ $expectedKey ]
+		$expected = $shouldMatch
 			? $this->getTargetToken( $expectedFunctionMarker, \T_STRING )
 			: false;
```
jrfnl: This is highly opinionated, but I would write this as follows to make it easier to grep (and stabilize):

```php
$expected = false;
if ( $shouldMatch === true) {
    $expected = $this->getTargetToken( $expectedFunctionMarker, \T_STRING );
}
```

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.inc
rodrigoprimo: Done. I replaced the four "complex parameters" test cases with a single `other_params_with_parentheses` test that has an array param before and a closure param after the target token.

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.inc
rodrigoprimo: Done

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.inc
rodrigoprimo: Done

jrfnl (commented):
Sorry forgot to post this yesterday

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.php
```diff
@@ -0,0 +1,422 @@
...
+
+namespace WordPressCS\WordPress\Tests\Helpers\ContextHelper;
+
+use WordPressCS\WordPress\Helpers\ContextHelper;
+use PHPCSUtils\TestUtils\UtilityMethodTestCase;
+
+/**
+ * Tests for the `ContextHelper::is_in_function_call()` utility method.
+ *
+ * @since 3.3.0
```
jrfnl: Note: we've released 3.3.0, so this version number still needs to be updated at a later point in time when we know what milestone this PR will be merged in.

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.php
```diff
@@ -0,0 +1,422 @@
...
+		$this->runIsInFunctionCallTest(
+			$marker,
+			$tokenType,
+			$expected[ self::NON_GLOBAL_NESTED ],
+			$expectedMarker,
+			self::PARAMETER_MAP[ self::NON_GLOBAL_NESTED ]
+		);
+	}
+
+	/**
```
jrfnl, line 257: Quick note, not a blocker - the test order seems a bit weird to me, with the below two tests in between the tests calling the helper method and the actual helper method definition.

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.php
```diff
@@ -0,0 +1,422 @@
...
+		$result            = ContextHelper::is_in_function_call(
+			self::$phpcsFile,
+			$insideFunctionPtr,
+			array()
+		);
+
+		$this->assertFalse( $result );
+	}
+
+	/**
```
jrfnl: And another quick note - **_do not address in this PR, leave it for a follow up PR later_** :
As the `ContextHelper::is_in_function_call()` is a helper function, maybe it should not contain assumptions about the case of the function names and should be made to handle things case-insensitively.

We discussed this in a call and @rodrigoprimo will create a follow up PR for this once this PR has been merged.

jrfnl:

WordPress/Tests/Helpers/ContextHelper/IsInFunctionCallUnitTest.php
rodrigoprimo: I just created the PR: https://github.com/WordPress/WordPress-Coding-Standards/pull/2706

Issue #2420, opened 2024-02-05
Outdated sniff: WordPress.WP.EnqueuedResourceParameters.NotInFooter


luistar15:
## Bug Description
[`$in_footer` is no longer a parameter of `wp_register_script`](https://developer.wordpress.org/reference/functions/wp_register_script/#parameters).
Now is part of `$args` parameter.

jrfnl:
@luistar15 Thanks for bringing this to our attention.

Just had a quick look: this change was made in WP 6.3, so will need a minimum WP based toggle. Additionally, what with `$args` being an array, this will now become much, much harder to detect.

dingo-d:
> Just had a quick look: this change was made in WP 6.3, so will need a minimum WP based toggle. Additionally, what with `$args` being an array, this will now become much, much harder to detect.

Even better, that argument can be a boolean or an array (which was done to avoid BC break). 

It seems that at least the array has specific keys so these can be checked but this definitely complicates things.

jrfnl:
True, the array has specific keys, but what with it being an array, chances of it not being defined in the function call, but elsewhere are much much higher (class constant, class property, passed in via a function parameter, declared within the same function, but not in the function call, conditionally declared)

To be honest, I think it will mean that the sniff will have to be removed as we can no longer reliably detect anything about this.

PR #2630, opened 2025-10-15
WP/EnqueuedResourceParameters: fix handling of non-lowercased `null` and fully qualified `\null` and `\false`

rodrigoprimo:
# Description

This PR fixes how the sniff handles non-lowercased `null` when passed as the value of the `$ver` parameter. Now the sniff consistently handles `null` regardless of the case and always returns a warning. Before, it would return a warning only for lower-cased `null` and an error for all other variations of `null`.

It also fixes how the sniff handles fully qualified `\false` and `\null`. Before this PR, passing `\false` or `\null` as the `$ver` parameter would result in a false negative. Now it correctly results in a warning for `\null` and an error for `\false`. For this fix, it was necessary to raise the minimum supported PHPCS version to 3.13.4 to benefit from a related change in how PHPCS tokenizes `\false` and `\null`.

Besides that, the PR includes a separate commit that fixes some mistakes in the code comments of this sniff.

## Suggested changelog entry

Changed:
- The minimum required `PHP_CodeSniffer` version to 3.13.4 (was 3.13.0).

Fixed:

- `WordPress.WP.EnqueuedResourceParameters`: correctly handles non-lowercase `null` when passed as the `$ver` parameter value.
- `WordPress.WP.EnqueuedResourceParameters`: now correctly handles fully qualified `\null` and `\false` values in the `$ver` parameter.

### DIFF
--- a/WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
+++ b/WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
@@ -54,14 +54,15 @@ final class EnqueuedResourceParametersSniff extends AbstractFunctionParameterSni
 	);
 
 	/**
-	 * False + the empty tokens array.
+	 * False + T_NS_SEPARATOR + the empty tokens array.
 	 *
 	 * This array is enriched with the $emptyTokens array in the register() method.
 	 *
 	 * @var array<int|string, int|string>
 	 */
 	private $false_tokens = array(
-		\T_FALSE => \T_FALSE,
+		\T_FALSE        => \T_FALSE,
+		\T_NS_SEPARATOR => \T_NS_SEPARATOR, // Needed to handle fully qualified \false (PHPCS 3.x).
 	);
 
 	/**
@@ -141,7 +142,7 @@ final class EnqueuedResourceParametersSniff extends AbstractFunctionParameterSni
 			}
 		}
 
-		if ( false === $version_param || 'null' === strtolower( $version_param['clean'] ) ) {
+		if ( false === $version_param || strtolower( ltrim( $version_param['clean'], '\\' ) ) === 'null' ) {
 			$type = 'script';
 			if ( strpos( $matched_content, '_style' ) !== false ) {
 				$type = 'style';
--- a/WordPress/Tests/WP/EnqueuedResourceParametersUnitTest.1.inc
+++ b/WordPress/Tests/WP/EnqueuedResourceParametersUnitTest.1.inc
@@ -98,3 +98,13 @@ wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array
 
 // Safeguard handling of non-lowercase `null`.
 wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), NULL, true ); // Warning - 0, false or NULL are not allowed.
+
+/*
+ * Safeguard handling of fully qualified \true, \false and \null.
+ * Also safeguard that adding T_NS_SEPARATOR to $false_tokens doesn't cause false positives due to problems in is_falsy().
+ */
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \FALSE, \true ); // Error - 0, false or NULL are not allowed.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \null, \TRUE ); // Warning - 0, false or NULL are not allowed.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \Null, true ); // Warning - 0, false or NULL are not allowed.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \true, \False ); // Ok.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \get_version(), \null ); // OK.
--- a/WordPress/Tests/WP/EnqueuedResourceParametersUnitTest.php
+++ b/WordPress/Tests/WP/EnqueuedResourceParametersUnitTest.php
@@ -31,22 +31,23 @@ final class EnqueuedResourceParametersUnitTest extends AbstractSniffUnitTest {
 		switch ( $testFile ) {
 			case 'EnqueuedResourceParametersUnitTest.1.inc':
 				return array(
-					6  => 1,
-					9  => 1,
-					10 => 1,
-					12 => 1,
-					13 => 1,
-					14 => 1,
-					22 => 1,
-					54 => 1,
-					57 => 1,
-					61 => 1,
-					82 => 1,
-					85 => 1,
-					89 => 1,
-					92 => 1,
-					95 => 1,
-					97 => 1,
+					6   => 1,
+					9   => 1,
+					10  => 1,
+					12  => 1,
+					13  => 1,
+					14  => 1,
+					22  => 1,
+					54  => 1,
+					57  => 1,
+					61  => 1,
+					82  => 1,
+					85  => 1,
+					89  => 1,
+					92  => 1,
+					95  => 1,
+					97  => 1,
+					106 => 1,
 				);
 
 			case 'EnqueuedResourceParametersUnitTest.2.inc':
@@ -79,6 +80,8 @@ final class EnqueuedResourceParametersUnitTest extends AbstractSniffUnitTest {
 					66  => 2,
 					77  => 1,
 					100 => 1,
+					107 => 1,
+					108 => 1,
 				);
 
 			default:


jrfnl:
> Fully qualified \null will be addressed in a subsequent PR together with fully qualified \false.

Why ? This feels to me like you are doubling the work for reviewers. Unless the changes are huge (which I don't expect them to be), it could/should just be a separate commit in this PR.

rodrigoprimo:
> Why ? This feels to me like you are doubling the work for reviewers. Unless the changes are huge (which I don't expect them to be), it could/should just be a separate commit in this PR.

I opted to separate precisely because I thought it would make the work for reviewers easier. The changes are indeed not huge and are related, but slightly different. I thought they were different enough to justify a separate PR.

Since you prefer to review them together, I added another commit to this PR with the fix for handling fully qualified `\null` and `\false` and edited the PR title and description. Adding this commit here uncovered an issue that I had missed.

My fix for handling fully qualified `\null` and `\false` only works on PHPCS >= 3.13.3, as this version changed how they are tokenized (https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/1206), and WPCS still supports PHPCS >= 3.13.0. I don't think it is worth changing the code here to also support how PHPCS tokenizes fully qualified `\null` and `\false` before 3.13.3. Is bumping the minimum required PHPCS version to 3.13.3 an option?

jrfnl:
> Is bumping the minimum required PHPCS version to 3.13.3 an option?

Totally and this change/fix is a good reason to justify the dependency bump.

rodrigoprimo:
Ok, I added a new commit to this PR bumping the minimum required PHPCS version to 3.13.4. Let me know if you prefer that this change is introduced in a separate PR and I will do that.

We need to bump to 3.13.4 instead of 3.13.3 due to the bug introduced in 3.13.3 and fixed in 3.13.4 that prevents the execution of sniff tests in external standards.

.github/CONTRIBUTING.md
```diff
@@ -62,7 +62,7 @@ When you introduce new `public` sniff properties, or your sniff extends a class
 
 ### Pre-requisites
 * WordPress-Coding-Standards
-* PHP_CodeSniffer 3.13.0 or higher
+* PHP_CodeSniffer 3.13.3 or higher
```
jrfnl:
```suggestion
* PHP_CodeSniffer 3.13.4 or higher
```

WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
```diff
@@ -18,7 +18,7 @@
  * This checks the enqueued 4th and 5th parameters to make sure the version and in_footer are set.
```
jrfnl: In light of named parameters, maybe the `4th` and `5th` references can be accompanied by the parameter name for clarity too ?

WordPress/Tests/WP/EnqueuedResourceParametersUnitTest.1.inc
```diff
@@ -98,3 +98,13 @@ wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array
...
+
+/*
+ * Safeguard handling of fully qualified \true, \false and \null.
+ * Also safeguard that adding T_NS_SEPARATOR to $false_tokens doesn't cause false positives due to problems in is_falsy().
+ */
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \FALSE, true ); // Error - 0, false or NULL are not allowed.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \null, true ); // Error - 0, false or NULL are not allowed.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \Null, true ); // Error - 0, false or NULL are not allowed.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \true, true ); // Ok.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \get_version(), true ); // OK.
```
jrfnl: For the fourth (`$ver`) parameter, these tests cover the cases sufficiently.

For the fifth parameter, we are not really looking at the value, but as it's typically a boolean anyway, we may as well have some case and FQN variations in these test case lines for the fifth parameter. Just as a safeguard in case how the fifth parameter is checked changes in the future.

Note: this is a non-blocking comment, but would be very nice to include in this PR.

WordPress/Tests/WP/EnqueuedResourceParametersUnitTest.1.inc
```diff
@@ -95,3 +95,16 @@ wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array
...
+// Safeguard handling of non-lowercase `null`.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), NULL, true ); // Error - 0, false or NULL are not allowed.
+
+/*
+ * Safeguard handling of fully qualified \true, \false and \null.
+ * Also safeguard that adding T_NS_SEPARATOR to $false_tokens doesn't cause false positives due to problems in is_falsy().
+ */
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \FALSE, true ); // Error - 0, false or NULL are not allowed.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \null, true ); // Error - 0, false or NULL are not allowed.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \Null, true ); // Error - 0, false or NULL are not allowed.
```
jrfnl:
```suggestion
wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \null, true ); // Warning - 0, false or NULL are not allowed.
wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), \Null, true ); // Warning - 0, false or NULL are not allowed.
```

WordPress/Tests/WP/EnqueuedResourceParametersUnitTest.1.inc
```diff
@@ -95,3 +95,16 @@ wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array
 wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), (double) 0, true ); // Error - 0, false or NULL are not allowed.
 
 wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), (binary) 0, true ); // Error - 0, false or NULL are not allowed.
+
+// Safeguard handling of non-lowercase `null`.
+wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), NULL, true ); // Error - 0, false or NULL are not allowed.
```
jrfnl:
```suggestion
wp_register_script( 'someScript-js', 'https://example.com/someScript.js' , array( 'jquery' ), NULL, true ); // Warning - 0, false or NULL are not allowed.
```

jrfnl (commented):
Just got some nitpicky remarks. Other than that, all good.

WordPress/Sniffs/WP/EnqueuedResourceParametersSniff.php
rodrigoprimo: I added your suggestion, but I made another modification to mention that the 5th parameter is only checked for `wp_register_script()` and `wp_enqueue_script()`. What do you think?

Adding the parameter name prompted another question. I wasn't sure if I should use `$in_footer` or the new `$args` name. Considering that the sniff currently only checks for booleans and not arrays and searches for the `$in_footer` parameter, I opted to use `$in_footer`. That being said, since the suggestion in #2420 is to remove this check from the sniff, I'm no longer sure if it is worth updating the docs to use the parameter name, as it will either be the old parameter name or the current parameter name that is not checked by the sniff. Thoughts?

jrfnl: > I added your suggestion, but I made another modification to mention that the 5th parameter is only checked for `wp_register_script()` and `wp_enqueue_script()`. What do you think?

LGTM

> Adding the parameter name prompted another question. I wasn't sure if I should use `$in_footer` or the new `$args` name. Considering that the sniff currently only checks for booleans and not arrays and searches for the `$in_footer` parameter, I opted to use `$in_footer`. That being said, since the suggestion in #2420 is to remove this check from the sniff, I'm no longer sure if it is worth updating the docs to use the parameter name, as it will either be the old parameter name or the current parameter name that is not checked by the sniff. Thoughts?

My thoughts are quite straight-forward for this:
* Class/inline documentation should reflect the **_current_** state of a sniff.
    As the sniff currently still analyses code with a presumption of an `$in_footer` parameter for the methods involved, that's what the documentation should reflect.
* This current PR is not the place to address #2420, so any decision making/considerations related to 2420 are null and void unless and until issue 2420 will be addressed.
    In other words: while relevant in the larger picture of tasks to be done for this sniff, issue 2420 is not relevant in the here and now of this PR.

rodrigoprimo: Thanks for your input. Sounds good to me.

PR #2631, opened 2025-10-16
WP/GetMetaSingle: simplify code using two class constants

rodrigoprimo:
# Description

Now that support for PHP < 7.2 has been dropped in PR #2614, we can simplify the sniff code by defining the two different function signature formats as class constants. This was not possible before, as PHP < 5.6 did not support defining class constants as arrays.

## Suggested changelog entry
_N/A_

### DIFF
--- a/WordPress/Sniffs/WP/GetMetaSingleSniff.php
+++ b/WordPress/Sniffs/WP/GetMetaSingleSniff.php
@@ -43,7 +43,7 @@ final class GetMetaSingleSniff extends AbstractFunctionParameterSniff {
 	 *
 	 * @var array<string, array<string, int|string>>
 	 */
-	const GENERIC_META_FUNCTIONS_FORMAT = array(
+	private const GENERIC_META_FUNCTIONS_FORMAT = array(
 		'condition'   => array(
 			'param_name' => 'meta_key',
 			'position'   => 3,
@@ -64,7 +64,7 @@ final class GetMetaSingleSniff extends AbstractFunctionParameterSniff {
 	 *
 	 * @var array<string, array<string, int|string>>
 	 */
-	const SPECIFIC_META_FUNCTIONS_FORMAT = array(
+	private const SPECIFIC_META_FUNCTIONS_FORMAT = array(
 		'condition'   => array(
 			'param_name' => 'key',
 			'position'   => 2,


WordPress/Sniffs/WP/GetMetaSingleSniff.php
```diff
@@ -33,6 +33,48 @@ final class GetMetaSingleSniff extends AbstractFunctionParameterSniff {
 	 */
 	const METRIC_NAME = 'get_*meta() function called with $single parameter';
 
+	/**
+	 * Relevant signature structure for specific meta functions.
+	 *
+	 * These functions use 'key' as the parameter name at position 2 and
+	 * 'single' as the parameter name at position 3.
+	 *
+	 * @since 3.2.0
```
dingo-d: Is the since tag correct? 3.2.0 has been released already.

rodrigoprimo: Thanks for the review, @dingo-d!

> Is the since tag correct? 3.2.0 has been released already.

Ops, my bad. It is fixed now.

WordPress/Sniffs/WP/GetMetaSingleSniff.php
```diff
@@ -33,6 +33,48 @@ final class GetMetaSingleSniff extends AbstractFunctionParameterSniff {
 	 */
 	const METRIC_NAME = 'get_*meta() function called with $single parameter';
 
+	/**
```
jrfnl, line 36: Can the order of these constants please be reversed ?

To me, the logical order would be to go from generic to more specific.
And even if that would be too arbitrary, we could use alphabetical order and that would still result in the same: the `GENERIC_META_FUNCTIONS_FORMAT` first, the `SPECIFIC_META_FUNCTIONS_FORMAT` constant second.

rodrigoprimo: Done, I pushed a new commit reversing the order.

jrfnl (approved):
Thanks @rodrigoprimo !

Only remark I have left is that as support for PHP < 7.2 has been dropped, these new class constants could be defined with visibility and in this case, I'd recommend `private` visibility as the info is only relevant to this sniff.

Not a blocker though, as the pre-existing constants don't have visibility modifiers yet, so I expect that to be changed in a future PR anyway.

Only thing to remember about this is that the pre-existing constants _must_ be declared as `public` and the visibility can't be changed until WPCS 4.0, while for new class constants like these, we can use the correct visibility from the get-go.

dingo-d (approved):

PR #2620, opened 2025-09-19
Sniffs: add tests for namespaced names

rodrigoprimo:
# Description

This PR is a continuation of #2581.

In preparation for PHPCS 4.0, which changes the tokenization of namespaced names, this PR adds tests with all variants of namespace names (unqualified, partially qualified, fully qualified, and namespace relative) to the following sniffs: `WordPress.Utils.I18nTextDomainFixer`, `WordPress.Security.NonceVerification`, `WordPress.NamingConventions.ValidHookName`, `WordPress.NamingConventions.PrefixAllGlobals`, `WordPress.PHP.NoSilencedErrors`, `WordPress.WP.CronInterval`, and `WordPress.WP.DiscouragedConstants`.

PR #2633, opened 2025-10-16
AbstractFunctionParameter sniffs: add tests for namespaced names

rodrigoprimo:
# Description

In preparation for PHPCS 4.0, which changes the tokenization of namespaced names, this PR adds tests with all forms of namespaced function calls (partially qualified, fully qualified, and namespace-relative using the 'namespace' keyword) as well as fully qualified global function calls to the sniffs that extend the `AbstractFunctionParameterSniff` class when necessary. Some test files already contained part or all of the required tests.

Tests for `WordPress.WP.EnqueuedResourceParameters` are not included in this PR as they would conflict with #2630. I will add tests for this sniff in a separate PR once #2630 is merged.

In some cases, it was necessary to move syntax error tests to a separate file before adding the new tests.

## Suggested changelog entry

_N/A_


rodrigoprimo:
Moved to draft pending the review of https://github.com/WordPress/WordPress-Coding-Standards/pull/2620

rodrigoprimo:
@jrfnl, since another PR was reviewed earlier today, I'm marking this one as ready for review now.

WordPress/Tests/WP/DeprecatedParametersUnitTest.inc
```diff
@@ -100,3 +100,12 @@ inject_ignored_hooked_blocks_metadata_attributes('', 'deprecated');
 // All will give an WARNING as they have been deprecated after WP 6.6.
 wp_render_elements_support_styles('deprecated');
 _wp_can_use_pcre_u('deprecated');
+
+/*
```
jrfnl: Placement-wise, this will break the ease of updating this sniff to account for new WP releases and/or new deprecations (as we use `array_fill()` logic in the data providers).

I'd like to suggest to either add these tests in a separate file, or - as the data providers are relatively straight-forward to update in this case (not a long list of number), to move these tests up to be above the "marker" on line 40.

jrfnl (approved):
@rodrigoprimo Thanks for all your work on this and your patience waiting for the review.

All good from me!

rodrigoprimo:
@GaryJones and @dingo-d, would either of you be available to review this PR? It already has one approval. Thanks.

GaryJones (approved):

rodrigoprimo:
As discussed in a chat, I squashed the commit applying changes discussed during the PR review into the original commit that adds tests for `WordPress.WP.DeprecatedParameters`.

Issue #2641, opened 2025-11-05
Security/ValidatedSanitizedInput: update error message to mention array_key_exists() and null coalesce operators


rodrigoprimo:
The `WordPress.Security.ValidatedSanitizedInput.InputNotValidated` error message currently suggests using `isset()` or `empty()` to check that superglobal array indexes exist:

> Detected usage of a possibly undefined superglobal array index: %s. Use isset() or empty() to check the index exists before using it

However, the sniff also recognizes additional validation methods that are not mentioned in the error message:

- `array_key_exists()` and its alias `key_exists()` (PR #1635)
- The null coalesce operator `??` and the null coalesce equal operator `??=` (PR #1684)

The current error message was written based on discussion in issue #1541. At that time, `array_key_exists()` support had already been added, but was likely not included due to an oversight. The null coalesce operators were added shortly after.

## Proposed change

Update the error message to:

> Detected usage of a possibly undefined superglobal array index: %s. Use isset(), empty(), array_key_exists(), or the ?? / ??= operators to ensure the index exists before using it

Notes:

- I opted to not include `key_exists()` as it is an `array_key_exists()` alias.
- The proposed message changes "check" to "ensure" because while `isset()`, `empty()`, `array_key_exists()`, and `??` check/validate if an index exists, the `??=` operator ensures an index exists by assigning a default value. To me, the word "ensure" better covers both cases, but I don't have a strong opinion here.

Should all of these validation methods be included in the error message?

jrfnl:
@rodrigoprimo We can always iterate on the wiki text, so feel free to just post it.

Looking at the diff you posted here, I'd suggest the following updates:

```diff
-## Validation
+## Checking for undefined array keys
```

I'd suggest changing that to "Validating received input" - keeping the term "validation" in the heading is important for searchability as that term is used in the sniff name.

```diff
+The `WordPress.Security.ValidatedSanitizedInput` sniff checks that array keys exist before accessing superglobal variables. When accessing input data from a superglobal array, you need to first verify that the array key you're trying to access exists. This prevents issues with undefined array keys and ensures the code handles missing input gracefully.
```
I'd remove the complete first sentence (`The ... variables.` - about the sniff) as it is incorrect and doesn't add any value.

"You'll get an error:" => "You should see the following error:"
[...]

PR #2642, opened 2025-11-06
Security/ValidatedSanitizedInput: improve `InputNotValidated` error message

rodrigoprimo:
# Description

The error message for `InputNotValidated` previously listed specific methods (isset() and empty()), but the list was incomplete. Rather than expanding it, this PR adopts a more generic message that won't require updates when new validation methods are added.

@jrfnl, I made a small modification to the phrase that you suggested by adding "that" after "check". This way the sentences reads a bit better to me, but let me know if you disagree.

> Detected usage of a possibly undefined superglobal array index: %s. Check that the array index exists before using it.

## Suggested changelog entry

Changed: 

`WordPress.Security.ValidatedSanitizedInput`: the `InputNotValidated` error message is now more generic.

Discussed in #2641


jrfnl (approved):
LGTM

PR #2643, opened 2025-11-06
AbstractClassRestrictionsSniff sniffs: add tests for namespaced names

rodrigoprimo:
# Description

In preparation for PHPCS 4.0, which changes the tokenization of namespaced names, this PR adds tests with all forms of namespaced function calls (partially qualified, fully qualified, and namespace-relative using the 'namespace' keyword) as well as fully qualified global function calls to the sniffs that extend the `AbstractClassRestrictionsSniff` class when necessary. Some test files already contained part or all of the required tests.

For `DeprecatedClasses`, I opted to add the tests in a separate file as the error lines `DeprecatedClassesUnitTest.1.inc` are generated dynamically. Since the tests were added to a separate file, I opted to include all possible namespace variations even though a few already exist in `DeprecatedClassesUnitTest.1.inc`. I didn't want to complicate things by having the namespaced names tests in different files.

## Suggested changelog entry
_N/A_

### DIFF
--- a/WordPress/Tests/DB/RestrictedClassesUnitTest.1.inc
+++ b/WordPress/Tests/DB/RestrictedClassesUnitTest.1.inc
@@ -30,7 +30,7 @@ class MyMysqli extends mysqli {}
 class YourMysqli extends \mysqli {}
 
 class OurMysqli implements mysqli {}
-class TheirMysqli implements \mysqli {}
+class TheirMysqli implements \MYSQLI {}
 
 $db5 = new PDO();
 $db6 = ( new PDO() )->exec();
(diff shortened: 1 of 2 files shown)


rodrigoprimo:
@jrfnl, this PR is now ready for review. I implemented the changes that we discussed for #2620. I also removed tests for `CapitalPDangit` and added tests for `RestrictedClasses`. This way, this PR is now about namespaced names tests for sniffs that extend `AbstractClassRestrictionsSniff`.

WordPress/Tests/WP/DeprecatedClassesUnitTest.2.inc
```diff
@@ -0,0 +1,20 @@
+<?php
+
+/*
+ * Safeguard correct handling of all types of namespaced class name references.
+ */
```
jrfnl, line 8: In the category of "don't make me think":

* Maybe make a note that there will only be errors, no warnings, due to the class names used in the code samples all having been deprecated in a WP version < 3 version before the current version ?
* Might also help to add a `// Error.` comment after the lines on which errors are expected.

WordPress/Tests/WP/ClassNameCaseUnitTest.inc
```diff
@@ -65,3 +65,21 @@ class NotYetDetected {
 	public function paramTypeDeclaration( wp_role $role ) {}
 	public function returnTypeDeclaration() : WP_TERM {}
 }
+
+/*
+ * Safeguard correct handling of namespaced class references (the namespace types not handled below are already handled
+ * in other tests above).
+ */
```
jrfnl, line 75: I think it would make sense to include some tests with `implements` here as the "class name" lists in the sniff also include interface names.

WordPress/Tests/DB/RestrictedClassesUnitTest.1.inc
```diff
@@ -115,3 +115,20 @@ $anon = new readonly class {
 $anon = new readonly class() extends PDOStatement {}; // Error.
 
 $anon = new #[MyAttribute] readonly class {};
+
+/*
+ * Safeguard correct handling of namespaced class references (the namespace types not handled below are already handled
+ * in other tests above).
+ */
```
jrfnl, line 122: As this sniff is the one also testing the abstract sniff, I think it would be prudent to also have tests with `implements` here.

rodrigoprimo:
Thanks for your review, @jrfnl. I addressed the points that you raised and, as we discussed, amended the original commits.

rodrigoprimo:

dingo-d (approved):

PR #2656, opened 2025-11-24
Update the minimum_wp_version to WP 6.6

jrfnl:
# Description
The minimum version should be three versions behind the latest WP release, so what with 6.9.0 slated for release on Dec 2nd, it should now become 6.6.

Includes updating the tests to match.

## Suggested changelog entry
Changed
- The default value for `minimum_wp_version`, as used by a [number of sniffs detecting usage of deprecated WP features](https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#various-sniffs-set-the-minimum-supported-wp-version), has been updated to `6.6`.

Previous: #2121, #2321. #2436, #2553

### DIFF
--- a/WordPress/Helpers/MinimumWPVersionTrait.php
+++ b/WordPress/Helpers/MinimumWPVersionTrait.php
@@ -79,7 +79,7 @@ trait MinimumWPVersionTrait {
 	 *
 	 * @var string WordPress version.
 	 */
-	private $default_minimum_wp_version = '6.5';
+	private $default_minimum_wp_version = '6.6';
 
 	/**
 	 * Overrule the minimum supported WordPress version with a command-line/config value.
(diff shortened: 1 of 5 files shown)


rodrigoprimo (commented):
Looks good to me!

The only question that I have is if this PR should also update [phpcs.xml.dist.simple](https://github.com/WordPress/WordPress-Coding-Standards/blob/da2054db75796de8bef5ff7904de0c2392aba9e3/phpcs.xml.dist.sample) to bump the value of the `minimum_wp_version` property to `6.6` and `testVersion` to `7.2-` (as [WP 6.6 dropped support for PHP 7.0 and 7.1](https://make.wordpress.org/core/2024/04/08/dropping-support-for-php-7-1/))? Or is the plan to address this separately?

rodrigoprimo:
> If you believe the setting and the example file should be tightly coupled, I suggest you submit a PR to update the release checklist proposing that and sync the example file while you are at it. 🤷🏻‍♀️

I opened a PR with those changes: https://github.com/WordPress/WordPress-Coding-Standards/pull/2662

Issue #1440, opened 2018-07-29
WordPress != WordPress-Core,WordPress-Docs,WordPress-Extra


GaryJones:
After refreshing https://github.com/GaryJones/pwa-wp/tree/phpcs-refresh, no violations are reported when running `phpcs`. All good.

If I change:

```xml
<rule ref="WordPress-Core"/>
<rule ref="WordPress-Docs"/>
<rule ref="WordPress-Extra"/>
```

to:

```xml
<rule ref="WordPress"/>
```

Then new violations are reported:

```
FILE: wp-includes/class-wp-service-workers.php
-----------------------------------------------------------------------------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 1 LINE
-----------------------------------------------------------------------------------------------------------------------------------------
 117 | ERROR | Missing wp_unslash() before sanitization. (WordPress.Security.ValidatedSanitizedInput.MissingUnslash)
 117 | ERROR | Detected usage of a non-sanitized input variable: $_SERVER
     |       | (WordPress.Security.ValidatedSanitizedInput.InputNotSanitized)
-----------------------------------------------------------------------------------------------------------------------------------------
```

Is that supposed to happen?

cc @westonruter

jrfnl:
This is a known non-issue.

When we moved/renamed the original badge of sniffs as discussed in #1157, most were added to the `Extra` ruleset in #1261, save three for which more discussion/bug fixes were needed. The sniff you refer to is one of those three.

This was annotated in the PR at the time.

> Note: There are three more candidates for this:
> * DB.DirectDatabaseQuery
> * DB.SlowDBQuery
> * Security.ValidatedSanitizedInput
> 
> I've not included the first two (yet) as those may need some more discussion first.
> The third one is not included as it throws a lot of noise messages and has quite some open issues surrounding it, which should be addressed first IMO.

See #1261

GaryJones:
Sorry @jrfnl, I don't follow.

Why does this violation appear when checking against `WordPress`, but not when checking against the three sub-standards together?

jrfnl:
> Why does this violation appear when checking against WordPress, but not when checking against the three sub-standards together?

Because the above mentioned three sniffs are not included in any of the three sub-standards.

The `WordPress` ruleset included the three sub-standards + every single sniff in the `Sniffs` directory, which is why these three will be included when you run against the `WordPress` standard.

[...]

PR #2661, opened 2025-11-26
Add note to README explaining that `WordPress` includes sniffs beyond the subsets.

leewillis77:
Adds a small explanation that running `WordPress` is not the same as running `WordPress-Core,WordPress-Docs,WordPress-Extra` as noted in https://github.com/WordPress/WordPress-Coding-Standards/issues/1440

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,8 @@ You can use the following as standard names when invoking `phpcs` to select snif
   - `WordPress-Extra` - extended ruleset with recommended best practices, not sufficiently covered in the WordPress core coding standards
     - includes `WordPress-Core`
 
+__Note:__ The `WordPress` standard includes some additional checks that are not included in any of the subsets, so running `WordPress-Core` + `WordPress-Docs` + `WordPress-Extra` subsets on their own will miss some checks that would be performed by running `WordPress`.
+
 ### Using a custom ruleset
 
 If you need to further customize the selection of sniffs for your project - you can create a custom ruleset file.


jrfnl:
Thanks @leewillis77, that looks like a valid addition.

I wonder whether people will raise the question "what is the difference" and whether it would make sense to answer that here.... Opinions ?

leewillis77:
> I wonder whether people will raise the question "what is the difference" and whether it would make sense to answer that here.... Opinions ?

There's no documentation on what sniffs are included in which standard as-is so I don't think this is necessary, but I guess it may be useful, maybe just link the text to https://github.com/WordPress/WordPress-Coding-Standards/tree/develop/WordPress/Sniffs ?

rodrigoprimo:
I'm not sure if that is what @jrfnl had in mind, but from my perspective, including in the note which sniffs are part of the `WordPress` standard and are not part of any of the subsets would be enough to answer the questions of what the difference is. What do you think?

leewillis77:
> I'm not sure if that is what @jrfnl had in mind, but from my perspective, including in the note which sniffs are part of the `WordPress` standard and are not part of any of the subsets would be enough to answer the questions of what the difference is. What do you think?

My personal perspective is that it's too detailed for that document, but happy to add it into the PR if that's requested.

leewillis77:
If we were going to add the information I'd suggest a footnote something like this:

https://github.com/leewillis77/WordPress-Coding-Standards/tree/footnote-for-wp-standard-checks?tab=readme-ov-file#rulesets

Let me know if you would like me to update the PR to include this change.

jrfnl:
Marking this PR as **_BLOCKED_** for the time being until the revived discussion in #1440 has been resolved.

Also see: https://github.com/WordPress/WordPress-Coding-Standards/issues/1440#issuecomment-3583648158

PR #2663, opened 2025-11-29
AbstractFunctionRestrictionsSniff: improve namespaced names tests

rodrigoprimo:
# Description

This PR improves the namespaced names tests for the sniffs that extend the `AbstractFunctionRestrictionsSniff` class based on what was discussed during the review of #2581 and #2620. The sniffs `WordPress.Security.EscapeOutput` and `WordPress.WP.AlternativeFunctions` are not included here as they are addressed separately in #2618 and #2617, respectively.

### WP/DiscouragedFunctions: rename test case file

Doing this to be able to move an intentional syntax error test to its own file.

### Tests: non-standard case function call tests for sniffs extending AbstractFunctionRestrictionsSniff

This commit ensures that the tests for all sniffs extending the `AbstractFunctionRestrictionsSniff` class contain a test with a non-standard case, an unqualified, and a fully qualified function call. In some cases, it was possible to update an existing test, and in other cases, it was necessary to add a new one.

For small test files, new tests were inserted alongside related existing tests (rather than appended at the end) to keep similar test cases grouped. The improved test organization should outweigh the added review complexity due to the need to update line numbers in test expectations.

Two sniffs that extend AbstractFunctionRestrictionsSniff will be addressed separately: `WordPress.Security.EscapeOutput` and `WordPress.WP.AlternativeFunctions`.

### Tests: add namespace\Sub\ tests for sniffs extending AbstractFunctionRestrictionsSniff

This commit adds `namespace\Sub\function_name()` test cases to all test files for sniffs extending `AbstractFunctionRestrictionsSniff`.

Two sniffs that extend `AbstractFunctionRestrictionsSniff` will be addressed separately: `WordPress.Security.EscapeOutput` and `WordPress.WP.AlternativeFunctions`.

### Tests: vary function names in namespace tests for sniffs extending AbstractFunctionRestrictionsSniff

As suggested in #2581, for sniffs that check multiple functions, the namespace name tests now use different function names rather than repeatedly testing the same function.

Two sniffs that extend AbstractFunctionRestrictionsSniff will be addressed separately: `WordPress.Security.EscapeOutput` and `WordPress.WP.AlternativeFunctions`.

## Suggested changelog entry

_N/A_


WordPress/Tests/WP/DeprecatedFunctionsUnitTest.1.inc
```diff
@@ -444,3 +444,4 @@ wp_add_editor_classic_theme_styles();
 /* ============ WP 6.9 ============ */
 seems_utf8();
 wp_print_auto_sizes_contain_css_fix();
+wp_PRINT_auto_SIZES_contain_CSS_fix();
```
jrfnl: Could you please move this test to test case file 2 ? And to prevent it needing an update once WP 7.3 comes out, maybe use one of the functions which is already an error for this specific test ?

WordPress/Tests/PHP/DiscouragedPHPFunctionsUnitTest.inc
jrfnl, line 1: Not sure these tests need to be moved. What about removing the blank line on line 2 ? That way you also shouldn't need to re-number (just add one extra line number).

WordPress/Tests/WP/DiscouragedFunctionsUnitTest.1.inc
jrfnl, line 1: I understand why you moved these (to keep similar tests together), but I'm not sure it is really needed.
Maybe just add the single new test at the bottom and reference that line 15-20 contain associated tests in the comment ?

jrfnl (commented):
Thanks for working on this @rodrigoprimo and for your patience.

Just left a few small remarks, nothing really blocking.

WordPress/Tests/WP/DeprecatedFunctionsUnitTest.1.inc
rodrigoprimo: Good point, I pushed a commit implementing this.

WordPress/Tests/PHP/DiscouragedPHPFunctionsUnitTest.inc
rodrigoprimo: As we discussed during a call, I opted not to implement this non-blocking change but I will keep in mind for future PRs.

WordPress/Tests/WP/DiscouragedFunctionsUnitTest.1.inc
rodrigoprimo: As we discussed during a call, I opted not to implement this non-blocking change but I will keep in mind for future PRs.

jrfnl (approved):

rodrigoprimo:
@GaryJones and @dingo-d, would either of you be available to review this PR? It already has one approval. Thanks.

GaryJones (approved):

PR #2673, opened 2025-12-16
WP/CapitalPDangit: add tests for namespaced names

rodrigoprimo:
# Description

In preparation for PHPCS 4.0, which changes the tokenization of namespaced names, this PR adds tests with all forms of namespaced function calls (partially qualified, fully qualified, and namespace-relative using the `namespace` keyword) to the `WordPress.WP.CapitalPDangit` sniff. The tests focus on calls to the `define()` function and namespaced variants that share the same name.

## Suggested changelog entry
_N/A_


jrfnl (approved):
Thanks for this PR @rodrigoprimo ! LGTM.

N.B.: I've verified that the other place where namespaced names could be problematic for PHPCS 4.0 (namespace declarations) is already sufficiently covered by tests.

dingo-d (approved):

PR #2487, opened 2024-09-17
Adds documentation for PregQuoteDelimiter

tikifez:

PR #2677, opened 2025-12-23
PHP/PregQuoteDelimiter: add XML documentation

rodrigoprimo:
# Description

This PR adds XML documentation for the `WordPress.PHP.PregQuoteDelimiter` sniff.

The documentation is based on the work started by @tikifez in #2487. I squashed the original commits and, in a separate commit, made subsequent changes based on the review left in #2487.

## Suggested changelog entry

_N/A_

Related to: #1722
Supersedes: #2487

### DIFF
--- a/WordPress/Docs/PHP/PregQuoteDelimiterStandard.xml
+++ b/WordPress/Docs/PHP/PregQuoteDelimiterStandard.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Preg Quote Delimiter"
+    >
+    <standard>
+    <![CDATA[
+    Passing the $delimiter parameter to `preg_quote()` is strongly recommended to ensure the regular expression delimiter is properly escaped if it occurs in the input string.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: The delimiter parameter is passed.">
+        <![CDATA[
+// $input = 'my #1 fan!';
+$quoted_input = preg_quote( $input, <em>'#'</em> );
+preg_match(
+    <em>'#^' . $quoted_input . '#i'</em>,
+    $post_content,
+    $matches
+);
+        ]]>
+        </code>
+        <code title="Invalid: The delimiter parameter is missing.">
+        <![CDATA[
+// $input = 'my #1 fan!';
+$quoted_input = preg_quote( $input<em></em> );
+preg_match(
+    <em>'#^' . $quoted_input . '#i'</em>,
+    $post_content,
+    $matches
+);
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>


WordPress/Docs/PHP/PregQuoteDelimiterStandard.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Preg Quote Delimiter"
+    >
+    <standard>
+    <![CDATA[
+    Passing the $delimiter parameter to preg_quote() is strongly recommended to ensure the delimiter used in the regex is properly escaped.
```
jrfnl:
```suggestion
    Passing the $delimiter parameter to preg_quote() is strongly recommended to ensure the regular expression delimiter is properly escaped if it occurs in the arbitrary text string passed to `preg_quote()`.
```

Just to be clear: The `$delimiter` is not used **_in_** the regex, it is used to deliminate the regex (outside boundaries). The whole point of passing the deliminator to `preg_quote()` is to prevent that character from ending up in the regex without escaping (which will often break the regex and can lead to security issues/fatal errors/etc).

This part of the original description was not wrong in that respect:
> If it is not passed, the default delimiter `/` is presumed, which is often wrong.

WordPress/Docs/PHP/PregQuoteDelimiterStandard.xml
```diff
@@ -0,0 +1,23 @@
...
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: The delimiter parameter is passed.">
+        <![CDATA[
+preg_quote( $dir, <em>'/'</em> );
```
jrfnl: While the code samples are technically correct, in my opinion, they are not useful for people to understand what this sniff is warning about.

May I suggest something along the lines of the following ? (feel free to deviate, but I hope this example can help shape further thought on this)

```suggestion
$quoted_input = preg_quote($user_input, <em>'#'</em>);
preg_match(
    <em>'#^' . $quoted_input . '#i'</em>,
    $post_content,
    $matches
);
```

Obviously, the "invalid" example will also need updating.

jrfnl (changes requested):
@rodrigoprimo Thanks for picking this up!

I do think this needs more work though to ensure that the documentation isn't just technically correct, but also helps people to understand why they should follow this best practice.

WordPress/Docs/PHP/PregQuoteDelimiterStandard.xml
rodrigoprimo: I implemented your suggestion with a small change to avoid using `preg_quote()` twice. 

> This part of the original description was not wrong in that respect:
>> If it is not passed, the default delimiter / is presumed, which is often wrong.

I'm not sure whether we are interpreting the above statement differently, but I don't think it is correct. I read it as saying that if the `$delimiter` parameter is not passed, the function uses `/` as the default value, which is not the case.

WordPress/Docs/PHP/PregQuoteDelimiterStandard.xml
rodrigoprimo: Thanks for this suggestion. I agree that it makes the importance of this sniff clearer.

rodrigoprimo:
Thanks for your review, @jrfnl. I implemented the changes you suggested, with a few minor modifications.

WordPress/Docs/PHP/PregQuoteDelimiterStandard.xml
jrfnl: > > This part of the original description was not wrong in that respect:
> > > If it is not passed, the default delimiter / is presumed, which is often wrong.
> 
> I'm not sure whether we are interpreting the above statement differently, but I don't think it is correct. I read it as saying that if the `$delimiter` parameter is not passed, the function uses `/` as the default value, which is not the case.

Ah, good point. All the more reason for the `$delimiter` needing to be passed.

WordPress/Docs/PHP/PregQuoteDelimiterStandard.xml
```diff
@@ -0,0 +1,33 @@
...
+preg_match(
+    <em>'#^' . $quoted_input . '#i'</em>,
+    $post_content,
+    $matches
+);
+        ]]>
+        </code>
+        <code title="Invalid: The delimiter parameter is missing.">
+        <![CDATA[
+$quoted_input = preg_quote( $input<em></em> );
```
jrfnl, line 26: Would this make it even more clear ?

```suggestion
// $input = 'my #1 fan!';
$quoted_input = preg_quote( $input<em></em> );
```

jrfnl (approved):
Thanks @rodrigoprimo ! Happy with it as it is now.

Just left one non-blocking suggestion inline to consider.

WordPress/Docs/PHP/PregQuoteDelimiterStandard.xml
rodrigoprimo: I think that helps make it even more clear. I just added this last suggestion.

rodrigoprimo:
Just documenting that I squashed all the commits into one without changes.

WordPress/Docs/PHP/PregQuoteDelimiterStandard.xml
jrfnl: Ah, but now the examples look unbalanced. If the above is added, it should be added to both sides of the comparison.

rodrigoprimo: Done, I added the same comment to the valid example. I incorrectly assumed that you preferred adding it just to the invalid example for some reason.

jrfnl: Thanks. I'll try to spell these things out more going forward.

jrfnl (approved):

dingo-d (approved):

PR #2678, opened 2025-12-23
CodeAnalysis/AssignmentInTernaryCondition: add XML documentation

rodrigoprimo:
# Description

This PR adds XML documentation for the `WordPress.CodeAnalysis.AssignmentInTernaryCondition` sniff.

The documentation is based on the work started by @Nic-Sevic in #2488. I squashed the original commits and, in a separate commit, made subsequent changes based on the review left in #2488.

## Suggested changelog entry

_N/A_

Related to: #1722
Supersedes: #2488
Closes #2488

### DIFF
--- a/WordPress/Docs/CodeAnalysis/AssignmentInTernaryConditionStandard.xml
+++ b/WordPress/Docs/CodeAnalysis/AssignmentInTernaryConditionStandard.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Assignment In Ternary Condition"
+    >
+    <standard>
+    <![CDATA[
+    Variables should not be assigned in the condition of a ternary expression. This is likely a mistake, as more often than not, a comparison was intended.
+
+    Side-note: This sniff can only detect assignments in ternary conditions when parentheses are used around either the ternary expression or its condition.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Comparison operator used in ternary condition.">
+        <![CDATA[
+echo ( $a <em>===</em> 'a' ) ? 'b' : 'c';
+        ]]>
+        </code>
+        <code title="Invalid: Assignment in ternary condition.">
+        <![CDATA[
+echo ( $a <em>=</em> 'a' ) ? 'b' : 'c';
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>


WordPress/Docs/CodeAnalysis/AssignmentInTernaryConditionStandard.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Assignment In Ternary Condition"
+    >
+    <standard>
+    <![CDATA[
+    Variables should not be assigned in the condition of a ternary expression. This is likely a mistake, as more often than not a comparison was intended.
+
+    Note: This sniff can only detect assignments when the ternary condition is wrapped in parentheses.
```
jrfnl: Not sure mentioning this in the docs has value. It is a limitation in the sniff due to the complexity of determining the start/end of the conditional part while taking operator precedence properly into account.

Also: _if_ it is mentioned, the mention should be correct. (yes, parentheses are needed, but no, they don't have to be around the condition.)

WordPress/Docs/CodeAnalysis/AssignmentInTernaryConditionStandard.xml
```diff
@@ -0,0 +1,25 @@
...
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Comparison operator used in ternary condition.">
+        <![CDATA[
+$mode = ( $a <em>===</em> 'a' ) ? 'b' : 'c';
+        ]]>
+        </code>
+        <code title="Invalid: Assignment in parenthesized ternary condition.">
```
jrfnl:
```suggestion
        <code title="Invalid: Assignment in ternary condition.">
```

WordPress/Docs/CodeAnalysis/AssignmentInTernaryConditionStandard.xml
```diff
@@ -0,0 +1,25 @@
...
+        <![CDATA[
+$mode = ( $a <em>=</em> 'a' ) ? 'b' : 'c';
```
jrfnl: Currently, the result of the ternary is being assigned to `$mode`. As the rule is about assignments (in ternary), that additional assignment makes it less straight-forward for a user to see which assignment is forbidden.

How about this ? (and if so, probably best to also change the "valid" example)

```suggestion
echo ( $a <em>=</em> 'a' ) ? 'b' : 'c';
```

jrfnl (commented):
@rodrigoprimo Thanks for picking this one up. Looking good. Left a couple of comments for you to think over.

WordPress/Docs/CodeAnalysis/AssignmentInTernaryConditionStandard.xml
rodrigoprimo: > Not sure mentioning this in the docs has value. It is a limitation in the sniff due to the complexity of determining the start/end of the conditional part while taking operator precedence properly into account.

I think mentioning this limitation could be helpful for users who wonder why the sniff didn't flag an assignment in a ternary condition when their code doesn't use parentheses. The documentation of the sniff is probably the first place they will check to try to clarify why the sniff is not flagging their code.

> Also: if it is mentioned, the mention should be correct. (yes, parentheses are needed, but no, they don't have to be around the condition.)

In that case, if you agree to keep the note, it can be rewritten as follows:

"Note: This sniff can only detect assignments when parentheses are used around the ternary expression or its condition."

WordPress/Docs/CodeAnalysis/AssignmentInTernaryConditionStandard.xml
rodrigoprimo: I implemented this change to both the valid and invalid examples.

rodrigoprimo:
Thanks for your review, @jrfnl. I believe I addressed all the points you raised, and this PR is ready for another check.

WordPress/Docs/CodeAnalysis/AssignmentInTernaryConditionStandard.xml
jrfnl: Okay, let's go for: "Side-note: This sniff can only detect assignments in ternary conditions when parentheses are used around either the ternary expression or its condition."

rodrigoprimo: Sounds good to me. I just pushed a new commit with this change.

jrfnl (approved):
Thanks @rodrigoprimo ! I believe this one is also ready for merge 🎉

jrfnl:
**Note for the second reviewer**: please **squash**-merge this PR.

WordPress/Docs/CodeAnalysis/AssignmentInTernaryConditionStandard.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Assignment In Ternary Condition"
+    >
+    <standard>
+    <![CDATA[
+    Variables should not be assigned in the condition of a ternary expression. This is likely a mistake, as more often than not a comparison was intended.
```
dingo-d:
```suggestion
    Variables should not be assigned in the condition of a ternary expression. This is likely a mistake, as more often than not, a comparison was intended.
```

dingo-d (changes requested):
Just one minor grammar quirk that stood out to me.

WordPress/Docs/CodeAnalysis/AssignmentInTernaryConditionStandard.xml
rodrigoprimo: Thanks for reviewing this PR, @dingo-d. I just pushed a new commit implementing your suggestion.

dingo-d (approved):

rodrigoprimo:
Just noting that I squashed the commits in this PR. I believe it is ready to be merged.

PR #2679, opened 2025-12-24
WP/GlobalVariablesOverride: add XML documentation

rodrigoprimo:
# Description

This PR adds XML documentation for the `WordPress.WP.GlobalVariablesOverride` sniff.

The documentation is based on the work started by @paulopmt1 in #2586. I squashed the original commits and, in a separate commit, made subsequent changes based on the review left in #2586.

## Suggested changelog entry

_N/A_

Related to: #1722
Supersedes: #2586
Closes #2586

### DIFF
--- a/WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
+++ b/WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
@@ -7,25 +7,25 @@
     <![CDATA[
     WordPress global variables should not be overridden.
 
-    Two WordPress global variables, `$content_width` and `$wp_cockneyreplace`, are 
+    Two WordPress global variables, `$content_width` and `$wp_cockneyreplace`, are
     specifically intended to be overridden by themes/plugins and are exempt from this rule.
     ]]>
     </standard>
     <code_comparison>
-        <code title="Valid: Different variable name.">
+        <code title="Valid: Using non-WordPress global names.">
         <![CDATA[
 <em>$my_query</em> = new WP_Query( $args );
 
 
-<em>$GLOBALS['my_data']</em> = "some data";
+<em>$GLOBALS['my_data']</em> = 'some data';
         ]]>
         </code>
-        <code title="Invalid: Global variable overridden.">
+        <code title="Invalid: WordPress global variable overridden.">
         <![CDATA[
 <em>global $wp_query;</em>
 <em>$wp_query</em> = new WP_Query( $args );
 
-<em>$GLOBALS['wp_query']</em> = new WP_Query( $args );
+<em>$GLOBALS['post']</em> = get_post( 1 );
         ]]>
         </code>
     </code_comparison>
--- a/WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
+++ b/WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Global Variables Override"
+    >
+    <standard>
+    <![CDATA[
+    WordPress global variables should not be overridden.
+
+    Two WordPress global variables, `$content_width` and `$wp_cockneyreplace`, are 
+    specifically intended to be overridden by themes/plugins and are exempt from this rule.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Different variable name.">
+        <![CDATA[
+<em>$my_query</em> = new WP_Query( $args );
+
+
+<em>$GLOBALS['my_data']</em> = "some data";
+        ]]>
+        </code>
+        <code title="Invalid: Global variable overridden.">
+        <![CDATA[
+<em>global $wp_query;</em>
+<em>$wp_query</em> = new WP_Query( $args );
+
+<em>$GLOBALS['wp_query']</em> = new WP_Query( $args );
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>
--- a/WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
+++ b/WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
@@ -12,7 +12,7 @@
     ]]>
     </standard>
     <code_comparison>
-        <code title="Valid: Using non-WordPress global names.">
+        <code title="Valid: Declaring global variables with names not reserved for use by WordPress itself.">
         <![CDATA[
 <em>$prefix_query</em> = new WP_Query( $args );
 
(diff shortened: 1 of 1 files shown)


WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Global Variables Override"
+    >
+    <standard>
+    <![CDATA[
+    WordPress global variables should not be overridden.
```
jrfnl: I believe we should make it clearer what this means as this is quite an important concept.

```suggestion
    Variables declared by WordPress in the global namespace should not be overridden by code extending WordPress.
```

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,32 @@
...
+
+    Two WordPress global variables, `$content_width` and `$wp_cockneyreplace`, are
+    specifically intended to be overridden by themes/plugins and are exempt from this rule.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using non-WordPress global names.">
```
jrfnl: Hmm... my suggestion got a bit long... let's have another good think about this.

```suggestion
        <code title="Valid: Using variable names in the global namespace which do not overlap with variable names in use by WordPress itself.">
```

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,32 @@
...
+        <![CDATA[
+<em>$my_query</em> = new WP_Query( $args );
+
+
+<em>$GLOBALS['my_data']</em> = 'some data';
```
jrfnl:
```suggestion
<em>$prefix_query</em> = new WP_Query( $args );


<em>$GLOBALS['prefix_data']</em> = 'some data';
```

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,32 @@
...
+        ]]>
+        </code>
+        <code title="Invalid: WordPress global variable overridden.">
```
jrfnl:
```suggestion
        <code title="Invalid: Overriding a WordPress defined global variable.">
```

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
```diff
@@ -0,0 +1,32 @@
...
+        <![CDATA[
+<em>global $wp_query;</em>
+<em>$wp_query</em> = new WP_Query( $args );
```
jrfnl: I think this code sample is not very realistic - the `global` keyword is rarely used in the global namespace (as it makes no sense to do so). This code snippet should be placed in the body of a function.

jrfnl (commented):
Thanks for picking this one up @rodrigoprimo ! Left some comments to consider inline.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
rodrigoprimo: What about "Valid: Using global variable names which do not overlap with global names defined by WordPress." or "Valid: Using global variable names not defined by WordPress."?

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
rodrigoprimo: Done

rodrigoprimo:
Thanks for your review, @jrfnl! I implemented your suggestions and proposed two shorter alternatives to the valid title. This PR is ready for another look when you get a chance.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
jrfnl: The thing is "global names" is not a thing. It's colloquial, shorthand phrasing, but not technically correct phrasing.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
jrfnl: The `<em>$GLOBALS['post']</em> = get_post( 1 );` should probably also be within the function body as using `$GLOBALS` when in the global namespace also doesn't make sense.

Maybe for the global namespace code sample, use `<em>$post</em> = get_post( 1 );` ?

jrfnl: Oh... just thinking... we could even show a nastier code sample:
```php
foreach ($selected_posts as <em>$post</em>) {
    // Do something.
}
```

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
rodrigoprimo: I wasn't aware that "global names" is colloquial and not technically correct.

After you pointed that out, I checked the [Variable scope](https://www.php.net/manual/en/language.variables.scope.php) page in the PHP documentation to see which terms are used there.

I think "Valid: Declaring global variables with names not defined by WordPress." is correct.

How does this sound? If it doesn't work, do you have another suggestion?

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
rodrigoprimo: I moved the `<em>$GLOBALS['post']</em>` invalid example to within the function body.

I also added the foreach loop that you suggested. I think that is a good addition. I opted to include a valid version of the foreach loop in the corresponding section.

Let me know what you think about the examples now.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
jrfnl: Thanks for thinking this over some more with me.

How about this ? "Valid: Declaring global variables with names not reserved for use by WordPress itself."

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
jrfnl: I think these look good, though the "line by line" comparison is a little off, but I don't think that's a real problem though.

WordPress/Docs/WP/GlobalVariablesOverrideStandard.xml
rodrigoprimo: I think that is good. I just pushed a commit with this new title.

jrfnl (approved):
Thanks @rodrigoprimo ! I think this one is ready for merge.

jrfnl:
**Note for the second reviewer**: please **squash**-merge this PR.

dingo-d (approved):

PR #2682, opened 2025-12-25
Arrays/ArrayDeclarationSpacing: add XML documentation

rodrigoprimo:
# Description

This PR adds XML documentation for the `WordPress.Arrays.ArrayDeclarationSpacing` sniff.

The documentation is based on the work started by @RafaelFunchal in #2489 and @mattgaldino in #2593. I squashed the original commits and, in a separate commit, made subsequent changes based on code review suggestions.

## Suggested changelog entry

_N/A_

Related to: #1722
Supersedes: #2489 and #2593
Closes #2489
Closes #2593

### DIFF
--- a/WordPress/Docs/Arrays/ArrayDeclarationSpacingStandard.xml
+++ b/WordPress/Docs/Arrays/ArrayDeclarationSpacingStandard.xml
@@ -1,0 +1,58 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Array Declaration Spacing"
+    >
+    <standard>
+    <![CDATA[
+    Multi-item arrays with explicit keys must be declared as multi-line arrays.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Multi-item array with explicit keys declared as multi-line.">
+        <![CDATA[
+$args = array(
+    <em>'cat'   => 1</em>,
+    <em>'paged' => 2</em>,
+);
+        ]]>
+        </code>
+        <code title="Invalid: Single line multi-item array with explicit keys.">
+        <![CDATA[
+$args = array( <em>'cat' => 1, 'paged' => 2</em> );
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Each item in a multi-line array must be on a new line.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Only one array item per line.">
+        <![CDATA[
+$post_types = array(
+    <em>'post'</em>,
+    <em>'page'</em>,
+);
+
+$args = array(
+    <em>'cat'   => 1</em>,
+    <em>'paged' => 2</em>,
+);
+        ]]>
+        </code>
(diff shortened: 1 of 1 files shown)


jrfnl (changes requested):
Thanks for taking this on @rodrigoprimo !

We've just discussed this PR in a call and while the second standard + code comparison are fine, the first is unclear.

The first block _should_ address the _"Arrays with keys containing more than one item, must be declared as multi-line arrays"_ rule, but the docs as they are, are mixing the first and the second rule for the first `<standard>` block as well as the first `<code_comparison>` block.

Want to have another look at this ?

---

Re: your review comment on a previous iterations of this PR - https://github.com/WordPress/WordPress-Coding-Standards/pull/2489#discussion_r1775242855 - which calls out that the sniff _says_ the above mentioned rule is about _associative_ arrays, while in reality the sniff just checked that there are keys, not that the keys are strings.

I agree we may need to update the sniff error message (and the docs to match), but that would first need an update to [the handbook](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#indentation) as this phrasing is taken from there.

In my opinion, that's a minor correction to the handbook which shouldn't need a Make post. The rule has been applied for "arrays with keys" for as long as this sniff exists, not for "associative arrays", so this would just be a case of aligning the handbook with the practiced reality.

rodrigoprimo:
Thanks for your review, @jrfnl.

> We've just discussed this PR in a call and while the second standard + code comparison are fine, the first is unclear.

You are right that I mixed both rules in the first block. I pushed a new commit with an attempt to clarify things. Please take a look when you get a chance.

> I agree we may need to update the sniff error message (and the docs to match), but that would first need an update to [the handbook](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#indentation) as this phrasing is taken from there.

Thanks for confirming my suspicion. I opened a PR in the handbook repository suggesting the change: https://github.com/WordPress/wpcs-docs/pull/156. I also prepared a change to the sniff error message that I will pull here if the handbook change is approved.

jrfnl (approved):
Thanks for that update @rodrigoprimo ! LGTM now.

WordPress/Docs/Arrays/ArrayDeclarationSpacingStandard.xml
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Array Declaration Spacing"
+    >
+    <standard>
+    <![CDATA[
+    Multi-item arrays using keys must be declared as multi-line arrays.
```
jrfnl: Thinking it over a little more - every array has keys, sometimes they are explicit, other times implicit.
So maybe this would be better ?

```suggestion
    Multi-item arrays with explicit keys must be declared as multi-line arrays.
```

rodrigoprimo: Good point, @jrfnl. I added your suggestion and also changed the valid and invalid titles to follow the same pattern, using "with explicit keys" instead of "using keys".

jrfnl (approved):

rodrigoprimo:
Just noting that I squashed the commits in this PR.

dingo-d (approved):

PR #2688, opened 2026-02-02
Arrays/ArrayDeclarationSpacing: replace "associative" with "explicit keys"

rodrigoprimo:
# Description

Updates one error message and docblocks in the `WordPress.Arrays.ArrayDeclarationSpacing` sniff to use "arrays with explicit keys" instead of "associative arrays" to align with the new PHP handbook terminology (see https://github.com/WordPress/wpcs-docs/pull/156, https://github.com/WordPress/wpcs-docs/pull/157, and https://github.com/WordPress/WordPress-Coding-Standards/pull/2682#pullrequestreview-3728288216).

The sniff has always checked for arrays with explicit keys (whether string or numeric), not just associative arrays with string keys. This change aligns the wording with the actual behavior.

I opted not to change the public property `allow_single_item_single_line_associative_arrays` and the error code `AssociativeArrayFound` to avoid breaking changes for users who reference them in their ruleset configurations. I'm unsure if the pros of changing (better alignment with terminology) outweigh the cons (breaking change).

## Suggested changelog entry

Changed:

`WordPress.Arrays.ArrayDeclarationSpacing`: the `AssociativeArrayFound` error message now uses "explicit keys" instead of "associative keys".

- https://github.com/WordPress/wpcs-docs/pull/156
- https://github.com/WordPress/wpcs-docs/pull/157
- https://github.com/WordPress/WordPress-Coding-Standards/pull/2682#pullrequestreview-3728288216

### DIFF
--- a/WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
+++ b/WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
@@ -137,9 +137,9 @@ final class ArrayDeclarationSpacingSniff extends Sniff {
 		if ( false === $array_has_keys ) {
 			return;
 		}
-		$error = 'When an array has explicit keys, each value should start on %s.';
+		$error = 'When an array is declared with explicit keys, each value should start on %s.';
 		if ( true === $this->allow_single_item_single_line_associative_arrays ) {
-			$error = 'When a multi-item array has explicit keys, each value should start on %s.';
+			$error = 'When a multi-item array is declared with explicit keys, each value should start on %s.';
 		}
 
 		/*
--- a/WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
+++ b/WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
@@ -19,21 +19,21 @@ use WordPressCS\WordPress\Sniff;
 /**
  * Enforces WordPress array spacing format.
  *
- * - Checks that associative arrays are multi-line.
+ * - Checks that arrays with explicit keys are multi-line.
  * - Checks that each array item in a multi-line array starts on a new line.
  *
  * @link https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#indentation
  *
  * @since 0.11.0 - The WordPress specific additional checks have now been split off
  *                 from the `WordPress.Arrays.ArrayDeclaration` sniff into this sniff.
- *               - Added sniffing & fixing for associative arrays.
+ *               - Added sniffing & fixing for arrays with explicit keys.
  * @since 0.12.0 Decoupled this sniff from the upstream sniff completely.
  *               This sniff now extends the WordPressCS native `Sniff` class instead.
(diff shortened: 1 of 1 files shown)


WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
```diff
@@ -137,9 +137,9 @@ protected function process_single_line_array( $stackPtr, $opener, $closer ) {
 		if ( false === $array_has_keys ) {
 			return;
 		}
-		$error = 'When an array uses associative keys, each value should start on %s.';
+		$error = 'When an array has explicit keys, each value should start on %s.';
```
jrfnl: Just a suggestion to consider:
```suggestion
		$error = 'When an array is declared with explicit keys, each value should start on %s.';
```

WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
```diff
@@ -137,9 +137,9 @@ protected function process_single_line_array( $stackPtr, $opener, $closer ) {
...
 		if ( true === $this->allow_single_item_single_line_associative_arrays ) {
-			$error = 'When a multi-item array uses associative keys, each value should start on %s.';
+			$error = 'When a multi-item array has explicit keys, each value should start on %s.';
```
jrfnl:
```suggestion
			$error = 'When a multi-item array is declared with explicit keys, each value should start on %s.';
```

jrfnl (commented):
Thanks for opening this PR @rodrigoprimo ! I left two small suggestions inline.

> I opted not to change the public property `allow_single_item_single_line_associative_arrays` and the error code `AssociativeArrayFound` to avoid breaking changes for users who reference them in their ruleset configurations. I'm unsure if the pros of changing (better alignment with terminology) outweigh the cons (breaking change).

I believe the property and the error code should be considered separately.

We could add a new property with a more appropriate name and deprecate the old property. This would not be a breaking change.
This would mean handling both properties until the 4.0 release, but that shouldn't be much of a problem, and then removing the `allow_single_item_single_line_associative_arrays` property in the 4.0 release.

We can't do this for the error code as that _would_ be a breaking change, so I agree that it's better not to change the error code at this time.
What you could consider, would be to open an issue for the 4.0 release to change the error code in that release.

rodrigoprimo:
> Thanks for opening this PR @rodrigoprimo ! I left two small suggestions inline.

Thanks, @jrfnl! I added the two suggestions that you made.

> I believe the property and the error code should be considered separately.

That works for me. I created an issue to get feedback on how to deprecate the property (https://github.com/WordPress/WordPress-Coding-Standards/issues/2691) and another to rename the error code before the WPCS 4.0 release (https://github.com/WordPress/WordPress-Coding-Standards/issues/2692).

dingo-d (approved):

jrfnl (approved):
Thanks for opening those issues.

This PR is good to go as far as I'm concerned.

rodrigoprimo:
Just documenting here that I updated the https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties wiki page to use "explicit keys" instead of "associative" as well in the section related to this sniff.

PR #2690, opened 2026-02-02
PHP/DevelopmentFunctions: add XML documentation

rodrigoprimo:
## Description

This PR adds XML documentation for the `WordPress.PHP.DevelopmentFunctions` sniff.

The documentation is based on the work started by @gogdzl in #2490. I used the original commits and then made subsequent changes, mostly based on the review left in #2490.

I suggest squashing those commits before merging. I'm opening the PR without doing that to make it easier to tell my changes apart from the original changes.

I opted to remove the function lists from the standard descriptions to avoid the maintenance burden of keeping docs in sync when functions are added or removed. Also, this approach is consistent with other similar sniffs like `DB/RestrictedFunctions` and `WP/DeprecatedFunctions` that also don't list all flagged functions.

## Suggested changelog entry

_N/A_

Related to: #1722
Supersedes: #2490
Closes: #2490

### DIFF
--- a/WordPress/Docs/PHP/DevelopmentFunctionsStandard.xml
+++ b/WordPress/Docs/PHP/DevelopmentFunctionsStandard.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Development Functions"
+    >
+    <standard>
+    <![CDATA[
+    Debug code should not normally be used in production.
+
+    Typically, this rule verifies if function calls to the PHP native `error_log()`, `var_dump()`, `var_export()`, `print_r()`, `trigger_error()`, `set_error_handler()`, `debug_backtrace`, `debug_print_backtrace` and `wp_debug_backtrace_summary()` functions are present in the code.
+    ]]>
+    </standard>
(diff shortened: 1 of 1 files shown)


jrfnl (commented):
Thanks for taking this on @rodrigoprimo !

When looking at these docs, I have a feeling that the code comparisons don't actually add any value in this case.

There is precedent for XML docs only containing `<standard>` blocks ([for example](https://github.com/PHPCSStandards/PHP_CodeSniffer/blob/4.x/src/Standards/Generic/Docs/Files/EndFileNoNewlineStandard.xml)), so I wonder if we should consider that for this sniff.

Maybe in combination with one or two examples of the discouraged functions in the `<standard>` description ? (not the complete list to prevent a maintenance nightmare as you rightly pointed out, but just a few examples).

jrfnl (approved):
LGTM

rodrigoprimo:
Thanks for the review, @jrfnl\!

I first removed the `code_comparison` blocks and added a couple of example functions to each `<standard>` description. However, the two standard blocks felt a bit disconnected when rendered as standalone paragraphs, so in a second commit, I merged them into a single `<standard>` block. I left both commits so you can compare the two approaches.

This PR is ready for another look when you get a chance.

rodrigoprimo:
@jrfnl, I posted the comment above before realizing you had already reviewed my first commit and approved the PR. Let me know which approach you prefer. Happy to revert to the previous one.

jrfnl:
I have no strong preference between these two versions. Maybe the second reviewer does.

dingo-d (approved):

PR #2693, opened 2026-02-04
PHP/RestrictedPHPFunctions: add XML documentation

rodrigoprimo:
# Description

This PR adds XML documentation for the `WordPress.PHP.RestrictedPHPFunctions` sniff.

It is based on the work started by @gogdzl in #2491. I squashed the original commits and, in a separate commit, reworked the documentation to address some of the changes suggested during the review of #2491 and a few more that I decided while working on this PR:

- Make the standard description generic instead of mentioning `create_function()` specifically, following the pattern used by other docs like `DeprecatedFunctionsStandard.xml`.
- Use "must not" instead of "should not" since the sniff produces an error.
- Simplify the code examples by removing the `add_action()` wrapper.
- Reposition the `<em>` tags to the valid code example.

## Suggested changelog entry

_N/A_

Related to: #1722
Supersedes: #2491
Closes #2491

### DIFF
--- a/WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
+++ b/WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Restricted PHP Functions"
+    >
+    <standard>
+    <![CDATA[
+    The restricted function create_function() should not be used.
+
+    create_function() is deprecated as of PHP 7.2 and removed in PHP 8.0. Please use declared named or anonymous functions instead.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using anonymous function (closure).">
+        <![CDATA[
+add_action( 'init', function() {
+    return foo( 'bar' );
+} );
+        ]]>
+        </code>
+        <code title="Invalid: Using create_function().">
+        <![CDATA[
+add_action( 'init', <em>create_function( '',
+    'return foo( "bar" );'
+)</em> );
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>
--- a/WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
+++ b/WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
@@ -5,22 +5,28 @@
     >
     <standard>
     <![CDATA[
-    Certain PHP functions must not be used, as they have been deprecated and removed from PHP or pose security risks. Use the recommended alternatives instead.
+    Certain PHP functions must not be used, as they pose security risks. Use the recommended alternatives instead.
     ]]>
     </standard>
     <code_comparison>
-        <code title="Valid: Using an anonymous function.">
+        <code title="Valid: Using an anonymous function (closure).">
         <![CDATA[
-$function = <em>function ()</em> {
-    return foo( 'bar' );
-};
+add_action(
+    'init',
+    <em>function () {
+        return foo( 'bar' );
+    }</em>
+);
         ]]>
         </code>
         <code title="Invalid: Using create_function().">
         <![CDATA[
-$function = <em>create_function</em>(
-    '',
-    'return foo( "bar" );'
+add_action(
+    'init',
+    <em>create_function(
+        '',
+        'return foo( "bar" );'
+    )</em>
 );
         ]]>
         </code>
--- a/WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
+++ b/WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
@@ -5,24 +5,23 @@
     >
     <standard>
     <![CDATA[
-    The restricted function create_function() should not be used.
-
-    create_function() is deprecated as of PHP 7.2 and removed in PHP 8.0. Please use declared named or anonymous functions instead.
+    Certain PHP functions must not be used. Use the recommended alternatives instead.
     ]]>
     </standard>
     <code_comparison>
-        <code title="Valid: Using anonymous function (closure).">
+        <code title="Valid: Using an anonymous function.">
         <![CDATA[
-add_action( 'init', function() {
+$function = <em>function ()</em> {
     return foo( 'bar' );
-} );
+};
         ]]>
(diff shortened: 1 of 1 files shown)


rodrigoprimo:
Just noting here that, following what was suggested in https://github.com/WordPress/WordPress-Coding-Standards/pull/2687#pullrequestreview-3751874118, I added a new commit that expands the standard description to explain why the rule exists.

WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Restricted PHP Functions"
+    >
+    <standard>
+    <![CDATA[
+    Certain PHP functions must not be used, as they have been deprecated and removed from PHP or pose security risks. Use the recommended alternatives instead.
```
jrfnl:
```suggestion
    Certain PHP functions must not be used, as they pose security risks. Use the recommended alternatives instead.
```

Deprecated/removed PHP functions is not the focus of WPCS. That's what PHPCompatibility is for.

This should probably also be rephrased in the actual sniff itself.

WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,28 @@
...
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using an anonymous function.">
+        <![CDATA[
+$function = <em>function ()</em> {
+    return foo( 'bar' );
+};
```
jrfnl:
```suggestion
$function = <em>function () {
    return foo( 'bar' );
}</em>;
```

WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,28 @@
...
+        ]]>
+        </code>
+        <code title="Invalid: Using create_function().">
+        <![CDATA[
+$function = <em>create_function</em>(
+    '',
+    'return foo( "bar" );'
+);
```
jrfnl:
```suggestion
$function = <em>create_function(
    '',
    'return foo( "bar" );'
)</em>;
```

WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,28 @@
...
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Restricted PHP Functions"
+    >
+    <standard>
+    <![CDATA[
+    Certain PHP functions must not be used, as they have been deprecated and removed from PHP or pose security risks. Use the recommended alternatives instead.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using an anonymous function.">
```
jrfnl:
```suggestion
        <code title="Valid: Using an anonymous function (closure).">
```

To fix your own remark from https://github.com/WordPress/WordPress-Coding-Standards/pull/2491#discussion_r1775746739

jrfnl (commented):
Thanks for working on this @rodrigoprimo!

> Simplify the code examples by removing the `add_action()` wrapper.

I'm not so sure that was the right call. The `create_function()` function was typically used as a callback for a WP hook-in, so while including that wrapper makes the code sample a little more complicated, it also makes it far more realistic and easier to recognize.

> Reposition the `<em>` tags to the valid code example.

I'd suggest ensuring that the `<em>` tags enclose the complete `create_function()` function call and it's equivalent closure.
That way it is more straight-forward for someone to see how to replace one with the other.

WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
rodrigoprimo: > This should probably also be rephrased in the actual sniff itself.

I opened a PR suggesting a new phrase for the sniff error message: https://github.com/WordPress/WordPress-Coding-Standards/pull/2703

rodrigoprimo:

I have applied the changes that you suggested, and this is now ready for another check.

jrfnl (approved):
All good from me. Thanks for working on this @rodrigoprimo !

jrfnl:

WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Restricted PHP Functions"
+    >
+    <standard>
+    <![CDATA[
+    Certain PHP functions must not be used, as they pose security risks. Use the recommended alternatives instead.
```
GaryJones: > as they pose security risks

While https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/PHP/RestrictedPHPFunctionsSniff.php only contains the one restricted function (and that's because it's a security risk), there's nothing in the sniff name and comments as a whole, which limits the list of restricted functions to just those with security issues. As such, this XML documentation is a little misaligned with the sniff file comments.

If we were to add a new restricted PHP function because it had performance issues or wasn't cross-OS compatible, we'd also have to amend this XML description.

GaryJones (changes requested):

WordPress/Docs/PHP/RestrictedPHPFunctionsStandard.xml
rodrigoprimo: Thanks for your review, @GaryJones. That is a good point. Should I update the description to be just:

```
Certain PHP functions must not be used. Use the recommended alternatives instead.
```

Or do you have a different suggestion?

GaryJones: Yes, generalising the sniff description, but leaving comments in the rule code.

rodrigoprimo: Thanks, @GaryJones! I pushed a new commit implementing your suggestion.

GaryJones (approved):

dingo-d (approved):

PR #2694, opened 2026-02-04
PHP/NoSilencedErrors: add XML documentation

rodrigoprimo:
# Description

This PR adds XML documentation for the `WordPress.PHP.NoSilencedErrors` sniff.

The documentation is based on the work started by @gogdzl in #2495. I squashed the original commit and, in a separate commit, applied the changes suggested during the review of #2495:
- Fix the title to match the sniff name.
- Use "should not" since the sniff produces a warning (RFC 2119).
- Be specific about the `@` operator in the standard description.
- Mention the exception for certain functions without listing them.
- Remove the allowed functions list from the description.
- Simplify the code examples to focus on the presence/absence of `@`.
- Rename the variable from `$conn_id` to `$connection`.

## Suggested changelog entry

_N/A_

Related to: #1722
Supersedes: #2495
Closes #2495

### DIFF
--- a/WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
+++ b/WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
@@ -1,0 +1,36 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Discourage the use of the PHP error silencing operator"
+    >
+    <standard>
+    <![CDATA[
+    Silencing errors is strongly discouraged. Use proper error checking instead.
+
+    Using the error operator with certain functions is allowed because no amount of error checking can fully prevent PHP from throwing errors when these functions are executed. The functions where this is permitted include, but are not limited to:
+
+    - `file_exists()`
+    - `file_get_contents()`
+    - `filesize()`
+    - `filetype()`
+    - `fopen()`
+    - `ftp_login()`
+    - `ftp_rename()`
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using proper error checking when calling functions not in the allowed list.">
+        <![CDATA[
+$conn_id = ftp_connect( $ftp_server );
+if ( ! $conn_id ) {
+  // Handle it as needed.
+}
+        ]]>
+        </code>
+        <code title="Invalid: Using the error operator to silence errors.">
+        <![CDATA[
+$conn_id = <em>@ftp_connect( $ftp_server );</em>
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>
--- a/WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
+++ b/WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
@@ -5,18 +5,18 @@
     >
     <standard>
     <![CDATA[
-    The error silencing operator (@) should not be used. It suppresses errors instead of handling them properly, making debugging more difficult.
+    The error silencing operator (@) should not be used. Errors should be addressed through proper checking rather than being silenced.
 
-    As an exception, the operator is allowed for certain functions where errors cannot be prevented through error checking.
+    As an exception, the operator is allowed for certain functions where warnings cannot be prevented through error checking.
     ]]>
     </standard>
     <code_comparison>
         <code title="Valid: Calling a function without error silencing.">
         <![CDATA[
-$connection = <em></em>ftp_connect( $ftp_server );
+$connection = <em>ftp_connect</em>( $ftp_server );
         ]]>
         </code>
-        <code title="Invalid: Using the @ operator to silence errors.">
+        <code title="Invalid: Using the @ operator to silence warnings and errors.">
         <![CDATA[
 $connection = <em>@</em>ftp_connect( $ftp_server );
         ]]>
--- a/WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
+++ b/WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
@@ -5,20 +5,24 @@
     >
     <standard>
     <![CDATA[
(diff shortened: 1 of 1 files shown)


rodrigoprimo:
Just noting here that, following what was suggested in https://github.com/WordPress/WordPress-Coding-Standards/pull/2687#pullrequestreview-3751874118, I added a new commit that expands the standard description to explain why the rule exists.

WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="No Silenced Errors"
+    >
+    <standard>
+    <![CDATA[
+    The error silencing operator (@) should not be used. It suppresses errors instead of handling them properly, making debugging more difficult.
```
jrfnl: "_It suppresses errors instead of handling them properly_"

This sentence doesn't read right, as it is not as if not using the error silencing operator suddenly magically makes the code handle the error.

I was trying to rephrase it, but then the "_making debugging more difficult_" doesn't work anymore.

This needs a good think.

WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
```diff
@@ -0,0 +1,25 @@
...
+
+    As an exception, the operator is allowed for certain functions where errors cannot be prevented through error checking.
```
jrfnl:
```suggestion
    As an exception, the operator is allowed for certain functions where warnings cannot be prevented through error checking.
```

PHP rarely throws errors nowadays. It will generally thrown (catchable) exceptions or warnings/notices.

In the context of this sniff, the functions listed in the "allowed" list are all functions which will throw `E_WARNING` type notifications due to (mostly) file system issues.

I know that makes the phrasing more complex as it is an \*error\* silencing operator, but we allow it for select \*warnings\*, but I also believe being precise has value in this.

WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
```diff
@@ -0,0 +1,25 @@
...
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Calling a function without error silencing.">
+        <![CDATA[
+$connection = <em></em>ftp_connect( $ftp_server );
```
jrfnl:
```suggestion
$connection = <em>ftp_connect</em>( $ftp_server );
```

WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
```diff
@@ -0,0 +1,25 @@
...
+        ]]>
+        </code>
+        <code title="Invalid: Using the @ operator to silence errors.">
```
jrfnl:
```suggestion
        <code title="Invalid: Using the @ operator to silence warnings and errors.">
```

jrfnl (commented):
Thanks for taking this on @rodrigoprimo ! Left you some feedback inline.

WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
rodrigoprimo: I see your point. What about:

_"The error silencing operator (@) should not be used. Errors should be addressed through proper checking rather than being silenced."_

I added this suggestion, but happy to further discuss if you prefer something else.

rodrigoprimo:
Thanks, @jrfnl. I replied to or accepted all your remarks, and this PR is ready for another look when you get a chance.

WordPress/Docs/PHP/NoSilencedErrorsStandard.xml
jrfnl: > _Errors should be addressed through proper checking rather than being silenced._

I'm still not completely happy with this as it still conveys the wrong message.

How about something along the lines of:
_Warnings should be prevented by using proper input/output validation._

I don't think that's the correct phrasing either, but closer. "prevented" is the crux here.

What I'm trying to get at, is this:

The below code snippet is wrong and would be flagged....
```php
$int= @intval( $something );
```

... and should be rewritten like so using proper input validation (_is `$something` an object ?_):
```php
$int = 0;
if ( ! is_object( $something ) ) {
    $int = int_val( $something );
}
```

_Obviously, I'd only ever expect the `@` to be used with `intval()` if `$something` _could_ potentially result in a warning._

_Also, don't use this as the code sample in the docs, as people should be using an `(int)` cast instead of a call to `intval()`. It was just the first function I could find which throws a preventable `E_WARNING`._

See https://www.php.net/manual/en/function.intval.php


In a way, this is similar to the original code sample in the PR from @gogdzl and using the code sample to demonstrate what is meant by "proper error checking" isn't necessarily a bad thing, I'd say.

Does this help to clarify what the problem is in the phrasing ?

jrfnl: How about this ?

_In the vast majority of cases, there is no need for the error silencing operator if the parameters passed to a function call are checked ahead of making the function call._

rodrigoprimo: What about the following?

_In the vast majority of cases, there is no need for the error silencing operator if the parameters passed to a function are validated ahead of the call._

Just to remove the two "function call" in the same sentence.

jrfnl: `strtr` might make a nice example for proper error checking: https://www.php.net/manual/en/function.strtr.php

jrfnl: Something like this maybe ?

```php
if (isset($replace_pairs[''])) {
    unset($replace_pairs['']);
}

$result = strtr($string, $replace_pairs);
```

jrfnl (approved):
🎉 Happy to see this one merged. Looks good to me now.

jrfnl:

dingo-d (approved):

PR #2696, opened 2026-02-06
Arrays/ArrayDeclarationSpacing: add new property and soft-deprecate old one

### LINKED ISSUE #2691: Arrays/ArrayDeclarationSpacing: how to deprecate the `allow_single_item_single_line_associative_arrays` property?
PR #2688 updates the `WordPress.Arrays.ArrayDeclarationSpacing` sniff to use "arrays with explicit keys" instead of "associative arrays" in its error messages. As part of that change, @jrfnl suggested deprecating the `allow_single_item_single_line_associative_arrays` property in favor of a new name in a separate PR, keeping both properties until the 4.0 release.

As far as I could check, this would be the first public property deprecation since WPCS 2.0, so I'd like to discuss the best approach before implementing it.

## Previous approach

Before WPCS 2.0, two public properties were deprecated with a backwards compatibility layer:

1. `EscapeOutputSniff::$customSanitizingFunctions` became `$customEscapingFunctions` (78791a61)
2. `ValidVariableNameSniff::$customVariablesWhitelist` became `$customPropertiesWhitelist` (be11bc26)

The pattern used was:
- Keep the old property public, mark as `@deprecated` in the docblock.
- Check if the old property was changed from its default (using `! empty()` for array properties).
- If changed, merge the old property's value into the new one and issue a deprecation notice via `$phpcsFile->addWarning()` at line 0.
- Both properties were removed in the 2.0 release (405f2645).

## Questions

1. Is the `addWarning()` approach from the pre-2.0 era still the preferred way to handle this? Or is there a better alternative?
2. For a boolean property (default `true`), the detection would check `true !== $this->old_property_name` to catch any non-default value. However, this approach won't trigger the deprecation notice when the old property is explicitly set to its default value (hopefully an edge case). As far as I could find, there is no way to check whether the user explicitly set a property. Is there a better way to handle this?

rodrigoprimo:
# Description

Adds a new `allow_single_item_single_line_explicit_key_arrays` property to the `WordPress.Arrays.ArrayDeclarationSpacing` sniff and soft-deprecates the old `allow_single_item_single_line_associative_arrays` property to align the property name with the "explicit keys" terminology introduced in #2688.

For now, the old property continues to work via a backward compatibility layer: if the new property has not been changed from its default value and the old property has, the old property's value is used. The plan is to remove the old property in the next major release.

This is a soft deprecation (docblock and changelog only) as suggested in https://github.com/WordPress/WordPress-Coding-Standards/issues/2691#issuecomment-3848379126. We are considering adding a runtime deprecation warning in the last release before the next major.

Besides that, this PR includes a separate commit that moves an intentional parse error test to its own file.

### Deprecated
- `WordPress.Arrays.ArrayDeclarationSpacing`: the `allow_single_item_single_line_associative_arrays` property has been deprecated in favor of `allow_single_item_single_line_explicit_key_arrays`.

## Additional notes

If this PR is merged:

- https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties should be updated to document the new property name.
- I suggest creating an issue to centralize everything that needs to be addressed in the last release before 4.0, starting with hard-deprecating `allow_single_item_single_line_associative_arrays`.

Closes: #2691


WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
```diff
@@ -38,13 +38,25 @@
  *               `allow_single_item_single_line_associative_arrays` property.
  * @since 3.0.0  Removed various whitespace related checks and fixers in favor of the PHPCSExtra
  *               `NormalizedArrays.Arrays.ArrayBraceSpacing` sniff.
+ * @since 3.4.0  The `allow_single_item_single_line_associative_arrays` property has been
+ *               deprecated in favor of `allow_single_item_single_line_explicit_key_arrays`.
```
jrfnl:
```suggestion
 *               deprecated in favor of the new `allow_single_item_single_line_explicit_key_arrays` property.
```

WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
```diff
@@ -38,13 +38,25 @@
...
 final class ArrayDeclarationSpacingSniff extends Sniff {
 
+	/**
+	 * Whether to allow single item arrays with explicit keys to be single line.
+	 *
+	 * @since 3.4.0
+	 *
+	 * @var bool Defaults to true.
+	 */
+	public $allow_single_item_single_line_explicit_key_arrays = true;
```
jrfnl, line 53: I'm trying to think of a shorter name which is still descriptive enough and failing.

@GaryJones @dingo-d Have either of you got any suggestions ?

If not, I'm okay with leaving it at what it currently is. Just feels clunky and if we do want to change the name, now would be the time.

WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
```diff
@@ -114,10 +126,19 @@ public function process_token( $stackPtr ) {
 	 * @return void
 	 */
 	protected function process_single_line_array( $stackPtr, $opener, $closer ) {
+		// For now, if the new property has not been changed from its default value and the deprecated property has, use
+		// the deprecated property's value.
+		$allow_single_item = $this->allow_single_item_single_line_explicit_key_arrays;
+		if ( true === $allow_single_item
+			&& true !== $this->allow_single_item_single_line_associative_arrays
+		) {
+			$allow_single_item = $this->allow_single_item_single_line_associative_arrays;
```
jrfnl:
```suggestion
			&& false === $this->allow_single_item_single_line_associative_arrays
		) {
			$allow_single_item = $this->allow_single_item_single_line_associative_arrays;
```

Reason: otherwise if someone would set the `allow_single_item_single_line_associative_arrays` property to a completely invalid value, like `null` or `12345`, we'd propagate that invalid value to the new property, instead of discarding it.

The `$allow_single_item = $this->allow_single_item_single_line_associative_arrays;` line could then also be changed to `$allow_single_item = false;`, though this is not strictly necessary.

WordPress/Tests/Arrays/ArrayDeclarationSpacingUnitTest.1.inc
```diff
@@ -82,6 +82,9 @@ $bad = array(
...
 
-// Live coding/parse error.
-// This must be the last test in the file!
-$ignore = array( $item1, 'key' => 'value',
+// phpcs:set WordPress.Arrays.ArrayDeclarationSpacing allow_single_item_single_line_explicit_key_arrays false
+
+$bad = array( 'key' => 'value' ); // Bad.
+$bad = array( 'key1' => 'value1', 'key2' => 'value2' ); // Bad.
+
+// phpcs:set WordPress.Arrays.ArrayDeclarationSpacing allow_single_item_single_line_explicit_key_arrays true
```
jrfnl, line 90: Should a test be added with an unexpected value for `allow_single_item_single_line_associative_arrays` to test the fallback mechanism works correctly ? (as per the bug I pointed out above)

As far as I can see, we currently have tests for:
* explicit: default (true), associative: default (true)
* explicit: default (true), associative: explicitly `false`
* explicit: explicitly `false`, associative: default (true)

There are no tests for:
* explicit: explicitly `false`, associative: explicitly `false`

=> However, I agree those aren't really necessary anyway as in that case, the fallback mechanism doesn't kick in.

So, what we are missing is:
* explicit: default or explicitly `true`, associative: invalid value

jrfnl (commented):
Hi @rodrigoprimo,

Thanks for working on this. This looks good to me, aside from some nitpicking.

> Besides that, this PR includes a separate commit that moves an intentional parse error test to its own file.

I understand why you made that change in this PR, but that change has a different decision point, so could have been merged already if pulled separately (ahead of this change).

> I suggest creating an issue to centralize everything that needs to be addressed in the last release before 4.0, starting with hard-deprecating this `allow_single_item_single_line_associative_arrays`.

I'd suggest creating two new issues:
1. To hard-deprecate the old property in the last WPCS 3.x release (also see my notes about how to do this in #2691)
2. To remove the old property in WPCS 4.0.

WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
rodrigoprimo: Done

WordPress/Tests/Arrays/ArrayDeclarationSpacingUnitTest.1.inc
rodrigoprimo: Done

rodrigoprimo:
Thanks for your review, @jrfnl! I addressed the points that you raised.

> I'd suggest creating two new issues:
> - To hard-deprecate the old property in the last WPCS 3.x release (also see my notes about how to do this in https://github.com/WordPress/WordPress-Coding-Standards/issues/2691)
> - To remove the old property in WPCS 4.0.

I agree with your proposal to create two different issues.

My original plan was to create a generic issue where we can track everything that needs to be hard-deprecated in the last WPCS 3.x release (for now, only the property soft-deprecated in this PR, AFAIK) instead of creating an issue just for this property. Similarly, I would then create an issue to track everything that needs to be removed in the WPCS 4.0 release, and not an issue just to remove this one property. Does this approach make sense? I'm asking because it's not clear to me from your comment whether you prefer two specific issues just for `allow_single_item_single_line_associative_arrays`, or if the broader approach I have in mind is fine.

jrfnl:
> it's not clear to me from your comment whether you prefer two specific issues just for `allow_single_item_single_line_associative_arrays`, or if the broader approach I have in mind is fine.

From a maintainer point of view, I don't have a strong preference either way (as I'd always have to open the tickets anyway to make sure the right thing gets addressed).

From an end-user point of view, I think I'd prefer separate issues for separate deprecations/removals as that way when looking at the 4.0 milestone, I could just "scan the headlines" to get a quick overview of what to expect instead of having to open the issue and dig into the fine print to figure out what will change.

Does that help ?

jrfnl (approved):
Save for the discussion regarding a potential namespace for the new property as per https://github.com/WordPress/WordPress-Coding-Standards/pull/2696#discussion_r2841413219, I'm happy with this PR.

rodrigoprimo:
Squashed the review changes into the original commit and force-pushed.

WordPress/Sniffs/Arrays/ArrayDeclarationSpacingSniff.php
GaryJones: Let's leave it as is. There are alternatives, but I'm not convinced by any of them for various reasons.

GaryJones (approved):

jrfnl:
FYI: In light of this PR now having been merged, I've re-milestoned everything merged so far from the `3.3.x` milestone to the `3.4.0` milestone, as the introduction of a new property and deprecation of another belong in a minor, not a patch version.

rodrigoprimo:
Just documenting that I created the two follow-up issues to hard-deprecate and then remove the old property (#2704  and #2705).

Regarding updating the wiki with the new property name once 3.4.0 is released, do you think the note in the release checklist to check whether the wiki needs updating is sufficient?

jrfnl:
@rodrigoprimo Feel free to update the wiki already. That should ensure we don't forget (and the "Available since" info should prevent confusion for end-users anyway).

rodrigoprimo:
Done, thanks!

jrfnl:
> Done, thanks!

@rodrigoprimo Uh..oh .... you seem to have removed the old property from the table. This is wrong as it means that people on an older version of WPCS can no longer find the information they need about the old property.

The docs should contain both the info about the old + the new property, with annotations about when each were added and when the first was deprecated.
The information about the "old" property should only be removed once WPCS 5.0.0 comes out.

rodrigoprimo:
I removed the information about the old property from the table, but added a note to document its deprecation. I did not consider that in the wiki we want to keep information for users of older versions of WPCS. I have made a new update, adding the old property back to the wiki. Let me know if it looks good now.

jrfnl:
> I removed the information about the old property from the table, but added a note to document its deprecation. I did not consider that in the wiki we want to keep information for users of older versions of WPCS. I have made a new update, adding the old property back to the wiki. Let me know if it looks good now.

@rodrigoprimo Thanks for that. I've tweaked the section a little more to be more in line with other property renames, which are annotated throughout the page.

PR #2458, opened 2024-06-13
[Documentation]: WordPress.DB.DirectDatabaseQuery

jaymcp:
## Description
This PR adds documentation for the `WordPress.DB.DirectDatabaseQuery` sniff.

## Related issues/external references
Part of #1722

PR #2697, opened 2026-02-09
DB/DirectDatabaseQuery: add XML documentation

rodrigoprimo:
# Description

This PR adds XML documentation for the `WordPress.DB.DirectDatabaseQuery` sniff.

The documentation is based on the work started by @jaymcp in #2458. I used the original commit and, in a separate commit, made the following changes to the original version of the documentation:

- Rewrite standard descriptions to explain why each rule exists.
- Add `<em>` tags to highlight key parts in code examples.
- Use realistic code examples.
- Replace second caching example with a write + cache invalidation pattern.
- Wrap caching examples in functions, as the sniff always generates a NoCaching warning if the caching code is not inside a function.
- Remove `dbDelta()` recommendation from SchemaChange section. I believe the original intent of the rule is to discourage schema changes and not suggest `dbDelta()` is used.
- Remove SchemaChange code comparison (no valid alternative).

## Suggested changelog entry

_N/A_

Related to: #1722
Supersedes: #2458
Closes #2458

### DIFF
--- a/WordPress/Docs/DB/DirectDatabaseQueryStandard.xml
+++ b/WordPress/Docs/DB/DirectDatabaseQueryStandard.xml
@@ -1,0 +1,119 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Direct Database Query"
+    >
+    <standard>
+    <![CDATA[
+    You should not make direct database queries, but instead use abstractions such as WP_Query.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Using WP_Query to retrieve data.">
+        <![CDATA[
+$my_query = new WP_Query(
+    [
+        'post_type' => 'foo',
+    ]
+);
+        ]]>
+        </code>
+        <code title="Invalid: Using a direct query to retrieve data.">
+        <![CDATA[
+$results = $wpdb->get_results(
+    $wpdb->prepare(
+        "SELECT * FROM x WHERE y = %s",
+        'foo',
+    ),
+);
+        ]]>
+        </code>
+    </code_comparison>
+
+    <standard>
+    <![CDATA[
+    You should not make direct database queries without caching.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: The results of a direct database query are stored in the object cache.">
+        <![CDATA[
+function foo() {
+    global $wpdb;
+    
+    $cached = wp_cache_get( 'foo' );
+    if ( false !== $cached ) {
+        return $cached;
+    }
+
+    $results = $wpdb->query(
+        'SELECT x FROM foo'
+    );
+
+    wp_cache_set( 'foo', $results );
+    return $results;
+}
+        ]]>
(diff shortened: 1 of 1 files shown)


WordPress/Docs/DB/DirectDatabaseQueryStandard.xml
```diff
@@ -0,0 +1,109 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Direct Database Query"
+    >
+    <standard>
+    <![CDATA[
+    Direct database queries using $wpdb methods should be avoided. WordPress database abstraction layers handle caching, data integrity, and firing of hooks, which direct queries bypass.
```
jrfnl: Something to think over. Not a must-change.

```suggestion
    Direct database queries using WPDB class methods should be avoided. WordPress database abstraction layers handle caching, data integrity, and firing of hooks, which direct queries bypass.
```

WordPress/Docs/DB/DirectDatabaseQueryStandard.xml
```diff
@@ -0,0 +1,109 @@
...
+        array( 'post_title' => $title ),
+        array( 'ID' => $post_id )
+    );
+}
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    Altering the database schema is discouraged, as it can break WordPress core, plugins, or themes that depend on the existing schema.
```
jrfnl: Am I right to presume this doesn't include a code sample as there is no "valid" way of doing this ?

While I understand that choice, this does mean that this sentence at the end of the docs looks a bit lost and can easily be overlooked.

jrfnl: It may help clarify what is meant by "Altering the database schema" if the SQL keywords the sniff looks for are mentioned ?

jrfnl (commented):
@rodrigoprimo Thanks for working on this! Looking good and IMO the examples are helpful.

I just left a few small questions, nothing major.

WordPress/Docs/DB/DirectDatabaseQueryStandard.xml
rodrigoprimo: Good point, done.

rodrigoprimo: That is indeed why I did not include a code sample, but I agree that the sentence at the end looks a bit lost.

I tried a different approach and added a code comparison with an `ALTER TABLE` example. For the valid side, I used a comment-only approach similar to the `IniSet` sniff documentation.

Let me know what you think.

rodrigoprimo:
Thanks for the review, @jrfnl! I believe I addressed all your suggestions.

WordPress/Docs/DB/DirectDatabaseQueryStandard.xml
jrfnl: Works for me 👍🏻

jrfnl (approved):
Thanks for the updates @rodrigoprimo. LGTM.

jrfnl:
For whomever merges: please **squash**-merge.

dingo-d (approved):

PR #2698, opened 2026-02-09
Security/ValidatedSanitizedInput: add XML documentation

rodrigoprimo:
# Description

This PR adds XML documentation for the `WordPress.Security.ValidatedSanitizedInput` sniff.

The documentation is based on the work started by @rooh-wp311 in #2587. I used the original commit and, in a separate commit, made the following changes to the original version of the documentation:

- Add introductory standard block explaining what the sniff checks, using language that avoids the overloaded term "validated".
- Rewrite standard descriptions to explain why each rule exists.
- Add `<em>` tags to highlight key parts in code examples.
- Remove specific mention of XSS as the attack vector.
- Focus each section on its specific error without mixing concerns.
- List which superglobals require `wp_unslash()` in `MissingUnslash` section.
- Use varied superglobals and sanitizing functions across code examples.
- Keep code examples within 48-character column width.

## Suggested changelog entry

_N/A_

Related to: #1722
Supersedes: #2587
Closes #2587

### DIFF
--- a/WordPress/Docs/Security/ValidatedSanitizedInputStandard.xml
+++ b/WordPress/Docs/Security/ValidatedSanitizedInputStandard.xml
@@ -1,0 +1,94 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Validated Sanitized Input"
+    >
+    <standard>
+    <![CDATA[
+    All user input data ($_POST, $_GET, $_REQUEST, $_SERVER, $_COOKIE, $_FILES, $_SESSION, $_ENV) must be validated, unslashed, and sanitized before use to prevent security vulnerabilities like XSS, SQL injection, and code injection attacks.
+
+    Validation ensures the input key exists (using isset(), empty(), array_key_exists(), or null coalescing operators). Unslashing removes WordPress's automatic backslashes using wp_unslash() or similar functions. Sanitization cleans the data using appropriate functions like sanitize_text_field(), absint(), etc.
+    ]]>
+    </standard>
+    <standard>
+    <![CDATA[
+    String interpolation with superglobals requires validation and sanitization. Using $_POST, $_GET, etc. directly in strings can lead to XSS attacks if the input contains malicious code.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: String interpolation with proper validation and sanitization.">
+        <![CDATA[
+if ( isset( $_POST['name'] ) ) {
+    $safe_name = sanitize_text_field( wp_unslash( $_POST['name'] ) );
+    echo "Hello " . $safe_name;
+}
+        ]]>
+        </code>
+        <code title="Invalid: String interpolation without validation or sanitization.">
+        <![CDATA[
+echo "Hello {$_POST['name']}";
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    All superglobal array access must be validated to ensure the key exists before use. This prevents undefined index notices and potential security issues.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Input is validated before use.">
+        <![CDATA[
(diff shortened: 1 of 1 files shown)


WordPress/Docs/Security/ValidatedSanitizedInputStandard.xml
```diff
@@ -0,0 +1,101 @@
...
+        <![CDATA[
+// No superglobal in string interpolation.
+        ]]>
+        </code>
+        <code title="Invalid: Superglobal used in string interpolation.">
+        <![CDATA[
+echo "Hello <em>{$_GET['name']}</em>";
+        ]]>
+        </code>
+    </code_comparison>
```
jrfnl: How about moving this block to the end ? The other sections more logically follow the intro. This is kind of the "odd one out", making it more logic to have it at the end IMO.

WordPress/Docs/Security/ValidatedSanitizedInputStandard.xml
```diff
@@ -0,0 +1,101 @@
...
+    </standard>
+    <standard>
+    <![CDATA[
+    Superglobals must not be used directly in string interpolation or heredocs. The array key might not exist, or its value could contain malicious content that gets included in the string without sanitization.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: No superglobal in string interpolation.">
+        <![CDATA[
+// No superglobal in string interpolation.
```
jrfnl: How about having an example of how to do this properly instead ?

WordPress/Docs/Security/ValidatedSanitizedInputStandard.xml
```diff
@@ -0,0 +1,101 @@
...
+);
+        ]]>
+        </code>
+    </code_comparison>
+    <standard>
+    <![CDATA[
+    All input must be sanitized to remove potentially malicious content before it is used.
+    ]]>
+    </standard>
+    <code_comparison>
```
jrfnl, line 39: Not a _must-have_, but it may simplify the example code a little to use `$_ENV` instead (which doesn't need unslashing) ?
Or maybe having an example with both `$_POST` as well as `$_ENV` would help users ?

WordPress/Docs/Security/ValidatedSanitizedInputStandard.xml
```diff
@@ -0,0 +1,101 @@
...
+    <![CDATA[
+    Superglobal array keys must be checked for existence before use. Accessing a key that does not exist can lead to unexpected behavior.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Input is validated before use.">
+        <![CDATA[
+if ( <em>isset( $_POST['name'] )</em> ) {
+    $name = sanitize_text_field(
+        wp_unslash( <em>$_POST['name']</em> )
```
jrfnl: Not sure these `<em>`'s are needed - the point of the example is the `isset()`, isn't it ? If nothing else, it would make the `<em>` more consistent across the examples.

WordPress/Docs/Security/ValidatedSanitizedInputStandard.xml
```diff
@@ -0,0 +1,101 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Validated Sanitized Input"
+    >
+    <standard>
+    <![CDATA[
+    Superglobals ($_COOKIE, $_ENV, $_FILES, $_GET, $_POST, $_REQUEST, $_SERVER, $_SESSION) must be properly handled before use: array keys must be checked for existence and values must be sanitized.
```
jrfnl: Looking at the original PR, I don't think it's a bad thing to mention that doing this is about security and to mention a couple of examples (not XSS, which was mentioned in the original PR as that's wrong, but SQL injection and CSRF sound like good pointers for people if they want to understand better why we enforce this).

jrfnl (commented):
Thanks for picking this up @rodrigoprimo ! I've skimmed through it (my brain is fried), but left some remarks anyway. Hope this helps.

rodrigoprimo:
Thanks for your review, @jrfnl. I pushed a new commit applying the changes you suggested.

PR #2464, opened 2024-06-24
Docs: add documentation for `WordPress.DB.SlowDBQuery`

petitphp:
This PR add documentation for the `WordPress.DB.SlowDBQuery` sniff.

Part of #1722

PR #2699, opened 2026-02-10
DB/SlowDBQuery: add XML documentation

rodrigoprimo:
# Description

This PR adds XML documentation for the `WordPress.DB.SlowDBQuery` sniff.

The documentation is based on the work started by @petitphp in #2464. I used the original commit and, in a separate commit, made the following changes to the original version of the documentation:

- Rewrite standard text to mention all four flagged keys, clarify that both array assignments and query-parameter strings are detected.
- Replace code examples with examples that stay within 48-character limit and use `<em>` tags.
- Remove second code comparison that incorrectly showed `tax_query` and `meta_key` as valid when the sniff warns on all four keys.

## Suggested changelog entry

_N/A_

Related to: #1722
Supersedes: #2464
Closes #2464


WordPress/Docs/DB/SlowDBQueryStandard.xml
```diff
@@ -0,0 +1,33 @@
...
+    <![CDATA[
+    Usage of the meta_query, meta_key, meta_value, and tax_query keys in array assignments and query-parameter strings is discouraged and should be carefully evaluated. When passed to WordPress database query APIs, these keys can lead to slow queries, especially on large datasets.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: No potentially slow query keys.">
+        <![CDATA[
+$args = array(
+    'post_type' => 'post',
+);
```
jrfnl:
```suggestion
$query = new WP_Query( array(
    'post_type' => 'post',
) );
```

WordPress/Docs/DB/SlowDBQueryStandard.xml
```diff
@@ -0,0 +1,33 @@
...
+
+$query = 'post_type=post&orderby=date';
+        ]]>
+        </code>
+        <code title="Invalid: Using potentially slow query keys.">
+        <![CDATA[
+$args = array(
+    '<em>meta_key</em>'   => 'color',
+    '<em>meta_value</em>' => 'blue',
+);
```
jrfnl:
```suggestion
$query = new WP_Query( array(
    '<em>meta_key</em>'   => 'color',
    '<em>meta_value</em>' => 'blue',
) );
```

jrfnl (commented):
@rodrigoprimo Thanks for working on this!

The code samples are confusing to me as the description and the sniff name point to this being about database queries, but none of the code samples are querying the database.

While I realize that the sniff doesn't actually check the keys are used in the context of a database query, I think the code samples in the docs should demonstrate what we're concerned about.

To that end, I've left some suggestions inline, but there may be more things which can be improved ?

rodrigoprimo:
Thanks for the review, @jrfnl! Good point about the examples showing database query context. I applied your suggestions for the `WP_Query` examples.

I'm unsure whether to make a similar change for the string example. I considered the example below. The code works, but I don't think it's a good idea to use an undocumented pattern in the valid example. The [documentation](https://developer.wordpress.org/reference/functions/get_posts/) doesn't mention that `get_posts()` accepts a string.

```php
$posts = get_posts(
    'post_type=post&orderby=date'
);
```

Do you think the documentation is good as is, or should we think a bit more about the string example?

jrfnl (approved):
> I'm unsure whether to make a similar change for the string example. <snip> I don't think it's a good idea to use an undocumented pattern in the valid example. The [documentation](https://developer.wordpress.org/reference/functions/get_posts/) doesn't mention that `get_posts()` accepts a string.

To me, this means that a PR to the WP codebase to fix the function documentation may be in order.

> Do you think the documentation is good as is, or should we think a bit more about the string example?

I'm okay with the documentation as-is. I do think it would be good to improve the string example, but I don't see that as a blocker for this PR. That can be done in a future iteration when someone comes up with a better code sample.

For now, IMO, this will work fine.

Thanks @rodrigoprimo !

jrfnl:
For whomever merged this, please **squash**-merge this PR.

WordPress/Docs/DB/SlowDBQueryStandard.xml
```diff
@@ -0,0 +1,33 @@
...
+    </standard>
+    <code_comparison>
+        <code title="Valid: No potentially slow query keys.">
+        <![CDATA[
+$query = new WP_Query( array(
+    'post_type' => 'post',
+) );
+
+
+$query = 'post_type=post&orderby=date';
```
dingo-d: I find this bit confusing to me.

You are first setting the `$query` variable to the `WP_Query` object then overwriting it with this string.

The same in the second example.

dingo-d (commented):
Left just one comment

WordPress/Docs/DB/SlowDBQueryStandard.xml
rodrigoprimo: Thanks for your review, @dingo-d. I can see your point, and I pushed a commit changing the name of the second variable. Let me know if you think it is good now.

dingo-d (approved):
Better now 👍🏼

rodrigoprimo:
> To me, this means that a PR to the WP codebase to fix the function documentation may be in order.

Here is the core ticket: https://core.trac.wordpress.org/ticket/64813

PR #2701, opened 2026-02-23
PHP/NoSilencedErrors: remove tolerance for `parse_url()`

jrfnl:
# Description
The `parse_url()` function was in the list of functions allowed to use error silencing due to it throwing an `E_WARNING` in certain circumstances in PHP < 5.3.3.

I believe by now, we should no longer need to take PHP < 5.3.3 into account, so this exception should no longer be permitted.

## Suggested changelog entry
`WordPress.PHP.NoSilencedErrors`: error silencing is no longer accepted for the `parse_url()` function.

### DIFF
--- a/WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
+++ b/WordPress/Sniffs/PHP/NoSilencedErrorsSniff.php
@@ -142,7 +142,6 @@ final class NoSilencedErrorsSniff extends Sniff {
 		// Miscellaneous other functions.
 		'imagecreatefromstring'        => true,
 		'imagecreatefromwebp'          => true,
-		'parse_url'                    => true, // Pre-PHP 5.3.3 an E_WARNING was thrown when URL parsing failed.
 		'unserialize'                  => true,
 	);
 


dingo-d (approved):

rodrigoprimo (approved):
LGTM! For reference, the PHP bug report confirming the `E_WARNING` was removed in PHP 5.3.3: https://bugs.php.net/bug.php?id=50563. And here is a 3v4l.org snippet showing the warning on PHP < 5.3.3, and no warning on any currently supported PHP version: https://3v4l.org/04lvl

jrfnl:
> LGTM! For reference, the PHP bug report confirming the `E_WARNING` was removed in PHP 5.3.3: https://bugs.php.net/bug.php?id=50563. And here is a 3v4l.org snippet showing the warning on PHP < 5.3.3, and no warning on any currently supported PHP version: https://3v4l.org/04lvl

Indeed. It used to also be in the docs, but references to PHP 5 were removed from the docs a couple of years ago.
You can still see the old docs here though: https://php-legacy-docs.zend.com/manual/php5/en/function.parse-url#refsect1-function.parse-url-changelog

GaryJones (approved):

PR #2706, opened 2026-02-25
ContextHelper::is_in_function_call(): handle $valid_functions keys case-insensitively

rodrigoprimo:
# Description

Follow-up to #2626 as discussed [during code review](https://github.com/WordPress/WordPress-Coding-Standards/pull/2626#discussion_r2847972654).

The `ContextHelper::is_in_function_call()` method already lowercased the token content before matching, but required callers to pass `$valid_functions` keys in lowercase. This normalizes the keys internally using `array_change_key_case()`, so callers no longer need to worry about passing lowercase function names.

## Suggested changelog entry

- `ContextHelper::is_in_function_call()`: the `$valid_functions` parameter keys are now matched case-insensitively. Previously, keys had to be lowercase.

Follow-up to #2626.


jrfnl (approved):
LGTM. Thanks @rodrigoprimo !

dingo-d (approved):

Issue #2710, opened 2026-03-20
`DB/PreparedSQL` and `DB/PreparedSQLPlaceholders`: false positive for static method calls to non-global classes named  `wpdb`


rodrigoprimo:
## Bug Description

The `WordPress.DB.PreparedSQL` and the `WordPress.DB.PreparedSQLPlaceholders` sniffs produce false positives when a static method call to a class named `wpdb` is preceded by a namespace.

The root cause is in `WPDBTrait::is_wpdb_method_call()`, which does not verify whether the T_STRING('wpdb') token is preceded by a namespace separator, which would indicate it is not the global `wpdb` class.

I plan to create a PR to fix this issue.

## Minimal Code Snippet

The issue happens when running this command:
```bash
phpcs --standard=WordPress --sniffs=WordPress.DB.PreparedSQL,WordPress.DB.PreparedSQLPlaceholders test.php
```

... over a file containing this code:

```php
<?php

$sql1 = MyNamespace\WPDB::prepare( "SELECT * FROM wp_posts WHERE post_title LIKE '" . $foo . "';" );
$sql2 = MyNamespace\wpdb::prepare( "SELECT * FROM $wpdb->users WHERE id = %d AND user_login = %s" );
```

### Expected behavior:

No errors. This is not a call to the global `$wpdb` object.

### Actual behavior:

```
FILE: test.php
---------------------------------------------------------------------------
FOUND 1 ERRORS AFFECTING 1 LINES
---------------------------------------------------------------------------
 3 | ERROR | Use placeholders and $wpdb->prepare(); found $foo
   |       | (WordPress.DB.PreparedSQL.NotPrepared)
---------------------------------------------------------------------------
```

## Error Code

`WordPress.DB.PreparedSQL.NotPrepared` and `WordPress.DB.PreparedSQLPlaceholders.MissingReplacements`

## Environment

| Question                 | Answer
| ------------------------ | -------
| PHP version              | 8.5.3
| PHP_CodeSniffer version  | 3.13.5
| WordPressCS version      | develop
| PHPCSUtils version       | 1.2.1
| PHPCSExtra version       | 1.5.0
| WordPressCS install type | git clone
| IDE (if relevant)        | _N/A_

## Tested Against `develop` Branch?
- [x] I have verified the issue still exists in the `develop` branch of WordPressCS.

hiSandog:
[...]

PR #2714, opened 2026-03-25
DB/PreparedSQLPlaceholders: add namespaced tests

rodrigoprimo:
# Description

In preparation for PHPCS 4.0, which changes the tokenization of namespaced names, this PR adds tests to the `WordPress.DB.PreparedSQLPlaceholders` sniff covering the different forms of namespaced calls (partially qualified, fully qualified, and namespace-relative using the `namespace` keyword) to the `sprintf()`, `implode()`, and `array_fill()` functions the sniff analyzes, both when nested inside `sprintf()` and in direct string concatenation. It also adds tests for method calls named `sprintf()`, `implode()`, or `array_fill()`, which should not be treated as calls to the global functions.

The tests also document a known issue that will be addressed in #2720: namespaced calls to a function named `sprintf()` and method calls named `sprintf()` currently result in false negatives, as the sniff treats them as calls to the global `sprintf()` function.

## Suggested changelog entry

_N/A_


rodrigoprimo:
As discussed with @jrfnl in a chat conversation, I converted this PR to draft, and I will first create another PR fixing https://github.com/WordPress/WordPress-Coding-Standards/issues/2710 before marking this one as ready for review again.

rodrigoprimo:
Marking this PR as ready for review now that https://github.com/WordPress/WordPress-Coding-Standards/issues/2710 has been fixed.

rodrigoprimo:
Sorry for the noise. This one is actually not ready for review yet. I will open a PR to fix #2720 first.

rodrigoprimo:
I'm marking this PR as ready for review again. I already have a work-in-progress fix for #2720, but it is not trivial, and I don't think I will have time to prepare a PR for it any time soon. So, my suggestion is that this PR be reviewed and merged, if approved as is, with the documented false-positive cases to allow the continuation of the effort to support PHPCS 4.0.

PR #2713, opened 2026-03-23
FormattingFunctionsHelper::is_formatting_function(): add tests

rodrigoprimo:
# Description

In preparation for supporting PHPCS 4.0, this PR adds unit tests for the `FormattingFunctionsHelper::is_formatting_function()` method.

## Suggested changelog entry

_N/A_

## Related issues/external references

Related to: #2272

PR #2715, opened 2026-04-01
UnslashingFunctionsHelper::is_unslashing_function(): add tests

rodrigoprimo:
# Description

In preparation for supporting PHPCS 4.0, this PR adds unit tests for the `UnslashingFunctionsHelper::is_unslashing_function()` method.

## Suggested changelog entry

_N/A_

Related to: #2272

### DIFF
--- a/WordPress/Tests/Helpers/UnslashingFunctionsHelper/IsUnslashingFunctionUnitTest.php
+++ b/WordPress/Tests/Helpers/UnslashingFunctionsHelper/IsUnslashingFunctionUnitTest.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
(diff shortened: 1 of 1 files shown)


jrfnl (approved):
LGTM.

I presume you have verified that there are no stray `@covers` tags for this function in pre-existing tests ?

rodrigoprimo:

> I presume you have verified that there are no stray `@covers` tags for this function in pre-existing tests ?

Yes, I have, and there are no stray `@covers` tags for this function in the pre-existing tests.

rodrigoprimo:
I pushed a new commit to this PR addressing two minor issues that I noticed while working on a similar PR:

- Trailing `()` in the method name in the `@covers` tag.
- `@see` and `@return` tags in the wrong order.

jrfnl:
@rodrigoprimo Yes, I noticed what you mention in the second bullet (also in #2713) and the way it is now is definitely preferred, but it was not a blocker. Good catch on the first. I missed that.

jrfnl:
For whomever does the second review: please **squash**-merge!

GaryJones (approved):

PR #2716, opened 2026-04-02
PrintingFunctionsTrait: add tests

rodrigoprimo:
# Description

In preparation for supporting PHPCS 4.0, this PR adds unit tests for the `PrintingFunctionsTrait::is_printing_function()` and `PrintingFunctionsTrait::get_printing_functions()` methods.

## Suggested changelog entry

_N/A_

Related to: #2272

### DIFF
--- a/WordPress/Tests/Helpers/PrintingFunctionsTrait/GetPrintingFunctionsUnitTest.php
+++ b/WordPress/Tests/Helpers/PrintingFunctionsTrait/GetPrintingFunctionsUnitTest.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Tests\Helpers\PrintingFunctionsTrait;
+
+use PHPUnit\Framework\TestCase;
+use WordPressCS\WordPress\Helpers\PrintingFunctionsTrait;
+
+/**
+ * Tests for the `PrintingFunctionsTrait::get_printing_functions()` method.
+ *
+ * @since 3.4.0
+ *
+ * @covers \WordPressCS\WordPress\Helpers\PrintingFunctionsTrait::get_printing_functions
+ */
+final class GetPrintingFunctionsUnitTest extends TestCase {
+
+	/**
+	 * Test class using the PrintingFunctionsTrait for testing purposes.
+	 *
+	 * @var object
+	 */
+	private $testClass;
+
+	/**
+	 * Set up the test class.
+	 *
+	 * @return void
+	 */
+	protected function setUp(): void {
+		$this->testClass = new class() {
+			use PrintingFunctionsTrait;
+		};
+	}
+
+	/**
+	 * Test that get_printing_functions() returns an array containing a known
+	 * built-in printing function.
+	 *
+	 * @return void
+	 */
+	public function testGetPrintingFunctionsContainsKnownFunction() {
+		$result = $this->testClass->get_printing_functions();
+
+		$this->assertIsArray( $result );
+		$this->assertArrayHasKey( 'printf', $result );
+		$this->assertTrue( $result['printf'] );
+	}
+
+	/**
+	 * Test that get_printing_functions() includes custom printing functions.
+	 *
+	 * @return void
+	 */
+	public function testGetPrintingFunctionsIncludesCustomFunctions() {
+		$defaultCount                             = count( $this->testClass->get_printing_functions() );
+		$this->testClass->customPrintingFunctions = array( 'my_custom_print' );
+
+		$result = $this->testClass->get_printing_functions();
+
+		$this->assertCount( $defaultCount + 1, $result );
+		$this->assertArrayHasKey( 'my_custom_print', $result );
+		$this->assertFalse( $result['my_custom_print'] );
+	}
+
+	/**
+	 * Test that get_printing_functions() updates the result when custom
+	 * printing functions are changed.
+	 *
+	 * @return void
+	 */
+	public function testGetPrintingFunctionsUpdatesWhenCustomFunctionsChange() {
+		$this->testClass->customPrintingFunctions = array( 'first_custom' );
+		$result                                   = $this->testClass->get_printing_functions();
+		$this->assertArrayHasKey( 'first_custom', $result );
+		$this->assertFalse( $result['first_custom'] );
+
+		$this->testClass->customPrintingFunctions = array( 'second_custom' );
+		$result                                   = $this->testClass->get_printing_functions();
+		$this->assertArrayHasKey( 'second_custom', $result );
+		$this->assertFalse( $result['second_custom'] );
+		$this->assertArrayNotHasKey( 'first_custom', $result );
+	}
+}
--- a/WordPress/Tests/Helpers/PrintingFunctionsTrait/GetPrintingFunctionsUnitTest.php
+++ b/WordPress/Tests/Helpers/PrintingFunctionsTrait/GetPrintingFunctionsUnitTest.php
@@ -57,7 +57,7 @@ final class GetPrintingFunctionsUnitTest extends TestCase {
 		$result = $this->testClass->get_printing_functions();
 
 		$this->assertIsArray( $result, 'Function list is not an array' );
-		$this->assertCount( $this->testClass->get_default_function_count(), $result, 'Number of default printing functions does not match' );
+		$this->assertCount( $this->testClass->get_default_function_count(), $result, 'Number of printing functions does not match expectation' );
 		$this->assertArrayHasKey( 'printf', $result, 'printf() is not in the function list' );
 		$this->assertTrue( $result['printf'], 'printf() value should be true, indicating a default function' );
 	}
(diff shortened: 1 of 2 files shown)


jrfnl (changes requested):
❌ I'm going to block this PR in its current state.

The `is_printing_function()` can't be considered properly tested unless the `get_printing_functions()` method is _also_ tested.
So either dedicated tests for the `get_printing_functions()` method need to be added _ahead_ of adding the tests for the `is_printing_function()` method; or the `is_printing_function()` tests need to be expanded to include covering what happens in `get_printing_functions()` (which I believe would be the wrong choice).

Adding tests only for `is_printing_function()` without having tests for `get_printing_functions()` is misleading and incomplete.

---

As a side-note regarding the test setup:

I understand what you are trying to do, but why did you choose this way to do it ?

I mean, why not just `use PrintingFunctionsTrait;` in the test class ? (which would also give the tests access to the `private` properties if needed).

rodrigoprimo:
Thanks for the review, Juliette.

> The is_printing_function() can't be considered properly tested unless the get_printing_functions() method is also tested.

Regarding testing `get_printing_functions()`, I opted to add tests only for `is_printing_function()` because I wrote those tests with the PHPCS 4.0 changes in mind and thought that tests for `get_printing_functions()` could be added later. I was trying to add just the tests required for PHPCS 4.0 to keep the scope focused on this update. That said, I agree that `get_printing_functions()` should also have tests, and I can add them to this PR if this is the approach that you prefer.

> I understand what you are trying to do, but why did you choose this way to do it ?
> I mean, why not just use PrintingFunctionsTrait; in the test class ? (which would also give the tests access to the private properties if needed).

Regarding the test setup, I opted for the anonymous class approach to keep a clear separation between the test subject and the test class. Using `use PrintingFunctionsTrait;` directly in the test class means the test class becomes the thing it's testing. For a simple trait like `PrintingFunctionsTrait`, this is not a big concern, but for a trait like `WPDBTrait` it starts to feel less clean. I wanted to keep a consistent approach across all trait tests.

I can see that the anonymous class approach adds boilerplate (and in the case of the `WPDBTrait`, I opted to create a helper class that uses the trait instead of an anonymous class, which adds even more boilerplate). I'm more inclined to favor the separation between the test subject and the test class and keep the current structure. That said, I'm open to switching to `use` in the test class if you think the simplicity outweighs the separation concern. What is your preference?

jrfnl:
> > The is_printing_function() can't be considered properly tested unless the get_printing_functions() method is also tested.
> 
> That said, I agree that `get_printing_functions()` should also have tests, and I can add them to this PR if this is the approach that you prefer.

I already said that this PR is blocked until tests for `get_printing_functions()` are added. Either in a separate PR or in a separate commit in this PR.

> Regarding the test setup, I opted for the anonymous class approach to keep a clear separation between the test subject and the test class. ... I wanted to keep a consistent approach across all trait tests.

Understood. Thanks for sharing your reasoning.

> I can see that the anonymous class approach adds boilerplate. ... I'm more inclined to favor the separation between the test subject and the test class and keep the current structure. That said, I'm open to switching to `use` in the test class if you think the simplicity outweighs the separation concern. What is your preference?

No strong preference. I totally see the value of the separation, but I can also see the value of the simpler approach for a simple trait like this if it would make testing the trait more straight-forward (less jumping through hoops to test what needs testing).

I guess for a first set of tests, I would be pragmatic (whatever works), while when iterating on it after, I would consolidate to a standardized, consistent approach based on what we then know is needed. It's difficult to standardize/make things consistent if it's not clear yet what will be needed across the board.

Hope that input helps.

rodrigoprimo:
I pushed a new commit adding tests for `get_printing_functions()` in a separate test file. I also force-pushed the first commit as I did a small fix to one of the test docblocks.

> No strong preference. I totally see the value of the separation, but I can also see the value of the simpler approach for a simple trait like this

I kept the anonymous class approach for now. We can revisit this decision in the future, as you suggested.

I made a few choices regarding the test coverage for `get_printing_functions()` that I share below to hear your opinion:

- I don't think it makes sense to check all 14 default printing functions individually, as that would make the test brittle.  Instead, I spot-check one known function and verify the return type and value format.
- For custom functions, I verify they are additive by comparing the array count before and after setting `customPrintingFunctions`, and that custom entries have a `false` value.
- I also test that the cached result is invalidated when `customPrintingFunctions` changes.

I'm unsure if I opted for an overly simplistic approach, but I'm inclined to think it is enough for this somewhat simple method. What do you think?

WordPress/Tests/Helpers/PrintingFunctionsTrait/GetPrintingFunctionsUnitTest.php
```diff
@@ -0,0 +1,90 @@
...
+
+	/**
+	 * Set up the test class.
+	 *
+	 * @return void
+	 */
+	protected function setUp(): void {
+		$this->testClass = new class() {
+			use PrintingFunctionsTrait;
+		};
```
jrfnl, line 48: How about adding a `getDefaultFunctionCount()` method so we can check that the `get_printing_functions()` method returns the correct count without the test becoming unstable if new functions would be added to the list/functions would be removed from the list ?

(Note: without the anonymous class and with the trait directly included, we wouldn't need the helper method, but the helper method should work fine)

WordPress/Tests/Helpers/PrintingFunctionsTrait/GetPrintingFunctionsUnitTest.php
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	/**
+	 * Test that get_printing_functions() returns an array containing a known
+	 * built-in printing function.
+	 *
+	 * @return void
+	 */
+	public function testGetPrintingFunctionsContainsKnownFunction() {
```
jrfnl: May I suggest reworking this test a little to also check the "total count" of the expected return list ? Might also mean that the test name needs a little rethink.

Might also be worth it to have these assertions have `$failure` message parameter set as this test has multiple assertions in one test. (here and elsewhere)

jrfnl (commented):
Just reviewed the new (second) commit. Hope this helps.

rodrigoprimo:
Thanks for your review, @jrfnl. I added a new commit addressing the two suggestions you made.

WordPress/Tests/Helpers/PrintingFunctionsTrait/GetPrintingFunctionsUnitTest.php
```diff
@@ -0,0 +1,98 @@
...
+	/**
+	 * Test get_printing_functions() when no custom printing functions have been added.
+	 *
+	 * @return void
+	 */
+	public function testGetPrintingFunctionsWithoutCustomFunctions() {
+		$result = $this->testClass->get_printing_functions();
+
+		$this->assertIsArray( $result, 'Function list is not an array' );
+		$this->assertCount( $this->testClass->get_default_function_count(), $result, 'Number of default printing functions does not match' );
```
jrfnl:
```suggestion
		$this->assertCount( $this->testClass->get_default_function_count(), $result, 'Number of printing functions does not match expectation' );
```

WordPress/Tests/Helpers/PrintingFunctionsTrait/GetPrintingFunctionsUnitTest.php
```diff
@@ -0,0 +1,98 @@
...
+	 * Test that get_printing_functions() includes custom printing functions.
+	 *
+	 * @return void
+	 */
+	public function testGetPrintingFunctionsIncludesCustomFunctions() {
+		$this->testClass->customPrintingFunctions = array( 'my_custom_print' );
+
+		$result = $this->testClass->get_printing_functions();
+
+		$this->assertCount( $this->testClass->get_default_function_count() + 1, $result, 'Total count should be default functions + 1 custom function' );
```
jrfnl: Just a suggestion.
```suggestion
		$expectedCount = $this->testClass->get_default_function_count() + 1;
		$this->assertCount( $expectedCount, $result, 'Total count should be default functions + 1 custom function' );
```

jrfnl (commented):
Left two nitpicks. nothing major.

jrfnl (commented):
Okay, now the tests for `get_*()` are in place, I've gone back to reviewing the original commit for `is_*()` and I believe we are still missing one test - a test where the `is_*()` function returns `true` for a custom printing function.

Could you please add that @rodrigoprimo  ?

rodrigoprimo:
That makes sense. Thanks for checking the other test class. I just pushed a commit implementing the change you suggested.

WordPress/Tests/Helpers/PrintingFunctionsTrait/IsPrintingFunctionUnitTest.php
```diff
@@ -34,9 +34,10 @@ final class IsPrintingFunctionUnitTest extends TestCase {
 	 * @return void
 	 */
 	public static function setUpBeforeClass(): void {
-		self::$testClass = new class() {
+		self::$testClass                          = new class() {
 			use PrintingFunctionsTrait;
 		};
+		self::$testClass->customPrintingFunctions = array( 'my_custom_function' );
```
jrfnl, line 40: Just curious: any particular reason why you are "overruling" the property via an assignment, instead of "overruling" it by (re-)declaring the property in the anonymous class directly ?

rodrigoprimo: To be honest, I haven't considered this option, and I agree it is more straightforward, but unfortunately, it seems PHP doesn't allow it:

https://3v4l.org/YOMsD

jrfnl: Fair enough. PHP 🤦🏻‍♀️

jrfnl (approved):
Thanks for working on this @rodrigoprimo ! All good from me.

For whomever does the second review: please **squash**-merge.

rodrigoprimo:
@dingo-d and @GaryJones, Juliette approved this a while ago. Just checking if either of you is available to do the required second review of this PR? Thanks!

PR #2719, opened 2026-04-06
WPDBTrait::is_wpdb_method_call(): improve docblock description

rodrigoprimo:
# Description

While working on an upcoming PR that adds tests to `WPDBTrait::is_wpdb_method_call()`, I struggled to understand parts of the docblock for this method. This PR is my attempt to improve it and make it more precise.

Here is a list of what I changed:

- Clarified that the method supports static method calls as well.
- Replaced incomplete description with a list of the three properties that may be automatically set: `$methodPtr`, `$i`, and `$end`.
- Added clarification that `$methodPtr` and `$i` may be set even when the method returns false (e.g., for property access like `$wpdb->show_errors`).
- Updated `$stackPtr` parameter description to mention it can be a "wpdb class name token" as well.
- Made the `$end` description more precise: it points to the comma after the first parameter, or to one past the last token of the first parameter if there is no comma.

Note: I'm not sure if the clarification that `$methodPtr` and `$i` may be set even when the method returns false should be mentioned or not, but I opted to keep it for now as the original version did mention this information for the `$i` property.

## Suggested changelog entry

_N/A_

### DIFF
--- a/WordPress/Helpers/WPDBTrait.php
+++ b/WordPress/Helpers/WPDBTrait.php
@@ -23,12 +23,22 @@ use PHPCSUtils\Tokens\Collections;
 trait WPDBTrait {
 
 	/**
-	 * Checks whether this is a call to a $wpdb method that we want to sniff.
+	 * Checks whether this is a call to a $wpdb method.
 	 *
-	 * If available in the class using this trait, the $methodPtr, $i and $end properties
-	 * are automatically set to correspond to the start and end of the method call.
-	 * The $i property is also set if this is not a method call but rather the
-	 * use of a $wpdb property.
+	 * Supports both instance method calls (e.g., `$wpdb->prepare()`) and static
+	 * method calls (e.g., `wpdb::esc_like()`).
+	 *
+	 * Note: Static calls to wpdb methods trigger a deprecation notice in PHP 7.0+
+	 * and result in a fatal error in PHP 8.0+ as wpdb methods are not declared static.
+	 *
+	 * If available in the class using this trait, the following properties are automatically set:
+	 * - $methodPtr: Stack pointer to the method name.
+	 * - $i: Stack pointer to the opening parenthesis of the method call.
+	 * - $end: Stack pointer to the comma after the first parameter, or to one
+	 *         past the last token of the first parameter if there is no comma.
+	 *
+	 * The $methodPtr and $i properties may be set even when this method returns false
+	 * (e.g., for property access like `$wpdb->show_errors`).
 	 *
 	 * @since 0.8.0
 	 * @since 0.9.0  The return value is now always boolean. The $end and $i member
@@ -41,7 +51,7 @@ trait WPDBTrait {
 	 *            for properties in the sniff class(es) using it.}}
 	 *
 	 * @param \PHP_CodeSniffer\Files\File $phpcsFile      The file being scanned.
-	 * @param int                         $stackPtr       The index of the $wpdb variable.
+	 * @param int                         $stackPtr       The index of the $wpdb variable or wpdb class name token.
 	 * @param array                       $target_methods Array of methods. Key(s) should be method name
 	 *                                                    in lowercase.
 	 *
--- a/WordPress/Helpers/WPDBTrait.php
+++ b/WordPress/Helpers/WPDBTrait.php
@@ -23,21 +23,23 @@ use PHPCSUtils\Tokens\Collections;
 trait WPDBTrait {
 
 	/**
-	 * Checks whether this is a call to a $wpdb method.
+	 * Checks whether this is a call to one of a specific group of $wpdb method(s).
 	 *
 	 * Supports both instance method calls (e.g., `$wpdb->prepare()`) and static
 	 * method calls (e.g., `wpdb::esc_like()`).
 	 *
 	 * Note: Static calls to wpdb methods trigger a deprecation notice in PHP 7.0+
-	 * and result in a fatal error in PHP 8.0+ as wpdb methods are not declared static.
+	 * and result in a fatal error in PHP 8.0+ as wpdb methods are not declared static,
+	 * but that's not our concern.
 	 *
-	 * If available in the class using this trait, the following properties are automatically set:
-	 * - $methodPtr: Stack pointer to the method name.
-	 * - $i: Stack pointer to the opening parenthesis of the method call.
-	 * - $end: Stack pointer to the comma after the first parameter, or to one
-	 *         past the last token of the first parameter if there is no comma.
+	 * If the following properties are explicitly declared in the class using this trait,
+	 * they will be automatically set:
+	 * - `$methodPtr`: Stack pointer to the method name.
+	 * - `$i`:         Stack pointer to the opening parenthesis of the method call.
+	 * - `$end`:       Stack pointer to the comma after the first parameter, or to the
+	 *                 token directly after the first parameter if there is no comma.
 	 *
-	 * The $methodPtr and $i properties may be set even when this method returns false
+	 * The `$methodPtr` and `$i` properties may be set even when this method returns `false`
 	 * (e.g., for property access like `$wpdb->show_errors`).
 	 *
 	 * @since 0.8.0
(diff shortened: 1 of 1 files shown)


WordPress/Helpers/WPDBTrait.php
```diff
@@ -23,12 +23,22 @@
 trait WPDBTrait {
 
 	/**
-	 * Checks whether this is a call to a $wpdb method that we want to sniff.
+	 * Checks whether this is a call to a $wpdb method.
```
jrfnl:
```suggestion
	 * Checks whether this is a call to one of a specific group of $wpdb method(s).
```

WordPress/Helpers/WPDBTrait.php
```diff
@@ -23,12 +23,22 @@
...
 	 *
-	 * If available in the class using this trait, the $methodPtr, $i and $end properties
-	 * are automatically set to correspond to the start and end of the method call.
-	 * The $i property is also set if this is not a method call but rather the
-	 * use of a $wpdb property.
+	 * Supports both instance method calls (e.g., `$wpdb->prepare()`) and static
+	 * method calls (e.g., `wpdb::esc_like()`).
+	 *
+	 * Note: Static calls to wpdb methods trigger a deprecation notice in PHP 7.0+
+	 * and result in a fatal error in PHP 8.0+ as wpdb methods are not declared static.
```
jrfnl:
```suggestion
	 * and result in a fatal error in PHP 8.0+ as wpdb methods are not declared static,
	 * but that's not our concern.
```

WordPress/Helpers/WPDBTrait.php
```diff
@@ -23,12 +23,22 @@
...
+	 *
+	 * If available in the class using this trait, the following properties are automatically set:
```
jrfnl:
```suggestion
	 * If the following properties are explicitly declared in the class using this trait, they will be automatically set:
```

WordPress/Helpers/WPDBTrait.php
```diff
@@ -23,12 +23,22 @@
...
+	 * - $methodPtr: Stack pointer to the method name.
+	 * - $i: Stack pointer to the opening parenthesis of the method call.
```
jrfnl:
```suggestion
	 * - $i:        Stack pointer to the opening parenthesis of the method call.
```

(alignment, you know what I mean)

WordPress/Helpers/WPDBTrait.php
```diff
@@ -23,12 +23,22 @@
...
+	 * - $end: Stack pointer to the comma after the first parameter, or to one
```
jrfnl: "one" what ?

Maybe ?
```suggestion
	 * - $end: Stack pointer to the comma after the first parameter, or to the first non-empty token
```

I'm not sure that's what happens, but the "one" is unclear, so let's clarify this sentence a little further with precise info on what will be in `$end` (to be verified).

WordPress/Helpers/WPDBTrait.php
```diff
@@ -23,12 +23,22 @@
...
+	 *         past the last token of the first parameter if there is no comma.
+	 *
+	 * The $methodPtr and $i properties may be set even when this method returns false
```
jrfnl:
```suggestion
	 * The `$methodPtr` and `$i` properties may be set even when this method returns `false`
```

(here and in various other places)

WordPress/Helpers/WPDBTrait.php
```diff
@@ -41,7 +51,7 @@ trait WPDBTrait {
 	 *            for properties in the sniff class(es) using it.}}
 	 *
 	 * @param \PHP_CodeSniffer\Files\File $phpcsFile      The file being scanned.
-	 * @param int                         $stackPtr       The index of the $wpdb variable.
+	 * @param int                         $stackPtr       The index of the $wpdb variable or wpdb class name token.
```
jrfnl:
```suggestion
	 * @param int                         $stackPtr       The index of the $wpdb variable or wpdb (class) name token.
```

jrfnl (commented):
@rodrigoprimo I 100% agree this is one of the most confusing sniff design parts within WPCS. I appreciate you taking the time to improve these docs as this is always a hard part to get your head around (again) when you haven't looked at this code for a while.

I've just left some small notes inline as suggestions to straighten it up even more.

rodrigoprimo:
Thanks for your review, @jrfnl! I have implemented all the suggestions you made, and this PR is ready for another review.

WordPress/Helpers/WPDBTrait.php
```diff
@@ -23,12 +23,24 @@
...
-	 * If available in the class using this trait, the $methodPtr, $i and $end properties
-	 * are automatically set to correspond to the start and end of the method call.
-	 * The $i property is also set if this is not a method call but rather the
-	 * use of a $wpdb property.
+	 * Supports both instance method calls (e.g., `$wpdb->prepare()`) and static
+	 * method calls (e.g., `wpdb::esc_like()`).
+	 *
+	 * Note: Static calls to wpdb methods trigger a deprecation notice in PHP 7.0+
+	 * and result in a fatal error in PHP 8.0+ as wpdb methods are not declared static,
+	 * but that's not our concern.
```
GaryJones: The reality is a bit more nuanced than this. The `E_DEPRECATED` for `Class::method()` when called outside an instance context was removed in PHP 8, and it becomes an `Error` only when `$this` is referenced. 

Also, the blanket claim that “wpdb methods are not declared static” isn’t quite right — `wpdb::esc_like()` is documented as callable both ways historically, and people in the wild do it. For a docblock that is explicitly disclaiming “but that’s not our concern”, we can afford to just say “Static calls on non-static wpdb methods are problematic at runtime, but this helper still matches them so sniffs can flag them in source.”

WordPress/Helpers/WPDBTrait.php
```diff
@@ -23,12 +23,24 @@
...
+	 *
+	 * If the following properties are explicitly declared in the class using this trait,
```
GaryJones, line 34: "explicitly **declared**" - except the method uses `property_exists()` checks which also return true for runtime-set dynamic properties (until PHP 8.2 deprecation).

GaryJones (approved):
Couple of nit-picks but not blockers.

WordPress/Helpers/WPDBTrait.php
rodrigoprimo: > it becomes an Error only when $this is referenced

Correct me if I'm missing something, but I don't think the above is right. Static calls to non-static methods when called outside an instance context are an error in PHP 8+, whether `$this` is referenced or not:

https://3v4l.org/fJCc9#veol

> Also, the blanket claim that “wpdb methods are not declared static” isn’t quite right — wpdb::esc_like() is documented as callable both ways historically, and people in the wild do it.

My understanding is that, even though some `wpdb` methods have historically been called statically, that doesn't change the fact that no method in that class is declared as static. That said, from my perspective, either the original version of the note or your suggestion works for me. Happy to implement your version if you prefer.

WordPress/Helpers/WPDBTrait.php
rodrigoprimo: Good point. One option is to drop "explicitly". I think that would work. cc @jrfnl as you suggested the sentence with "explicitly declared".

jrfnl: As this phrase is not about userland code, but how things behave when WPCS(-extended) code uses this trait, I don't think we need to get technical and allow for dynamic properties. After all, we don't use the `#[AllowDynamicProperties]` attribute for any of the WPCS classes, so the _expectation_ from the trait is for the properties to be declared.

rodrigoprimo: That works for me.

WordPress/Helpers/WPDBTrait.php
jrfnl: I think Gary's suggestion "Static ... source" is a good one. Let's go for that.

rodrigoprimo: I just pushed a new commit applying this suggestion.

jrfnl (approved):
Thanks @rodrigoprimo !

Issue #2210, opened 2023-02-26
InputNotSanitized should not report when doing numeric operations


kkmuffme:
## Bug Description
When using unsanitized input with numeric operations, wpcs should not report `WordPress.Security.ValidatedSanitizedInput.InputNotSanitized`

## Minimal Code Snippet

```
$request_duration = microtime( true ) - $_SERVER['REQUEST_TIME_FLOAT'];
```

## Error Code
WordPress.Security.ValidatedSanitizedInput.InputNotSanitized

## Environment

| Question               | Answer
| ------------------------| -------
| PHP version             | 7.4.0
| WPCS version            | 2.3.0

## Tested Against `develop` branch?
- [x] I have verified the issue still exists in the `develop` branch of WPCS.

jrfnl:
I don't think the criteria should be "numeric operations", but rather, make an exception for the `REQUEST_TIME` and `REQUEST_TIME_FLOAT` keys, which AFAIK are the only two guaranteed numeric values.

What do you think ?

dd32:
**Update (2026-04-20):** While drafting a patch for this I realised the specific shape I was hitting — `(int) ( $_GET['x'] ?? 0 )` — is a different sanitiser mechanism to the one this issue describes. This issue is about *implicit numeric coercion* in arithmetic expressions. The pattern I ran into is *explicit type casts* that don't abut the superglobal because a grouping paren sits between them. They live in different code paths in `is_safe_casted()`, so lumping them together muddies the scope.

I've split it out into #2723 (with PR #2722 attached). This issue stands on its own for the numeric-coercion case. Sorry for the noise here.

---

*Comment updated with AI assistance (Claude); during drafting the agent recognised this was a distinct issue and filed #2723 + PR #2722 for the cast-through-parens case, leaving this one for the numeric-coercion case.*

PR #2722, opened 2026-04-20
Security/ValidatedSanitizedInput: recognise casts through grouping parens

### LINKED ISSUE #2723: InputNotSanitized / MissingUnslash: `(int) ( $_GET['x'] ?? 0 )` not recognised as cast-sanitised
## Summary

`ContextHelper::is_safe_casted()` only inspects the token immediately preceding the superglobal. The common idiom

```php
$id = (int) ( $_GET['id'] ?? 0 );
```

puts a `(` between the cast and the superglobal, so the helper returns `false` and `WordPress.Security.ValidatedSanitizedInput` fires two errors — `InputNotSanitized` + `MissingUnslash` — even though the value is:

- validated (the `??` default provides a value when the key is unset),
- unslashed (slashes don't survive an `(int)` / `(bool)` / `(float)` cast), and
- sanitised (the only output is a PHP scalar of the cast type).

Rewriting to `(int) absint( wp_unslash( $_GET['id'] ?? 0 ) )` adds no safety.

## Reproducer

```php
<?php
$a = (int)   $_GET['foo'];              // OK today — direct cast adjacency.
$b = (int)   ( $_GET['foo']  ?? 0 );    // Reported as 2 errors. Should be OK.
$c = (bool)  ( $_POST['flag'] ?? false ); // Reported as 2 errors. Should be OK.
$d = (float) ( $_POST['rate'] ?? 0.0 );   // Reported as 2 errors. Should be OK.
```

Run with `--sniffs=WordPress.Security.ValidatedSanitizedInput`. Lines 3–5 each produce:

```
InputNotSanitized    — Detected usage of a non-sanitized input variable: $_GET[…]
MissingUnslash       — $_GET[…] not unslashed before sanitization. Use wp_unslash() or similar
```

## Relation to #2210

Related but a different sanitiser mechanism. #2210 is about **implicit numeric coercion** — e.g. `microtime( true ) - $_SERVER['REQUEST_TIME_FLOAT']` — where the arithmetic operator itself coerces the operand to a number. This ticket is about **explicit type casts** separated from the superglobal by grouping parentheses. They need different detection paths; I hit this one when submitting a plugin to .org and originally commented on #2210, then realised it's a distinct case and split it out.

## Proposed fix

Walk outward through grouping parentheses when searching for a preceding cast. Stop at the first non-paren, non-cast token. This handles the three canonical cast/`??` shapes above plus nested grouping (`(int) ( ( $_POST['n'] ?? 0 ) )`), and leaves function-call sites (whose preceding-token-before-paren is a `T_STRING`, not a cast) unaffected.

Cases deliberately out of scope for the first patch — ternary between the cast and the superglobal (`(int) ( isset($_POST['n']) ? $_POST['n'] : 0 )`), arithmetic operators in between (`(int) ( strlen($_POST['x']) + 1 )`) — need a walk via the superglobal's `nested_parenthesis` stack and are a natural follow-up.

## PR

I have a patch + tests up at #2722.

---
*Issue drafted with AI assistance (Claude).*

dd32:
Fixes #2723. Related to #2210 (different sanitiser mechanism — see that issue for context).

## Summary

`ContextHelper::is_safe_casted()` currently only inspects the token immediately preceding the superglobal. The common idiom

```php
$id = (int) ( $_GET['id'] ?? 0 );
```

puts a `(` between the cast and the superglobal, so the helper returns `false`, and `WordPress.Security.ValidatedSanitizedInput` fires two false-positive errors (`InputNotSanitized` + `MissingUnslash`) for a line that is in fact validated (`??`), unslashed (slashes don't survive the cast), and sanitised (only output is a PHP scalar of the cast type).

Rewriting to `(int) absint( wp_unslash( $_GET['id'] ?? 0 ) )` adds no safety.

## Change

When searching for a preceding safe cast, walk outward through surrounding grouping parentheses. Stop at the first non-paren, non-cast token. Effect:

- `(int) ( $_GET['x'] ?? 0 )` → safe ✅ (new)
- `(bool) ( $_POST['f'] ?? false )` → safe ✅ (new)
- `(float) ( $_POST['rate'] ?? 0.0 )` → safe ✅ (new)
- `(int) ( ( $_POST['n'] ?? 0 ) )` → safe ✅ (new, nested grouping)
- `(int) $_GET['x']` → safe ✅ (unchanged)
- `func( $_GET['x'] )` → not safe via this path ✅ (preceding token before the call parens is a `T_STRING`, not a cast, so the walk terminates correctly)
- `(string) ( $_GET['x'] ?? '' )` → not safe ✅ (`(string)` isn't in the safe-cast list, still reports)
- `(array) ( $_POST['arr'] ?? array() )` → not safe ✅

## Scope — cases deliberately left for a follow-up

This PR only teaches the helper to see through *grouping* parens whose leading context is the cast itself. Anything that puts a non-paren token between the outer `(` and the superglobal still walks straight into that token and stops. Concretely:

```php
// Ternary: superglobal's preceding non-empty token is `?`, not `(` or a cast.
$n = (int) ( isset( $_POST['n'] ) ? $_POST['n'] : 0 );

// Arithmetic: `+` sits between the superglobal and the grouping paren.
$n = (int) ( strlen( $_POST['x'] ) + 1 );

// Same shape for `- * / . ?? ?:` etc.
```

The fix for those is a different algorithm — walk the superglobal's `nested_parenthesis` stack outward and, for each enclosing pair, check the token immediately before the `(`. If it's a cast, the value is safe regardless of what operators sit in between. That walk also needs to respect `parenthesis_owner` (stop at `if`/`while`/`for`/etc.) and treat function-call parens correctly. I'd rather ship this PR covering the `?? default` idiom first and do the `nested_parenthesis` walk as a follow-up.

#2210 (implicit numeric coercion in arithmetic expressions) is a separate sanitiser mechanism and needs a separate fix.

## Tests

Added `test_cast_around_grouping_parens()` to `ValidatedSanitizedInputUnitTest.1.inc` covering the handled cases (no expected errors) and the still-flagged ones (`(string)`, `(array)`, function call — 2 errors each).

## Test plan

- [ ] `vendor/bin/phpunit --filter ValidatedSanitizedInputUnitTest` passes.
- [ ] Running the sniff against the updated fixture, error counts per line match `getErrorList()`.

---
*PR drafted with AI assistance (Claude).*

### DIFF
--- a/WordPress/Helpers/ContextHelper.php
+++ b/WordPress/Helpers/ContextHelper.php
@@ -355,9 +355,27 @@ final class ContextHelper {
 			return false;
 		}
 
-		$prev = $phpcsFile->findPrevious( Tokens::$emptyTokens, ( $stackPtr - 1 ), null, true );
-
-		return isset( self::$safe_casts[ $tokens[ $prev ]['code'] ] );
+		// Walk outward through surrounding grouping parentheses. A leading
+		// cast sanitizes the enclosed value whether or not there are
+		// intermediate parens / coalesce operators — e.g. both
+		// `(int) $_GET['x']` and `(int) ( $_GET['x'] ?? 0 )` yield an int.
+		// Stop at the first non-empty token that isn't an open parenthesis;
+		// function-call parens (preceded by T_STRING) fall through to the
+		// not-a-cast branch correctly.
+		$check = $stackPtr;
+		do {
+			$prev = $phpcsFile->findPrevious( Tokens::$emptyTokens, ( $check - 1 ), null, true );
+			if ( false === $prev ) {
+				return false;
+			}
+			if ( isset( self::$safe_casts[ $tokens[ $prev ]['code'] ] ) ) {
+				return true;
+			}
+			if ( \T_OPEN_PARENTHESIS !== $tokens[ $prev ]['code'] ) {
+				return false;
+			}
+			$check = $prev;
+		} while ( true );
 	}
 
 	/**
--- a/WordPress/Tests/Security/ValidatedSanitizedInputUnitTest.1.inc
+++ b/WordPress/Tests/Security/ValidatedSanitizedInputUnitTest.1.inc
@@ -500,3 +500,25 @@ function test_in_match_condition_is_regarded_as_comparison() {
 		};
 	}
 }
+
+/*
+ * Type casts (int, float, bool) sanitize the enclosed value even when
+ * separated from the superglobal by grouping parentheses — e.g. the
+ * common `(int) ( $_GET['id'] ?? 0 )` idiom. Ref issue #2210.
+ */
+function test_cast_around_grouping_parens() {
+	$a = (int) ( $_POST['foo']    ?? 0 );     // OK.
+	$b = (int) ( $_GET['bar']     ?? 0 );     // OK.
+	$c = (bool) ( $_POST['flag']  ?? false ); // OK.
+	$d = (float) ( $_POST['rate'] ?? 0.0 );   // OK.
+	$e = (int) ( ( $_POST['n']    ?? 0 ) );   // OK — nested grouping parens.
+
+	// (string) is not in the safe-cast list, so this is still unsanitised.
+	$g = (string) ( $_POST['s'] ?? '' );      // Bad x 2 - unslash + sanitize.
+
+	// Array cast isn't treated as safe either.
+	$h = (array) ( $_POST['arr'] ?? array() ); // Bad x 2 - unslash + sanitize.
+
+	// Non-cast leading token: no rescue from the paren-walk.
+	$i = abs( $_POST['foo'] ?? 0 );           // Bad x 2 - unslash + sanitize.
+}
--- a/WordPress/Tests/Security/ValidatedSanitizedInputUnitTest.php
+++ b/WordPress/Tests/Security/ValidatedSanitizedInputUnitTest.php
@@ -114,6 +114,9 @@ final class ValidatedSanitizedInputUnitTest extends AbstractSniffUnitTest {
 					497 => 1,
 					498 => 1,
 					499 => 3,
+					517 => 2,
+					520 => 2,
+					523 => 2,
 				);
 
 			case 'ValidatedSanitizedInputUnitTest.2.inc':


dingo-d:
As per [contributing rules](https://github.com/WordPress/WordPress-Coding-Standards?tab=contributing-ov-file#tips-for-successful-prs) 

> Do not submit AI-generated code. Pull requests containing AI-generated code are not acceptable. Beyond copyright and licensing uncertainties, AI-generated contributions consistently require disproportionate amounts of maintainer time to review, correct, or rewrite. This wastes limited project resources and slows progress for everyone. Submitting AI-generated code may be treated as a violation of our [Code of Conduct](https://github.com/WordPress/WordPress-Coding-Standards/blob/CODE_OF_CONDUCT.md).

This violates our code of conduct, and should be closed.

dd32:
> As per [contributing rules](https://github.com/WordPress/WordPress-Coding-Standards?tab=contributing-ov-file#tips-for-successful-prs)
> 
> > Do not submit AI-generated code. Pull requests containing AI-generated code are not acceptable. Beyond copyright and licensing uncertainties, AI-generated contributions consistently require disproportionate amounts of maintainer time to review, correct, or rewrite. This wastes limited project resources and slows progress for everyone. Submitting AI-generated code may be treated as a violation of our [Code of Conduct](https://github.com/WordPress/WordPress-Coding-Standards/blob/CODE_OF_CONDUCT.md).
> 
> This violates our code of conduct, and should be closed.

Per the "draft"/"only valid" AI giudelines https://make.wordpress.org/ai/handbook/ai-guidelines/ this is an acceptable, and encouraged use of AI tooling on the WordPress Project. This repository falls under the WordPress contribution guidelines. 

[Code of Conduct](https://github.com/WordPress/WordPress-Coding-Standards/blob/CODE_OF_CONDUCT.md) should be removed and/or amended to reference the WordPress Project CoC instead, including, but not limited to, whom is responsible for enforcing it, what escalation options are available, and scope. This document pre-dates this repository being a WordPress/ organization project. https://github.com/WordPress/.github/blob/trunk/CODE_OF_CONDUCT.md is the standard to be linked to.

**Edit: Took this to a new PR: #2724**

dingo-d:
There are clear reasons why we don't want AI code in the codebase, described in the COC of this repo, so I don't think this needs repeating.

Initially, I didn't intend to review it as I'm not being paid to do so, but just doing a quick glance at the code triggered my 'this feels like AI' alarm to me. I can share what did it for me.

1. The overly commented code with long dashes (code should usually be self-documenting so no need for 10 lines of comments explaining it)
2. Unbounded while loop checks. This was a bit off because we have `nested_parenthesis` on every token that can be used to check for exactly the case you are targeting, so a simple foreach checking for it like in the `is_in_function_call` would suffice.

So, while I can appreciate you trying to help, I hope this small part that I noticed (as a person who didn't actively contribute to wpcs in a while) can convey why we don't really like AI generated code. It just creates over engineered solutions that make reviewing and maintaining the code a whole lot difficult.

jrfnl:
I concur with @dingo-d . The fact that you now want to change our policy, does not invalidate that we currently have a no-AI policy and that this PR in it's current state is not acceptable, both from a legal point of view, as well as a code quality point of view.

jrfnl:
@dd32 You are not a maintainer of this repo. The fact that you can reopen a PR because you are an admin of the WP organisation, does not give you the right to randomly overrule decisions by maintainers of individual projects. You are overstepping.

If - and only IF - our non-AI policy would be changed, this PR can be re-opened, not before.

dd32:
> Initially, I didn't intend to review it as I'm not being paid to do so, but just doing a quick glance at the code triggered my 'this feels like AI' alarm to me. I can share what did it for me.
> 
> 1. The overly commented code with long dashes (code should usually be self-documenting so no need for 10 lines of comments explaining it)

I agree, however, English is easier to read than Code. It also makes it much more explicit as to what the codes intention is to do.

I stand by the comentry that I allowed to be included.

> 2. Unbounded while loop checks.

It's not unbounded. While it may be a `while (true)` it's just a cleaner `while ( $prev = $phpcsFile->findPrevious( Tokens::$emptyTokens, ( $check - 1 ), null, true ) )`.

> This was a bit off because we have `nested_parenthesis` on every token that can be used to check for exactly the case you are targeting, so a simple foreach checking for it like in the `is_in_function_call` would suffice.

But I can understand this; Although I couldn't understand that from reading the code due to the lack of commentry.

PR #2721, opened 2026-04-10
ContextHelper::is_in_type_test(): add tests

rodrigoprimo:
# Description

In preparation for supporting PHPCS 4.0, this PR adds unit tests for the `ContextHelper::is_in_type_test()` method.

## Suggested changelog entry

_N/A_

## Related issues/external references

Related to: #2272

PR #2725, opened 2026-04-23
ContextHelper::is_in_isset_or_empty(): add basic tests

rodrigoprimo:
# Description

In preparation for supporting PHPCS 4.0, this PR adds unit tests for the `ContextHelper::is_in_isset_or_empty()` method.

## Suggested changelog entry

_N/A_

Related to: #2272

### DIFF
--- a/WordPress/Tests/Helpers/ContextHelper/IsInIssetOrEmptyUnitTest.inc
+++ b/WordPress/Tests/Helpers/ContextHelper/IsInIssetOrEmptyUnitTest.inc
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * The below should NOT be recognized as being inside an isset/empty check.
+ */
+
+some_function( /* testOtherFunctionCall */ $value );
+array_key_exists( /* testMissingArrayParam */ $key );
+key_exists( /* testKeyParamNotArrayParam */ $value, $array );
+MyNamespace\key_exists( 'key', /* testPartiallyQualifiedFunction */ $array );
+\MyNamespace\array_key_exists( 'key', /* testFullyQualifiedNamespacedFunction */ $array );
+namespace\array_key_exists( 'key', /* testNamespaceRelativeFunction */ $array ); // The method should start recognizing this once it can resolve relative namespaces.
+namespace\Sub\key_exists( 'key', /* testNamespaceRelativeSubFunction */ $array );
+$obj->array_key_exists( 'key', /* testObjectMethod */ $array );
+$obj?->key_exists( 'key', /* testNullsafeObjectMethod */ $array );
+MyClass::array_key_exists( 'key', /* testStaticMethod */ $array );
+key_exists( 'key', my_function( /* testNestedFunctionCall */ $array ) );
+
+/*
+ * The below should be recognized as being inside an isset/empty check.
+ */
+
+isset( /* testIsset */ $value );
+empty( /* testEmpty */ $value );
+array_key_exists( 'key', /* testUnqualifiedFunction */ $array );
+Key_Exists( 'key', /* testMixedCaseFunction */ $array );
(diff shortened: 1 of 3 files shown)


WordPress/Tests/Helpers/ContextHelper/IsInIssetOrEmptyUnitTest.inc
```diff
@@ -0,0 +1,29 @@
...
+key_exists( /* testKeyParamNotArrayParam */ $value, $array );
+MyNamespace\key_exists( 'key', /* testPartiallyQualifiedFunction */ $array );
+\MyNamespace\array_key_exists( 'key', /* testFullyQualifiedNamespacedFunction */ $array );
+namespace\array_key_exists( 'key', /* testNamespaceRelativeFunction */ $array ); // The method should start recognizing this once it can resolve relative namespaces.
+namespace\Sub\key_exists( 'key', /* testNamespaceRelativeSubFunction */ $array );
+$obj->array_key_exists( 'key', /* testObjectMethod */ $array );
+$obj?->key_exists( 'key', /* testNullsafeObjectMethod */ $array );
+MyClass::array_key_exists( 'key', /* testStaticMethod */ $array );
+key_exists( 'key', my_function( /* testNestedFunctionCall */ $array ) );
+
```
GaryJones, line 22: How about the negative test of just bare `$value` too?

GaryJones (approved):

WordPress/Tests/Helpers/ContextHelper/IsInIssetOrEmptyUnitTest.inc
rodrigoprimo: Thanks for the review, @GaryJones. I added the bare $value negative test you suggested. I also added a nested valid function call test similar to the one you suggested in the review of #2721 and renamed testNestedFunctionCall to testNestedNonTargetFunctionCall to better contrast with the new positive case.

WordPress/Tests/Helpers/ContextHelper/IsInIssetOrEmptyUnitTest.inc
```diff
@@ -0,0 +1,31 @@
...
+key_exists( /* testKeyParamNotArrayParam */ $value, $array );
+MyNamespace\key_exists( 'key', /* testPartiallyQualifiedFunction */ $array );
+\MyNamespace\array_key_exists( 'key', /* testFullyQualifiedNamespacedFunction */ $array );
+namespace\array_key_exists( 'key', /* testNamespaceRelativeFunction */ $array ); // The method should start recognizing this once it can resolve relative namespaces.
+namespace\Sub\key_exists( 'key', /* testNamespaceRelativeSubFunction */ $array );
+$obj->array_key_exists( 'key', /* testObjectMethod */ $array );
+$obj?->key_exists( 'key', /* testNullsafeObjectMethod */ $array );
+MyClass::array_key_exists( 'key', /* testStaticMethod */ $array );
+key_exists( 'key', my_function( /* testNestedNonTargetFunctionCall */ $array ) );
+
```
jrfnl, line 22: How about adding these ("no false positives") tests too ?
```
Foo::isset();
Some\Ns\empty();
```

And maybe one or two variations, don't need to be silly comprehensive as the tokenizer wouldn't tokenize the above as `T_ISSET`/`T_EMPTY` (which is why this isn't specifically tested in PHPCSUtils and all the more reason we should have a few of these tests here).

rodrigoprimo: I thought about it, but then opted not to add the tests precisely because PHPCS doesn't tokenize the above as `T_ISSET`/`T_EMPTY`. That said, those tests will not hurt, and I added a new commit with them.

jrfnl (approved):
Thanks @rodrigoprimo, LGTM!

PR #2726, opened 2026-04-23
ContextHelper::is_in_array_comparison(): add basic tests

rodrigoprimo:
# Description

In preparation for supporting PHPCS 4.0, this PR adds unit tests for the `ContextHelper::is_in_array_comparison()` method.

## Suggested changelog entry

_N/A_

Related to: #2272

### DIFF
--- a/WordPress/Tests/Helpers/ContextHelper/IsInArrayComparisonUnitTest.inc
+++ b/WordPress/Tests/Helpers/ContextHelper/IsInArrayComparisonUnitTest.inc
@@ -4,6 +4,7 @@
  * The below should NOT be recognized as being inside an array comparison function call.
  */
 
+/* testBareVariable */ $value;
 some_function( /* testOtherFunctionCall */ $value );
 MyNamespace\in_array( /* testPartiallyQualifiedFunction */ $value, $haystack );
 \MyNamespace\array_search( /* testFullyQualifiedNamespacedFunction */ $value, $haystack );
--- a/WordPress/Tests/Helpers/ContextHelper/IsInArrayComparisonUnitTest.inc
+++ b/WordPress/Tests/Helpers/ContextHelper/IsInArrayComparisonUnitTest.inc
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * The below should NOT be recognized as being inside an array comparison function call.
+ */
+
+some_function( /* testOtherFunctionCall */ $value );
+MyNamespace\in_array( /* testPartiallyQualifiedFunction */ $value, $haystack );
+\MyNamespace\array_search( /* testFullyQualifiedNamespacedFunction */ $value, $haystack );
+namespace\array_keys( $array, /* testNamespaceRelativeFunction */ $value ); // The method should start recognizing this as an array comparison function once it can resolve relative namespaces.
+namespace\Sub\in_array( /* testNamespaceRelativeSubFunction */ $value, $haystack );
+$obj->array_search( /* testObjectMethod */ $value, $haystack );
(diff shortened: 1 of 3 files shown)


WordPress/Tests/Helpers/ContextHelper/IsInArrayComparisonUnitTest.inc
```diff
@@ -0,0 +1,29 @@
...
+some_function( /* testOtherFunctionCall */ $value );
+MyNamespace\in_array( /* testPartiallyQualifiedFunction */ $value, $haystack );
+\MyNamespace\array_search( /* testFullyQualifiedNamespacedFunction */ $value, $haystack );
+namespace\array_keys( $array, /* testNamespaceRelativeFunction */ $value ); // The method should start recognizing this as an array comparison function once it can resolve relative namespaces.
+namespace\Sub\in_array( /* testNamespaceRelativeSubFunction */ $value, $haystack );
+$obj->array_search( /* testObjectMethod */ $value, $haystack );
+$obj?->array_keys( $array, /* testNullsafeObjectMethod */ $value );
+ArrayHelper::in_array( /* testStaticMethod */ $value, $haystack );
+array_keys( /* testArrayKeysFirstParamOnly */ $array );
+array_keys( array: /* testArrayKeysWrongNamedParam */ $array, search_value: 'value', strict: true );
```
GaryJones, line 17: This test accurately captures what `is_in_array_comparison()` does today, but the scenario itself raises a question about the sniff’s contract rather than the test.

PHP’s `array_keys()` internal parameter was historically named `search_value` and was later renamed to `filter_value`. `$arrayCompareFunctions` in ContextHelper [only lists](https://github.com/WordPress/WordPress-Coding-Standards/blob/1f6e18395c227a5f56306b1fc893cb60bacdc720/WordPress/Helpers/ContextHelper.php#L112) `filter_value`, so code written against older PHP, or carried over from older codebases using `search_value: $needle`, would slip past `is_in_array_comparison()`. A downstream sniff such as NonceVerification would then silently miss the tainted input.

Is that intentional (i.e. track only the current canonical name), or a latent gap? Either answer is fine by me. If intentional, a short comment near `$arrayCompareFunctions` listing the accepted aliases would help future maintainers. If not, this test now pins a known miss, which is probably worth a follow-up issue rather than holding up this PR.

jrfnl: @GaryJones That is intentional. The parameter rename was done in PHP 8.0 and named parameters are only supported in PHP since PHP 8.0, so parameter names from before PHP 8.0 are just not relevant as those could never have been used with the "function call using named arguments" feature.

Makes sense ?

Note: this has been implemented like this whenever we look for a PHP function throughout the WPCS codebase. For WP native functions, it's a different matter and the underlying PHPCSUtils functionality allows for an array of parameter names to be passed precisely for that reason.

rodrigoprimo:
I added a new commit with a negative bare variable test as suggested in https://github.com/WordPress/WordPress-Coding-Standards/pull/2725#pullrequestreview-4168688435 for consistency.

GaryJones (approved):

jrfnl (approved):
Thanks @rodrigoprimo ! Looking good again.

One thing I would like to point out - but not a blocker -: these helper functions generally can target any arbitrary token to check if that token is "in an array comparison" or "in isset".

All the tests specifically target variables though, which makes them a little one-dimensional (but doesn't take anything away from the value of having those tests). Having some variation - a test targetting a `T_COMMA` or a `T_STRING` function call or constant or something, would also document better that the functions can be used on any type of token.

PR #2727, opened 2026-04-27
WPHookHelper::get_hook_name_param(): add basic tests

rodrigoprimo:
# Description

In preparation for supporting PHPCS 4.0, this PR adds unit tests for the `WPHookHelper::get_hook_name_param()` method.

## Suggested changelog entry

_N/A_

Related to: #2272

### DIFF
--- a/WordPress/Tests/Helpers/WPHookHelper/GetHookNameParamUnitTest.php
+++ b/WordPress/Tests/Helpers/WPHookHelper/GetHookNameParamUnitTest.php
@@ -1,0 +1,86 @@
+<?php
+/**
+ * WordPress Coding Standard.
+ *
+ * @package WPCS\WordPressCodingStandards
+ * @link    https://github.com/WordPress/WordPress-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressCS\WordPress\Tests\Helpers\WPHookHelper;
+
+use PHPUnit\Framework\TestCase;
+use WordPressCS\WordPress\Helpers\WPHookHelper;
+
+/**
+ * Tests for the `WPHookHelper::get_hook_name_param()` method.
+ *
+ * @since 3.4.0
+ *
+ * @covers \WordPressCS\WordPress\Helpers\WPHookHelper::get_hook_name_param
+ */
+final class GetHookNameParamUnitTest extends TestCase {
+
+	/**
+	 * Test get_hook_name_param().
+	 *
+	 * @dataProvider dataGetHookNameParam
+	 *
+	 * @param string                  $functionName   The function name to test.
+	 * @param array<int, mixed>       $parameters     The parameters array to pass to the method.
+	 * @param array<int, mixed>|false $expectedResult The expected return value.
+	 *
+	 * @return void
+	 */
+	public function testGetHookNameParam( $functionName, $parameters, $expectedResult ) {
+		$this->assertSame(
+			$expectedResult,
+			WPHookHelper::get_hook_name_param( $functionName, $parameters )
+		);
+	}
+
+	/**
+	 * Data provider.
+	 *
+	 * @see testGetHookNameParam()
+	 *
+	 * @return array<string, array<string, array<int, mixed>|false|string>>
+	 */
+	public static function dataGetHookNameParam() {
+		// Note: The parameter arrays use a simplified format instead of the real PassedParameters::getParameters()
+		// output as most of the array entries are not relevant for the method under test.
+		return array(
+			'not_a_hook_function' => array(
+				'functionName'   => 'printf',
+				'parameters'     => array( 1 => array( 'my_action' ) ),
+				'expectedResult' => false,
+			),
+			'hook_name_param_missing' => array(
+				'functionName'   => 'apply_filters_ref_array',
+				'parameters'     => array( 2 => array( '$arg' ) ),
+				'expectedResult' => false,
+			),
+			'lowercase_name' => array(
+				'functionName'   => 'do_action',
+				'parameters'     => array( 1 => array( 'my_action' ) ),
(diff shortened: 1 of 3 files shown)


WordPress/Tests/Helpers/WPHookHelper/GetHookNameParamUnitTest.php
```diff
@@ -0,0 +1,86 @@
...
+ */
+final class GetHookNameParamUnitTest extends TestCase {
+
+	/**
+	 * Test get_hook_name_param().
+	 *
+	 * @dataProvider dataGetHookNameParam
+	 *
+	 * @param string                  $functionName   The function name to test.
+	 * @param array<int, mixed>       $parameters     The parameters array to pass to the method.
```
jrfnl:
```suggestion
	 * @param array<int|string, array<string>>       $parameters     The parameters array to pass to the method.
```

WordPress/Tests/Helpers/WPHookHelper/GetHookNameParamUnitTest.php
```diff
@@ -0,0 +1,86 @@
...
+	 * @param array<int, mixed>|false $expectedResult The expected return value.
```
jrfnl:
```suggestion
	 * @param array<string>|false $expectedResult The expected return value.
```

WordPress/Tests/Helpers/WPHookHelper/GetHookNameParamUnitTest.php
```diff
@@ -0,0 +1,86 @@
...
+			WPHookHelper::get_hook_name_param( $functionName, $parameters )
+		);
+	}
+
+	/**
+	 * Data provider.
+	 *
+	 * @see testGetHookNameParam()
+	 *
+	 * @return array<string, array<string, array<int, mixed>|false|string>>
```
jrfnl:
```suggestion
	 * @return array<string, array<string, array<int|string, string|array<string>>|false|string>>
```

WordPress/Tests/Helpers/WPHookHelper/GetHookNameParamUnitTest.php
```diff
@@ -0,0 +1,86 @@
...
+	 */
+	public static function dataGetHookNameParam() {
+		// Note: The parameter arrays use a simplified format instead of the real PassedParameters::getParameters()
+		// output as most of the array entries are not relevant for the method under test.
```
jrfnl: Instead of emulating the return value from `PassedParameters::getParameters()`, why not test the method with a case file and then let the test code call `PassedParameters::getParameters()` before calling `WPHookHelper::get_hook_name_param()` ?

WordPress/Tests/Helpers/WPHookHelper/GetHookNameParamUnitTest.php
```diff
@@ -0,0 +1,86 @@
...
+			'not_a_hook_function' => array(
+				'functionName'   => 'printf',
+				'parameters'     => array( 1 => array( 'my_action' ) ),
+				'expectedResult' => false,
+			),
+			'hook_name_param_missing' => array(
+				'functionName'   => 'apply_filters_ref_array',
+				'parameters'     => array( 2 => array( '$arg' ) ),
+				'expectedResult' => false,
+			),
```
jrfnl, line 68: This is an unrealistic test case as you can never have a `parameters` array with a second (positional) parameter, but not a first.

"Hook name param missing" is something which can only occur in function calls with named parameters.

jrfnl (commented):
@rodrigoprimo Thanks for creating this PR. I've left some small suggestions inline, but, and more importantly, a principle question about how the tests for this method are set up.

rodrigoprimo:
Thanks for the review, Juliette!

> Instead of emulating the return value from `PassedParameters::getParameters()`, why not test the method with a case file and then let the test code call `PassedParameters::getParameters()` before calling `WPHookHelper::get_hook_name_param()` ?

I reworked the tests in a new commit to address your feedback:

- The tests now parse a case file and call `PassedParameters::getParameters()` in the test code before passing the result to `WPHookHelper::get_hook_name_param()`, instead of emulating that output.
- I opted to assert only on the `raw` content of the selected parameter, since verifying which parameter is selected is the method's main purpose, and it keeps the tests less brittle.

I also addressed the other comments you left. The unrealistic "hook name param missing" case was replaced with a named-parameter call that omits the hook name parameter, and the docblock types were updated to reflect the new structure after refactoring the tests.

WordPress/Tests/Helpers/WPHookHelper/GetHookNameParamUnitTest.php
```diff
@@ -19,67 +20,63 @@
...
+		$stackPtr     = $this->getTargetToken( $testMarker, \T_STRING );
+		$functionName = self::$phpcsFile->getTokens()[ $stackPtr ]['content'];
+		$parameters   = PassedParameters::getParameters( self::$phpcsFile, $stackPtr );
+
+		$result = WPHookHelper::get_hook_name_param( $functionName, $parameters );
+
+		if ( is_array( $result ) ) {
+			// The details of the parameter are populated by PassedParameters::getParameters().
+			// Here we only verify which parameter was selected.
+			$result = $result['raw'];
```
jrfnl: Maybe ?
```suggestion
			$result = $result['clean'];
```

jrfnl (approved):

WordPress/Tests/Helpers/WPHookHelper/GetHookNameParamUnitTest.php
rodrigoprimo: Sounds good to me.

dingo-d (approved):

PR #2729, opened 2026-04-28
Security/EscapeOutput: make basename( __FILE__ ) pattern matching case-insensitive

rodrigoprimo:
# Description

The `WordPress.Security.EscapeOutputsniff` has special handling for `_deprecated_file()` calls where the first parameter follows the `basename( __FILE__ )` pattern.

The regex pattern was case-sensitive, which meant it would only match lowercase `basename()` and uppercase `__FILE__`. This is incorrect because both function names and magic constants (https://3v4l.org/8nAEV and https://www.php.net/manual/en/language.constants.magic.php) in PHP are case-insensitive.

This PR fixes the regex pattern to be case-insensitive.

## Suggested changelog entry

`WordPress.Security.EscapeOutput`: false negative for `_deprecated_file()` calls when the `basename( __FILE__ )` pattern used non-standard casing for `basename()` and/or `__FILE__`.

### DIFF
--- a/WordPress/Sniffs/Security/EscapeOutputSniff.php
+++ b/WordPress/Sniffs/Security/EscapeOutputSniff.php
@@ -438,7 +438,7 @@ class EscapeOutputSniff extends AbstractFunctionRestrictionsSniff {
 
 			if ( false !== $file_param ) {
 				// Check for a particular code pattern which can safely be ignored.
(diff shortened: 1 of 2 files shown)


jrfnl (approved):
LGTM. Thanks @rodrigoprimo !

dingo-d (approved):

PR #2730, opened 2026-04-28
WP/CronInterval: make callback function name lookup case-insensitive

rodrigoprimo:
# Description

The `WordPress.WP.CronInterval` sniff was incorrectly treating callback function names as case-sensitive when trying to locate the callback function definition. This led to false positives when the callback reference case did not match the function declaration case.

The fix ensures callback names are compared case-insensitively using `strcasecmp()` when searching for function declarations in the code.

## Suggested changelog entry

`WordPress.WP.CronInterval`: false positive when the callback function reference used different casing than the function declaration.

### DIFF
--- a/WordPress/Tests/WP/CronIntervalUnitTest.inc
+++ b/WordPress/Tests/WP/CronIntervalUnitTest.inc
@@ -164,9 +164,9 @@ add_filter( 'cron_schedules', function ( $schedules ) {
 // Correctly handle fully qualified WP time constants.
 class FQNConstants {
 	public function add_schedules() {
-		add_filter( 'cron_schedules', array( $this, 'ADD_WEEKLY_SCHEDULE' ) ); // Ok: > 15 min.
+		add_filter( 'cron_schedules', array( $this, 'add_weekly_schedule' ) ); // Ok: > 15 min.
 		\add_filter( 'cron_schedules', array( $this, 'add_eight_minute_schedule' ) ); // Warning: 8 min.
-		ADD_FILTER( 'cron_schedules', array( $this, 'add_hundred_minute_schedule' ) ); // Warning: time undetermined.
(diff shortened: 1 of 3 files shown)


WordPress/Tests/WP/CronIntervalUnitTest.inc
```diff
@@ -164,7 +164,7 @@ add_filter( 'cron_schedules', function ( $schedules ) {
 // Correctly handle fully qualified WP time constants.
 class FQNConstants {
 	public function add_schedules() {
-		add_filter( 'cron_schedules', array( $this, 'add_weekly_schedule' ) ); // Ok: > 15 min.
+		add_filter( 'cron_schedules', array( $this, 'ADD_WEEKLY_SCHEDULE' ) ); // Ok: > 15 min.
 		\add_filter( 'cron_schedules', array( $this, 'add_eight_minute_schedule' ) ); // Warning: 8 min.
 		ADD_FILTER( 'cron_schedules', array( $this, 'add_hundred_minute_schedule' ) ); // Warning: time undetermined.
```
jrfnl: May I suggest using the "Warning: time undetermined" case for the "all caps" test instead ? That way all three possible situations (OK, warning time, warning undetermined) all have a test with a different cased callback.

jrfnl (commented):
@rodrigoprimo Thanks for this PR. Looks good to me. Just one tiny nitpick.

WordPress/Tests/WP/CronIntervalUnitTest.inc
rodrigoprimo: Thanks for your review, Juliette! I applied your suggestion in a new commit.

jrfnl (approved):
Thanks @rodrigoprimo. All good from me.

GaryJones (approved):

PR #2724, opened 2026-04-20
Update project CoC to match WordPress Project & update out-of-date AI guidelines.

dd32:
# Description

This PR:
 - Updates the repository to include the proper WordPress project CoC.
 - Removes outdated AI guidelines which are incompatible with the project CoC.
 - Removes the outdated funding section, a vestige of the coding standards being a separate project.

Any discussion related to the validity of [the guidelines](https://make.wordpress.org/ai/handbook/ai-guidelines/) should likely be discussed with the AI team, via https://make.wordpress.org/ai/2026/02/01/ai-guidelines-for-wordpress/ or #core-ai on Slack.

Partially reverts #2598

## Suggested changelog entry

None required. Alternatively: Welcome the usage of AI tools.

## Related issues/external references

Related: https://github.com/WordPress/WordPress-Coding-Standards/pull/2722#issuecomment-4278388568

PR #2734, opened 2026-05-08
Remove outdated Funding section

dd32:
# Description

Removes the Funding section from the README and the `.github/FUNDING.yml` file.

This is a vestige of the Coding Standards being a separate project, and is no longer accurate now that WordPressCS is maintained as part of the WordPress project.

## Suggested changelog entry

None required.

Partially supersedes #2724

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,0 @@
-custom: "https://opencollective.com/php_codesniffer"
--- a/README.md
+++ b/README.md
@@ -253,10 +253,6 @@ At this moment, WordPressCS offer the following tools:
 
 See [CONTRIBUTING](.github/CONTRIBUTING.md), including information about [unit testing](.github/CONTRIBUTING.md#unit-testing) the standard.
 
-## Funding
-
-If you want to sponsor the work on WordPressCS, you can do so by donating to the [PHP_CodeSniffer Open Collective](https://opencollective.com/php_codesniffer).
-
 ## License
 
 See [LICENSE](LICENSE) (MIT).


jrfnl:
> This is a vestige of the Coding Standards being a separate project, and is no longer accurate now that WordPressCS is maintained as part of the WordPress project.

Actually no, it is not a "vestige of the Coding Standard being a separate project" and it never was.

We introduced this after WPCS 3.0 was released as the amount of time needed to maintain WPCS without funding was completely silly and we announced this as such as well: https://make.wordpress.org/core/2023/08/21/wordpresscs-3-0-0-is-now-available/

In my opinion, there is no reason at all to remove this. WPCS heavily relies on PHPCS and the related side-projects and encouraging people to fund those projects helps keep WPCS maintainable.

dd32:
> > This is a vestige of the Coding Standards being a separate project, and is no longer accurate now that WordPressCS is maintained as part of the WordPress project.
> 
> Actually no, it is not a "vestige of the Coding Standard being a separate project" and it never was.

Correct, I thought it pre-dated the Github move (the `WordPress-Coding-Standards` org to `WordPress`), but that happened in 2019, not 2023.

Addition of the Funding sections in 2023 doesn't seem like it was appropriate, to me, as any WordPress / WordPress.org projects should be funded through the WordPress Foundation, or suggesting that donations be made to that foundation.

There's a caveat that I'll mention that wp-cli still has listed sponsors, for historical reasons, although I don't believe additional ones are being added, AFAIK. https://make.wordpress.org/cli/  https://make.wordpress.org/cli/sponsors/

jrfnl:
@dd32 

1. The Charter of the WordPress Foundation doesn't allow it to fund people.
2. The Foundation has never in it's lifetime actually spend funds on software projects in the WordPress ecosphere. Instead it has squandered funds intended for the WP ecosphere on arbitrary external projects - case in point, the $100.000 to the Internet Archive.

More than anything, the WP Foundation appears to be the funding backbone of WordCamps. While worthwhile, that doesn't help WPCS.

Given the above, can you please explain how funding the WordPress Foundation will help make the maintenance of WPCS more sustainable ? (which is what the 2023 Make post was about).

PR #2738, opened 2026-06-05
PHP/TypeCasts: add XML documentation

rodrigoprimo:
# Description

This PR adds XML documentation for the `WordPress.PHP.TypeCasts` sniff.

The documentation is based on the work started by @brentwilson-clariio in #2591. I used the original commits and, in a separate commit, made the changes discussed in the review of #2591, plus I opted to replace the single bulleted standard block with three separate standard blocks, one per error/warning, and context indicating when each cast was deprecated and/or removed from PHP.

Regarding the `(unset)` cast, the sniff message suggests using `unset()` instead, so I kept that in the documentation from the original PR. But I'm not sure about it, as the behavior is different. As far as I can see, `(unset) $a` returns `null` while `unset($a)` unsets the variable `$a`.

I suggest squashing the commits before merging. I'm opening the PR without doing that to make it easier to tell my changes apart from the original changes.

## Suggested changelog entry

_N/A_

Related to: #1722
Supersedes: #2591
Closes #2591

### DIFF
--- a/WordPress/Docs/PHP/TypeCastsStandard.xml
+++ b/WordPress/Docs/PHP/TypeCastsStandard.xml
@@ -1,0 +1,56 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Type Casts"
+>
+    <standard>
+    <![CDATA[
+      Enforce normalized and safe type casts.
+
+      • Use (float) instead of legacy float casts.
+      • Do not use the (unset) cast. Use unset().
+      • Avoid the (binary) cast (discouraged).
+    ]]>
+    </standard>
+
+    <code_comparison>
+        <code title="Valid: Using normalized float casts.">
+        <![CDATA[
+$price = <em>(float)</em> $value;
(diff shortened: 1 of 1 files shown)


WordPress/Docs/PHP/TypeCastsStandard.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Type Casts"
+    >
+    <standard>
+    <![CDATA[
+    The (float) cast must be used instead of the (double) and (real) casts. The (real) cast was deprecated in PHP 7.4 and removed in PHP 8.0. The (double) cast was deprecated in PHP 8.5.
```
GaryJones: Are the parentheses needed in this English string? `The float cast...` reads just as well as `The (float) cast...`. We don't need to demonstrate the syntax for how to cast when talking about casting.

Same applies for title attributes below.

rodrigoprimo: Thanks for your review, @GaryJones. I don't have a strong preference. I went ahead and pushed a new commit applying your suggestion.

GaryJones (approved):

PR #2752, opened 2026-07-06
ConstantsHelper::is_use_of_global_constant(): add tests

rodrigoprimo:
# Description

In preparation for supporting PHPCS 4.0, this PR adds unit tests for the `ConstantsHelper::is_use_of_global_constant()` method.

It also removes the now-redundant `@covers` tag for this method from `EscapeOutputUnitTest.php`, as the method is now covered by its own dedicated test.

## Suggested changelog entry

_N/A_

Related to: #2272

### DIFF
--- a/WordPress/Tests/Helpers/ConstantsHelper/IsUseOfGlobalConstantUnitTest.inc
+++ b/WordPress/Tests/Helpers/ConstantsHelper/IsUseOfGlobalConstantUnitTest.inc
@@ -1,0 +1,134 @@
+<?php
+
+/*
+ * The below should NOT be recognized as use of a global constant.
+ */
+
+/* testVariableAssignment */
+$a = 'foo';
+
+/* testFunctionCall */
+some_function();
+
+/* testFunctionDeclaration */
+function PHP_OS() {}
+
+/* testClassInstantiation */
+$obj = new PHP_OS();
+
+/* testStaticMethodCall */
+PHP_OS::someMethod();
+
+/* testStaticClassConstantAccess */
+PHP_OS::SOME_CONST;
+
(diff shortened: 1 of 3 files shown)


WordPress/Tests/Helpers/ConstantsHelper/IsUseOfGlobalConstantUnitTest.inc
```diff
@@ -0,0 +1,134 @@
...
+use const SomeNamespace\{/* testUseConstStatementGrouped */ PHP_OS, PHP_VERSION_ID};
+
+use Foo\{Bar, /* testUseStatementGrouped */ PHP_OS};
+
+/*
+ * The below should be recognized as use of a global constant.
+ */
+
+/* testEchoStatement */
+echo PHP_OS;
```
dingo-d, line 120: Should we add `(function() { echo FOO; })();` as a test, or is this redundant?

rodrigoprimo: Thanks for reviewing this PR, @dingo-d. As far as I can see, such a test would not exercise any part of the code that is not already covered by the existing tests. The enclosing scope doesn't change the behavior of the tested method.

That said, is there anything in particular that you are concerned about that is currently not covered by the existing tests, or am I missing the reason that would justify this test that you are suggesting?

dingo-d: Nah, was just looking at the test cases and wanted to see if I can come with some additional odd possible edge cases 😄

